Issue #23, opened 2019-09-05
On startup, the userlib is required for derbySproc


prb112:
On startup, the userlib is required for derbySproc.  This hard requirement is no longer valid as the derby-sproc is no longer used in R4. 

The derby bootstrap (in FHIR Servlet Context Listener) checks for `userlib` in the server and stops startup if the bootstrap fails.

prb112:
PR - https://github.com/IBM/FHIR/pull/24

PR #24, opened 2019-09-05
On startup, the userlib is required for derbySproc #23

prb112:
- removed hard dependency on userlib
 - removed reference ato derby-sproc*.jar


prb112:
```

Issue #26, opened 2019-09-06
Swagger Issues


lmsurpre:
The FHIRSwaggerGenerator for R4 still has some issues:
* looks for XMLElement annotation even though we don't use that any more
* only adds inhereted properties when the parent is a resource
* misses all enum due to using wrong delimiter for nested classes
* missing logic for handling Boolean
* choice elements get serialized without their typeSuffices
* object definitions are no longer scoped by their container classes

PR #27, opened 2019-09-06
issue #26 - fix swagger issues

lmsurpre:
* use proper delimiter for working with subclasses; "$" not "."
* use @Required annotation to determine which fields are required
* add logic for boolean
* look for all parent classes, not just resources (SwaggerGenerator only; OpenAPIGenerator was already working)
* update handling of choice elements


fhir-swagger-generator/src/main/java/com/ibm/watson/health/fhir/openapi/generator/FHIROpenApiGenerator.java
```diff
@@ -828,13 +817,27 @@ private static void generateProperty(StructureDefinition structureDefinition, Cl
...
-        String description = null;
-        if (elementDefinition != null) {
-            description = elementDefinition.getDefinition().getValue();
+        String elementName = ModelSupport.getElementName(field);
+
+        if (ModelSupport.isChoiceElement(modelClass, elementName)) {
+            Set<Class<?>> choiceElementTypes = ModelSupport.getChoiceElementTypes(modelClass, elementName);
+            ElementDefinition elementDefinition = getElementDefinition(structureDefinition, modelClass, elementName + "[x]");
+            String description = elementDefinition.getDefinition().getValue();
+            for (Class<?> choiceType : choiceElementTypes) {
```
albertwang-ibm: I found the following result in the generated openapi document, and seems we combine the field name with the type names to create multiple fields, not sure if we can use "oneof" in this case... 

                           "subjectCodeableConcept": {
                                "$ref": "#/components/schemas/CodeableConcept",
                                "description": "A code or group definition that describes the intended subject of the activity being defined."
                            },
                            "subjectReference": {
                                "$ref": "#/components/schemas/Reference",
                                "description": "A code or group definition that describes the intended subject of the activity being defined."
                            },

lmsurpre: Combining the field name with the typeName actually produces the proper element name for choice elements  (i.e. if the type is Reference we want a serialized field name of "subjectReference" not "subject").

I considered using "oneOf", but to me it looks like that can only be used at the top level, meaning you'd need to repeat the full list of properties for each permutation of choice elements.  That just isn't feasible when there are multiple choice elements in the same resource type.

lmsurpre: I also verified that this representation of choice elements is consistent with the JSON Schema distributed with the specification.  They only us "oneOf" in places where a field can be any resource (e.g. Bundle.entry.resource).

Issue #28, opened 2019-09-06
Conformance Registry


JohnTimm:
Create a registry that holds conformance resources (e.g. structure definitions, value sets, etc.) that other components can use for various purposes.

lmsurpre:
seems to be working

PR #29, opened 2019-09-06
Issue #28 - FHIR conformance registry + US core provider

JohnTimm:


lmsurpre:
from john:
> the meat is in FHIRConformanceRegistry, ConformanceResource, ConformanceUtil

lmsurpre:
Adding all those spec conformance resources makes our project quite a bit larger.

For the fhir-examples project, we kept it separate from the server build since its just a test dependency.
Since this one is a runtime dependency, I guess it makes sense to include it in the main build.

I was a little worried about 
1. additional build time
2. additional size of jar / war

but the build completed pretty fast:
```[INFO] fhir-conformance ................................... SUCCESS [  9.629 s]
[INFO] fhir-us-core ....................................... SUCCESS [  3.963 s]```

The built jar is ~10M for me:
`-rw-r--r--   1 lmsurpre  staff   9.9M Sep  9 16:30 fhir-conformance-4.0.0-SNAPSHOT.jar`

Not sure what Travis/Maven limits look like, but for comparison our fhir-examples project takes ~20 seconds to build and weighs in about 33MB.

lmsurpre:
I wonder if we should use some naming convention for IGs to group them alpahbetically in the project list. Maybe "fhir-ig-us-core"?

That way you can introduce projects for "fhir-ig-davinci", "fhir-ig-mcode", etc. in the future,
similar to how we use "fhir-operation-x" to group the operations.

prb112:
@lmsurpre we can include it in the parent module and exclude it as necessary.

JohnTimm:
> I wonder if we should use some naming convention for IGs to group them alpahbetically in the project list. Maybe insert "ig" so it becomes "fhir-ig-us-core"?
> 
> That way you can introduce projects for "fhir-ig-davinci", "fhir-ig-mcode", etc. in the future,
> similar to how we use "fhir-operation-x" to group the operations.

My only concern with this is that many of these profiles and extensions are "loose" in that they are not affiliated with a specific IG. If we want to group them together, then I would vote for:

fhir-conformance-spec
fhir-conformance-us-core
fhir-conformance-carin-bb
fhir-conformance-davinci
fhir-conformance-mcode

lmsurpre:
Only the spec ones are "loose", right?  I assumed that would stay in the base "fhir-conformance" project and stay separate from the ig ones.  But I suppose it doesn't matter too much, was just looking for a pattern we could follow and its easier to change the names now than when everyone has it imported.

JohnTimm:
Introduced fhir-conformance-spec project
Renamed fhir-us-core to fhir-conformance-us-core
Moved definitions/ and resources/ directory contents from fhir-conformance to fhir-conformance-spec
Additional minor cleanup

JohnTimm:
> Only the spec ones are "loose", right? I assumed that would stay in the base "fhir-conformance" project and stay separate from the ig ones. But I suppose it doesn't matter too much, was just looking for a pattern we could follow and its easier to change the names now than when everyone has it imported.

Yes, I think that is true. I am going to revert my decision to rename fhir-conformance-* and go with your suggestion. The loose conformance resources will be in fhir-conformance and everything else will be in a fhir-ig-* project.

PR #34, opened 2019-09-09
issue #26 - include containing classes in element definition ids to avoid conflicts

lmsurpre:
In DSTU2 we prefixed classes with the name of the containing type.
However, in R4, we moved to a nested class structure and so the simple names are no longer unique.
Furthermore, there are now some BackboneElements that share a local name
within the same resource (e.g. ImplementationGuide.Definition.resource
and ImplementationGuide.Manifest.resource) and so we need to use the
full path of contained resource simple names to ensure uniqueness.

### DIFF
--- a/fhir-openapi/pom.xml
+++ b/fhir-openapi/pom.xml
@@ -14,7 +14,6 @@
     <packaging>war</packaging>
 
     <build>
-        <finalName>${project.name}</finalName>
         <plugins>
             <plugin>
                 <artifactId>maven-war-plugin</artifactId>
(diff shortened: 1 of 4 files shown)


JohnTimm:
I think we should use "." between the class names as it follows the official XML schema for naming the backbone classes.

lmsurpre:
> Should we use # as separator for the inner classes?
I don't think "#" is a good separator in names because it has special meaning in paths, but just pushed a new commit which uses "-" as a separator...see if you like that.

> or can we simply use .getName() instead of .getSimpleName?
We could have started from that and adjusted it, but I went with a simple loop to construct the name isntead.

lmsurpre:
added one more commit to fix the name of the webapp built from the fhir-openapi project

PR #40, opened 2019-09-10
Issue #32 Enhance Bulk JavaBatch to use single COS object.

albertwang-ibm:

### DIFF
--- a/fhir-bulkimportexport/src/main/java/com/ibm/waston/health/fhir/bulkcommon/COSUtils.java
+++ b/fhir-bulkimportexport/src/main/java/com/ibm/waston/health/fhir/bulkcommon/COSUtils.java
@@ -1,0 +1,126 @@
+/**
+ * (C) Copyright IBM Corp. 2019
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.ibm.waston.health.fhir.bulkcommon;
+
+import java.io.InputStream;
+import java.util.List;
+import java.util.logging.Logger;
+
+import com.ibm.cloud.objectstorage.ClientConfiguration;
+import com.ibm.cloud.objectstorage.SDKGlobalConfiguration;
+import com.ibm.cloud.objectstorage.auth.AWSCredentials;
+import com.ibm.cloud.objectstorage.auth.AWSStaticCredentialsProvider;
+import com.ibm.cloud.objectstorage.auth.BasicAWSCredentials;
+import com.ibm.cloud.objectstorage.client.builder.AwsClientBuilder.EndpointConfiguration;
+import com.ibm.cloud.objectstorage.oauth.BasicIBMOAuthCredentials;
+import com.ibm.cloud.objectstorage.services.s3.AmazonS3;
+import com.ibm.cloud.objectstorage.services.s3.AmazonS3ClientBuilder;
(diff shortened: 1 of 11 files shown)


albertwang-ibm:
Created custom checkpointAlgorithm to leverage IBM COS multipe-parts upload API to use single cos object for the bulk exporting. In the newly added custom algorithm, we check the to-be-pushed data size to make sure it's bigger or equal to the minimal part size of COS multipe-part upload(5M bytes, except for the last part).   And if cosbucketobjectname is configured, then single COS object with the configured name will be used for exporting, otherwise, multiple objects will be used.

fhir-bulkimportexport/src/main/java/com/ibm/waston/health/fhir/bulkcommon/COSUtils.java
```diff
@@ -0,0 +1,126 @@
...
+        } catch (Exception sdke) {
+            log("startPartUpload", "Upload start Error: " + sdke.getMessage());
+            throw sdke;
+        }
+    }
+    
+    
+    public static AmazonS3 getCosClient(String cosCredentialIbm, String cosApiKeyProperty, String cosSrvinstId, 
+            String cosEndpintUrl, String cosLocation) {
+        SDKGlobalConfiguration.IAM_ENDPOINT = "https://iam.cloud.ibm.com/oidc/token";
```
lmsurpre: probably a comment for last PR rather than this one, but use of ALL CAPS here makes it look like a constant when apparently its not

lmsurpre:
> if cosbucketobjectname is configured, then single COS object with the configured name will be used for exporting, otherwise, multiple objects will be used.

the export is supposed to put different resource types into different objects, are you still doing that?

albertwang-ibm:
> > if cosbucketobjectname is configured, then single COS object with the configured name will be used for exporting, otherwise, multiple objects will be used.
> 
> the export is supposed to put different resource types into different objects, are you still doing that?

yes, we only support one ResourceType in one job. and the caller has fully control of the object name(s).  if using one single object (by setting cosbucketobjectname) for the uploading(One single COS object - can be up to TBs, is big enough for holding all the exported data).  If we want to use multiple objects, then cos.bucket.pathprefix is used to control the prefix of the generated objects names' the name will be something like prefix/xxxxx-xxxxx-xxx(UUID). per discussed, I will change to use jobid-resourcetype-sequence number as the object name for the multi-object solution.

albertwang-ibm:
added job parameter for the page size, and also make code changes to remove the cosbatchsize limit for chunk export,  and also use the new naming pattern jobid/resourcetype_##.ndjson for the cos objects.

fhir-bulkimportexport/src/main/java/com/ibm/waston/health/fhir/bulkcommon/COSUtils.java
lmsurpre: albert pointed out that this comes from the COS SDK, not from us

albertwang-ibm: this constant is defined by COS SDK.

lmsurpre:
LGTM

Issue #66, opened 2019-09-14
Configure server to override "self" URIs with external path


lmsurpre:
**Is your feature request related to a problem? Please describe.**
We're running the FHIR Server behind a proxy and that proxy modifies the request URLs.

Request come in to an endpoint like `https://[host]/wh-fhir-dev/api/v4` but the server is returning URIs like `https://[host]/fhir-server/api/v4/$export-status?job=9`.

Is there a way to use the origin request URL in the server URLs returned by the server?

I think it affects at least:
* search result `self` uri
* Bundle entry fullUrl values
* bulkimportexport polling URLs

**Describe the solution you'd like**
https://[host]/wh-fhir-dev/api/v4
Either:
* a setting the server will use as the "base" in all self uris; or
* a way to tell the server which HTTP header has the original request URI (so it can use that to construct self uris)

albertwang-ibm:
if we have to set/change a new config, why don't we simply change the existing contextRoot in server.xml instead of adding another config? BTW, usually proxy server doesn't pass the original request url to the internal servers.

prb112:
I think Lee is saying that we add a new configuration. 
The context root doesn't always have the Proxy origin.

albertwang-ibm:
yeah, if we want to change hostname part also, then we may need a new configure. actually, can we configure proxy to do the translaction if the hostname is different?

albertwang-ibm:
ideally, proxy can change the response also ...

albertwang-ibm:
but I agree that adding a new config in fhir server side sounds like a simpler solution.

prb112:
I'll try to fix as-soon-as-possible

prb112:
PR https://github.com/IBM/FHIR/pull/659

lmsurpre:
#659 removes the bulkdata-specific hostname/path properties that we invented, but I still think we need:
A. a way to infer the uri of the original request from headers; and/or
B. a fhir-server-config property to overwrite the request uri  (relates to #660)

prb112:
```
{
    "__comment": "FHIR Server configuration",
    "fhirServer": {
        "core": {
            "truststoreLocation": "resources/security/fhirTruststore.jks",
            "truststorePassword": "change-password",
            "tenantIdHeaderName": "X-FHIR-TENANT-ID",
            "datastoreIdHeaderName": "X-FHIR-DSID",
            "originalRequestUriHeaderName" : "IBM-PROXY",
            "checkReferenceTypes": false
        },
[...]

PR #79, opened 2019-09-15
Issue 67 - fix ndjson workaround absolute url issue

lmsurpre:
* fix ndjson implementation
* return absolute url for job status checks

the latter is just a quickfix.  we need https://github.com/IBM/FHIR/issues/66 for the proper fix.

### DIFF
--- a/fhir-bulkimportexport/src/main/java/com/ibm/watson/health/fhir/bulkexport/ChunkProcessor.java
+++ b/fhir-bulkimportexport/src/main/java/com/ibm/watson/health/fhir/bulkexport/ChunkProcessor.java
@@ -6,14 +6,19 @@
 
 package com.ibm.watson.health.fhir.bulkexport;
 
+import java.io.StringWriter;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.batch.api.chunk.ItemProcessor;
 import javax.batch.runtime.context.JobContext;
 import javax.inject.Inject;
 
+import com.ibm.watson.health.fhir.model.format.Format;
+import com.ibm.watson.health.fhir.model.generator.FHIRGenerator;
+import com.ibm.watson.health.fhir.model.generator.exception.FHIRGeneratorException;
 import com.ibm.watson.health.fhir.model.resource.Resource;
 
 /**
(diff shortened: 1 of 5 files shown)


fhir-bulkimportexport/src/main/java/com/ibm/watson/health/fhir/bulkexport/ChunkProcessor.java
```diff
@@ -53,11 +58,24 @@ public Object processItem(Object arg0) {
         int dataSize = 0;
                         
         for (Resource res : resources) {
-            String ndJsonLine = res.toString().replace("\r", "").replace("\n", "");
+            String ndJsonLine = null;
+            StringWriter writer = new StringWriter();
+            try{
+                FHIRGenerator.generator(Format.JSON).generate(res, writer);
```
prb112: Since we are using a non-closing outputstream to generate, the start upload could occur first, and stream out to the OutputStream to COS.

fhir-operation-bulkdata/src/main/java/com/ibm/watson/health/fhir/operation/bullkdata/client/BulkDataClient.java
```diff
@@ -188,7 +188,8 @@ public String submit(MediaType outputFormat, Instant since, List<String> types,
         // From the response
         String jobId = Integer.toString(response.getInstanceId());
 
-        return "/wh-fhir-dev/api/v4/$export-status?job=" + jobId;
+        String hostname = properties.get(BulkDataConfigUtil.SERVER_HOSTNAME);
+        return "https://" + hostname + "/wh-fhir-dev/api/v4/$export-status?job=" + jobId;
```
prb112: Cool - I think we could <config> wh-fhir-dev

fhir-bulkimportexport/src/main/java/com/ibm/watson/health/fhir/bulkexport/ChunkProcessor.java
albertwang-ibm: yes, the process was added simply for formatting the output and also for the purpose of trying a chunck with all components (read, process. write) in it. and I do have plan to drop the process, and use write to push the resources to COS directly. I will have a separated PR for this.

fhir-operation-bulkdata/src/main/java/com/ibm/watson/health/fhir/operation/bullkdata/client/BulkDataClient.java
albertwang-ibm: because /wh-fhir-dev/ could also be changed, e.g, to make it also works in local test env, I think we should store both host name and  "/wh-fhir-dev/api/v4/$export-status?job=" together and use the following pattern:

return "https://" + exportUrl + jobId;

I can make the change in my next PR

prb112: @albertwang-ibm I raised PR80 for this

Issue #76, opened 2019-09-15
Bulk should accept _outputFormat requests with encoded "+"


lmsurpre:
**Describe the bug**
When _outputFormat is submitted as a query parameter, the UriInfo injected in FHIRResource replaces the "+" char with a space.

This is the same issue we hit with the FHIR "_format" query parameter, but in a different part of the code.

**To Reproduce**
`{{http}}://{{host}}:{{port}}/{{basePath}}/$export?_outputFormat=application/fhir+ndjson&_since=2019-01-01T08:21:26.94-04:00&_type=Patient`

**Expected behavior**
Not sure if its liberty or JAX-RS or CXF at fault here, but we need to work around the issue for this to work.

prb112:
It's coming from Liberty. I can take this one.

PR #82, opened 2019-09-16
Bulk should accept _outputFormat requests with encoded "+" #76

prb112:
- fix for decode=false on URL's query parameters


fhir-server-test/src/test/java/com/ibm/watson/health/fhir/server/test/ExportOperationTest.java
```diff
@@ -47,7 +47,7 @@
 
     public static final String FORMAT = "application/fhir+ndjson";
 
-    public static final boolean ON = true;
+    public static final boolean ON = false;
```
lmsurpre: what is ON?

prb112: ON constant was my way of not overriding DEBUG in an integration test.  I don't think it will pass otherwise. *it flips the test to ON or OFF.

PR #129, opened 2019-09-20
introduce github action for performing maven build on pull_requests

lmsurpre:
Also includes a change to configure the fhir-examples dependency version in the fhir-parent pom.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -13,7 +13,13 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 1.8
+    - name: Build tools
+      run: mvn install --file fhir-tools/pom.xml
     - name: Build samples
       run: mvn install --file fhir-examples/pom.xml
     - name: Build with Maven
-      run: mvn package --file fhir-parent/pom.xml
+      run: mvn install --file fhir-parent/pom.xml
+    - name: Build sample generator
+      run: mvn package --file fhir-examples-generator/pom.xml
+    - name: Build benchmark
+      run: mvn package --file fhir-benchmark/pom.xml
--- a/fhir-parent/pom.xml
+++ b/fhir-parent/pom.xml
@@ -622,12 +622,4 @@
             </properties>
         </profile>
     </profiles>
-
-    <repositories>
-        <repository>
-            <id>github</id>
-            <name>GitHub lmsurpre Apache Maven Packages</name>
-            <url>https://maven.pkg.github.com/lmsurpre/fhir</url>
-        </repository>
-    </repositories>
 </project>
--- a/fhir-persistence-jdbc/pom.xml
+++ b/fhir-persistence-jdbc/pom.xml
@@ -30,6 +30,7 @@
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>fhir-examples</artifactId>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>${project.groupId}</groupId>
--- a/fhir-persistence/pom.xml
+++ b/fhir-persistence/pom.xml
@@ -45,6 +45,7 @@
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>fhir-examples</artifactId>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.testng</groupId>
--- a/fhir-search/pom.xml
+++ b/fhir-search/pom.xml
@@ -44,6 +44,7 @@
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>fhir-examples</artifactId>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.testng</groupId>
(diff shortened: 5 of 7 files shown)


.github/workflows/maven.yml
```diff
@@ -0,0 +1,19 @@
...
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    - name: Build samples
+      run: mvn install --file fhir-examples/pom.xml
+    - name: Build with Maven
+      run: mvn package --file fhir-parent/pom.xml
```
prb112: I think we should build fhir-tools as well? thoughts?

fhir-parent/pom.xml
```diff
@@ -616,4 +622,12 @@
             </properties>
         </profile>
     </profiles>
+
+    <repositories>
+        <repository>
+            <id>github</id>
+            <name>GitHub lmsurpre Apache Maven Packages</name>
+            <url>https://maven.pkg.github.com/lmsurpre/fhir</url>
```
prb112: should we put to ibm/fhir?

fhir-persistence-jdbc/pom.xml
```diff
@@ -30,8 +30,6 @@
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>fhir-examples</artifactId>
-            <version>${project.version}</version>
-            <scope>test</scope>
```
prb112: why is this changing the scope?

fhir-persistence/pom.xml
```diff
@@ -45,8 +45,6 @@
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>fhir-examples</artifactId>
-            <version>${project.version}</version>
-            <scope>test</scope>
```
prb112: change of scope is ...

fhir-search/pom.xml
```diff
@@ -44,16 +44,14 @@
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>fhir-examples</artifactId>
-            <version>${project.version}</version>
-            <scope>test</scope>
```
prb112: change of scope?

fhir-server-test/pom.xml
```diff
@@ -1,12 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- THIS PRODUCT CONTAINS RESTRICTED MATERIALS OF IBM 5724-H88, 5724-J08, 
-    5724-I63, 5655-W65, COPYRIGHT International Business Machines Corp., 2014 
-    All Rights Reserved * Licensed Materials - Property of IBM US Government 
-    Users Restricted Rights - Use, duplication or disclosure restricted by GSA 
-    ADP Schedule Contract with IBM Corp. -->
```
prb112: good find

fhir-server-test/pom.xml
```diff
@@ -61,8 +55,6 @@
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>fhir-examples</artifactId>
-            <version>${project.version}</version>
-            <scope>test</scope>
```
prb112: scope change?

fhir-validation/pom.xml
```diff
@@ -38,8 +38,6 @@
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>fhir-examples</artifactId>
-            <version>${project.version}</version>
-            <scope>test</scope>
```
prb112: scope change?

.github/workflows/maven.yml
lmsurpre: I was actually looking forward to removing the step that builds fhir-examples to speed it up a tad.  I guess it boils down to what we're hoping to achieve with this...ideally there would be some way to only build the maven projects that actually have changes (either directly or in their dependencies).
However, since I don't know how to do that, I'm not opposed to building "everything" for now.  Should we build the fhir-benchmark project while we're at it, too?  Know of any others that aren't hooked into the main server build?

fhir-parent/pom.xml
lmsurpre: good catch.  this pipeline doesn't even put to the package registry yet and so this PR shouldn't have included this.  will clean it out now.

fhir-persistence-jdbc/pom.xml
lmsurpre: its not.  we're just inhereting the scope from the parent now.  i think.
but i'll put the `scope` element back in to make sure (its more clear this way anyway).

.github/workflows/maven.yml
lmsurpre: OK, I pushed a change to build everything (including fhir-tools, fhir-examples-generator, and fhir-benchmark.

fhir-persistence-jdbc/pom.xml
lmsurpre: added scope=test back

PR #138, opened 2019-09-23
Remove renaming note and reduce unneccessary section header

lmsurpre:


lmsurpre:
I killed the CI build on purpose...no need to run for README-only change

Issue #118, opened 2019-09-20
Delete nonexist resource should return 200 instead of 404


lmsurpre:
https://touchstone.aegis.net/touchstone/scriptexecution?exec=201908231432318334944596&qn=/FHIR4-0-0-Advanced/Patient/05-delete/fhir-api-r4-patient-delete-02-conditional-single#WAITING_OPER

https://touchstone.aegis.net/touchstone/scriptexecution?exec=201908231432318334944596&qn=/FHIR4-0-0-Advanced/Patient/05-delete/fhir-api-r4-patient-delete-01-base#WAITING_OPER

lmsurpre:
there's two related scenarios:
1. delete a resource that has never existed
2. delete a resource that did exist but has already been deleted 

Currently I think we return a 404 (Not Found) for both cases.  Sounds like Touchstone wants us to return a 200 or a 204 for scenario 2.

But lets continue returning a 404 for scenario 1.

lmsurpre:
Albert found that Touchstone is expected a 200 or 204 in scenario 1 as well and the spec seems to agree:

https://www.hl7.org/fhir/http.html#delete

    Upon successful deletion, or if the resource does not exist at all, the server should return either a 200 OK if the response contains a payload, or a 204 No Content with no response payload, or a 202 Accepted if the server wishes to be non-commital about the outcome of the delete.

lmsurpre:
verified by the passing test at https://touchstone.aegis.net/touchstone/scriptexecution?exec=201909302354331933965186&qn=/FHIR4-0-0-Advanced/Patient/05-delete/fhir-api-r4-patient-delete-02-conditional-single#WAITING_OPER

Issue #116, opened 2019-09-20
UpdateCreate response code should be 201 instead of 200


lmsurpre:
https://touchstone.aegis.net/touchstone/scriptexecution?exec=201908231432318334944596&qn=/FHIR4-0-0-Advanced/Patient/04-update/fhir-api-r4-patient-update-03-create#WAITING_OPER

PR #155, opened 2019-09-26
issues #116 and #118 update response codes to appease touchstone

albertwang-ibm:
#116 UpdateCreate response code should be 201 instead of 200
#118 Deleting a nonexistent resource should return a 200


albertwang-ibm:
Fixes for both #116 and #118

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -545,7 +545,8 @@ public Response conditionalDelete(@PathParam("type") String type) throws Excepti
             status = e.getHttpStatus();
             return exceptionResponse(e);
         } catch (FHIRPersistenceResourceNotFoundException e) {
-            status = Response.Status.NOT_FOUND;
+            // Return 200 instead of 404 to pass TouchStone test
+            status = Response.Status.OK;
```
lmsurpre: can we do a 204 instead?   or touchstone is only happy with 200?

albertwang-ibm: because we have operation outcome, so I chose to use 200.
Touchstone can accept:
(1) 200 with operation outcome.
(2) 204 not content.

Issue #57, opened 2019-09-13
Extra EAR projects are created needlessly on Maven Project import


prb112:
Extra EAR projects are created needlessly on Maven Project import

PR #156, opened 2019-09-26
Fix for ear creation #57

prb112:


prb112:
This is an investigation only.  If it works, we can consider merging it.  I'm sharing here to not lose track.

prb112:
Albert reported that this works. 

@lmsurpre when you get a chance can you verify too.

Issue #117, opened 2019-09-20
Create resource if conditionUpdate not found


lmsurpre:

albertwang-ibm:
double checked the test case again, seems it doesn't require the prefer header, I'm going to work on this.

albertwang-ibm:
The other test cases in the same test scripts do require prefer-header support

albertwang-ibm:
According to the spec, conditional update a resource without match and without id in the input resource, server should create the resource with server assigned resource id and with 201 response code instead of return 400 error.
Conditional update a resource with 1 match but with a id in the input resource which is different from the id of the search result, server should return 400 error.

PR #158, opened 2019-09-26
issue #117 Create resource if conditionalUpdate not found

albertwang-ibm:
According to the spec, 
(1) Conditional update a resource without match and without id in the input resource, server should create the resource with server assigned resource id and with 201 response code instead of return 400 error.
(2) Conditional update a resource with 1 match but with a id in the input resource which is different from the id of the search result, server should return 400 error.

### DIFF
--- a/fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
@@ -1228,12 +1228,6 @@ public class FHIRResource implements FHIRResourceHelpers {
                             + "' does not match type specified in request URI: " + type;
                     throw buildRestException(msg, Status.BAD_REQUEST, IssueType.ValueSet.INVALID);
                 }
-
-                // Make sure the resource has an 'id' attribute.
-                if (newResource.getId() == null) {
-                    String msg = "Input resource must contain an 'id' attribute.";
-                    throw buildRestException(msg, Status.BAD_REQUEST, IssueType.ValueSet.INVALID);
-                }
             }
 
             // Next, if a conditional update was invoked then use the search criteria to find the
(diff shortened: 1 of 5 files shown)


fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -1290,6 +1299,12 @@ private FHIRRestOperationResponse doPatchOrUpdate(String type, String id, FHIRPa
                     String msg = "The 'id' parameter is required for an update/pach operation.";
                     throw buildRestException(msg, Status.BAD_REQUEST, IssueType.ValueSet.REQUIRED);
                 }
+                
+                // Make sure the resource has an 'id' attribute.
+                if (newResource.getId() == null) {
```
prb112: out of curiosity - at this point?  why would `newResource.getId() == null`?

albertwang-ibm: yeah, it can happen if there is no resource id in the request body. according to the spec, in this case, there should always be a resource id in the request body, otherwise, we should 400.

prb112: OK. 

Should ValueSet.Required be set on 1306?  like line 1300?

prb112:
Reviewed e2e.  master has an update for the failing test which increases the latch timeout to 10 seconds.

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
albertwang-ibm: I think Invalid should also be fine, because it means resource without an id in the request body is invalid content...

prb112: Right - just double checking for consistency.

albertwang-ibm:
because the pipeline test failure is not related to the code change, so Iet me merge now.

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
albertwang-ibm: I think invalid should also be fine, it means the resource without id in the request body is a invalid contents.

Issue #162, opened 2019-09-27
conditionalUpdateOneMatch without resourceId in the input resource causes 500 error


albertwang-ibm:
we should use the found resource id for the input resource to prevent the null point exception.

Sep 27 13:59:12 fhir-server-r4-59cbb6cc48-nczhm fhir-server-r4  Caught exception while processing 'update/patch' request.
java.lang.NullPointerException
	at com.ibm.fhir.server.resources.FHIRResource.doPatchOrUpdate(FHIRResource.java:1346)
	at com.ibm.fhir.server.resources.FHIRResource.doUpdate(FHIRResource.java:1194)
	at com.ibm.fhir.server.resources.FHIRResource.conditionalUpdate(FHIRResource.java:331)
        ...
ep 27 13:59:12 fhir-server-r4-59cbb6cc48-nczhm fhir-server-r4  Performing a rollback of transaction on current thread... 
Sep 27 13:59:12 fhir-server-r4-59cbb6cc48-nczhm fhir-server-r4  Completed request[0.075 secs]: tenantId:[6d35976d-99b7-4842-92f5-ad48f7dd1274] dsId:[default] user:[&lt;unauthenticated&gt;] method:[PUT] uri:[https://us-south.wh-fhir.dev.cloud.ibm.com/fhir-server/api/v4/Patient?family=Lastname8737847339223297751388446&amp;given=Firstname8737847339223297751388446 status:[500]

Issue #161, opened 2019-09-27
Content-Type of the response doesn't has the character-set in the request


albertwang-ibm:
There is "Accept-Charset: utf-8" or "Accept: application/fhir+xml;charset=UTF-8" in the request header, but in the "Content-Type" response header, there was no "utf-8" in it.

https://touchstone.aegis.net/touchstone/scriptexecution?exec=201909271137469410753433&qn=/FHIR4-0-0-Advanced/Patient/02-vread/fhir-api-r4-patient-vread-01-found#WAITING_OPER

PR #163, opened 2019-09-30
issue #161 #167 fixes for missing charset in content-type response header

albertwang-ibm:
Issue #161:
(1) Added support for "Accept-Charset" of the request header, if there is charset defined here, then we added the charset to the content-type header of the response;
(2) If the json/xml(but not the charset) format is override by _format http query parameter, then we add the charset if any defined in either the "Accept" or "Accept-Charset" header of the request. 

Issue #167:
Added support to override charset via _format.

issue #162:
nullpointer exception caused by null resourceID


fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRHttpServletRequestWrapper.java
```diff
@@ -45,6 +47,9 @@
     public static final String UTF16 = "utf-16";
     public static final String DEFAULT_ACCEPT_HEADER_VALUE = FHIRMediaType.APPLICATION_FHIR_JSON;
     public static final String HEADER_X_METHOD_OVERRIDE = "X-Method-Override";
+    public static final String CHARSET = "charset";
+    public static final String ACCEPT = "Accept";
+    public static final String ACCEPTCHARSET = "Accept-Charset";
```
lmsurpre: maybe `ACCEPT_CHARSET` instead...very minor

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRHttpServletRequestWrapper.java
```diff
@@ -315,6 +341,19 @@ private String possiblyMapQueryParameterValue(String headerName, String value) {
                 else if(delegateQueryString.contains("_format=application/fhir+xml")) {
                     value = FHIRMediaType.APPLICATION_FHIR_XML;
                 }
+                
+                // if there is charset defined in _format, then take it.
+                if (delegateQueryString.contains("_format=application/fhir+json")
```
prb112: doubling up on checks on the contains method.  
embed in existing else/if blocks. 
use indexOf on the delegateQueryString to get the initial index, and check to see if the following char is ';' or non-existant

something like this... 
int loc = delegateQueryString.indexOf("_format="); 
then 
if (loc > 0){
  // change between the four cases. 
  
}

prb112: per discussion the _format decode/false in this case also works

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRHttpServletRequestWrapper.java
albertwang-ibm: good point, let me change

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRHttpServletRequestWrapper.java
albertwang-ibm: I don't use indexOf, because we have to consider if charset is the first or the last format or in the middle of the _format string

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRHttpServletRequestWrapper.java
albertwang-ibm: Done!

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRHttpServletRequestWrapper.java
albertwang-ibm: changed to use existing contains if else blocks.

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRHttpServletRequestWrapper.java
```diff
@@ -305,18 +331,31 @@ private String possiblyMapQueryParameterValue(String headerName, String value) {
...
                     value = FHIRMediaType.APPLICATION_FHIR_JSON;
+                    addCharset = true;
                 }
                 else if(delegateQueryString.contains("_format=application/fhir+xml")) {
                     value = FHIRMediaType.APPLICATION_FHIR_XML;
+                    addCharset = true;
+                }
+                
+                // if there is charset defined in _format, then take it.
+                if (addCharset) {
```
prb112: makes sense to me

prb112:
Chatted with Albert offline to understand logic - the changes make sense as we both came to understand the changes and the existing code.

Issue #182, opened 2019-10-01
false negative on ref-1 constraint for reference to contained resource within bundle entry


lmsurpre:
**Describe the bug**
I created a transaction bundle using Synthea, but it fails our validator with the following error:
> ref-1: SHALL have a contained resource if a local reference is provided, expression: Bundle.entry[0].resource.generalPractitioner[0]

The reference looks ok to me, so I think its a bug with our Validator.

**To Reproduce**
Validate a bundle that has an entry with a reference that points to a contained resource like this:
```
    Bundle bundle = parser.parse(ExamplesUtil.reader("json/ibm/minimal/Bundle-1.json"));
    Patient patient = parser.parse(ExamplesUtil.reader("json/ibm/minimal/Patient-1.json"));
    Practitioner practitioner = parser.parse(ExamplesUtil.reader("json/ibm/minimal/Practitioner-1.json"));
        
    patient = patient.toBuilder()
                         .contained(practitioner.toBuilder().id(Id.of("test")).build())
                         .generalPractitioner(Reference.builder().reference(String.of("#test")).build())
                         .build();
        
    bundle = bundle.toBuilder()
                       .entry(Entry.builder().resource(patient).build())
                       .build();
```

**Expected behavior**
ref-1 should properly evaluate the constraint

**Additional context**
We had a hard time figuring out the proper semantics of the `%resource` variable within the FHIRPath expression.  I thought we had considered all the cases, but apparently not.

JohnTimm:
We had a hard time figuring out the proper semantics because the spec is ambiguous.

lmsurpre:
In that GForge they say they are going to update ref-1 to use %rootResource, but given this example of a Bundle.entry.resource with a link to a contained resource (Bundle.entry.resource.contained) I'm not sure thats right.

Where do they define `%rootResource`?  Maybe ping Graham on Zulip to make sure they considered this?

lmsurpre:
OK, its actually defined in the FHIR specification at https://build.fhir.org/fhirpath.html#variables

> When a DomainResource contains another resource, and that contained resource is the focus (%resource) then %rootResource refers to the container resource.

I guess that the presence of the focus resource (%resource) within a bundle is immaterial...we just need to make sure `%rootResource` resolves to the resource that contains it (and not the real "root" which would be the parent Bundle).

PR #183, opened 2019-10-01
issue #182 - update FHIRValidator.getResource to handle contained resources properly

lmsurpre:
Previously, we always returned the root of the FHIRPath tree of nodes for datatype constraints, but this was wrong for contained resources within a Bundle entry.

Also added unit tests for validating various constraints within a Bundle entry, including:
* Bundle in Bundle
* local reference to contained resource within a bundle
* local reference to a peer contained resource within a bundle

Also converted FHIRValidatorTest to be a unit test and removed the unnecessary outputStream wrapper.


lmsurpre:
closed in favor of #186

Issue #197, opened 2019-10-05
use JsonBuilderFactory and JsonReaderFactory to improve performance


albertwang-ibm:
Json.createObjectBuilder and Json.createReader cause performance issues, we need to change to use according Factory to avoid this issue

PR #199, opened 2019-10-07
issue #197 use JsonBuilder and JsonReaderFactory to improve performance

albertwang-ibm:


albertwang-ibm:
not sure if it's related, but I did find that the integration tests finished within 2 minutes which seems faster than before.

fhir-config/src/main/java/com/ibm/fhir/config/mock/MockPropertyGroup.java
```diff
@@ -9,13 +9,16 @@
 import java.io.ByteArrayInputStream;
 
 import javax.json.Json;
+import javax.json.JsonException;
 import javax.json.JsonObject;
 import javax.json.JsonReader;
+import javax.json.JsonReaderFactory;
 
 import com.google.gson.Gson;
```
prb112: is Gson still necessary in this case?

Issue #201, opened 2019-10-07
FHIRParserException message should include the path


lmsurpre:
While updating some DSTU2 resources to R4, I hit a few issues where the parser was rejecting the request.  Unfortunately, the error output from my test doesn't include the path that caused the issue:

```
com.ibm.fhir.model.parser.exception.FHIRParserException: Expected: JsonArray but found: STRING  [probeId=9-1b-78-d-0e9c3608-4821-43e8-aa6f-f0be5d04b65a]  [probeId=9-1b-78-d-f549dd93-96c6-4ef1-9185-d3f9fb6c4f3f]
    at com.ibm.fhir.model.parser.FHIRJsonParser.parseAndFilter(FHIRJsonParser.java:76)
    at com.ibm.fhir.model.parser.FHIRJsonParser.parse(FHIRJsonParser.java:68)
    at com.ibm.fhir.persistence.search.test.AbstractPLSearchTest.readResource(AbstractPLSearchTest.java:221)
    at com.ibm.fhir.persistence.search.test.AbstractSearchDateTest.testCreateBasicResource(AbstractSearchDateTest.java:21)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:498)
    at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:86)
    at org.testng.internal.Invoker.invokeMethod(Invoker.java:643)
    at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:820)
    at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1128)
    at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:129)
    at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:112)
    at org.testng.TestRunner.privateRun(TestRunner.java:782)
    at org.testng.TestRunner.run(TestRunner.java:632)
    at org.testng.SuiteRunner.runTest(SuiteRunner.java:366)
    at org.testng.SuiteRunner.runSequentially(SuiteRunner.java:361)
    at org.testng.SuiteRunner.privateRun(SuiteRunner.java:319)
[...]

PR #202, opened 2019-10-07
issue #201 - append the path to the message in FHIRParserException

lmsurpre:
Also, stop wrapping FHIRParserException in another FHIRParserException in FHIRJsonParser


fhir-core/src/main/java/com/ibm/fhir/exception/FHIRException.java
```diff
@@ -85,8 +85,8 @@ public String getUniqueId() {
     }
 
     @Override
-    public String getMessage() {
-        String superMsg = super.getMessage();
+    public String toString() {
```
lmsurpre: by overwriting toString instead of getMessage, it becomes possible to call getMessage to get just the message (without the probeId).  full ramifications of this change are not yet known/addressed.

Issue #210, opened 2019-10-08
WebSocket IT Fails intermittently 


prb112:
WebSocket Listener for Tyrus Threading issues

```
[ERROR] Tests run: 311, Failures: 1, Errors: 0, Skipped: 2, Time elapsed: 334.835 s <<< FAILURE! - in TestSuite
[ERROR] testCreatePatient(com.ibm.fhir.server.test.WebSocketNotificationsTest)  Time elapsed: 5.244 s  <<< FAILURE!
java.lang.AssertionError: 
	at com.ibm.fhir.server.test.WebSocketNotificationsTest.testCreatePatient(WebSocketNotificationsTest.java:68)
```

PR #216, opened 2019-10-08
WebSocket IT Fails intermittently #210

prb112:


fhir-server-test/src/test/java/com/ibm/fhir/server/test/FHIRServerTestBase.java
```diff
@@ -272,6 +272,15 @@ public void beforeRequest(Map<String, List<String>> headers) {
...
             Session session = container.connectToServer(endpoint, config, new URI(webSocketURL));
+            
+            // Add a Delay
+            int count = 10;
+            while ( !session.isOpen() && count > 0) {
+                System.out.println(">>> " + count + " waiting");
+                Thread.currentThread().sleep(1000l);
+                count--;
+            }
+
```
albertwang-ibm: seems the test client always got and printed out the message correctly, so I would assume the websocket connection worked well(already opened and connected) in the failed test,  so not really sure if adding the check here can help or not...

But we definitely should try this fix!

Issue #184, opened 2019-10-01
CI: Tests report failure when messages are sent back. 


prb112:
CI tests report failure when messages in WSS are sent back. 

Timeout Failure waiting for Messages to be sent back
```
[ERROR]   WebSocketNotificationsTest.testCreatePatient:59
```

websocket failure:  https://github.com/IBM/FHIR/pull/149/checks?check_run_id=243809676

prb112:
PR https://github.com/IBM/FHIR/pull/185

PR #231, opened 2019-10-09
issue #184 #210 websocket test timeout issue

albertwang-ibm:


albertwang-ibm:
Per discussion with Paul, we found that websocket tests sometimes received old events from previous testng test right before the websocket testng test, so we think we some delay to the tests should be able to avoid this and could possible make the test pass.

Issue #226, opened 2019-10-09
Refactor JDBCParameterBuilder to use visitor pattern


lmsurpre:
It should be possible to avoid reflection here in order to extract parameter values from the underlying FHIRPathNodes being indexed.

We should continue to support both FHIRPathElementNode and FHIRPathPrimitiveValue and reject anything else.  Unsupported element types should return an empty list of Parameters just like we do today.

PR #237, opened 2019-10-10
issue #226 - introduce JDBCParameterBuildingVisitor and refactor code to use it

lmsurpre:
1. Introduced javadoc on Visitable interface
2. Updated CodeBuilder and
CodeGenerator to add Generated annotation and
javadoc to Visitor and AbstractVisitor methods
3. Updated Visitor interface to include primitive subtypes and exclude
BackboneElement subtypes
4. Replaced AbstractVisitor with DefaultVisitor
    * no longer abstract
    * takes a "visitChildren" boolean to allow subclasses to set the
most
reasonable default
    * introduces defaultAction methods, making it simple to define
global
default actions across all visit methods
    * subtype visits now delegate to their supertype, making it possible
to modify types and their subtypes in a single method, while still
allowing subclasses to override the behavior for specific subtypes
5. Refactored all uses of AbstractVisitor over to the new / more
powerful DefaultVisitor
6. Introduced JDBCParameterBuildingVisitor which extends DefaultVisitor and
implements parameter extraction without the use of reflection
7. Refactored FHIRPersistenceJDBCNormalizedImpl.extractSearchParameters
to use the new ParameterBuilder, while adding separate support for search
parameter expressions that resolve to FHIRPathPrimitiveValue nodes (like
Patient-deceased)


fhir-tools/src/main/java/com/ibm/fhir/tools/CodeGenerator.java
```diff
@@ -306,17 +306,60 @@ private void generateAbstractVisitorClass(String basePath) {
...
+        cb.javadoc("value of the passed {@code visitChildren} boolean to control whether or not to");
+        cb.javadoc("visit the children of the Resource or Element being visited.");
+        cb.javadoc("");
+        cb.javadoc("Subclasses can override the default behavior in a number of places, including:");
+        cb.javadoc("<ul>",false);
+        cb.javadoc("<li>preVisit methods to control whether a given Resource or Element gets visited",false);
+        cb.javadoc("<li>visitStart methods to provide setup behavior prior to the visit",false);
+        cb.javadoc("<li>defaultAction methods to perform some common action on all visited Resources and Elements",false);
+        cb.javadoc("<li>specific visit methods to perform unique behavior that varies by the type being visited",false);
+        cb.javadoc("<li>visitEnd methods to provide initial cleanup behavior after a Resource or Element has been visited",false);
```
prb112: should the li tags be closed?

fhir-tools/src/main/java/com/ibm/fhir/tools/CodeGenerator.java
```diff
@@ -3285,6 +3381,14 @@ private boolean isBackboneElement(JsonObject elementDefinition) {
         List<JsonObject> types = getTypes(elementDefinition);
         return !types.isEmpty() && "BackboneElement".equals(types.get(0).getString("code", null));
     }
+    
+    private boolean isBackboneSubtype(String className) {
```
prb112: If I follow this correctly, it'll tell you if it's 

com.ibm.fhir.model.type.test.test
vs
com.ibm.fhir.model.type.test2

com.ibm.fhir.model.other.test3

test2 would return false? 
test1 would return true
test 3 would return true?

fhir-model/src/main/java/com/ibm/fhir/model/visitor/DeepCopyingVisitor.java
```diff
@@ -1,175 +0,0 @@
...
-
-/**
- * Copy the value of each element within a Resource/Element to a new element with the same values.
- *  
- * Note: this class is NOT threadsafe.  Only one object should be visited at a time.
- * 
- * @author lmsurpre
- * @param <T> The type to copy. Only visitables of this type should be visited.
- */
-public class DeepCopyingVisitor<T extends Visitable> extends AbstractVisitor {
```
prb112: asked confirming question internally where did this class go?

prb112: it's CopyingVisitor

fhir-tools/src/main/java/com/ibm/fhir/tools/CodeGenerator.java
lmsurpre: its proper html, but not proper xhtml.  it is consistent with our use of unclosed <p> tag in the model javadoc.  i don't feel too strongly, but when john and I discussed that a while ago we sort of agreed that the optional closing tags don't add a whole lot.

fhir-tools/src/main/java/com/ibm/fhir/tools/CodeGenerator.java
lmsurpre: specifically what we're trying to avoid here is having visit methods for every single *nested* BackboneElement subtype in the model.

the only reason for stripping the `com.ibm.fhir.model.type` is because we want to avoid fully qualified types like `com.ibm.fhir.model.type.String` from coming back true.

com.ibm.fhir.model.type.Test.Test1 would return true (and thus get skipped)
com.ibm.fhir.model.type.Test2 would return false
com.ibm.fhir.model.resource.Test3 would return true, but I'm not sure it matters much since this is a private method thats only called with fullyqualified names in the com.ibm.fhir.model.type package

lmsurpre: i'll just add some javadoc to explain to expected usage

prb112: Cool.  I was attempting to follow :)

fhir-tools/src/main/java/com/ibm/fhir/tools/CodeGenerator.java
prb112: just an observation :)

prb112: my mind goes to balance, and I wanted to double check

fhir-tools/src/main/java/com/ibm/fhir/tools/CodeGenerator.java
lmsurpre: updated

JohnTimm:
We may want to consider moving the @Required annotation to the bottom of the list. Least important to most important (top to bottom):

@Summary
@Binding
@Choice
@Required
Element someElement;

Thoughts?

Thanks,

JT

lmsurpre:
> We may want to consider moving the @Required annotation to the bottom of the list.
done

fhir-model/src/main/java/com/ibm/fhir/model/resource/DomainResource.java
```diff
@@ -14,6 +14,7 @@
 import javax.annotation.Generated;
 
 import com.ibm.fhir.model.annotation.Constraint;
+import com.ibm.fhir.model.annotation.Summary;
```
prb112: not used, but understand how/why it's imported

fhir-model/src/main/java/com/ibm/fhir/model/type/Age.java
```diff
@@ -13,6 +13,7 @@
 
 import com.ibm.fhir.model.annotation.Binding;
 import com.ibm.fhir.model.annotation.Constraint;
+import com.ibm.fhir.model.annotation.Summary;
```
prb112: Unused import

fhir-model/src/main/java/com/ibm/fhir/model/type/SimpleQuantity.java
```diff
@@ -12,6 +12,7 @@
 import javax.annotation.Generated;
 
 import com.ibm.fhir.model.annotation.Constraint;
+import com.ibm.fhir.model.annotation.Summary;
```
prb112: unused

fhir-model/src/main/java/com/ibm/fhir/model/type/Distance.java
```diff
@@ -13,6 +13,7 @@
 
 import com.ibm.fhir.model.annotation.Binding;
 import com.ibm.fhir.model.annotation.Constraint;
+import com.ibm.fhir.model.annotation.Summary;
```
prb112: single use import

fhir-model/src/main/java/com/ibm/fhir/model/type/Count.java
```diff
@@ -12,6 +12,7 @@
 import javax.annotation.Generated;
 
 import com.ibm.fhir.model.annotation.Constraint;
+import com.ibm.fhir.model.annotation.Summary;
```
prb112: all alone...

Issue #92, opened 2019-09-18
Cleanup fhir-persistence-jdbc tests


lmsurpre:
We should either revive the tests under `com.ibm.watson.health.fhir.persistence.jdbc.search.test` or create something comparable.
The idea is to systematically define search parameter definitions for each combination of SearchParameter type and target element type.

Perhaps some of the other tests are worth reviving as well.  If not, then they should be deleted.

Acceptance criteria 1:  Unit tests (under either fhir-search or fhir-persistence or both) cover search parameters of each search parameter type and each fhir data type.
Acceptance criteria 2:  Unit tests that are not executed are deleted (or have a clear comment about why they are still needed but aren't included in the build)

lmsurpre:
related to #93

PR #247, opened 2019-10-13
issue #92 - re-enable JDBC search tests

lmsurpre:
1. correct issue where all JDBC search test tenants had the same extensions file (the date one); while some of the content is similar, each tenant has its own unique configuration (where all defined search parameters are of the type that matches the tenant name)
2. introduce extension-search-parameters-valuetypes.json files for the remaining JDBC search test tenants (one per search param type)
3. add null checking and include root cause in logging messages while loading these value types
4. move the remaining JDBC Search test json files to fhir-examples
5. update DerbyInitializer to use the new schema behavior from Robin
6. fix bug in FHIRPersistenceJDBCNormalizedImpl for detecting integer searches
7. enable the JDBC Search tests in the default JDBC TestNG suite


fhir-examples/src/main/resources/json/ibm/basic/BasicReference.json
```diff
@@ -49,25 +49,25 @@
       }
     },
     {
-      "url": "http://example.org/ReferenceId",
+      "url": "http://example.org/Reference-id",
```
prb112: is this a pattern I should follow if I run into extensions?

fhir-examples/src/main/resources/json/ibm/basic/BasicToken.json
```diff
@@ -57,15 +54,13 @@
       }
     },
     {
-      "__comment": "If the system is present, and there is no code, then this is understood to mean that there is no suitable code in the system in which to represent the code.",
```
prb112: are these removed as no longer functional or the details are in the test?

fhir-persistence-jdbc/src/test/java/com/ibm/fhir/persistence/jdbc/test/util/DerbyInitializer.java
```diff
@@ -116,14 +113,7 @@ public void bootstrapDb(boolean reset) throws FHIRPersistenceDBConnectException,
         }
         else {
             System.out.println("Bootstrapping database");
-            final String url = DERBY_TRANSLATOR.getUrl(dbProps);
-            try (Connection connection = DriverManager.getConnection(url + ";create=true")) {
```
prb112: removing this makes create=false.  I think in this case don't you want to set to true? I can't think of the exact reason we hit this before, but just worth checking that create=true is specified in DerbyFhirDatabase.

fhir-examples/src/main/resources/json/ibm/basic/BasicReference.json
lmsurpre: doesn't really matter; just needs to match whats in the corresponding SearchParameter in extension-search-parameters.json

fhir-persistence-jdbc/src/test/java/com/ibm/fhir/persistence/jdbc/test/util/DerbyInitializer.java
lmsurpre: `new DerbyFhirDatabase()` is how Robin was instantiating the database R4JDBCExamplesTest, so I was just trying to be consistent with that.

fhir-examples/src/main/resources/json/ibm/basic/BasicToken.json
lmsurpre: in past (DSTU2 timeframe) our parser was lenient by default and ignored unexpected elements.  now its strict by default and so these cause issues.
john tells me there is still "special" support for a field called `fhir_comments` but indicated that he might like to remove that in the future and so I stayed away.

PR #282, opened 2019-10-17
issue #92 - refactor SearchUtil.parseQueryParams and enable tests

lmsurpre:
1. remove queryString from parseQueryParameters methods

Previously, all variants of SearchUtil.parseQueryParameters wanted the
queryString in
addition to the Map of queryParameters. This doesn't make much sense
because the queryParameters map should have everything we need.

Now we can just pass the queryParameters. Changes needed to enable this
include:
* removal of queryStringIndex from SortParameter (just pass them in the
right order in the _sort param values list)
* moved "_sort + _include/_revinclude" check into parseQueryParameters
method where we have all the parameters

2. Enable the new Abstract PL tests.

Removed the commented out classes in testng.xml and added new section
for the
following tests:
* com.ibm.fhir.persistence.jdbc.test.FHIRDBDAOTest
* com.ibm.fhir.persistence.jdbc.test.JDBCNormDeleteTest
* com.ibm.fhir.persistence.jdbc.test.JDBCNormCompartmentTest
* com.ibm.fhir.persistence.jdbc.test.JDBCNormMultiResourceTest
* com.ibm.fhir.persistence.jdbc.test.JDBCNormSortTest
* com.ibm.fhir.persistence.jdbc.test.JDBCNormWholeSystemSearchTest

3. Get everything passing

### DIFF
--- a/fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
+++ b/fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
@@ -80,12 +80,10 @@ public class SearchUtil {
     private static final String UNSUPPORTED_EXCEPTION = "Search Parameter includes an unsupported operation or bad expression : [%s] [%s] [%s]";
 
     // Exception Strings
-    private static final String INVALID_SORT_SEARCH_EXCEPTION = "Sort parameters cannot be processed with null queryString.";
     private static final String UNSUPPORTED_SEARCH_PARAMETERS_EXCEPTION = " %s is not supported as a sort parameter on a system-level search.";
     private static final String MODIFIER_NOT_ALLOWED_WITH_CHAINED_EXCEPTION = "Modifier: '%s' not allowed on chained parameter";
     private static final String TYPE_NOT_ALLOWED_WITH_CHAINED_PARAMETER_EXCEPTION = "Type: '%s' not allowed on chained parameter";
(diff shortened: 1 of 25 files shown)


fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -650,6 +655,7 @@ public static FHIRSearchContext parseQueryParameters(Class<?> resourceType, Map<
                     parameterValue.setPrefix(prefix);
                 }
 
+                // DONT USE DateTime here
```
prb112: yeah - my current WIP should address this.

lmsurpre: cool.  forgot i had added that comment there (probably when we were on webmeeting discussing it).  will leave in for now and you can remove it in yours

lmsurpre:
> Was the unparsed query string used for debug / logging at some point?

Previously, it was used in parseSortParameter (and a tiny bit in the parseInclusionParameter).  But I'm pretty sure there was flawed thinking there and it seems safe to remove it.

prb112:
> > Was the unparsed query string used for debug / logging at some point?
> 
> Previously, it was used in parseSortParameter (and a tiny bit in the parseInclusionParameter). But I'm pretty sure there was flawed thinking there and it seems safe to remove it.

We could always get it from the FHIRSearchContext which packs in the HttpServletRequest object. If logging is needed in the future.

Issue #93, opened 2019-09-18
Remove all references to the "basic" schema


lmsurpre:
We used to have multiple relational schemas:
1. a "basic" schema that puts all resource types into the same tables
2. a "normalized" schema that uses a set of tables per resource type

At the implementation level, the normalized schema extends the basic schema, which makes things more complicated than they need to be.  We never updated the "basic" schema for R4 and so this is needless complexity.

Acceptance criteria 1:  Class hierarchy in fhir-persistence-jdbc should be simplified and without reference to the basic schema
Acceptance criteria 2:  Configuration and documentation related to the "basic" schema should be removed

PR #286, opened 2019-10-17
Issue #93 - delete JDBC "basic" schema impl classes

lmsurpre:
1. Remove schemaType config parameter and simplify all places using it (docs still need updating)

2. Remove all classes dedicated to the "Basic" schema implementation

3. Simplify the class hierarchy by merging the "Normalized" impl behavior into the level above it (since now there is only one impl, the extra layer is not needed and we don't need to use "normalized" to distinguish this impl from any other)

For example:
* JDBCNormalizedQueryBuilder function was moved to JDBCQueryBuilder (replacing "basic" schema logic as needed)
* ParameterNormalizedDAO merged into ParameterDAO; ParameterBasicDAO
is now deleted.
* ParameterDAONormalizedImpl merged into ParameterDAOImpl;
ParameterDAOBasicImpl is now deleted.
* ResourceNormalizedDAO merged into ResourceDAO; ResourceBasicDAO is
now deleted.
* ResourceDAONormalizedImpl merged into ResourceDAOImpl;
ResourceDAOBasicImpl is now deleted.
* FHIRDbDAONormalizedImpl merged into FHIRDbDAO; FHIRDbDAOBasicImpl
is now deleted.


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/api/ParameterDAO.java
```diff
@@ -1,24 +1,27 @@
...
+import java.util.Map;
 
+import com.ibm.fhir.persistence.exception.FHIRPersistenceException;
 import com.ibm.fhir.persistence.jdbc.dto.Parameter;
 import com.ibm.fhir.persistence.jdbc.exception.FHIRPersistenceDBConnectException;
 import com.ibm.fhir.persistence.jdbc.exception.FHIRPersistenceDataAccessException;
 
 /**
- * This Data Access Object interface defines methods for creating, updating, and retrieving rows in the FHIR Parameter table.
- * @author markd
```
albertwang-ibm: should we remove the author?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/api/ParameterDAO.java
```diff
@@ -1,24 +1,27 @@
...
- *
+ * This Data Access Object interface defines methods for creating, updating, 
+ * and retrieving rows in the IBM FHIR Server parameter-related tables.
  */
 public interface ParameterDAO extends FHIRDbDAO {
-
+
```
albertwang-ibm: spaces

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/api/ResourceDAO.java
```diff
@@ -8,28 +8,23 @@
...
+import com.ibm.fhir.persistence.jdbc.dto.Parameter;
 import com.ibm.fhir.persistence.jdbc.dto.Resource;
 import com.ibm.fhir.persistence.jdbc.exception.FHIRPersistenceDBConnectException;
 import com.ibm.fhir.persistence.jdbc.exception.FHIRPersistenceDataAccessException;
+import com.ibm.fhir.persistence.jdbc.util.SqlQueryData;
+
 
 /**
- * This Data Access Object interface provides methods creating, updating, and retrieving rows in the FHIR Resource table.
- * @author markd
```
albertwang-ibm: seems we need a rule about how to change the author info, either remove the old author or add the new author to the list or simply don't keep any author info to the codes ... when I modify existing files, I don't add my name to the author list, not sure if this is a bad practice...

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/api/ParameterDAO.java
lmsurpre: Eclipse does this.  We're wildly inconsistent as to whether empty lines have spaces that indent it to the length of the previous/next line or not.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/api/ParameterDAO.java
lmsurpre: IMHO we shouldn't be using @author tags everywhere.  I've touched a lot of classes and I don't add my name everywhere.  I tried to remove it from places that have changed since the original authoring.  But we can discuss as a group.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/api/ResourceDAO.java
lmsurpre: same

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/FHIRDbDAOImpl.java
```diff
@@ -43,10 +44,10 @@
  * @author markd
  *
  */
-public class FHIRDbDAOBasicImpl<T> implements FHIRDbDAO {
+public class FHIRDbDAOImpl implements FHIRDbDAO {
```
albertwang-ibm: I like the change, it's not always good to use template.

lmsurpre: yeah, from what i can tell this particular type argument was never even actually used.  it only caused every single reference to it to need a @suppressWarnings annotation :-)

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/api/ParameterDAO.java
lmsurpre: we just discussed this as a team and agreed that the `@author` tags add limited value beyond what git history already provides. furthermore, they get stale over time and become maintenance burden.  we will remove them incrmentally over time (as we edit classes) and this PR is consistent with that approach

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/api/ParameterDAO.java
lmsurpre: let me know if you find a way to have eclipse not add the spaces on newlines...

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/api/ResourceDAO.java
lmsurpre: we just discussed this as a team and agreed that the `@author` tags add limited value beyond what git history already provides. furthermore, they get stale over time and become maintenance burden.  we will remove them incrmentally over time (as we edit classes) and this PR is consistent with that approach

Issue #294, opened 2019-10-18
Support _summary for resource read


albertwang-ibm:
in addition to search, we need to support _summary for reading of a single resource.

PR #299, opened 2019-10-21
issue #294  support _summary for resource read

albertwang-ibm:
We already support _summary for search.
This enhancement is for _summary (true, text, data, false) of single resource read with resource type and resource id:
   GET [base]/[type]/[id] {?_summary=text}


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -627,13 +627,41 @@ private ResourceDAO getResourceDao() {
...
+                summaryElements = JsonSupport.getSummaryDataElementNames(resourceType);
+                break;
+            default:
+                break;
+                
+            }
+
+            if (summaryElements != null) {
+                elements = new ArrayList<String>();
+                elements.addAll(summaryElements);
```
prb112: maybe directly add elements.addAll during the case in the switchStatement?

fhir-server-test/src/test/java/com/ibm/fhir/server/test/ServerSpecTest.java
```diff
@@ -686,4 +687,94 @@ public void testConditionalUpdateObservation2() throws Exception {
...
+    public void testReadPatientSummary() {
+        WebTarget target = getWebTarget();
+        Response response = target.path("Patient/" + savedPatient.getId().getValue())
+                .queryParam("_summary", "true")
+                .request(FHIRMediaType.APPLICATION_FHIR_JSON).get();
+        assertResponse(response, Response.Status.OK.getStatusCode());
+        Patient responsePatient = response.readEntity(Patient.class);
+        Coding subsettedTag =
+                Coding.builder().system(uri("http://terminology.hl7.org/CodeSystem/v3-ObservationValue")).code(Code.of("SUBSETTED")).display(string("subsetted")).build();
+        assertTrue(FHIRUtil.hasTag(responsePatient, subsettedTag));
```
prb112: I know we are deprecating FHIRUtil is there an alternative in ModelSupport?

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -610,8 +610,8 @@ public Response read(@PathParam("type") String type, @PathParam("id") String id)
 
         try {
             checkInitComplete();
-
-            Resource resource = doRead(type, id, true, false, null, null);
+            MultivaluedMap<String, String> queryParameters = uriInfo.getQueryParameters();
```
prb112: This should be available in the FHIRPersistenceContext.get().getFHIRPersistenceEvent(PROPNAME_REQUEST_PROPERTIES)

I think this is probably the preferred way, rather than add to the signature method.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
albertwang-ibm: I want to keep element as null instead of an empty ArrayList if there is no _summary.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/ServerSpecTest.java
albertwang-ibm: no, we don't have such util Functions in ModelSupport yet, seems all tag helper functions are in FhirUtil only at present.

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
albertwang-ibm: not really sure what's your meaning... 
According to the function comments of doSearch, seems requestProperties is used for the http headers only, not sure how it can be used for the query parameters... 

 * @param queryParameters
     *            a Map containing the query parameters from the request URL
     * @param requestProperties
     *            additional request properties which supplement the HTTP headers associated with this request

prb112: I'm suggesting that there is an existing parameter/property. Or should be extended to add the property to the context, rather than put in the method.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
prb112: right, that's the default case.

albertwang-ibm: do you mean to add these following in each case:
                elements = new ArrayList<String>();
                elements.addAll(JsonSupport.getSummaryxxxxNames(resourceType));
?

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
prb112: Discussed with Albert, no problem with code, just a pattern conversation, resolved to leave as implemented in PR

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
prb112: elements = new ArrayList(); this would be before the case. 
then you would call elements.AddAll once-line per each

albertwang-ibm: I want a null elements when  _summary=false etc instead of an empty Arraylist, and null is used in filtering code logic ...   this was why I didn't  do "elements = new ArrayList()" right before the switch ...

prb112: ah! OK. Then I approve.

Issue #134, opened 2019-09-23
Remove or refactor fhir-replication-api


lmsurpre:
fhir-replication-api defines an API for "replication events" that can be used to broadcast create/update/delete interactions to a messaging/event engine (e.g. Kafka).
To assist with reliable delivery of messages, the server writes a fhir-replication record to the database within the managed transaction.
However, when there is nothing processing these records (i.e. no separate component deleting rows in this table like in a simple deploy), it will grow unbounded.
We should either remove this project or refactor it so that it can be disabled.
Additionally, some of the metadata fields on these tables are not appropriate for a general-purpose fhir server implementation.

prb112:
PR https://github.com/IBM/FHIR/pull/301

Issue #270, opened 2019-10-15
Document the schema migration process on the project wiki


lmsurpre:
We used to use liquibase.  Now we use custom database-util logic.  I need to learn this part of our code and I'll create some documentation as I do it.

lmsurpre:
I also left a TODO in the FHIRServerUserGuide for this content as well.

prb112:
Dockerfile used to test this is merged in this PR https://github.com/IBM/FHIR/pull/413/files/fc3635fd3e250dada4ff3979f517e3944bf9ff71#diff-b0ee41879dbc3cde079556c7613ea929

prb112:
PR https://github.com/IBM/FHIR/pull/452  cleans up documentation related to the docker, which is used to test this issue/migration

prb112:
PR https://github.com/IBM/FHIR/pull/782

PR #301, opened 2019-10-22
Issue 134 - Remove fhir-replication-api

prb112:
issue #134


fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/jdbc/ConnectionStub.java
```diff
@@ -44,14 +44,12 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
 
     @Override
     public Statement createStatement() throws SQLException {
-        // Stub Only
-        return null;
+        return new StatementStub(this);
```
albertwang-ibm: it's weird that my local looks correct - the same as you change. 
and also it's a big surprise to find that the line numbers are totally different from my local...

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/jdbc/ConnectionStub.java
```diff
@@ -44,14 +44,12 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
...
     }
 
     @Override
     public PreparedStatement prepareStatement(String sql) throws SQLException {
-        // Stub Only
-        return null;
+        return new PreparedStatementStub(this);
     }
```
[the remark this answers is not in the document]
prb112: right it should very much be the same as before ... I removed auto-generated comments

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/jdbc/ConnectionStub.java
prb112: same -  auto generated comments, I added the statement stub back in between a prior commit and this one.  There were 100s of lines of auto-generated comments

fhir-persistence-schema/src/main/resources/add_resource_template.sql
```diff
@@ -14,44 +14,26 @@
...
-      IN p_changed_by                    VARCHAR(64),
-      IN p_correlation_token             VARCHAR(36),
-      IN p_tenant_data_id                VARCHAR(36),
-      IN p_reason                        VARCHAR(255),
-      IN p_event                         VARCHAR(32),
-      IN p_site_id                       VARCHAR(255),
-      IN p_study_id                      VARCHAR(255),
-      IN p_service_id                    VARCHAR(32),
-      IN p_patient_id                    VARCHAR(255),
-      IN p_version                           INT,
```
albertwang-ibm: Are those dropped fields used for replication only?

prb112: yes, that's right

fhir-model/src/test/java/com/ibm/fhir/model/test/TestUtil.java
```diff
@@ -142,7 +142,7 @@ public static InputStream resolveFileLocation(String fileName) throws Exception
         
         // If we didn't find the file yet, then look on the classpath.
         String resourceName = (fileName.startsWith("json/spec/") ? fileName : "json/spec/" + fileName);
-        InputStream is = Thread.currentThread().getContextClassLoader().getResourceAsStream(resourceName);
+        InputStream is = TestUtil.class.getResourceAsStream(resourceName);
```
prb112: @senthilbak

fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractWholeSystemSearchTest.java
```diff
@@ -70,16 +72,16 @@ public void testSearchAllUsingId() throws Exception {
         assertTrue(resources.size() > 0);
     }
     
-    @Test
+    @Test(singleThreaded = true)
```
lmsurpre: does it need to be singleThreaded?  maybe leave comment to explain why?

prb112: No, I thought I had removed that after my tests.  thanks for identifying, intend to remove.

.gitignore
```diff
@@ -26,3 +26,6 @@ test-output/
 /server-logs/
 /integration-test-results/
 /integration-test-results.zip
+
+# ignores removed projects
+/fhir-replication-api/**/*
```
lmsurpre: do we need this?

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/db2/Db2CheckCompatibility.java
```diff
@@ -18,15 +18,14 @@
 import com.ibm.fhir.database.utils.version.SchemaConstants;
 
 /**
- * @author rarnold
  *
  */
```
lmsurpre: The empty javadoc will look a little funny.  Should we just delete it?  Or maybe attempt a very simple summary just to avoid the empty comment?

.gitignore
prb112: Yes - it drives me nuts when we don't - general practice is to keep open Eclipse projects from leaking back in

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/db2/Db2CheckCompatibility.java
prb112: There are many examples of this.. completely up to how we approach - there are probably dozens

.gitignore
lmsurpre: ok, i don't think i agree.  i personally don't like seeing remnants of project history...prefer a clean gitignore.  i don't think we'd "leak project files back in given that all eclipse project files are ignored already.

prb112: If you want, I'll remove it.  best practice in my opinion is to add the gitignore

lmsurpre:
This looks good to me, but we'll need to consider how to deploy this on our cloud instance.

Since we havn't had an R4 release yet, this one assumes a fresh database.

To perform the migration, I think we'd need to delete the two fhir_replication_* tables and update the stored procedure to reduce the number of input variables.

In the longer term, we have https://github.com/IBM/FHIR/issues/270 for investigating and documenting our schema change procedure.

Issue #305, opened 2019-10-24
Failing dateTime searches with Java 11


lmsurpre:
**Describe the bug**
I did some refactoring to our AbstractWholeSystemSearchTest and, after that, it repeatedly fails when run with Java 11 (but works with Java 8).
After some investigation, I think the issue is in JDBCQueryBuilder.processDateParm.  We have a TODO here like "TODO: clean up handling of date parameters in the query predicates" and it seems to be converting the passed DateTime into an Instant and then converting that Instant into a Date.
My current theory is that we're losing some precision during this process and that this is somehow causing the query to fail in Java 11.

**To Reproduce**
Perform a search on a search parameter of type date and pass a full Instant.  For example:
```
    List<Resource> resources = runQueryTest(Resource.class, "_lastUpdated", savedResource.getMeta().getLastUpdated().getValue().toString());
    assertNotNull(resources);
    assertEquals(resources.size(), 1, "Number of resources returned");
```

**Expected behavior**
Since its using the lastUpdated time from the savedResource, it should return the saved resource when you search on that.

**Additional context**
@prb112 noticed that our `fhir-persistence-jdbc` AbstractWholeSystemSearchTest was failing when he used Java 11.  He included some related changes in https://github.com/IBM/FHIR/pull/301#discussion_r337768214 which we thought might address the issue.  However, I was able to reproduce the issue (without his fix) and I think there is a deeper issue with respect to our dateTime handling in JDBCQueryBuilder.

PR #307, opened 2019-10-25
issue #305 - handle more date param values as ranges

lmsurpre:
Now all date param values will be handled as an implicit range unless
there are microseconds given. For example:
* `01:01:01` -> `[01:01:01, 01:01:02)`
* `01:01:01.1` -> `01:01:01.100000`
* `01:01:01.111111` -> `01:01:01.111111`

Once we support passing times without seconds, this should also adjust the implicit range by adding a whole minute:
* `01:01` -> `[01:01, 01:02)`

I considered guessing at the passed precision by looking for the last
non-zero digit, but our DateTime actually doesn't store this type of precision detail.


lmsurpre:
actually, i just realized that if we're gonna use ranges like this at the second / millisecond level, we probably need to do something similar at the minute level as well (since seconds are optional).  and maybe hours too, although minutes are required when hours are passed and so we'd be back to guessing at the precision.

PR #279, opened 2019-10-16
issue #92 - Cleanup fhir-persistence test suite

lmsurpre:
1. Remove outdated persistence layer tests
2. Introduce dedicated test for Compartment search
3. Rename AllResources test to WholeSystemSearch test
4. Move the readResource helper to FHIRModelTestBase; now we have
    * readLocalResource for reading resources from src/test folders of
the project under test; and
    * readExampleResource for reading resources from fhir-examples
5. Move FHIRModelTestBase back to fhir-model
6. Update all persistence layer tests to use fhir-examples examples and
delete the examples under fhir-persistence

Signed-off-by: lmsurpre <lmsurpre@us.ibm.com>

Issue #273, opened 2019-10-16
URI should default to complete match (not starts-with)


lmsurpre:
**Describe the bug**
https://localhost:9443/fhir-server/api/v4/Patient?favorite-uri=http://www.ibm.com results in SQL like:
```sql
... P5.LOGICAL_RESOURCE_ID = R.LOGICAL_RESOURCE_ID AND 
(P5.PARAMETER_NAME_ID=3000 AND (P5.STR_VALUE LIKE ? ESCAPE '+'))

bindVariables=[http://www.ibm.com%]
```

**To Reproduce**
perform a search on a search parameter of type uri (e.g. `favorite-uri` in the tenant1 test config) and check the results

**Expected behavior**
From https://www.hl7.org/fhir/search.html#uri
> By default, matches are precise (e.g. case, accent, and escape) sensitive, and the entire URI must match.

lmsurpre:
Start by adding a test for this to AbstractSearchURITest

lmsurpre:
Also, don't forget to update Conformance.md if needed

prb112:
PR https://github.com/IBM/FHIR/pull/447

PR #315, opened 2019-10-26
Issue 314 - improve PL test coverage and fix _revinclude issue

lmsurpre:
1. introduce PagingTest (Abstract and JDBC) to verify PL paging behavior
    for both search and history interactions

2. re-introduce IncludeRevincludeTest (Abstract and JDBC) to verify
_include/_revinclude behavior. we previously had
AbstractQueryIncludeTest but we had commented it out in our transition
to R4 and then I deleted it as part of pull request #279. this PR
revives this old test and updates it to align with our new testing
patterns.

3. added a single _revinclude test for each search parameter type to
test the combination of search parameters and the _revinclude parameter

4. minor cleanup in FHIRResource to highlight potential issue if we have
more resources than we can fit in Integer.MAX_VALUE (>2 billion)

5. minor doc updates to Conformance.md

6. formatted SQL in javadoc in both InclusionQuerySegmentAggregator
(added tags) and JDBCQueryBuilder (complete overhaul). we need to be
more careful when applying auto-formatting rules so that we don't break
this formatting again.

7. update InclusionQuerySegmentAggregator impl (and doc) to use IN
instead of `=` when building the WHERE clause for _revinclude searches.

Previous SQL:
```
WHERE P1.STR_VALUE='Patient/' ||
    (SELECT LR.LOGICAL_ID FROM ...
```

New SQL:
```
WHERE P1.STR_VALUE IN
    (SELECT 'Patient/' || LR.LOGICAL_ID FROM ...
```


docs/Conformance.md
```diff
@@ -23,12 +23,12 @@ Search parameters defined in the specification can be found by browsing the R4 F
 In addition, the following search parameters are supported on all resources:
 * `_id`
 * `_lastUpdated`
+* `_tag`
 * `_profile`
 * `_security`
 * `_source`
-* `_tag`
```
albertwang-ibm: not sure why move up _tag, the old list is and ordered list.

docs/Conformance.md
```diff
@@ -54,7 +56,7 @@ FHIR search modifiers are described at https://www.hl7.org/fhir/R4/search.html#m
 |--------------------------|-------------------|-----------------------------------------------------|
 |String                 |`:exact`,`:contains`,`:missing` |Performs a "starts with" search that is case-insensitive and accent-insensitive|
 |Reference              |`:[type]`,`:missing`            |Performs an exact match search|
-|URI                    |`:below`,`:missing`             |Performs a "starts with" search|
+|URI                    |`:below`,`:missing`             |Performs a "starts with" search (issue #273)|
```
albertwang-ibm: do we really need the issue number here?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/InclusionQuerySegmentAggregator.java
```diff
@@ -233,10 +237,10 @@ private void processRevIncludeParameters(StringBuilder queryString, List<Object>
...
             // P1.PARAMETER_NAME_ID=xx AND 
             queryString.append("P1.PARAMETER_NAME_ID=").append(this.getParameterNameId(includeParm.getSearchParameter())).append(" AND ");
-            // P1.STR_VALUE='Patient/'  ||
-            queryString.append("P1.STR_VALUE='").append(includeParm.getSearchParameterTargetType()).append("/' || ");
-            // (SELECT LR.LOGICAL_ID
-            queryString.append("(SELECT LR.LOGICAL_ID ");
+            // P1.STR_VALUE IN 
+            queryString.append("P1.STR_VALUE IN ");
+            // (SELECT 'Patient/' || LR.LOGICAL_ID
+            queryString.append("(SELECT '").append(includeParm.getSearchParameterTargetType()).append("/' || LR.LOGICAL_ID ");
```
albertwang-ibm: does this mean _revinclude never work well before?

fhir-persistence-jdbc/src/test/java/com/ibm/fhir/persistence/jdbc/test/JDBCIncludeRevincludeTest.java
```diff
@@ -0,0 +1,40 @@
...
+            derbyInit = new DerbyInitializer(this.testProps);
+            derbyInit.bootstrapDb(false);
+        }
+    }
+    
+    @Override
+    public FHIRPersistence getPersistenceImpl() throws Exception {
+        return new FHIRPersistenceJDBCImpl(this.testProps);
+    }
+}
```
albertwang-ibm: honestly, I don't fually understand why we need AbstractIncludeRevincludeTest etc and put them in the persistence project instead of the jdbc project, the same for other persistence tests, Does this mean in addition to jdbc, we can also support others?

fhir-persistence/src/test/java/com/ibm/fhir/persistence/search/test/AbstractSearchURITest.java
```diff
@@ -49,14 +56,25 @@ public void testSearchURI_uri_chained() throws Exception {
         assertSearchReturnsComposition("subject:Basic.uri", "http://hl7.org/fhir/DSTU2");
         assertSearchReturnsComposition("subject:Basic.uri", "urn:uuid:53fefa32-1111-2222-3333-55ee120877b7");
         
+        // https://github.com/IBM/FHIR/issues/273
         // Matches are supposed to be precise (e.g. case, accent, and escape sensitive), but aren't
-        // 
+        
 //        assertSearchDoesntReturnComposition("subject:Basic.uri", "http://HL7.org/FHIR/dstu2");
 //        assertSearchDoesntReturnComposition("subject:Basic.uri", "urn:uuid:53FEFA32-1111-2222-3333-55EE120877B7");
```
albertwang-ibm: why not just remove them?

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -3720,20 +3721,25 @@ private Interaction buildInteractionStatement(TypeRestfulInteraction value) {
      */
     private Bundle createSearchBundle(List<Resource> resources, FHIRSearchContext searchContext)
         throws FHIROperationException {
+        
+        // throws if we have a count of more than 2,147,483,647 resources
+        UnsignedInt totalCount = com.ibm.fhir.model.type.UnsignedInt.of(Math.toIntExact(searchContext.getTotalCount()));
```
albertwang-ibm: I like this check, even thought I would believe the DB server query will most likely timeout if there are such huge number of matched resource ...

docs/Conformance.md
lmsurpre: I reordered it to match https://www.hl7.org/fhir/search.html

docs/Conformance.md
lmsurpre: the intention is to help us remember to update it as part of #273 or, if we forget, for folks to know its stale when they see #273 is resolved

fhir-persistence/src/test/java/com/ibm/fhir/persistence/search/test/AbstractSearchURITest.java
lmsurpre: because we want exactly these lines after we implement #273 ...this is also why i left the link to that issue just above the comment

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/InclusionQuerySegmentAggregator.java
lmsurpre: correct.  seems it was broken since the beginning...we just lacked test coverage to show it.

fhir-persistence-jdbc/src/test/java/com/ibm/fhir/persistence/jdbc/test/JDBCIncludeRevincludeTest.java
lmsurpre: yes.  the persistence layer tests were designed this way because they are done more at the functional level (black box testing)...the logic in these tests should be applicable to any persistence layer implementation, not just our JDBC implementation.  we've used a few other implementations internally (Cloudant, HBase, etc), but we've only used the JDBC one in R4 and so thats the only one we've open-sourced.

PR #332, opened 2019-10-29
issue #129 cleanup and timestamp test for core FHIRUtilities.java

albertwang-ibm:


fhir-core/src/test/java/com/ibm/fhir/core/test/TimestampTest.java
```diff
@@ -0,0 +1,28 @@
...
+import org.testng.annotations.Test;
+
+import com.ibm.fhir.core.FHIRUtilities;
+
+public class TimestampTest {
+    
+    @Test
+    public void test1() throws Exception {
+        
+        String strTimeStamp = FHIRUtilities.formatTimestamp(new Date(System.currentTimeMillis()));
```
lmsurpre: you could just use `new Date()`...it creates it from System.currentTimeMillis() by default
```suggestion
        String strTimeStamp = FHIRUtilities.formatTimestamp(new Date());
```

fhir-core/src/test/java/com/ibm/fhir/core/test/mains/DateTestMain.java
```diff
@@ -1,96 +0,0 @@
...
-
-public class DateTestMain {
-    // TODO remove when we are happy we can permanently delete the FHIRUtilities methods under test
-//    public static void main(String[] args) throws Exception {
-//        List<XMLGregorianCalendar> calendars = Arrays.asList(
-//            parseDateTime("2015-12-31T20:00:00-04:00", false),   // timezone other than GMT
-//            parseDateTime("2015-12-31T20:00:00-05:00", false),   // different timezone other than GMT
-//            parseDateTime("2016-01-01", false),                  // date
-//            parseDateTime("2016-01", false),                     // year month
-//            parseDateTime("2016", false)                         // year
```
lmsurpre: do we have somewhere else that tests the parsing of values like this into our DateTime object?

albertwang-ibm: parseDateTime is defined in the model project now, so I think we should unitest there instead of in the core project.

fhir-core/src/test/java/com/ibm/fhir/core/test/TimestampTest.java
albertwang-ibm: done! I didn't commit your proposal but made the change in my local directly and push. because otherwise, I will have to fix the DCO issue again.

fhir-core/src/test/java/com/ibm/fhir/core/test/mains/DateTestMain.java
albertwang-ibm: OK, sounds good!

albertwang-ibm: @lmsurpre just added test cases for FHIR date, time types

albertwang-ibm: done

Issue #329, opened 2019-10-29
Invalid SQL generated by ResourceDAOImpl.searchByIds when list of ids is empty


JohnTimm:
```
Oct 29, 2019 9:47:51 AM com.ibm.fhir.persistence.jdbc.dao.impl.FHIRDbDAOImpl severe
SEVERE: Failure retrieving FHIR Resources. SQL=SELECT R.RESOURCE_ID, R.LOGICAL_RESOURCE_ID, R.VERSION_ID, R.LAST_UPDATED, R.IS_DELETED, R.DATA, LR.LOGICAL_ID FROM Basic_RESOURCES R, Basic_LOGICAL_RESOURCES LR WHERE R.LOGICAL_RESOURCE_ID = LR.LOGICAL_RESOURCE_ID AND R.RESOURCE_ID IN () [probeId=c0-a8-1-59-cf866f97-f7a8-4970-ab78-3f00379fa70e]
java.sql.SQLSyntaxErrorException: Syntax error: Encountered ")" at line 1, column 238.
```

lmsurpre:
looks good

PR #333, opened 2019-10-30
Issue #329 - skip resource query if pageSize == 0

JohnTimm:

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
@@ -425,6 +425,10 @@ public class FHIRPersistenceJDBCImpl implements FHIRPersistence, FHIRPersistence
                         return resultBuilder.success(false).build();
                     }
                 }
+                
+                if (searchContext.getPageSize() == 0) {
+                    searchContext.setSummaryParameter(SummaryValueSet.COUNT);
+                }
                                 
                 // For _summary=count, we don't need to return any resource
                 if (searchResultCount > 0 && !SummaryValueSet.COUNT.equals(searchContext.getSummaryParameter())) {
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
@@ -425,13 +425,9 @@ public class FHIRPersistenceJDBCImpl implements FHIRPersistence, FHIRPersistence
                         return resultBuilder.success(false).build();
                     }
                 }
-                
-                if (searchContext.getPageSize() == 0) {
-                    searchContext.setSummaryParameter(SummaryValueSet.COUNT);
-                }
                                 
-                // For _summary=count, we don't need to return any resource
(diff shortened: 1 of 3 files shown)


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -425,6 +425,10 @@ public FHIRPersistenceJDBCImpl(Properties configProps, IConnectionProvider cp) t
                         return resultBuilder.success(false).build();
                     }
                 }
+                
+                if (searchContext.getPageSize() == 0) {
```
albertwang-ibm: Do you want to check page size or page number?
BTW, do we check the page size and number in the API layer to prevent them from being set to equal or less than 0?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -425,6 +425,10 @@ public FHIRPersistenceJDBCImpl(Properties configProps, IConnectionProvider cp) t
...
+                    searchContext.setSummaryParameter(SummaryValueSet.COUNT);
+                }
```
albertwang-ibm: This will add _summary=count to the self(search) link of the result.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
JohnTimm: Yes, see validatePagingContext. We validate pageSize and pageNumber. The logic you have highlighted is to check pageSize after validation and if pageSize == 0, then set the summary parameter to COUNT (which is what the spec states is the correct behavior).

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
JohnTimm: I think it's okay as it is implicit behavior per the spec but I'll defer to @lmsurpre for his thoughts.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
albertwang-ibm: I see, thanks for the clarification!

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
albertwang-ibm: I also think it's OK if this is aligned with the spec.

JohnTimm: I changed it to check pageSize before executing the resource query instead of setting the summary parameter.

lmsurpre: FWIW, I have no problem with this setting _summary=count in the self link.  I'm fine with whatever got merged.

PR #337, opened 2019-10-30
Issues #304 and #305 - remove JAXB and add java 11 to github CI

lmsurpre:
* remove remaining JAX-B references
* used matrixed build to test with both java 8 and java 11
* Update CXF to 3.3.3 for Java 11 support (CXF 3.1 is for Java 8.  Version 3.3 is the first version to support Java 11.)
* update pre-integration-test.sh to hide extraneous output
* updated labels for github action setup-java step
* explicitly call Base64.Encoder.encodeToString (previously I was using Base64.Encoder.encode which returns an int...)
* added action jobs for adoptopenjdk8 and adoptopenjdk 11
* Update to WebSocketNotificationTest
* added print env steps
* fix for integration-tests.sh
* added logging statement for date types in ParameterVisitorBatchDAO
* fix for potential liberty issue
* Update derby to 10.14.2.0 
* updated CodeGenerator and regenerated DateTime and Instant so that these will never have precision beyond 6 fractional second decimals (to ensure consistency with values parsed from strings)


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ParameterVisitorBatchDAO.java
```diff
@@ -238,6 +245,12 @@ public void dateValue(String parameterName, Timestamp date, Timestamp dateStart,
...
             if (isBase) {
+                // store in the base (resource) table
+                if (logger.isLoggable(Level.FINE)) {
+                    logger.fine("baseDateValue: " + parameterName + "[" + parameterNameId + "], " 
+                            + date + " [" + dateStart + ", " + dateEnd + "]");
+                }
+                
                 // Insert record into the base level date attribute table
                 resourceDates.setInt(1, parameterNameId);
                 resourceDates.setTimestamp(2, date);
```
prb112: `TIMESTAMP('2007-09-24-15.53.37.2162474',9);`
We could set this as 7 instead of the 9 you found.

prb112: a couple more examples below like this.

lmsurpre: sorry, i don't know if i understand the suggestion

lmsurpre: after further investigation, derby DDL doesn't support specifying TIMESTAMP precision.
still, I've added support for precision for Db2 (and other future impls) to the PR.

Issue #348, opened 2019-11-01
monitoring-program-name search for medicationknoledge resource doesn't work


albertwang-ibm:
create a medicationknowledge resource with:
“monitoringProgram” : [ {
   “name” : “TouchTEST”
 } ]

and then search via .../api/v4/MedicationKnowledge?monitoring-program-name=TouchTEST returns empty search result.

According touchstone error:
https://touchstone.aegis.net/touchstone/scriptexecution?exec=201910250959499075020483&qn=/FHIR4-0-0-Basic/I-O/MedicationKnowledge/Client%20Assigned%20Id/MedicationKnowledge-client-id-json#WAITING_OPER

prb112:
The search parameter with `monitoring-program-name` uses a type of `token`.  The ParameterVisitorBatchDAO expects token. 

Search Parameter Snippet: 
```
      "code" : "monitoring-program-name",
      "base" : ["MedicationKnowledge"],
      "type" : "token",
      "expression" : "MedicationKnowledge.monitoringProgram.name",
      "xpath" : "f:MedicationKnowledge/f:monitoringProgram/f:name",
```

[Search Details](https://www.hl7.org/fhir/medicationknowledge.html) 
```
monitoring-program-name | token | Name of the reviewing program | MedicationKnowledge.monitoringProgram.name |  
-- | -- | -- | -- | --
monitoring-program-type | token | Type of program under which the medication is monitored | MedicationKnowledge.monitoringProgram.type
```

There might be a few possible explanations: 
1 - The spec does not agree with the XPath (not the name field is not a token rather a string) 
2 - The example resource provided in Touchstone does not agree with the specification. 

The difference between the prior runs and this run is the refactoring in the Persistence layer - jdbc.

lmsurpre:
verified via touchstone

PR #351, opened 2019-11-01
issue #348 - add extraction support for multi-type search params

lmsurpre:
string, uri, and canonical types can be extracted as multiple different
search parameter types

to address issue #348, we need to store the value in the valueCode when
a string data type is extracted for a search parameter of type token

while there, I flushed out ParameterExtractionTest to cover all valid
combinations of data type + search parameter type (but just one test per
combo)

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
@@ -6,6 +6,8 @@
 
 package com.ibm.fhir.persistence.jdbc.util;
 
+import static com.ibm.fhir.model.type.code.SearchParamType.*;
+
 import java.math.BigDecimal;
 import java.sql.Timestamp;
 import java.util.ArrayList;
@@ -14,20 +16,24 @@ import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import com.ibm.fhir.model.resource.SearchParameter;
 import com.ibm.fhir.model.type.Address;
 import com.ibm.fhir.model.type.CodeableConcept;
 import com.ibm.fhir.model.type.Coding;
 import com.ibm.fhir.model.type.ContactPoint;
 import com.ibm.fhir.model.type.Date;
 import com.ibm.fhir.model.type.DateTime;
+import com.ibm.fhir.model.type.Element;
 import com.ibm.fhir.model.type.HumanName;
 import com.ibm.fhir.model.type.Identifier;
+import com.ibm.fhir.model.type.Money;
 import com.ibm.fhir.model.type.Period;
 import com.ibm.fhir.model.type.Quantity;
 import com.ibm.fhir.model.type.Range;
 import com.ibm.fhir.model.type.Reference;
 import com.ibm.fhir.model.type.Timing;
 import com.ibm.fhir.model.type.Uri;
+import com.ibm.fhir.model.type.code.SearchParamType;
 import com.ibm.fhir.model.visitor.DefaultVisitor;
 import com.ibm.fhir.model.visitor.Visitable;
 import com.ibm.fhir.persistence.jdbc.dto.Parameter;
@@ -35,13 +41,16 @@ import com.ibm.fhir.persistence.jdbc.dto.Parameter.TimeType;
 
 /**
  * This class is the JDBC persistence layer implementation for transforming SearchParameters into Parameter Data Transfer Objects.
- * 
- * <p>Call {@code Element.accept} with this visitor to add zero to many Parameters to the result list and invoke {@code getResult} 
+ *
+ * <p>Call {@code Element.accept} with this visitor to add zero to many Parameters to the result list and invoke {@code getResult}
  * to get the current list of extracted Parameter objects.
+ *
+ * <p>Note: this class DOES NOT set the resourceType on the underlying JDBC Parameter objects it creates;
+ * that is a responsibility of the caller.
  */
 public class JDBCParameterBuildingVisitor extends DefaultVisitor {
     private static final Logger log = Logger.getLogger(JDBCParameterBuildingVisitor.class.getName());
-    
+
     public static final String EXCEPTION_MSG = "Unexpected error while processing parameter [%s] with value [%s]";
     public static final String EXCEPTION_MSG_NAME_ONLY = "Unexpected error while processing parameter [%s]";
 
@@ -50,13 +59,21 @@ public class JDBCParameterBuildingVisitor extends DefaultVisitor {
     // Derby: https://db.apache.org/derby/docs/10.0/manuals/reference/sqlj271.html
     private static final Timestamp SMALLEST_TIMESTAMP = Timestamp.valueOf("0001-01-01 00:00:00.000000");
     private static final Timestamp LARGEST_TIMESTAMP = Timestamp.valueOf("9999-12-31 23:59:59.999999");
-    
-    private String name;
+
+    // We only need the SearchParameter type and code, so just store those directly as members
+    private final String searchParamCode;
+    private final SearchParamType searchParamType;
+
+    /**
+     * The result of the visit(s)
+     */
     private List<Parameter> result;
-    
-    public JDBCParameterBuildingVisitor(String searchParameterName) {
+
+    public JDBCParameterBuildingVisitor(SearchParameter searchParameter) {
         super(false);
-        this.name = searchParameterName;
+        this.searchParamCode = searchParameter.getCode().getValue();
+        this.searchParamType = searchParameter.getType();
+
         result = new ArrayList<Parameter>();
     }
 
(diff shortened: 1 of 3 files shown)


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
```diff
@@ -78,11 +95,14 @@ public boolean visit(String elementName, int elementIndex, Visitable visitable)
...
+
     @Override
     public boolean visit(java.lang.String elementName, int elementIndex, com.ibm.fhir.model.type.Boolean _boolean) {
         Parameter p = new Parameter();
-        p.setName(name);
+        if (!TOKEN.equals(searchParamType)) {
+            throw invalidComboException(searchParamType, _boolean);
+        }
+        p.setName(searchParamCode);
         p.setValueSystem("http://hl7.org/fhir/special-values");
```
albertwang-ibm: should this be: http://hl7.org/fhir/ValueSet/special-values

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
```diff
@@ -251,44 +324,50 @@ public boolean visit(java.lang.String elementName, int elementIndex, ContactPoin
...
             result.add(p);
         }
         return false;
     }
     @Override
     public boolean visit(java.lang.String elementName, int elementIndex, Identifier identifier) {
+        if (!TOKEN.equals(searchParamType)) {
+            throw invalidComboException(searchParamType, identifier);
+        }
         if (identifier != null && identifier.getValue() != null) {
```
albertwang-ibm: why don't do token check for Identifier?  https://www.hl7.org/fhir/search.html#token

fhir-persistence-jdbc/src/test/java/com/ibm/fhir/persistence/jdbc/test/util/ParameterExtractionTest.java
```diff
@@ -89,7 +472,6 @@ public void testRange() throws FHIRPersistenceProcessorException {
     private String timestampToString(java.sql.Timestamp tstamp) {
         // do not use Timestamp#toString() because it converts to local timezone.
         // We need it rendered in UTC
-        return tstamp.toInstant().atZone(ZoneOffset.UTC).format(DATE_TIME_FORMATTER);
+        return tstamp.toInstant().atZone(ZoneOffset.UTC).format(TIMESTAMP_FORMATTER);
     }
-
 }
```
albertwang-ibm: optionally, even though his may have been covered by those new persistence layer tests. but seems still better to add a e2e test in server integration tests to cover the touchstone medicationknoledge search case mentioned in issue #348.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
```diff
@@ -153,73 +205,88 @@ public boolean visit(java.lang.String elementName, int elementIndex, com.ibm.fhi
     @Override
     public boolean visit(String elementName, int elementIndex, com.ibm.fhir.model.type.String value) {
         Parameter p = new Parameter();
-        p.setName(name);
-        p.setValueString(value.getValue());
+        if (STRING.equals(searchParamType)) {
+            p.setValueString(value.getValue());
+        } else if (TOKEN.equals(searchParamType)) {
```
prb112: in the case of the monitoring program, it'll store the value now in the setValueCode and be inserted as such.  

Does any existing parameters need to be updated? I think they are registered in a parameters table, and drives the query through to the TOKEN table for the Resource type.  in this case, it may be OK as there are no running instances, but may need to be carefully considered on how to update the parameter relationship to point to the right table

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
```diff
@@ -460,15 +570,20 @@ private void setDateValues(Parameter p, java.time.Instant start, java.time.Insta
...
      * @return inclusiveEndTime
      */
     private Timestamp convertToInclusiveEnd(Timestamp exlusiveEndTime) {
         // Our current schema uses the db2/derby default of 6 decimal places (1000 nanoseconds) for fractional seconds.
         return Timestamp.from(exlusiveEndTime.toInstant().minusNanos(1000));
     }
+
+    private IllegalArgumentException invalidComboException(SearchParamType paramType, Element value) {
+        return new IllegalArgumentException("Data type '" + value.getClass().getSimpleName() + "' is not supported "
+                + "for SearchParameter of type '" + paramType.getValue() + "'");
```
prb112: I think the Search Parameter should be added here... the combination will help provide a trajectory to resolve.

prb112: Also, not a super fan of IllegalArgumentException.   why the choice here?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
prb112: further should this be a constant?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
lmsurpre: i don't understand the question

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
lmsurpre: I actually think it should be available from the Boolean class, perhaps as an annotation.  See https://github.com/IBM/FHIR/issues/353

lmsurpre: https://www.hl7.org/fhir/search.html#token lists the exact value I use here:
> The implicit system for boolean values is http://hl7.org/fhir/special-values but this is never actually used

Its possible that they just never updated that with the ValueSet URL you sent (which definitely has true and false).  However, in that case the correct value to use would be `http://terminology.hl7.org/CodeSystem/special-values` and not the one you suggested here.

lmsurpre: I opened https://gforge.hl7.org/gf/project/fhir/tracker/?action=TrackerItemEdit&tracker_item_id=25124 and will update the PR to use this new value instead

lmsurpre:
```suggestion
        p.setValueSystem("http://terminology.hl7.org/CodeSystem/special-values");
```

fhir-persistence-jdbc/src/test/java/com/ibm/fhir/persistence/jdbc/test/util/ParameterExtractionTest.java
lmsurpre: I think the unit tests are sufficient in this case.  My goal is to be very comprehensive in the unit tests, but i don't think we need 100% coverage for the e2e suite.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
lmsurpre: I think not.  The changes how we're extracting the values, but shouldn't affect the actual parameter names or types at all.  The query was always looking in the X_TOKEN_VALUES table...but now we will actually store the value there :-)

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
lmsurpre: > Also, not a super fan of IllegalArgumentException. why the choice here?
Unless we want to change the visitor interface to throw a checked exception, it needs to be a RunTime exception and IllegalArgumentException is the best fit.

lmsurpre:
```suggestion
                + "for SearchParameter '" + searchParamCode + "' of type '" + paramType.getValue() + "'");
```

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
albertwang-ibm: ha, my mistake, the token check is exactly there

Issue #355, opened 2019-11-04
FHIRPath `as` operation (and function) should return an empty list if the operand cannot be converted


lmsurpre:
**Describe the bug**
The `as` function (and operator) is used quite heavilly in FHIR search expressions. For example:
* Condition-onset-age:  `Condition.onset.as(Age) | Condition.onset.as(Range)`
* Patient-death-date:  `Patient.deceased as dateTime`

But when we execute these against the model, our implementation returns the operand, even when its of the wrong type.

**To Reproduce**
```
/*
 * (C) Copyright IBM Corp. 2019
 *
 * SPDX-License-Identifier: Apache-2.0
 */
package com.ibm.fhir.model.path.test;

import static com.ibm.fhir.model.type.String.string;
import static org.testng.Assert.assertEquals;

import java.util.Collection;

import org.testng.annotations.Test;

import com.ibm.fhir.model.path.FHIRPathNode;
import com.ibm.fhir.model.path.evaluator.FHIRPathEvaluator;
import com.ibm.fhir.model.resource.Condition;
import com.ibm.fhir.model.resource.Patient;
import com.ibm.fhir.model.type.Boolean;
import com.ibm.fhir.model.type.DateTime;
import com.ibm.fhir.model.type.Reference;

public class FHIRPathAsTest {
    @Test
    void testAsOperation() throws Exception {
        Patient patient = Patient.builder()
                                 .deceased(Boolean.TRUE)
                                 .build();

        FHIRPathEvaluator evaluator = FHIRPathEvaluator.evaluator();
        Collection<FHIRPathNode> result = evaluator.evaluate(patient, "Patient.deceased as dateTime");

        assertEquals(result.size(), 0, "Number of selected nodes");
    }
    
    @Test
    void testAsFunction() throws Exception {
        Condition condition = Condition.builder()
                                       .subject(Reference.builder().display(string("dummy reference")).build())
                                       .onset(DateTime.now())
                                       .build();

        FHIRPathEvaluator evaluator = FHIRPathEvaluator.evaluator();
        Collection<FHIRPathNode> result = evaluator.evaluate(condition, "Condition.onset.as(Age) | Condition.onset.as(Range)");

        assertEquals(result.size(), 0, "Number of selected nodes");
    }
}
```

**Expected behavior**
Per the spec:
[...]

PR #356, opened 2019-11-04
issue #355 - assert baseType is not null before recursing

lmsurpre:
In the currently implementation, FHIRPathType.isAssignableFrom will
always return true.

For example, if we check is FHIR_DATE_TIME assignable from FHIR_BOOLEAN,
it will recurse until finally we had a baseType of null:
* is FHIR_DATE_TIME assignable from FHIR_ELEMENT
* is FHIR_DATE_TIME assignable from FHIR_ANY
* is FHIR_DATE_TIME assignable from null

And, since everything is assignable from null, it will always be true.

This PR changes that by returning false when we get to baseType == null.

Additionally:
* removal of jdk11 profile for fhir-examples (to match earlier change with rest of project)
* a logging improvement in FHIRPersistenceJDBCImpl.extractSearchParameters and
* a stricter implementation of `as` wrt error conditions.

### DIFF
--- a/fhir-examples/pom.xml
+++ b/fhir-examples/pom.xml
@@ -223,16 +223,6 @@
                 </repository>
             </distributionManagement>
         </profile>
-        
-        <profile>
-            <id>switch-jdk-11</id>
-            <!-- preference is towards 11, when 11-->
-            <activation>
-                <jdk>!1.8</jdk>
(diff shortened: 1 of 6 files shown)


lmsurpre:
closed in favor of #357

Issue #350, opened 2019-11-01
build pipeline fails ocassionally in server integration test


albertwang-ibm:
We found that timeout happens from time to time when processing complete-mock - ExplanationOfBenefit-1.json.

The [CI Workflow](https://github.com/IBM/FHIR/pull/332/checks?check_run_id=285174293) generated the following error on one of the `complete-mock` resources - `OK         ibm/complete-mock/ExplanationOfBenefit-1.json`.

```
 [ERROR] processExamples(com.ibm.fhir.server.test.R4ExampleServerTest)  Time elapsed: 130.08 s  <<< FAILURE!
javax.ws.rs.client.ResponseProcessingException: Problem with reading the data, class com.ibm.fhir.model.resource.ExplanationOfBenefit, ContentType: application/fhir+json.
```

The resource generates the following parameters:

```
Starting Test -> testExtractAll
0|care-team|=[eVOYEME6M8]
1|claim|=[ldKETF2ZcD]
2|coverage|=[QCuLLknY32]
Value DateTime: [2019-08-03T00:12:34.062-04:00]
3|created|=[2019-08-03T00:12:34.062-04:00]
4|detail-udi|=[WhMqzPWY0J]
5|disposition|=[CzARoc8inI]
6|encounter|=[FsfGHHCM4v]
7|enterer|=[FJTi3mIU49]
8|facility|=[819y6m8cV7]
9|identifier|=[null]
10|item-udi|=[XU7oL5ld4n]
11|patient|=[Jad8F8OL2Z]
12|payee|=[gnk1pVCcYx]
13|procedure-udi|=[ZA8nSONiyf]
14|provider|=[NvE9EamfCR]
15|status|=[cancelled]
16|subdetail-udi|=[JSHDCyCWuh]
17|_id|=[YSbxPVnN2s]
Value DateTime: [2019-08-03T00:12:34.152-04:00]
18|_lastUpdated|=[2019-08-03T00:12:34.152-04:00]
19|_profile|=[pjq89TOc13]
20|_security|=['true']
21|_source|=[aTBkgna3DR]
22|_tag|=['true',ibm/complete-mock]
End of Test -> testExtractAll
PASSED: testExtractAll
```

We've hit this issue with both openjdk11 and openjdk8, although it seems more common with openjdk11.

prb112:
I did some rough math and lots of tests 1000 loops of the minimal spec test.  It resulted in 25 to 25 seconds to complete each loop. 

A few causes could be: 
[...]

PR #360, opened 2019-11-05
issue #350 fix for pipeline async http client timeout issue

albertwang-ibm:
use.async.http.conduit was set to true which caused the async client timeout issue which we fuond both in build pipeline and in local. 
This fix changed it to false to use the standard http client instead, and all Patch tests in server integration tests run well with this setting.


lmsurpre:
I'm very confused by the comment that was there previously:
> Tell Apache CXF to use the Async HTTP conduit for PATCH operation as the default HTTP conduit does not support PATCH

Is that just totally bogus?

albertwang-ibm:
> I'm very confused by the comment that was there previously:
> 
> > Tell Apache CXF to use the Async HTTP conduit for PATCH operation as the default HTTP conduit does not support PATCH
> 
> Is that just totally bogus?

it was also confusing for me, and this is why I kept the comments there instead of simply removing it... I would guess maybe a old cxf/httpclient lib was used when the comments was made.

PR #369, opened 2019-11-06
Issue 260 - introduce handling config and use strict by default 

lmsurpre:
previously we used lenient by default, so I needed to update a number of
tests as well

also updated FHIRServerUserGuide with the new config property and
deleted the stale/outdated version of this (and Conformance.md) under
docs

also updated SearchUtil.parseSortParameter to handle comma-separated list of values

also includes a fix for #370 such that
1. the `as` operator will operate over a non-singleton collection; and
2. the `as` and `is` operators will return an empty list if the left operand is an empty list

and finally:
1. added SQL to the logging statements in ResourceDAOImpl searchForIds
and searchByIds methods; and
2. updated javadoc in SortedQuerySegmentAggregator to preserve
formatting of SQL snippets

*this pull request updates both fhir-core and fhir-config, so don't forget to update the shared libs in your local deployments*


prb112:
There are a few errors on the CI run 
```
[ERROR] Tests run: 335, Failures: 14, Errors: 0, Skipped: 0
```

fhir-core/src/main/java/com/ibm/fhir/core/HTTPHandlingPreference.java
```diff
@@ -0,0 +1,42 @@
...
+ * (C) Copyright IBM Corp. 2019
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.ibm.fhir.core;
+
+/**
+ * An enumeration of handling preference codes.
+ * 
+ * @see https://tools.ietf.org/html/rfc7240#section-4.4
```
prb112: in the future, we should consider using an a href - check @see form 2 --> https://docs.oracle.com/javase/8/docs/technotes/tools/windows/javadoc.html#CHDJGIJB

> Form 2. The @see <a href="URL#value">label</a> form adds a link as defined by URL#value. The URL#value parameter is a relative or absolute URL. The javadoc command distinguishes this from other cases by searching for a less-than symbol (<) as the first character. For example, @see <a href="spec.html#section">Java Spec</a>

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
```diff
@@ -400,7 +400,7 @@ public Integer readResourceTypeId(String resourceType) throws FHIRPersistenceDBC
             resultSet = stmt.executeQuery();
             dbCallDuration = (System.nanoTime()-dbCallStartTime)/1e6;
             if (log.isLoggable(Level.FINE)) {
-                log.fine("DB search for ids complete. executionTime=" + dbCallDuration + "ms");
+                log.fine("DB search for ids complete. " + queryData + "  executionTime=" + dbCallDuration + "ms");
```
prb112: I think a comment should be added as to what is logged and why, and also indicate that PHI could be logged here. 

Also should a `[` and `]` be used to delineate the queryData output? It might already be wrapped in the output of toString.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
```diff
@@ -455,7 +455,7 @@ public Integer readResourceTypeId(String resourceType) throws FHIRPersistenceDBC
             resultSet = stmt.executeQuery();
             dbCallDuration = (System.nanoTime()-dbCallStartTime)/1e6;
             if (log.isLoggable(Level.FINE)) {
-                log.fine("DB search by ids complete. executionTime=" + dbCallDuration + "ms");
+                log.fine("DB search by ids complete. SQL=" + idQuery.toString() + "  executionTime=" + dbCallDuration + "ms");
```
prb112: Should a `[` and `]` be used to delineate the idQuery output?

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -627,13 +615,17 @@ public static FHIRSearchContext parseQueryParameters(Class<?> resourceType, Map<
                         parameter.getValues().addAll(queryParameterValues);
                         parameters.add(parameter);
                     }
-                }
-
+                } // end else
             } catch (Exception e) {
-                throw SearchExceptionUtil.buildNewParseParameterException(name, e);
+                if (lenient) {
+                    // log and continue
```
prb112: I don't think this should be in an exception block. 
595 should set a boolean indicating a problem with parsing and in the finally block, it should throw or log. 
on 623 the continue is not necessary you are at the end of the branch.

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -2088,22 +2092,56 @@ public Bundle doSearch(String type, String compartment, String compartmentId,
...
+     * 
+     * @throws IllegalArgumentException if the mode is STRICT and the Prefer header contains an invalid handling value.
+     */
     private boolean isLenient(Map<String, String> requestProperties) {
-        boolean lenient = true;
-
-        String handlingStringValue = getHeaderValue(requestProperties, "Prefer", "handling");
-        if ("strict".equals(handlingStringValue)) {
-            lenient = false;
+        boolean isLenient = false; // Assign it to appease the compiler
```
prb112: to appease the compiler in this case, you need to set isLenient in the default branch of the below, however, I assume Handling.from ensures no other case.

fhir-server/src/main/java/com/ibm/fhir/server/util/Handling.java
```diff
@@ -0,0 +1,54 @@
+/*
+ * (C) Copyright IBM Corp. 2019
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.ibm.fhir.server.util;
+
+/**
+ * Allowed options for the fhirServer/core/handling config property
+ * @see https://tools.ietf.org/html/rfc7240#section-4.4
```
prb112: same comment from earlier - a href please

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -2088,22 +2092,56 @@ public Bundle doSearch(String type, String compartment, String compartmentId,
...
+        String stringVal = FHIRConfigHelper.getStringProperty(FHIRConfiguration.PROPERTY_HANDLING, Handling.STRICT.value());
+        Handling handlingConfig = Handling.from(stringVal);
```
prb112: up to 3 calls here... and the same 3 calls in the case statement. 

I don't think you need this if you just test handlingConfig as a string, and pass in the found Handling.LENIENT at the end.

fhir-server/src/main/java/com/ibm/fhir/server/util/Handling.java
```diff
@@ -0,0 +1,54 @@
...
+        throw new IllegalArgumentException(value);
+    }
+
+    /**
+     * The String value of this enumeration
+     */
+    public String value() {
+        return value;
+    }
+}
```
prb112: This comment isn't so much about this PR (I think) 

Below this line... 
What is the `Unchanged Files with check annotations` ?  is that where the above PR failed?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
lmsurpre: yeah, queryData.toString is like `[queryString=" + queryString + ", bindVariables=" + bindVariables + "]"`

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
lmsurpre: in this case we didn't have a SqlQueryData object, so yeah, wrapping the query in square brackets could help to delineate it.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
lmsurpre: there's a lot of places that throw within this try block.  in all cases we want the same behavior...if there was an issue parsing the query parameter then move on to the next one

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
lmsurpre: correct.  we could either initialize it or create a default block.  neither is actually needed because its an enum and we have cases for all possible values.

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
lmsurpre: different calls.  this one is reading the config preference.  within the case we're reading the Prefer header handling value.

fhir-server/src/main/java/com/ibm/fhir/server/util/Handling.java
lmsurpre: where do you see `Unchanged Files with check annotations`?

Pretty sure PR failed because I updated the default behavior and didn't update the e2e tests.

prb112: I pinged you an image. just in case we aren't seeing the same thing.  this was more curiosity as I don't think those files changed.

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
prb112: yes and no.  you are using the value twice in a branch which might as well be one set of branches in the switch statement.

fhir-server/src/main/java/com/ibm/fhir/server/util/Handling.java
lmsurpre:
```suggestion
 * @see <a href="https://tools.ietf.org/html/rfc7240#section-4.4">href="https://tools.ietf.org/html/rfc7240#section-4.4</a>
```

fhir-core/src/main/java/com/ibm/fhir/core/HTTPHandlingPreference.java
lmsurpre:
```suggestion
 * @see <a href="https://tools.ietf.org/html/rfc7240#section-4.4">href="https://tools.ietf.org/html/rfc7240#section-4.4</a>
```

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
prb112:
```suggestion
                log.fine("DB search by ids complete. SQL=[" + idQuery.toString() + "]  executionTime=" + dbCallDuration + "ms");
```

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
prb112: messy... and understandable.  mind adding your comment as a comment in the Exception block.  ?

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
prb112: design choice in my opinion and we can mark as resolved

lmsurpre: I think the only line thats repeated is `String handlingStringValue = getHeaderValue(requestProperties, HEADERNAME_PREFER, "handling");` and I didn't want to do that before the switch because we don't need it for LEIENT_ONLY or STRICT_ONLY cases.

fhir-model/src/main/java/com/ibm/fhir/model/path/evaluator/FHIRPathEvaluator.java
```diff
@@ -242,22 +242,24 @@ private void reset() {
...
+            String identifier = typeName.getText();
+            FHIRPathType type = FHIRPathType.from(identifier);
+            if (type == null) {
+                throw new IllegalArgumentException(String.format("Argument '%s' cannot be resolved to a valid type identifier", identifier));
+            }
+            FHIRPathNode node = getSingleton(currentContext);
+            if (type.isAssignableFrom(node.type())) {
+                return SINGLETON_TRUE;
             }
             return SINGLETON_FALSE;
```
prb112: If I'm looking at this correctly 261 and 262 and 264 could be collapsed into a single expression?

fhir-model/src/main/java/com/ibm/fhir/model/path/evaluator/FHIRPathEvaluator.java
```diff
@@ -880,31 +882,35 @@ private IllegalArgumentException unexpectedNumberOfArguments(int arity, String f
             Collection<FHIRPathNode> nodes = visit(ctx.expression());
             String operator = ctx.getChild(1).getText();
             
-            Collection<FHIRPathNode> result = "is".equals(operator) ? SINGLETON_FALSE : empty();
+            Collection<FHIRPathNode> result = "is".equals(operator) ? SINGLETON_FALSE : new ArrayList<>();
```
prb112: Collections.emptyList() ?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
```diff
@@ -432,12 +432,12 @@ public boolean visit(java.lang.String elementName, int elementIndex, Quantity qu
             // see https://gforge.hl7.org/gf/project/fhir/tracker/?action=TrackerItemEdit&tracker_item_id=19597
             if (quantity.getCode() != null) {
                 p.setValueCode(quantity.getCode().getValue());
+                if (quantity.getSystem() != null) {
+                    p.setValueSystem(quantity.getSystem().getValue());
```
prb112: Looking at this branching it occurs to me... 
should Parameter object have a default getSystem ?  

How do we handle null? sorry if I am re-treading

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -512,23 +512,22 @@ private static SearchParameter getSearchParameterInternal(Map<String, Map<String
                     log.fine(String.format(UNSUPPOTED_EXPR_NULL, parameter.getType(), parameter.getCode().getValue()));
                 }
             }
-
         }
-
+
```
prb112: extra space?

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -512,23 +512,22 @@ private static SearchParameter getSearchParameterInternal(Map<String, Map<String
...
+        return parseQueryParameters(resourceType, queryParameters, false);
     }
 
     public static FHIRSearchContext parseQueryParameters(Class<?> resourceType, Map<String, List<String>> queryParameters, boolean lenient)
         throws Exception {
         
         FHIRSearchContext context = FHIRSearchContextFactory.createSearchContext();
         context.setLenient(lenient);
-        List<Parameter> parameters = new ArrayList<>();
+        List<Parameter> parameters = new ArrayList<>();
```
prb112: space at end?

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -627,13 +614,21 @@ public static FHIRSearchContext parseQueryParameters(Class<?> resourceType, Map<
                         parameter.getValues().addAll(queryParameterValues);
                         parameters.add(parameter);
                     }
+                } // end else
+            }  catch (FHIRSearchException se) {
+                // There's a number of places that throw within this try block. In all cases we want the same behavior:
+                // If we're in lenient mode and there was an issue parsing the query parameter then log and move on to the next one.
```
prb112: thank you

fhir-server-test/src/test/java/com/ibm/fhir/cli/test/FHIRCliTest.java
```diff
@@ -208,17 +208,25 @@ public void testConditionalDeletePatientError() throws Exception {
     
     @Test(dependsOnMethods={"testCreatePatient"})
     public void testConditionalCreatePatient() throws Exception {
-        assertNotNull(patientId);
         runTest("testConditionalCreatePatient", "-p", propsFile(), "--operation", "conditional-create", "--resource", testData("Patient_MookieBetts.json"), "-qp", "_id=" + patientId);
-        // According to https://www.hl7.org/fhir/R4/http.html#create , condition create against an existing resource should return 200
+        // According to https://www.hl7.org/fhir/R4/http.html#ccreate , conditional create against an existing resource should return 200
```
prb112: is the extra c in ccreate intentional?

fhir-server-test/src/test/java/com/ibm/fhir/server/test/BundleTest.java
```diff
@@ -829,14 +827,12 @@ public void testBatchSearchWithEncoding() throws Exception {
         
         Bundle responseBundle = getEntityWithExtraWork(response,method);
         
-        assertResponseBundle(responseBundle, BundleType.BATCH_RESPONSE, 2);
+        assertResponseBundle(responseBundle, BundleType.BATCH_RESPONSE, 1);
```
prb112: Why is this 1?  Is it suppose to be 1? 

I got sucked into this before where the code was wrong.

docs/readme.md
```diff
@@ -78,11 +78,11 @@ Only one column and content area used. (RIGHT column not used)
...
+This is the only one that should be edited.
 
-The other files are theme. 
+The other files are theme.
 
-# Links 
+# Links
 [Gatsby Theme: Carbon](https://github.com/carbon-design-system/gatsby-theme-carbon)
 [Gatsby Theme Carbon: Guides](https://gatsby-theme-carbon.now.sh/guides/)
 [Carbon Design](https://www.carbondesignsystem.com/components/link/code/)
```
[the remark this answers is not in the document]
prb112: I don't use Eclipse for md files. I use vscode

fhir-model/src/main/java/com/ibm/fhir/model/path/evaluator/FHIRPathEvaluator.java
lmsurpre: we need ability to append to it

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
lmsurpre: the DAO layer inserts a default system if its null, but my understanding is its just a database perf thing.  a quantity SHOULD have a code and a system, but isn't required to.

albertwang-ibm:
seems a few build error and integration test errors on both jdk8 and 11.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
lmsurpre: courtesy of eclipse i guess.  we have these all over our codebase.

docs/readme.md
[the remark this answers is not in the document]
albertwang-ibm: cool, I like vscode. I use it also for node.js and javascript.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
lmsurpre:
```suggestion
        List<Parameter> parameters = new ArrayList<>();
```

fhir-server-test/src/test/java/com/ibm/fhir/cli/test/FHIRCliTest.java
lmsurpre: yes.  now it links to 'conditional create' and not just 'create'

docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -1190,6 +1190,7 @@ This section contains reference information about each of the configuration prop
...
 |`fhirServer/core/tenantIdHeaderName`|string|The name of the request header that will be used to specify the tenant-id for each incoming FHIR REST API request. For headers with semicolon-delimited parts, setting a header name like `<headerName>:<partName>` will select the value from the part of header `<headerName>`'s value with a name of `<partName>` (e.g. setting `X-Test:part1` would select `someValue` from the header `X-Test: part1=someValue;part2=someOtherValue`).|
 |`fhirServer/core/dataSourceIdHeaderName`|string|The name of the request header that will be used to specify the datastore-id for each incoming FHIR REST API request. For headers with semicolon-delimited parts, setting a header name like `<headerName>:<partName>` will select the value from the part of header `<headerName>`'s value with a name of `<partName>` (e.g. setting `X-Test:part1` would select `someValue` from the header `X-Test: part1=someValue;part2=someOtherValue`).|
+|`fhirServer/core/handling`|string|The default handling mode of the server (`strict | lenient`) and whether to honor client handling preferences passed through the Prefer header or not (`strict vs strict-only`)|
```
albertwang-ibm: maybe prefer-handling is more accurate?

docs/readme.md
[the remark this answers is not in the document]
lmsurpre: the trailing whitespaces don't really bother me, but atom edit automatically squashes these.  i mostly just use atom for resolving merge conflincts (because i can't stand those in eclipse).

docs/src/pages/guides/FHIRServerUsersGuide.md
lmsurpre: well this is the server config.  the client can pass their preference in the Prefer header, but this config setting is for controlling what the server uses.  could be `handling-mode` or some such, but i thought just `handling` was fine.

fhir-model/src/main/java/com/ibm/fhir/model/path/evaluator/FHIRPathEvaluator.java
lmsurpre: yeah, it could just be:
`return type.isAssignableFrom(node.type()) ? SINGLETON_TRUE : SINGLETON_FALSE;`

about the same to me, but i'm happy to make that change

fhir-server-test/src/test/java/com/ibm/fhir/server/test/BundleTest.java
lmsurpre: I removed the entry from lines 820 and 821, so now we expect 1 instead of 2.

Specifically, now that we're strict by default this second request was failing due to the invalid "Jeter" parameter name.  I don't think that was needed for the test (which is ensuring that the `%26` gets properly unencoded) and so I removed it.

fhir-model/src/main/java/com/ibm/fhir/model/path/evaluator/FHIRPathEvaluator.java
prb112: Just an observation, no change necessary.

prb112: I do believe in the Knuth principle of a single return

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
prb112: Won't let me commit your suggestion

fhir-server-test/src/test/java/com/ibm/fhir/server/test/BundleTest.java
prb112: as an Annoying RedSox fan... this makes me happy

** and double checking the intent

fhir-model/src/main/java/com/ibm/fhir/model/path/evaluator/FHIRPathEvaluator.java
```diff
@@ -878,6 +882,10 @@ private IllegalArgumentException unexpectedNumberOfArguments(int arity, String f
             indentLevel++;
             
             Collection<FHIRPathNode> nodes = visit(ctx.expression());
+            if (nodes.isEmpty()) {
+                // early exit
```
JohnTimm: indentLevel-- should go right before the return empty()

albertwang-ibm:
still build errors

fhir-model/src/main/java/com/ibm/fhir/model/path/evaluator/FHIRPathEvaluator.java
lmsurpre: i thought i had pushed this change, but i must have missed it in my commit.  now addressed.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
lmsurpre: done

fhir-model/src/main/java/com/ibm/fhir/model/path/evaluator/FHIRPathEvaluator.java
lmsurpre: updated it as suggested

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
lmsurpre: done

Issue #383, opened 2019-11-08
Fail to store fhir resource to db2


albertwang-ibm:
Steps:
(1) Created a new db2 instance with FHIR schema deployed.
(2) Created a new tenant and pointed fhir server to use the new db2 instance.
(3) Run server integration tests 
Results:
Failed to create resources

PR #384, opened 2019-11-08
issue #383 fix db2 fhir resource creation issue

albertwang-ibm:

In add_any_resource storedprocedure, the codes insert record to the resource table with "TENANT_ID" column, but in the resource table creation codes, the "TENANT_ID" column is missed, this fix is to add the missed "TENANT_ID" column to the table creation codes.
I choose to add it instead of dropping it from the storedprocedure for the purpose of being compatible with existing deployments.

### DIFF
--- a/fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirResourceGroup.java
+++ b/fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirResourceGroup.java
@@ -186,7 +186,6 @@ public class FhirResourceGroup {
      * data,
      * last_updated,
      * is_deleted,
-  tenant_id                 INT             NOT NULL,
   resource_id            BIGINT             NOT NULL,
   logical_resource_id    BIGINT             NOT NULL,
   version_id                INT             NOT NULL,
@@ -245,8 +244,6 @@ public class FhirResourceGroup {
 
     /**
      * Add the STR_VALUES table for the given resource name prefix
-
-  tenant_id                INT             NOT NULL,
   parameter_name_id        INT             NOT NULL,
   str_value            VARCHAR(511 OCTETS),
   str_value_lcase      VARCHAR(511 OCTETS),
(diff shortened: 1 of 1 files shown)


fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirResourceGroup.java
```diff
@@ -181,12 +223,13 @@ public void addResources(List<IDatabaseObject> group, String prefix) {
...
-                // Start Backwards Compatibility 
+                // Start Backwards Compatibility
                 .addVarcharColumn(TX_CORRELATION_ID, 36, true)
                 .addVarcharColumn(CHANGED_BY, 64, true)
-                .addVarcharColumn(CORRELATION_TOKEN, 36, true) 
+                .addVarcharColumn(CORRELATION_TOKEN, 36, true)
                 .addVarcharColumn(REASON, 255, true)
                 .addVarcharColumn(SERVICE_ID, 32, true)
+                .addVarcharColumn(TENANT_ID, 36, true)
                 // End Backwards Compatibility
```
prb112: out of curiosity why is TENANT_ID added back?  are we setting this on INSERT?

albertwang-ibm: yes, it's in the storedprocedure, I choose to add it instead of removing it because existing envs.

prb112: I suspect changing the stored procedure is probably preferred. Is the TENANT_ID actually used?

prb112: got it - because it's not null

prb112: Looks good!

albertwang-ibm: I don't really think it's used, I think we should further clean the codes later, I kept it there just because the storeprocedure with tenant_id works well with our cloud deployment, and we don't have a clear path or db schema migration yet, so just want to play safe for now.

PR #395, opened 2019-11-11
issue #394 fix nullpointer issue in parameter builder and visitor

albertwang-ibm:


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
```diff
@@ -560,6 +569,9 @@ private void setDateValues(Parameter p, DateTime dateTime) {
      * @param end
      */
     private void setDateValues(Parameter p, java.time.Instant start, java.time.Instant end) {
+        if (start == null) {
+            return;
+        }
```
lmsurpre: what if start is not null, but end is?

also, maybe update the javadoc to say what happens for a null start or end?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
```diff
@@ -560,6 +569,9 @@ private void setDateValues(Parameter p, DateTime dateTime) {
      * @param end
      */
     private void setDateValues(Parameter p, java.time.Instant start, java.time.Instant end) {
+        if (start == null) {
```
lmsurpre:
```suggestion
        if (start == null || end == null) {
```

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
lmsurpre: maybe it should be more like this?
```
        if (start != null) {
            Timestamp startTime = Timestamp.from(start);
            p.setValueDateStart(startTime);
            p.setValueDate(startTime);
            p.setTimeType(TimeType.UNKNOWN);
        }

        if (end != null) {
            Timestamp implicitEndExclusive = Timestamp.from(end);
            // TODO: Is it possible to avoid this by using <= or BETWEEN instead of < when constructing the query?
            Timestamp implicitEndInclusive = convertToInclusiveEnd(implicitEndExclusive);
            p.setValueDateEnd(implicitEndInclusive);
        }
```

        Timestamp implicitEndExclusive = Timestamp.from(end);
        // TODO: Is it possible to avoid this by using <= or BETWEEN instead of < when constructing the query?
        Timestamp implicitEndInclusive = convertToInclusiveEnd(implicitEndExclusive);
        p.setValueDateEnd(implicitEndInclusive);
```

lmsurpre:
don't we also need a null check on either 
* `visit(java.lang.String elementName, int elementIndex, com.ibm.fhir.model.type.DateTime dateTime)`; or 
* `setDateValues(Parameter p, DateTime dateTime)` ?

lmsurpre:
maybe also add null tests in ParameterExtractionTest.java ?

PR #397, opened 2019-11-11
Patch for 395

lmsurpre:
### commit 1
    issue #395 - add guards for missing values

    privitive elements with no values (only extensions) are not common, but
    they are allowed and we shouldn't fail on them
    
    also note: i updated the `Coding` (and `CodeableConcept`) extraction so
    that a Coding won't get indexed unless it has a code value.
    
    also updated parent-pom jdbc-all-tests profile to enable setting of
    JDBC testType independent of the model testType

### commit 2
    I found that the existing profiles (`model-all-tests` and
    `jdbc-all-tests`) didn't actual set any system properties at runtime. I
    fixed this by
    1. introducing `fhir-model.testType` and
    `fhir-persistence-jdbc.testType` parameters with a default value of
    MINIMAL
    2. updating the profile definitions to set these properties when
    activated
    3. updating surefire-plugin config in `fhir-model/pom.xml` and
    `fhir-persistence-jdbc/pom.xml` to use these properties to set the
    `com.ibm.fhir.model.spec.test.R4ExamplesDriver.testType` and
    `com.ibm.fhir.persistence.jdbc.test.spec.R4JDBCExamplesTest.testType`
    system properties respectively.
    
    Also, minor cleanup in ExtractorValidator.

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
@@ -104,136 +104,155 @@ public class JDBCParameterBuildingVisitor extends DefaultVisitor {
 
     @Override
     public boolean visit(java.lang.String elementName, int elementIndex, com.ibm.fhir.model.type.Boolean _boolean) {
-        Parameter p = new Parameter();
-        if (!TOKEN.equals(searchParamType)) {
-            throw invalidComboException(searchParamType, _boolean);
+        if (_boolean.hasValue()) {
+            Parameter p = new Parameter();
+            if (!TOKEN.equals(searchParamType)) {
+                throw invalidComboException(searchParamType, _boolean);
+            }
+            p.setName(searchParamCode);
+            p.setValueSystem("http://terminology.hl7.org/CodeSystem/special-values");
+            if (_boolean.getValue()) {
+                p.setValueCode("true");
+            } else {
+                p.setValueCode("false");
+            }
+            result.add(p);
         }
-        p.setName(searchParamCode);
-        p.setValueSystem("http://terminology.hl7.org/CodeSystem/special-values");
-        if (_boolean.getValue() != null && _boolean.getValue()) {
-            p.setValueCode("true");
-        } else {
-            p.setValueCode("false");
-        }
-        result.add(p);
         return false;
     }
     @Override
     public boolean visit(java.lang.String elementName, int elementIndex, com.ibm.fhir.model.type.Canonical canonical) {
-        Parameter p = new Parameter();
-        if (!REFERENCE.equals(searchParamType) && !URI.equals(searchParamType)) {
-            throw invalidComboException(searchParamType, canonical);
+        if (canonical.hasValue()) {
+            Parameter p = new Parameter();
+            if (!REFERENCE.equals(searchParamType) && !URI.equals(searchParamType)) {
+                throw invalidComboException(searchParamType, canonical);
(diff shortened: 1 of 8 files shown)


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
```diff
@@ -104,136 +104,155 @@ public boolean visit(String elementName, int elementIndex, Visitable visitable)
...
-        if (!TOKEN.equals(searchParamType)) {
-            throw invalidComboException(searchParamType, _boolean);
-        }
-        p.setName(searchParamCode);
-        p.setValueSystem("http://terminology.hl7.org/CodeSystem/special-values");
-        if (_boolean.getValue() != null && _boolean.getValue()) {
-            p.setValueCode("true");
-        } else {
-            p.setValueCode("false");
+        if (_boolean.hasValue()) {
```
albertwang-ibm: can _boolean be null?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
```diff
@@ -104,136 +104,155 @@ public boolean visit(String elementName, int elementIndex, Visitable visitable)
...
     }
     @Override
     public boolean visit(java.lang.String elementName, int elementIndex, com.ibm.fhir.model.type.Instant instant) {
-        if (instant == null || instant.getValue() == null) {
-            return false;
-        }
-        Parameter p = new Parameter();
-        if (!DATE.equals(searchParamType)) {
-            throw invalidComboException(searchParamType, instant);
+        if (instant.hasValue()) {
```
albertwang-ibm: so, don't have to do the null check anymore?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
lmsurpre: technically it could if someone invoked the method that way, but it will never be null from FHIRPersistenceJDBCImpl.extractSearchParameters because its invoked via double dispatch and the datatype visit methods have null checks in AbstractVisitable:
```
protected void accept(java.lang.String value, java.lang.String elementName, Visitor visitor) {
        if (value != null) {
            visitor.visit(elementName, value);
        }
    }
```

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
albertwang-ibm: seems the jdbc tests passed with all the examples in the pipeline, so, I think it should be good
Nov 11, 2019 7:32:22 PM com.ibm.fhir.model.spec.test.R4ExamplesDriver processIndex
INFO: Overall success rate = 4338/4338 = 100%. Took 541359 ms

lmsurpre: i removed the `if (instant == null)` part to be consistent will.  per my response at https://github.com/IBM/FHIR/pull/397/files#r344880278 I don't think we need to check whether the passed Element is null, but we DO need to check that it has a nonNull value.  I favored `hasValue()` over `getValue() != null` for that.

PR #398, opened 2019-11-11
issue-370 - fixup 1c6498f1b41b9f790ef8673dc45e2048dd447cce

lmsurpre:
In pull request https://github.com/IBM/FHIR/pull/376 I missed that `!isSingleton(nodes)` would be true for empty collections and so I didn't get the intended behavior.
Now `is` will evaluate to SINGLETON_FALSE for in both the operation and function contexts.

### DIFF
--- a/fhir-model/src/main/java/com/ibm/fhir/model/path/evaluator/FHIRPathEvaluator.java
+++ b/fhir-model/src/main/java/com/ibm/fhir/model/path/evaluator/FHIRPathEvaluator.java
@@ -904,14 +904,14 @@ public class FHIRPathEvaluator {
 
             switch (operator) {
             case "is":
-                if (!isSingleton(nodes)) {
+                if (nodes.size() > 1) {
                     throw new IllegalArgumentException(String.format("Input collection has %d items, but only 1 is allowed", nodes.size()));
                 } else if (!nodes.isEmpty()) {
                     FHIRPathNode node = getSingleton(nodes);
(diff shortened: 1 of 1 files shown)


fhir-model/src/main/java/com/ibm/fhir/model/path/evaluator/FHIRPathEvaluator.java
```diff
@@ -904,14 +904,14 @@ private IllegalArgumentException unexpectedNumberOfArguments(int arity, String f
 
             switch (operator) {
             case "is":
-                if (!isSingleton(nodes)) {
+                if (nodes.size() > 1) {
```
albertwang-ibm: the description in the exception messages seems doesn't align with the codes... 
so, only 0 and 1 are allowed or really only 1 is allowed?

lmsurpre: its a bit fuzzy.  the spec says only 1 is allowed, but in practice we need to allow 0.  i think its ok that the error message says only 1 is allowed because now its only shown in cases where the size > 1.

lmsurpre: see https://gforge.hl7.org/gf/project/fhir/tracker/?action=TrackerItemEdit&tracker_item_id=25189 for more info

PR #399, opened 2019-11-11
conditionally execute more tests from the github action

lmsurpre:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,12 +46,37 @@ jobs:
       env:
         BASE: origin/${{ github['base_ref'] }}
       run: |
-        paths=( fhir-search/src/main fhir-persistence/src/main fhir-persistence-schema/src/main fhir-persistence-jdbc )
-        if [ -n "$(git --no-pager diff --name-only ${BASE} ${GITHUB_SHA} -- ${paths[@]})" ]; then
+        jdbc_paths=(
+          fhir-search/src/main
+          fhir-persistence/src/main
+          fhir-persistence-schema/src/main
+          fhir-persistence-jdbc
+        )
+        model_paths=(
+          fhir-model
+        )
+        declare -a PROFILES_ARR
+
+        if [ -n "$(git --no-pager diff --name-only ${BASE} ${GITHUB_SHA} -- ${jdbc_paths[@]})" ]; then
           echo "Changes affect fhir-persistence-jdbc, running ALL"
-          mvn -B -T2C package --file fhir-parent/pom.xml -P jdbc-all-tests --no-transfer-progress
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -43,7 +43,16 @@ jobs:
...
+        if [ -n "$(git --no-pager diff --name-only ${BASE} ${GITHUB_SHA} -- fhir-search fhir-persistence fhir-persistence-jdbc)" ]; then
+          echo "Changes affect fhir-persistence-jdbc, running ALL"
+          mvn -B -T2C package --file fhir-parent/pom.xml -P jdbc-all-tests --no-transfer-progress
+        else
+          echo "No changes affect fhir-persistence-jdbc, running MINIMAL"
+          mvn -B -T2C package --file fhir-parent/pom.xml --no-transfer-progress
+        fi
   e2e-tests:
     runs-on: ubuntu-latest
     strategy:
```
albertwang-ibm: should we add fhir-persistence-schema also?

fhir-persistence-jdbc/src/test/java/com/ibm/fhir/persistence/jdbc/test/spec/Main.java
```diff
@@ -94,9 +94,6 @@
     // The thread pool used for concurrent request processing
     private ExecutorService pool;
 
-    // number of ms to wait for the thread pool to shut down
-    private long poolShutdownWaitTime = 60000;
-
     private List<FileExpectation> fileExpectations = new ArrayList<>();
 
     // mode of operation
```
albertwang-ibm: does this mean poolShutdownWaitTime has never been used?

.github/workflows/build.yml
lmsurpre: sure.  included in update

fhir-persistence-jdbc/src/test/java/com/ibm/fhir/persistence/jdbc/test/spec/Main.java
lmsurpre: correct, eclipse warned that this value is never used and so i removed it.

however, to be completely honest, i was just looking for an edit i could make in jdbc to verify the conditional logic.

.github/workflows/build.yml
```diff
@@ -43,7 +43,42 @@ jobs:
     - name: Build samples
       run: mvn -B install --file fhir-examples/pom.xml --no-transfer-progress
     - name: Build parent with tests
-      run: mvn -B -T2C package --file fhir-parent/pom.xml -P jdbc-all-tests --no-transfer-progress
+      env:
```
prb112: excellent

Issue #400, opened 2019-11-12
jdbc performance test tool enhancement


albertwang-ibm:
To allow us to do performance tests against derby network and db2.
(1) Add derby network support with the simple connection pool.
(2) Enhance db2 test to support multiple connections with the simple connection pool.
(3) Provide sample configuration file for the performance tests.
(4) Provide document about how to use the tool.

PR #405, opened 2019-11-13
issue #400 #406 minor changes for 400 and add simpe testng performance test for search

albertwang-ibm:

### DIFF
--- a/fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchPerformaceTest.java
+++ b/fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchPerformaceTest.java
@@ -1,0 +1,488 @@
+/*
+ * (C) Copyright IBM Corp. 2017,2019
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.ibm.fhir.server.test;
+
+import static com.ibm.fhir.model.type.String.string;
+import static com.ibm.fhir.model.type.Uri.uri;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertNull;
+import static org.testng.Assert.assertTrue;
+
+import java.lang.reflect.Method;
+import java.util.List;
+
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.client.WebTarget;
+import javax.ws.rs.core.Response;
+
+import org.testng.annotations.Test;
+
+import com.ibm.fhir.core.FHIRMediaType;
+import com.ibm.fhir.model.resource.Bundle;
+import com.ibm.fhir.model.resource.Observation;
+import com.ibm.fhir.model.resource.Patient;
+import com.ibm.fhir.model.test.TestUtil;
+import com.ibm.fhir.model.type.Code;
+import com.ibm.fhir.model.type.Coding;
+import com.ibm.fhir.model.type.code.AdministrativeGender;
+import com.ibm.fhir.model.util.FHIRUtil;
+
+public class SearchPerformaceTest extends FHIRServerTestBase {
+
+    private static final boolean DEBUG_SEARCH = false;
+    private String patientId;
+    private String observationId;
+    private Boolean compartmentSearchSupported = null;
+    // Controls which tenant to use for the test.
+    private final String tanantName = "perform2";
+    // Controls how many observations and patients to create for the test.
+    // Using invocationCount of testng can cause the testng report grows too big, so
+    // this config is used to make sure all test users are created in one testng step.
+    private final int numOfPatientObservationsToCreate = 100;
+
+    /**
+     * Retrieve the server's conformance statement to determine the status of certain runtime options.
+     *
+     * @throws Exception
+     */
+    @Test
+    public void retrieveConfig() throws Exception {
+        compartmentSearchSupported = isComparmentSearchSupported();
+        System.out.println("Compartment-based search supported?: "
+                + compartmentSearchSupported.toString());
+    }
+
+    @Test(groups = { "server-search-performance" }, invocationCount = 1)
+    public void testCreatePatientAndObservation() throws Exception {
+        WebTarget target = getWebTarget();
+        for (int i = 0; i < numOfPatientObservationsToCreate; i++) {
+            // Build a new Patient and then call the 'create' API.
+            Patient patient = TestUtil.readLocalResource("Patient_JohnDoe.json");
+
+            patient = patient.toBuilder().gender(AdministrativeGender.MALE).build();
+            Entity<Patient> entity =
+                    Entity.entity(patient, FHIRMediaType.APPLICATION_FHIR_JSON);
+            Response response =
+                    target.path("Patient").request()
+                    .header("X-FHIR-TENANT-ID", tanantName)
+                    .post(entity, Response.class);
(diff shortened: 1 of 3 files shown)


fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchPerformaceTest.java
```diff
@@ -0,0 +1,488 @@
...
+     * @throws Exception
+     */
+    @Test
+    public void retrieveConfig() throws Exception {
+        compartmentSearchSupported = isComparmentSearchSupported();
+        System.out.println("Compartment-based search supported?: "
+                + compartmentSearchSupported.toString());
+    }
+
+    @Test(groups = { "server-search-performance" }, invocationCount = 1)
```
lmsurpre: so to bump up the number of resources you just bump `invocationCount`?

albertwang-ibm: please refer to my comments at the top of the file, actually, I'm thinking to remove this...

albertwang-ibm: // Controls how many observations and patients to create for the test.
    // Using invocationCount of testng can cause the testng report grows too big, so
    // this config is used to make sure all test users are created in one testng step.
    private final int numOfPatientObservationsToCreate = 100;

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchPerformaceTest.java
```diff
@@ -0,0 +1,488 @@
...
+                } else if (entry.getResource() instanceof Patient) {
+                    patient = (Patient) entry.getResource();
+                }
+            }
+        }
+        assertNotNull(observation);
+        assertTrue(FHIRUtil.hasTag(observation, subsettedTag));
+        // Verify that only the requested elements (and "mandatory elements") are
+        // present in the returned Observation.
+        Method[] observationMethods = Observation.class.getDeclaredMethods();
```
prb112: Is Reflection required here? ... I get why it's here, but it feels a little dirty. 
FHIRPath may be another option which would simplify. 

Also I get this is testCode.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchPerformaceTest.java
```diff
@@ -0,0 +1,488 @@
...
+    // this config is used to make sure all test users are created in one testng step.
+    private final int numOfPatientObservationsToCreate = 100;
+
+    /**
+     * Retrieve the server's conformance statement to determine the status of certain runtime options.
+     *
+     * @throws Exception
+     */
+    @Test
+    public void retrieveConfig() throws Exception {
```
prb112: I think we should remove the System.out.println... This will be printed out in other tests as well.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchPerformaceTest.java
albertwang-ibm: those codes were copied from SearchTTest and modified based on that

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchPerformaceTest.java
```diff
@@ -0,0 +1,488 @@
...
+        compartmentSearchSupported = isComparmentSearchSupported();
+        System.out.println("Compartment-based search supported?: "
+                + compartmentSearchSupported.toString());
+    }
+
+    @Test(groups = { "server-search-performance" }, invocationCount = 1)
+    public void testCreatePatientAndObservation() throws Exception {
+        WebTarget target = getWebTarget();
+        for (int i = 0; i < numOfPatientObservationsToCreate; i++) {
+            // Build a new Patient and then call the 'create' API.
```
prb112: Is this for the same patient? or separate patients?

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchPerformaceTest.java
prb112: OK - makes sense... I get it's test Code.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchPerformaceTest.java
```diff
@@ -0,0 +1,488 @@
...
+import com.ibm.fhir.model.resource.Bundle;
+import com.ibm.fhir.model.resource.Observation;
+import com.ibm.fhir.model.resource.Patient;
+import com.ibm.fhir.model.test.TestUtil;
+import com.ibm.fhir.model.type.Code;
+import com.ibm.fhir.model.type.Coding;
+import com.ibm.fhir.model.type.code.AdministrativeGender;
+import com.ibm.fhir.model.util.FHIRUtil;
+
+public class SearchPerformaceTest extends FHIRServerTestBase {
```
prb112: is this added to the testng.xml? `fhir-server-test//src/test/java/testng.xml` 

Should it be added?

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchPerformaceTest.java
lmsurpre: thx, i missed that.  then yes, lets remove invocationCount from here

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchPerformaceTest.java
albertwang-ibm: I don't think we should add this to pipeline, the default 100 is really too small, and use a big value can cause the test run hours to finish. so, I think we'd better run such tests local only...

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchPerformaceTest.java
albertwang-ibm: this will create different observations and patients with different ids, but actually only the last patient id will be used in some of the search tests which uses patient id.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchPerformaceTest.java
albertwang-ibm: good idea, let me remove it

albertwang-ibm: done

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchPerformaceTest.java
prb112: That's what I wanted to confirm.  Thanks!

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchPerformaceTest.java
prb112: great

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchPerformaceTest.java
albertwang-ibm: done

Issue #385, opened 2019-11-08
Fix or disable erroring tests and enable FHIRPathSpecTest in the build


lmsurpre:
In  #374 I introduced a new TestNG test class for invoking the spec-provided FHIRPath tests.

I added skips for the easy stuff (e.g. when we throw UnsupportedOperationException), but we still have about 166 test cases failing.

FHIR does not use all of FHIRPath and so getting to 100% passing is a non-goal for our 4.0.0 release.
However, there are definitely some things in here that need to get cleaned up.

[test-output.zip](https://github.com/IBM/FHIR/files/3825632/test-output.zip)

lmsurpre:
Are #387 or #388 started yet?
I checked out master and I see that our results are looking pretty good:
`Tests run: 679, Failures: 0, Skips: 206`

However, there are a few test cases commented out in `tests-fhir-r4.xml`.
I assume `tests-fhir-r4-old.xml` was the original test file and we've updated it based on what we support.
I recommend removing `tests-fhir-r4-old.xml` altogether (git has the history) and adding comments to the commented out test cases to indicate why they are commented out.
For example, I see a few with comments like `<!-- testWrong(patient(), "Encounter.name.given"); -->`...does that mean if it doesn't have this comment then the test is likely valid?

What does it mean when we remove the "expected output" (output element) from the test xml?  Does that just cause us to skip the test?

JohnTimm:
I didn't add `tests-fhir-r4-old.xml`. I don't know where that file came from but I'm happy to remove it.

JohnTimm:
I don't know the meaning of Grahame's comments.

senthilbak:
The skipped tests(skip=199, total=678) at this time are due to unsupported functionality that will come post R1.

PR #411, opened 2019-11-14
Update FHIRPath implementation to accommodate new grammar version

JohnTimm:
Presently, FHIR R4 points at a FHIRPath snapshot (v1.2.0) from September of 2018: http://hl7.org/fhirpath/2018Sep/index.html

This is the version we used to generate our original FHIRPath engine from.
However, the FHIRPath grammar has evolved and the other FHIRPath engines have stayed up-to-date with the changes, meaning we probably need to as well.

This PR updates our version of the FHIRPath grammar to the version shipped with http://build.fhir.org/ig/HL7/FHIRPath/branches/master/N1 and updates our logic around it in order to address issues related to epic #385

As part of the FHIRPath spec version upgrade, there were also changes to the FHIRPath data model and the logic used to handle FHIR.Quantity and System.Quantity nodes.


fhir-model/src/main/java/com/ibm/fhir/model/path/FHIRPathParser.java
```diff
@@ -1,59 +1,69 @@
...
+import org.antlr.v4.runtime.*;
+import org.antlr.v4.runtime.misc.*;
+import org.antlr.v4.runtime.tree.*;
+import java.util.List;
+import java.util.Iterator;
+import java.util.ArrayList;
 
-@SuppressWarnings({ "all", "warnings", "unchecked", "unused", "cast" })
+@SuppressWarnings({"all", "warnings", "unchecked", "unused", "cast"})
 public class FHIRPathParser extends Parser {
```
lmsurpre: would be nice if antlr generated our @Generated annotation....is it worth adding it?

fhir-validation/pom.xml
```diff
@@ -40,6 +40,12 @@
             <artifactId>fhir-examples</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>fhir-ig-us-core</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
```
lmsurpre: circular dependency, but i think its ok because its just test scope.  but still might be better to have us-core-specific tests under the us-core project.

fhir-validation/src/main/java/com/ibm/fhir/validation/util/ConstraintGenerator.java
```diff
@@ -39,6 +39,8 @@
  * A class used to generate FHIRPath expressions from a profile
  */
 public class ConstraintGenerator {
+    public static boolean DEBUG = false;
```
lmsurpre: can we use proper logging instead?

fhir-validation/src/test/java/com/ibm/fhir/validation/test/BloodPressureObservationTest.java
```diff
@@ -0,0 +1,112 @@
...
+import com.ibm.fhir.model.type.Meta;
+import com.ibm.fhir.model.type.Quantity;
+import com.ibm.fhir.model.type.Reference;
+import com.ibm.fhir.model.type.Uri;
+import com.ibm.fhir.model.type.code.ObservationStatus;
+import com.ibm.fhir.validation.FHIRValidator;
+import com.ibm.fhir.validation.util.ConstraintGenerator;
+
+public class BloodPressureObservationTest {
+    public static void main(String[] args) throws Exception {
```
lmsurpre: can it be a proper TestNG test?

fhir-model/src/main/java/com/ibm/fhir/model/path/FHIRPathBaseVisitor.java
```diff
@@ -1,498 +1,301 @@
-/*
- * (C) Copyright IBM Corp. 2019
- *
- * SPDX-License-Identifier: Apache-2.0
- */
-
+// Generated from FHIRPath.g4 by ANTLR 4.7.2
 package com.ibm.fhir.model.path;
```
lmsurpre: maybe add our `@Generated` annotation?

fhir-model/src/main/java/com/ibm/fhir/model/path/FHIRPathParser.java
JohnTimm: I'm going to leave the ANTLR generated code as is.

fhir-model/src/main/java/com/ibm/fhir/model/path/FHIRPathBaseVisitor.java
JohnTimm: I'm going to leave the ANTLR generated code as is.

fhir-validation/pom.xml
JohnTimm: I think this should be okay given the test scope. I'm using us-core IG to validate the validator / constraint generator. Once that work has been wrapped up for 4.0.0, I may consider moving it out into a different project.

fhir-validation/src/test/java/com/ibm/fhir/validation/test/BloodPressureObservationTest.java
JohnTimm: Converted it to a TestNG test.

fhir-validation/src/main/java/com/ibm/fhir/validation/util/ConstraintGenerator.java
JohnTimm: Changed the class to use java.util.logging instead of global DEBUG flag.

Issue #414, opened 2019-11-14
Enhance server integration tests to support both derby and db2


albertwang-ibm:
If change to use db2 as data store for fhir server, some server integration tests fail. we need to enhance those test codes to make sure them working well on both derby and db2.

PR #415, opened 2019-11-14
issue #414 enhance server integration tests to support both derby and db2

albertwang-ibm:


fhir-server-test/src/test/java/com/ibm/fhir/server/test/SortingTest.java
```diff
@@ -279,7 +279,7 @@ public void testCreateObservation5() throws Exception {
     // Patient?gender=male&_sort:asc=family
     @Test(groups = { "server-search" }, dependsOnMethods = { "testCreatePatient1",
             "testCreatePatient2", "testCreatePatient3", "testCreatePatient4",
-            "testCreatePatient5" })
+    "testCreatePatient5" })
```
lmsurpre: why this bad format change?

albertwang-ibm: this is by my eclipse auto actions during saving, I would think this is actually the correct format, because ")" at the end is the most top level.

lmsurpre: why would line 281 be indented and not 282?

lmsurpre: and this is why i don't like auto-formatting...

fhir-server-test/src/test/java/com/ibm/fhir/server/test/R4ExampleServerTest.java
```diff
@@ -16,21 +16,21 @@
...
-        
+
         // Process each of the examples using the provided ExampleRequestProcessor. We want to
         // validate first before we try and send to FHIR
         final R4ExamplesDriver driver = new R4ExamplesDriver();
         driver.setMetrics(dm);
         driver.setValidator(new ValidationProcessor());
-        driver.setProcessor(new ExampleRequestProcessor(this, null, dm, 1));
-        
+        driver.setProcessor(new ExampleRequestProcessor(this, "default", dm, 1));
```
lmsurpre: remind me again what this is for and why we need to pass default now, when null worked before?

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchTest.java
```diff
@@ -154,52 +167,52 @@ public void testCreatePersonWithManyIdentifiers() throws Exception {
...
+    //    // Uncomment in order to test the limits when using the 'normalized' model with
+    //    // Db2
+    //    @Test(groups = { "server-search" })
+    //    public void testCreatePersonWithTooManyLinks() throws Exception {
+    //        // Build a new Person with a single identifier and then call call 'create-on-update'.
+    //        // This will ensure that the search parameters are added to the cache before we insert one with a thousand of
+    //        //them.
+    //        String identifier = UUID.randomUUID().toString();
+    //        String patientRef = "Patient/" + identifier;
+    //        Person person = createPersonWithIdentifierAndLink(identifier, patientRef);
```
lmsurpre: Not a fan of the double-commenting.
We discussed that these limits are no longer valid and so I think the tests should be remove.
If we add limits like these back in later, we can introduce the tests then.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchTest.java
```diff
@@ -154,52 +167,52 @@ public void testCreatePersonWithManyIdentifiers() throws Exception {
...
-                      .build();
+                .id(Id.of(id))
+                .identifier(Identifier.builder()
+                        .value(string(identifier))
+                        .system(uri("test"))
+                        .build())
+                .link(Link.builder()
+                        .target(Reference.builder().reference(string(patientRef)).build())
+                        .build())
+                .build();
```
lmsurpre: I dislike this formatting change too

fhir-server-test/src/test/java/com/ibm/fhir/server/test/R4ExampleServerTest.java
albertwang-ibm: null didn't work for db2 test, I believe we had never test it on db2 before with null (at least after Robin's change) for db2 multi-tenant support, and I believe that's caused by that currently we need a db2 tenant key for  for each tenant in the database(s), and there is no tenant key for a null tenant.  the mapping is done in Robin's tenant management table in each database...

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchTest.java
albertwang-ibm: sounds good, will remove these 2 already commented out tests.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SortingTest.java
albertwang-ibm: 281 is for "}", but any way, I can make it better

albertwang-ibm: done

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchTest.java
albertwang-ibm: done

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchTest.java
albertwang-ibm: done

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchTest.java
```diff
@@ -228,7 +203,8 @@ private Person addIdentifiers(Person person, int numberOfIdentifiers) {
         return personBuilder.build();
     }
 
-    @Test(groups = { "server-search" }, dependsOnMethods = { "testCreatePatient" })
+    @Test(groups = { "server-search" }, dependsOnMethods = {
+    "testCreatePatient" })
```
lmsurpre: still don't like this formatting change

albertwang-ibm: ha, good catch. I have just fixed this and also removed the auto save action which has caused this. thanks!

albertwang-ibm:
fixed the formatting issue that Lee mentioned and pipeline test passed, so merge it now

Issue #410, opened 2019-11-14
value-quantity search and sort don't work as expected


albertwang-ibm:
Create several observations , some of them with value-quantity(125.0mmHg), and some of them without value-quantity, the following search tests got wrong results: 
(1) (verified that this was actually fixed in PR369) search observations equal to any value, less than any value or greater than any value always returns all the observations(with or without value-quantity). e.g: 
        queryParam("value-quantity", "25.0||mmHg")
        queryParam("value-quantity", "le12.0||mmHg")
        queryParam("value-quantity", "gt823.0||mmHg")
 (2) sorting by value-quantity doesn't work, the returned entries are not sorted at all
       queryParam("_sort:asc", "component-value-quantity")

lmsurpre:
Where are the units?  Are they coded units (code field) or display units (unit field)?

lmsurpre:
see https://gforge.hl7.org/gf/project/fhir/tracker/?action=TrackerItemEdit&tracker_item_id=19597 for why that matters.  currently we only index the unit field if the coded units are missing.

lmsurpre:
I chatted with Albert on this one and I think its a duplicate of https://github.com/IBM/FHIR/issues/370 ...albert is going to sync with master to confirm.

albertwang-ibm:
after sync with master, update the server config and test again, the first issue is gone, but the 1nd issue is not fixed.

lmsurpre:
I split the sort issue into #418.

Also, we merged pull request #416 to  address the potential issue wrt https://gforge.hl7.org/gf/project/fhir/tracker/?action=TrackerItemEdit&tracker_item_id=19597.

Based on those, I'm moving this to under review...please re-open if you think we need any further changes on this one.

Issue #257, opened 2019-10-15
Number and Quantity searches do not consider implicit ranges / significant digits


prb112:
Numeric searches may not consider implicit ranges / significant digits. The FHIR user schema uses doubles. 

The FHIR Specification uses ranges as described in https://hl7.org/fhir/search.html#prefix:
> The range interpretation is provided for decimals and dates. Searches are always performed on values that are implicitly or explicitly a range. For instance, the number 2.0 has an implicit range of 1.95 to 2.05, and the date 2015-08-12 has an implicit range of the all the time during that day. If the target value is a Range, a Period, or a Timing, then the target is explicitly a range.

Notes from the spec: 
> a number search is used against a resource element that stores a simple integer (e.g. ImmunizationRecommendation.recommendation.doseNumber), and the search parameter is not expressed using the exponential forms, and does not include any non-zero digits after a decimal point, the significance issues cancel out and searching is based on exact matches. Note that if there are non-zero digits after a decimal point, there cannot be any matches
> When a comparison prefix in the set lgt, lt, ge, le, sa & eb is provided, the implicit precision of the number is ignored, and they are treated as if they have arbitrarily high precision
> The way search parameters operate in resources is not the same as whether two numbers are equal to each other in a mathematical sense
> Searching on decimals involves an implicit range. The number of significant digits of the implicit range is the number of digits specified in the search parameter value, excluding leading zeros. So 100 and 1.00e2 both have the same number of significant digits - three
> sa (starts-after) and eb (ends-before) are not used with integer values but are used for decimals. 

So its clear that the search parameter value should be interpreted as a range, unless:
- A. the target of the search is a simple integer type and the search parameter is given as a whole number with no decimal point;
    - For A, I assume that the data type of the element being search must be integer, so it doesn't apply for decimals that just happen to have whole number values (including the "count" specialization?).
- B. a prefix of gt, lt, ge, le, sa, or eb is used

The following questions for the Investigation are:
- Should a decimal target value be interpreted as a range or not? Theory is: Yes, a target decimal should be handled like a range. Based on: 
    > Searches are always performed on values that are implicitly or explicitly a range.
- leads to some interesting results if both the search value and the target value are interpreted as ranges. For example, a search for value=5 would NOT match any values in [4.5,5.5) except for exactly 5 because in a range search, "eq" is interpreted to mean that the target range is fully contained within the search range.

Note that this would require us to handle integers different from decimals.

[...]

PR #416, opened 2019-11-14
issue #410 - index the display unit field even when a coded unit is present

lmsurpre:
see https://chat.fhir.org/#narrow/stream/179166-implementers/topic/Quantity.20search.20and.20display.20units/near/180734154

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
@@ -61,9 +61,10 @@ public class JDBCParameterBuildingVisitor extends DefaultVisitor {
     public static final String EXCEPTION_MSG_NAME_ONLY = "Unexpected error while processing parameter [%s]";
 
     // Datetime Limits from
-    // DB2: https://www.ibm.com/support/knowledgecenter/en/SSEPGG_10.5.0/com.ibm.db2.luw.sql.ref.doc/doc/r0001029.html
+    // DB2: https://www.ibm.com/support/knowledgecenter/en/SSEPGG_11.5.0/com.ibm.db2.luw.sql.ref.doc/doc/r0001029.html
     // Derby: https://db.apache.org/derby/docs/10.0/manuals/reference/sqlj271.html
     private static final Timestamp SMALLEST_TIMESTAMP = Timestamp.valueOf("0001-01-01 00:00:00.000000");
+    // 23:59:59.999999 used instead of 24:00:00.000000 to ensure it could be represented in FHIR if needed
     private static final Timestamp LARGEST_TIMESTAMP = Timestamp.valueOf("9999-12-31 23:59:59.999999");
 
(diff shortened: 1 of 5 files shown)


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
```diff
@@ -452,21 +453,31 @@ public boolean visit(java.lang.String elementName, int elementIndex, Quantity qu
         if (!QUANTITY.equals(searchParamType)) {
             throw invalidComboException(searchParamType, quantity);
         }
-        if (quantity != null && quantity.getValue() != null && quantity.getValue().getValue() != null &&
-                (quantity.getCode() != null || quantity.getUnit() != null)) {
-            Parameter p = new Parameter();
-            p.setName(searchParamCode);
-            p.setValueNumber(quantity.getValue().getValue());
+        if (quantity.getValue() != null && quantity.getValue().hasValue()) {
```
albertwang-ibm: so, don't have to do null check for quantity any more?

fhir-persistence/src/test/java/com/ibm/fhir/persistence/search/test/AbstractSearchQuantityTest.java
```diff
@@ -37,14 +37,10 @@ protected void setTenant() throws Exception {
...
-        // DSTU2 does not say if this is allowed or not, but we do not support it.
-        // In more recent versions, they clarified that it should work:  https://build.fhir.org/search.html#quantity
-//        assertSearchReturnsSavedResource("Quantity", "25");
-        
-        // I think this should return the resource but it currently doesn't.
         // https://gforge.hl7.org/gf/project/fhir/tracker/?action=TrackerItemEdit&tracker_item_id=19597
-//        assertSearchReturnsSavedResource("Quantity", "25||sec");
+        assertSearchReturnsSavedResource("Quantity", "25||sec");
         
         assertSearchDoesntReturnSavedResource("Quantity", "24.4999||s");
```
albertwang-ibm: according to the spec https://www.hl7.org/fhir/search.html,  GET [base]/Observation?value-quantity=5.4||mg  means "Search for all the observations with a value of 5.4(+/-0.05) mg where the unit - either the code (code) or the stated human unit (unit) are "mg",   do we do the +/- value range check also?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
lmsurpre: right, this makes it consistent with the other visit methods...they all assume a non-null argument and this should always be the case from FHIRPersistenceJDBCImpl due to the double-dispatch logic

fhir-persistence/src/test/java/com/ibm/fhir/persistence/search/test/AbstractSearchQuantityTest.java
lmsurpre: No, that is https://github.com/IBM/FHIR/issues/257

Issue #418, opened 2019-11-15
Bad assumption in SQL generated by query with sort parameters


lmsurpre:
**Describe the bug**
Albert found that the sort tests are not working for him.  I wasn't able to reproduce the issue on Derby, but after looking at the SQL being generated, I think we have an issue.

The search query is currently getting split into two parts:
1. searchForId which uses all the search parameters to select the resources and sorts the list of resource ids by the sort parameters
```sql
SELECT R.RESOURCE_ID,MIN(S1.QUANTITY_VALUE) FROM Observation_RESOURCES R 
	JOIN Observation_LOGICAL_RESOURCES LR ON R.LOGICAL_RESOURCE_ID=LR.LOGICAL_RESOURCE_ID AND R.RESOURCE_ID = LR.CURRENT_RESOURCE_ID   
	LEFT OUTER JOIN Observation_QUANTITY_VALUES S1 ON (S1.PARAMETER_NAME_ID=401 AND S1.LOGICAL_RESOURCE_ID = R.LOGICAL_RESOURCE_ID) 
	WHERE R.IS_DELETED <> 'Y' AND R.LOGICAL_RESOURCE_ID IN 
(SELECT LOGICAL_RESOURCE_ID FROM Observation_TOKEN_VALUES  
	WHERE (PARAMETER_NAME_ID=216 AND ((TOKEN_VALUE = ?)))) AND ...)
GROUP BY R.RESOURCE_ID 
ORDER BY MIN(S1.QUANTITY_VALUE) asc 
NULLS LAST OFFSET 0 ROWS FETCH NEXT 50 ROWS ONLY
```
2. searchById which returns the actual resources
```sql
SELECT R.RESOURCE_ID, R.LOGICAL_RESOURCE_ID, R.VERSION_ID, R.LAST_UPDATED, R.IS_DELETED, R.DATA, LR.LOGICAL_ID FROM Observation_RESOURCES R, Observation_LOGICAL_RESOURCES LR WHERE R.LOGICAL_RESOURCE_ID = LR.LOGICAL_RESOURCE_ID AND R.RESOURCE_ID IN (1727,1784,1774,...)] 
```

Derby seems to always return the resources in the same order provided in that IN clause, but in general we can't rely on that.

**To Reproduce**
Only reproducable in Db2?

**Expected behavior**
The sort query should be written in a way to ensure the results are always in the proper order.

**Additional context**
[...]

PR #419, opened 2019-11-15
issue #418 - add order by clause to searchByIds to ensure sorted results

lmsurpre:
1. Updated the SearchPerformanceTest so that the observations have
components with values...just to easily create many observations which
helps verify the search behavior.  This change could be removed if we
want.

2. Deleted unused variants of searchForIds and searchByIds. I think
these came from the original "basic" schema that was removed for
https://github.com/IBM/FHIR/issues/93 but they were no longer in use and
should be safe to delete.

3. Introduced ORDER BY CASE clause to ResourceDAOImpl.searchByIds.
I found that DERBY is always returning the results in the same order as
the resource ids provided in the "IN" clause, but in general we
shouldn't rely on that.  I tested the CASE statements on derby with up
to 1000 results, but we need to run a similar test on Db2 to ensure it
works there as well.


albertwang-ibm:
still a lit bit surprised to see that we search and find the ids first and then search those ids to get the results, just curious, why don't we get the full results in the first search.

lmsurpre:
@JohnTimm can you address Albert's comment?  Pretty sure it was like this before I got in here, but not sure on the reasoning.  Does it help to leave out the `Data` blob column while doing the filtering?

Issue #396, opened 2019-11-11
derby deaklock if read/create fhir resources in multiple thread simultaneously


albertwang-ibm:
This issue happened during jdbc performance test for Derby Network Server, and because the performance test for embedded Derby server only use a single shared connection and each thread never commit the changes, so it works fine there. and there is no such issue for DB2.

The root cause of this issue is the missing indexes(for CURRENT_RESOURCE_ID and LOGICAL_ID) in all the XXX_LOGICAL_RESOURCES tables, and these 2 columns are widely used in the where clause of the select sql for search of all fhir resources. If there is any column used in where clause and not indexed, derby will lock the whole table instead of the required row(s) only, this has caused the dead lock issue we saw in the test.

The fix is to add the missing indexes for the tables.

```
Nov 11, 2019 1:40:30 PM com.ibm.fhir.model.spec.test.R4ExamplesDriver processIndex
INFO: Processing index file 'json/performance.txt' with format 'JSON'
..Nov 11, 2019 1:40:32 PM com.ibm.fhir.database.utils.common.JdbcConnectionProvider getConnection
INFO: Opening connection to database: jdbc:derby://localhost:1527/fhirdb
Nov 11, 2019 1:40:32 PM com.ibm.fhir.database.utils.common.JdbcConnectionProvider getConnection
INFO: Opening connection to database: jdbc:derby://localhost:1527/fhirdb
Nov 11, 2019 1:40:53 PM com.ibm.fhir.persistence.jdbc.dao.impl.FHIRDbDAOImpl severe
SEVERE: Failure retrieving FHIR objects. SQL=SELECT R.RESOURCE_ID, R.LOGICAL_RESOURCE_ID, R.VERSION_ID, R.LAST_UPDATED, R.IS_DELETED, R.DATA, LR.LOGICAL_ID FROM Account_RESOURCES R, Account_LOGICAL_RESOURCES LR WHERE LR.LOGICAL_ID = ? AND R.RESOURCE_ID = LR.CURRENT_RESOURCE_ID  searchArgs=[2159fec5-8b38-4908-ba0f-6b79c22258f5] [probeId=c0-a8-1-be-cb9f757a-81d0-40aa-b5bd-02f8caf7fc81]
java.sql.SQLTransactionRollbackException: A lock could not be obtained due to a deadlock, cycle of locks and waiters is:
Lock : ROW, ACCOUNT_LOGICAL_RESOURCES, (1,25)
[...]

PR #423, opened 2019-11-18
issue #396 fix for derby dead lock issue

albertwang-ibm:
Added the missing indexes to avoid the derby dead look issue.
After adding the indexes, the deadlock issue is gone in jdbc derby network performance test.


fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirResourceGroup.java
```diff
@@ -159,6 +159,11 @@ public void addLogicalResources(List<IDatabaseObject> group, String prefix) {
                 .setTablespace(fhirTablespace)
                 .addPrivileges(resourceTablePrivileges)
                 .enableAccessControl(this.sessionVariable)
+                // Add indexes to avoid dead lock issue of derby
+                // Derby requires all columns used in where clause to be indexed, otherwise whole table lock will be
+                // used instead of row lock, which can cause dead lock issue frequently during concurrent accesses.
+                .addIndex(IDX + tableName + CURRENT_RESOURCE_ID, CURRENT_RESOURCE_ID)
```
prb112: Should this be a conditional deployment of the indices?  
E.g. use a translator / adapter to identify the index as derby only. 

Also, for the query, what query failed?

prb112: these are not for re-dev, just discussion here to get on same page.

albertwang-ibm: I thought about this(making the change for derby only) also, but I believe the added new indexes should be able to improve performance for db2 also, because these 2 columns are really heavily used in sql of almost all resource search scenarios, so hopefully we can benefit from indexing them... 

The dead-locked sql is "SELECT R.RESOURCE_ID, R.LOGICAL_RESOURCE_ID, R.VERSION_ID, R.LAST_UPDATED, R.IS_DELETED, R.DATA, LR.LOGICAL_ID FROM Observation_RESOURCES R, Observation_LOGICAL_RESOURCES LR WHERE LR.LOGICAL_ID = ? AND R.RESOURCE_ID = LR.CURRENT_RESOURCE_ID"

prb112: Mind updating 162, to say also improve db2.  It'll keep someone from coming in and changing removing if derby goes away.

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirResourceGroup.java
```diff
@@ -159,6 +159,11 @@ public void addLogicalResources(List<IDatabaseObject> group, String prefix) {
                 .setTablespace(fhirTablespace)
                 .addPrivileges(resourceTablePrivileges)
                 .enableAccessControl(this.sessionVariable)
+                // Add indexes to avoid dead lock issue of derby
```
prb112:
```suggestion
                // Add indexes to avoid dead lock issue of derby, and improve Db2 performance
```

Issue #255, opened 2019-10-15
_id and _lastUpdated should not be stored as pivoted name-value pair values


prb112:
The _id and _lastUpdated attributes are stored as first-class columns in the <resourceType>_logical_resources and <resourceType>_resources tables. Therefore, we should not store these attributes in the name-value pair pivoted _values tables, and search queries should be written to use the columns directly, which is more efficient (and logical).

**Steps**
- Investigate Schema Status
- Investigate Persistence Layer changes and Search changes
- Implement change to support and use _id and _lastUpdated uniquely 
- Confirm common ids don't collide on the the Search (e.g. Practioner/1 doesn't  collide with Patient/1)

**Related to**:  
- Global Search Performance

**Original Issue**:  `_id and _lastUpdated should not be stored as pivoted name-value pair values #324`

prb112:
PR https://github.com/IBM/FHIR/pull/506

Working on improving the AND/OR conditioning for `_lastUpdated`

PR #430, opened 2019-11-19
WIP:  issue #255 - start handling even more datetimes as ranges

lmsurpre:
on call with Paul and John, we identified about 7 other related things
that need to get done, but hopefully these changes are a step in the right direcation

TODO:
* remove DATE column and always use DATE_START and DATE_END (even for instants)
* handle date search parameter query params as instants when prefixed by lt, gt, le, or ge
* handle date search parameter query params as implicit ranges in all other cases (when no fractional seconds are passed)

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
@@ -48,11 +48,8 @@ import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
 import com.ibm.fhir.model.resource.Location;
-import com.ibm.fhir.model.resource.OperationOutcome.Issue;
 import com.ibm.fhir.model.resource.SearchParameter;
 import com.ibm.fhir.model.type.Code;
-import com.ibm.fhir.model.type.code.IssueSeverity;
-import com.ibm.fhir.model.type.code.IssueType;
 import com.ibm.fhir.model.util.ModelSupport;
 import com.ibm.fhir.persistence.exception.FHIRPersistenceException;
 import com.ibm.fhir.persistence.exception.FHIRPersistenceNotSupportedException;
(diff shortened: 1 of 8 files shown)


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -794,62 +774,36 @@ private SqlQueryData processDateParm(Class<?> resourceType, Parameter queryParm,
...
+            // If the search parameter value includes fractional seconds and the target value is an instant, then use equals
+            if (!value.getValueDate().isPartial() && QueryBuilderUtil.hasSubSeconds(value.getValueDate().getValue()) && isInstantSearch) {
+                whereClauseSegment.append(LEFT_PAREN);
+                whereClauseSegment.append(tableAlias + DOT).append(DATE_VALUE)
+                                  .append(operator.value())
+                                  .append(BIND_VAR);
+                bindVariables.add(Timestamp.from(datetime));
+                whereClauseSegment.append(RIGHT_PAREN);
+                whereClauseSegment.append(RIGHT_PAREN);
             } else {
```
lmsurpre: i think this whole logic branch can get deleted if we add that "inclusive" boolean into the db and handle all datetime searches (except `_lastUpdated`) through DATE_START and DATE_END

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QueryBuilderUtil.java
```diff
@@ -149,6 +155,20 @@ private static ZonedDateTime addOneMinuteOrSecond(ZonedDateTime dateTime) {
         }
         return dateTime;
     }
+    private static ZonedDateTime minusOneMicroSecond(ZonedDateTime dateTime) {
+        return dateTime.minusNanos(1000);
+    }
```
lmsurpre: this goes away if we can either normalize the end times to always be inclusive or exclusive OR make schema edits to know the difference.

lmsurpre: on our call today, we decided to attempt the latter.

prb112:
I believe I've captured all the changes in the PR in PR 499
Leaving open for Lee to confirm

PR #439, opened 2019-11-22
Issue 438

prb112:


JohnTimm:
Did you add any integration tests to exercise _id search parameter? Also, would it make sense to do some performance loads on this as well?

prb112:
> Are you planning to add any integration tests to exercise _id search parameter? Also, would it make sense to do some performance loads on this as well?

I think there are some integration tests for _id, I'll confirm and check
I'd like to do a performance load as well on it.

<< checking and waiting to merge>>

PR #441, opened 2019-11-24
more swagger tweaks and update OpenApiGenerator to match

lmsurpre:
1. Updated FHIROpenApiGenerator to match the recent changes to
FHIRSwaggerGenerator (e.g. one file per resource type)

2. utilize the Liberty mpOpenAPI-1.0 feature to validate the updated
OpenAPI definition; this helped to find and a number of issues that were
applicable to both the OpenAPI and the Swagger 2.0 definitions, includ
ing:

* Problem:  Complex datatypes and wildcard choice elements are
referencing non-existent primitive elements.
    * Solution:  add isApplicableForClass guard to the loops for adding
the choice type properties and the data type and definitions...this
simplifies the schema by skipping both special types like Dosage and
ElementDefinition and metadata types like ContactDetail, Contributor,
DataRequirement, and ParameterDefinition (unless they are explicitly
referenced from the resource type being generated). Although technically
these types are allowed values for the Extension.value[x] element that
is always applicable, they are almost never desired here.

* Problem:  String definitions for Element.id and Extension.url are not
properly inlined.
    * Found and fixed the issue; I had accidentally left a conditional
clause (`&& !"id".equals(elementName) && !"url".equals(elementName)`)
which needed to be removed for proper functioning.

* Problem:  Date and DateTime are not properly inlined; instead we have
stale references*
    * Found and fixed the issue; the logic for these 2 primitives (and
only these primitives) was using `modelClass` where it should have used
`fieldClass`.

3. To appease the Liberty OpenAPI validation, path parameters `id` and
`vid` are now defined inline in the OpenAPI. I did *not* make the
corresponding change to the Swagger 2.0 generator, but its possible to
make this change there as well if needed.

### DIFF
--- a/fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
+++ b/fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
@@ -7,6 +7,7 @@
 package com.ibm.fhir.openapi.generator;
 
 import java.io.File;
+import java.io.FileWriter;
 import java.io.InputStream;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
@@ -14,10 +15,6 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
-import java.math.BigDecimal;
-import java.time.LocalTime;
-import java.time.ZonedDateTime;
-import java.time.temporal.TemporalAccessor;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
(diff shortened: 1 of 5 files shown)


fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
```diff
@@ -63,112 +116,234 @@
...
+        info.add("description", "The capabilities interaction retrieves the information about a server's capabilities - which portions of the HL7 FHIR specification it supports.");
         info.add("version", "4.0.0");
         swagger.add("info", info);
 
-        /**
-         * "servers": [ { "url": "/fhir-server/api/v4" } ]
-         */
-        JsonArrayBuilder servers = factory.createArrayBuilder();
-        JsonObjectBuilder server = factory.createObjectBuilder();
-        server.add("url", "/fhir-server/api/v4");
```
albertwang-ibm: should we hard coded or get from server.xml?  I think maybe we should allow to configure/change rootcontext in server.xml.

fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
```diff
@@ -63,112 +116,234 @@
...
-
-        swagger.add("components", components);
-
+        
+        // generate definition for all inner classes inside the top level resources.
+        for (String innerClassName : getAllResourceInnerClasses()) {
+            Class<?> parentClass = Class.forName(RESOURCEPACKAGENAME + "." + innerClassName.split("\\$")[0]);
+            if (CapabilityStatement.class.equals(parentClass)) {
+                Class<?> innerModelClass = Class.forName(RESOURCEPACKAGENAME + "." + innerClassName);
+                // System.out.println("Resource:  " + className);
```
albertwang-ibm: can we remove the commented out System.out.println? there are a few in this file.

fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
```diff
@@ -185,6 +360,7 @@ public static void main(String[] args) throws Exception {
     private static void populateStructureDefinitionMap(Map<Class<?>, StructureDefinition> structureDefinitionMap,
             String structureDefinitionFile) throws Exception {
         InputStream stream = FHIROpenApiGenerator.class.getClassLoader().getResourceAsStream(structureDefinitionFile);
```
albertwang-ibm: should we close the inputStream?

fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
lmsurpre: its still hard-coded, but i moved it to a constant at the top of the file and added a comment

fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
lmsurpre: done

fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
lmsurpre: done

Issue #321, opened 2019-10-28
fhir-audit uses jackson databind


prb112:
fhir-audit uses jackson databind

The project's style is to use the java dependencies from jakarta-ee. 
The jackson pattern needs to be refactored and changed to use the default pattern.

PR #442, opened 2019-11-25
change out data-bind for jakarta libraries #321

prb112:
- Remove jackson databind libraries
- Add json library to pom.xml and classpath
- Add Parser, Writer, Builder for EventStreamsCredentials and update
Environment
- Added Parser, Writer, Builder to CadfReason, CadfMapItem,
CadfGeolocation, CadfEndpoint, CadfCredential, CadfAttachment,
CadfReporterStep
- Clean up copyrights and formatting
- Removed CadfParser and followed the fhir-model pattern (inside the
Model classes)
- Highlight in comment that the DisabledAuditLogService has No
Implementation
- breakout the enumerations into a separate set of files/package
- Update Beans - ApiParameters, Batch, ConfigData, Data, Context
- Cleanup formatting in AuditLogEntry
- Update corresponding tests
- Added CadfMetric and CadfMeasurement as split from CadfEvent
- Update the RestAuditLogger to meet the new format of the Builders and
parsers
- Add tests and add audit test utility
- Update testng.xml


prb112:
> The code changes looks OK for me, just 2 questions to discuss:
> (1) Should we consider to use something like jakarta json bingding (http://json-b.net) to convert java objects to/from json messages, not really sure if it's the good approach for us to do the job all by ourselves.
> (2) Have theses changes been tested using eventstream and ibm sql query service in cloud? we have guidance in the user guide document about how to query the logs using sql query service, per my understanding, the log format is not changed, so we don't need to update the guide, Is this correct?

1 - Per discussion. I did consider using json binding, and I knew this method would work quickly. 
2 - I have not tested with eventstream and ibm sql query service in cloud. (it's an internal change and should not require an update to the guide).

Issue #256, opened 2019-10-15
Updates to search sort for R4


lmsurpre:
In DSTU2, we used the `:asc` and `:desc` modifiers for controlling sort order.
In R4, they've replaced these with the `-` prefix so that you can now control use multiple sorts on a single parameter.

from https://www.hl7.org/fhir/search.html#sort
```
The client can indicate which order to return the results by using the parameter _sort, which can contain a comma-separated list of sort rules in priority order:

GET [base]/Observation?_sort=status,-date,category

Each item in the comma separated list is a search parameter, optionally with a '-' prefix. The prefix indicates decreasing order; in its absence, the parameter is applied in increasing order. 
```

lmsurpre:
see AbstractQuerySortTest

prb112:
PR https://github.com/IBM/FHIR/pull/443

PR #443, opened 2019-11-26
Updates to search sort for R4 #256

prb112:
- Break the Sort behavior out of SearchUtil
- Implement the sort package with Sort and Sort.Direction
- Change direction indicator to -, absence and + for internal
representation
- Update signature in SearchUtil and Sort
- Update the SortParameter to support the new toString pattern
- Add coverage tests
- Updated the UriBuilder and added a SortUriTest to test the behavior
- Update the Server Tests with ascending and descending


fhir-search/src/test/java/com/ibm/fhir/search/sort/test/SortParameterParseTest.java
```diff
@@ -83,76 +116,71 @@ public void testUnknownSortParm_strict() throws Exception {
...
         searchContext = SearchUtil.parseQueryParameters(resourceType, queryParameters);
         assertNotNull(searchContext);
 
         // Do sort parameter validation
         assertNotNull(searchContext.getSortParameters());
         assertEquals(1, searchContext.getSortParameters().size());
         SortParameter sortParm = searchContext.getSortParameters().get(0);
-        assertEquals(sortParmName, sortParm.getName());
-        assertEquals(direction, sortParm.getDirection());
+        assertEquals(sortParmName, sortParm.getCode());
```
albertwang-ibm: (1) not sure why change Name to Code for search parameter.  
(2) seems in the test codes, Name is still used, e.g, sortParmName instead of sortParmCode.

fhir-search/src/main/java/com/ibm/fhir/search/uri/UriBuilder.java
```diff
@@ -75,20 +72,22 @@ public UriBuilder requestUri(String requestUriString) {
...
         queryString.append(SearchConstants.EQUALS_CHAR);
         queryString.append(context.getPageSize());
-        queryString.append(SearchConstants.AND_CHAR);
 
         // the URLs is converted from a for loop to a functional paradigm. The functional
         // paradigm is consistent with JOINING multiple strings
-        Function<Parameter, String> serializeSearchParmToQueryString = p -> serializeSearchParmToQueryString(p);
-        queryString.append(context.getSearchParameters().stream().map(serializeSearchParmToQueryString).collect(Collectors.joining(SearchConstants.AND_CHAR_STR)));
-
+        if (!context.getSearchParameters().isEmpty()) {
```
albertwang-ibm: can context.getSearchParameters() be empty at this point? but anyway, I like this added check.

fhir-search/src/test/java/com/ibm/fhir/search/sort/test/SortParameterParseTest.java
prb112: it should always be code, and not name... I'll change to code in these cases

fhir-search/src/main/java/com/ibm/fhir/search/uri/UriBuilder.java
prb112: It should never be null. (I tested )

fhir-search/src/test/java/com/ibm/fhir/search/sort/test/SortParameterParseTest.java
prb112: fixed in the next commit... (2) ... (1)

Issue #389, opened 2019-11-08
FHIRPathEvaluator throws NullPointerException for certain expressions


lmsurpre:
**Describe the bug**
```
com.ibm.fhir.model.path.exception.FHIRPathException: java.lang.NullPointerException  [probeId=7f-0-0-1-5eba9c55-c4d5-4fb4-a889-78ba981f15ca]
	at com.ibm.fhir.model.path.evaluator.FHIRPathEvaluator.evaluate(FHIRPathEvaluator.java:109)
	at com.ibm.fhir.model.path.evaluator.FHIRPathEvaluator.evaluate(FHIRPathEvaluator.java:99)
	at com.ibm.fhir.model.path.evaluator.FHIRPathEvaluator.evaluate(FHIRPathEvaluator.java:95)
	at com.ibm.fhir.model.path.test.FHIRPathSpecTest.executeTest(FHIRPathSpecTest.java:79)
	at sun.reflect.GeneratedMethodAccessor4.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:86)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:643)
	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:820)
	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1128)
	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:129)
	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:112)
	at org.testng.TestRunner.privateRun(TestRunner.java:782)
	at org.testng.TestRunner.run(TestRunner.java:632)
	at org.testng.SuiteRunner.runTest(SuiteRunner.java:366)
	at org.testng.SuiteRunner.runSequentially(SuiteRunner.java:361)
	at org.testng.SuiteRunner.privateRun(SuiteRunner.java:319)
	at org.testng.SuiteRunner.run(SuiteRunner.java:268)
	at org.testng.SuiteRunnerWorker.runSuite(SuiteRunnerWorker.java:52)
	at org.testng.SuiteRunnerWorker.run(SuiteRunnerWorker.java:86)
[...]

PR #446, opened 2019-11-27
Issue #389 - updated FHIRPath engine per spec tests

JohnTimm:


fhir-model/src/main/java/com/ibm/fhir/model/path/FHIRPathDateTimeValue.java
```diff
@@ -220,12 +255,16 @@ public String toString() {
...
-    public static void main(String[] args) {
-        System.out.println(LocalDate.from(Year.now()));
-    }
-
     @Override
     public void accept(FHIRPathNodeVisitor visitor) {
         visitor.visit(this);
     }
+
+    public static void main(String[] args) {
```
prb112: Move to Unit Test?

fhir-model/src/main/java/com/ibm/fhir/model/path/FHIRPathDateValue.java
```diff
@@ -194,7 +201,7 @@ public String toString() {
     }
     
     public static void main(String[] args) {
-        System.out.println(LocalDate.from(Year.now()));
+        System.out.println(Year.from(LocalDate.now()));
```
prb112: Doesn't seem like a test? or test code?

fhir-model/src/main/java/com/ibm/fhir/model/path/FHIRPathQuantityNode.java
```diff
@@ -168,24 +172,59 @@ public int compareTo(FHIRPathNode other) {
...
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (!(obj instanceof FHIRPathNode)) {
+            return false;
+        }
+        FHIRPathNode other = (FHIRPathNode) obj;
+        if (!isComparableTo(other)) {
```
prb112: Block 207 on repeats in `isComparableTo`

fhir-model/src/main/java/com/ibm/fhir/model/path/FHIRPathQuantityValue.java
```diff
@@ -132,24 +132,25 @@ public FHIRPathQuantityValue subtract(FHIRPathQuantityValue quantityValue) {
...
         }
-        if (other instanceof FHIRPathQuantityValue) {
-            return value.compareTo(((FHIRPathQuantityValue) other).value());
+        if (other instanceof FHIRPathQuantityValue || other.getValue() instanceof FHIRPathQuantityValue) {
+            FHIRPathQuantityValue quantityValue = (other instanceof FHIRPathQuantityValue) ? (FHIRPathQuantityValue) other : (FHIRPathQuantityValue) other.getValue();
+            return value.compareTo(quantityValue.value());
         }
-        return value.compareTo(((FHIRPathQuantityValue) other.getValue()).value());
+        FHIRPathNumberValue numberValue = (other instanceof FHIRPathNumberValue) ? (FHIRPathNumberValue) other : (FHIRPathNumberValue) other.getValue();
+        return value.compareTo(numberValue.decimal());
```
prb112: hum...  there seems to be similar conversions and checks in the blocks

fhir-model/src/main/java/com/ibm/fhir/model/path/FHIRPathType.java
```diff
@@ -347,6 +348,20 @@
...
+        javaTypeMap.put(java.lang.Boolean.class, SYSTEM_BOOLEAN);
+        javaTypeMap.put(java.lang.Integer.class, SYSTEM_INTEGER);
+        javaTypeMap.put(java.math.BigDecimal.class, SYSTEM_DECIMAL);
+        javaTypeMap.put(java.lang.String.class, SYSTEM_STRING);
+        javaTypeMap.put(java.time.LocalTime.class, SYSTEM_TIME);
+        javaTypeMap.put(java.time.ZonedDateTime.class, SYSTEM_DATE_TIME);
+        javaTypeMap.put(java.time.LocalDate.class, SYSTEM_DATE_TIME);
+        javaTypeMap.put(java.time.YearMonth.class, SYSTEM_DATE_TIME);
+        javaTypeMap.put(java.time.Year.class, SYSTEM_DATE_TIME);
+        return javaTypeMap;
```
prb112: Should this be an unmodifiable map? 

how does it differ from buildTypeMap?  does buildTypeMap need to respond with this map?

fhir-model/src/main/java/com/ibm/fhir/model/path/FHIRPathType.java
```diff
@@ -469,4 +487,8 @@ public static boolean isSystemType(FHIRPathType type) {
     public static boolean isMetamodelType(FHIRPathType type) {
         return METAMODEL_TYPES.contains(type);
     }
+    
+    public static boolean isJavaType(Class<?> clazz) {
+        return clazz.getName().startsWith("java");
```
prb112: small suggestion here.... starts with`java.` may be more precise vs `javax`

fhir-model/src/main/java/com/ibm/fhir/model/path/FHIRPathTypeInfoNode.java
```diff
@@ -16,6 +16,7 @@
     protected FHIRPathTypeInfoNode(Builder builder) {
         super(builder);
         this.typeInfo = Objects.requireNonNull(builder.typeInfo);
+        System.out.println(children.size());
```
prb112: Logger? or debug code sneaking through.

fhir-model/src/main/java/com/ibm/fhir/model/path/function/SubstringFunction.java
```diff
@@ -45,7 +45,7 @@ public int getMaxArity() {
         
         int start = getInteger(arguments.get(0));
         
-        if (start > value.length() - 1) {
+        if (start < 0 || start > value.length() - 1) {
```
prb112: what's the behavior if you have a negative?

fhir-model/src/test/java/com/ibm/fhir/model/path/test/FHIRPathAsTest.java
```diff
@@ -33,9 +35,14 @@ void testAsOperation() throws Exception {
...
-        
-        assertEquals(result.size(), 0, "Number of selected nodes");
+        try {
+//          Collection<FHIRPathNode> result = evaluator.evaluate(patient, "Patient.deceased as dateTime");
+            evaluator.evaluate(patient, "Patient.deceased as dateTime");
+            fail();
+        } catch (FHIRPathException e) {
+            
+        }
+//      assertEquals(result.size(), 0, "Number of selected nodes");
```
prb112: What does this test expect as an outcome? The assertEquals comment and use of fail() are confusing my understanding ... 

if the intent is to throw an Exception, add an assertTrue(true, "expected to reach here");

fhir-model/src/main/java/com/ibm/fhir/model/path/util/FHIRPathUtil.java
```diff
@@ -257,12 +358,12 @@ public static Temporal getTemporal(TemporalAccessor temporalAccessor) {
             return yearMonth.atDay(1);
         } else if (temporalAccessor instanceof LocalDate) {
             return (LocalDate) temporalAccessor;
+        } else if (temporalAccessor instanceof LocalDateTime) {
+            return (LocalDateTime) temporalAccessor;
         } else if (temporalAccessor instanceof ZonedDateTime) {
             return (ZonedDateTime) temporalAccessor;
         } else if (temporalAccessor instanceof LocalTime) {
             return (LocalTime) temporalAccessor;
```
prb112: This pattern is repeated in multiple spots in JDBCBuilder... only mentioning here that we should probably be consistent and use this as the reference point

fhir-model/src/main/java/com/ibm/fhir/model/path/evaluator/FHIRPathEvaluator.java
```diff
@@ -254,20 +267,18 @@ private void reset() {
...
             if (currentContext.isEmpty()) {
-                // early exit
                 return SINGLETON_FALSE;
             } else if (currentContext.size() > 1) {
                 throw new IllegalArgumentException(String.format("Input collection has %d items, but only 1 is allowed", currentContext.size()));
             }
             
             ExpressionContext typeName = arguments.iterator().next();
-            String identifier = typeName.getText();
+            String identifier = typeName.getText().replace("`", "");
```
prb112: double-check: are there any complicated double escapes? 
eg... java \\\ means a single instance...

fhir-model/src/main/java/com/ibm/fhir/model/path/evaluator/FHIRPathEvaluator.java
```diff
@@ -277,7 +288,7 @@ private void reset() {
             }
             Collection<FHIRPathNode> result = new ArrayList<>();
             ExpressionContext typeName = arguments.get(0);
-            String identifier = typeName.getText();
+            String identifier = typeName.getText().replace("`", "");
```
prb112: duplicate line as the above... only commenting in case there is an explanation or change necessary.

fhir-model/src/main/java/com/ibm/fhir/model/path/evaluator/FHIRPathEvaluator.java
```diff
@@ -1209,7 +1267,7 @@ public boolean hasExternalConstant(String name) {
     
     public static void main(String[] args) throws Exception {
         FHIRPathEvaluator.DEBUG = true;
-        Collection<FHIRPathNode> result = FHIRPathEvaluator.evaluator().evaluate("@2016-12-31 < @2017-01-01");
+        Collection<FHIRPathNode> result = FHIRPathEvaluator.evaluator().evaluate("'a'-'b' = 'ab'");
```
prb112: test code in the main branch

prb112:
```
 [ERROR] Failures: 
[ERROR]   R4ExampleServerTest.processExamples:34->FHIRServerTestBase.assertResponse:446 expected:<201> but was:<400>
[ERROR]   SearchTest.testSearchObservationCodeGTSystem:820 expected [true] but found [false]
[ERROR]   SearchTest.testSearchObservationCodeLTSystem:756 expected [true] but found [false]
[ERROR]   SearchTest.testSearchObservationCodeSystem:691 expected [true] but found [false]
[ERROR]   SearchTest.test_SearchObservationCodeGTSystem:851 expected [true] but found [false]
[ERROR]   SearchTest.test_SearchObservationCodeLTSystem:787 expected [true] but found [false]
[ERROR]   SearchTest.test_SearchObservationCodeSystem:724 expected [true] but found [false]
[ERROR]   SortingTest.testSortValueQuantityAscending:613->assertTrueNaturalOrdering:668 expected [true] but found [false]
```
Some failures

prb112:
The JDK 8 failures might be only the parseInt problem... 

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project fhir-model: Compilation failure
[ERROR] /home/runner/work/FHIR/FHIR/fhir-model/src/main/java/com/ibm/fhir/model/path/evaluator/FHIRPathEvaluator.java:[940,38] no suitable method found for parseInt(java.lang.String,int,int,int)
[ERROR]     method java.lang.Integer.parseInt(java.lang.String,int) is not applicable
[ERROR]       (actual and formal argument lists differ in length)
[ERROR]     method java.lang.Integer.parseInt(java.lang.String) is not applicable
[ERROR]       (actual and formal argument lists differ in length)
[ERROR] -> [Help 1]
```

fhir-model/src/main/java/com/ibm/fhir/model/path/FHIRPathDateTimeValue.java
JohnTimm: Removed

fhir-model/src/main/java/com/ibm/fhir/model/path/FHIRPathDateValue.java
JohnTimm: Removed

fhir-model/src/main/java/com/ibm/fhir/model/path/FHIRPathQuantityNode.java
JohnTimm: isComparableTo check to see whether we can compare. The block in equals checks to see if compareTo result == 0.

fhir-model/src/main/java/com/ibm/fhir/model/path/FHIRPathQuantityValue.java
JohnTimm: compareTo is protected by isComparableTo (even though isComparableTo is always called before compareTo in the FHIRPath evaluator)

fhir-model/src/main/java/com/ibm/fhir/model/path/FHIRPathType.java
JohnTimm: Changed all maps and sets to their "unmodifiable" counterparts. There are 3 distinct Map<Class<?>, FHIRPathType> maps. One holds FHIR model class to FHIRPathType mapping, one holds FHIRPath metamodel class to FHIRPathType mapping and one holds java types (used in the model) to their FHIRPath type equivalent.

fhir-model/src/main/java/com/ibm/fhir/model/path/FHIRPathType.java
JohnTimm: Changed to use `java.` prefix. I only care about "java" in this case, but it would work more generally if I supported both `java.` and `javax.` prefixes.

fhir-model/src/main/java/com/ibm/fhir/model/path/FHIRPathTypeInfoNode.java
JohnTimm: Removed

fhir-model/src/main/java/com/ibm/fhir/model/path/function/SubstringFunction.java
JohnTimm: Per the spec, we should return empty set if the start index is out of range.

fhir-model/src/test/java/com/ibm/fhir/model/path/test/FHIRPathAsTest.java
JohnTimm: I changed this back to the original code after relaxing the `as` operator behavior.

fhir-model/src/main/java/com/ibm/fhir/model/path/util/FHIRPathUtil.java
JohnTimm: Some sort of Map<Class<?>, Function> might be cleaner. I'll leave it for now.

fhir-model/src/main/java/com/ibm/fhir/model/path/evaluator/FHIRPathEvaluator.java
JohnTimm: In FHIRPath the tick mark is the escape character for when special characters are used in an identifier. The visitIdentifier method takes care of this for most cases but I had to add it in here for when I get the raw text direction.

fhir-model/src/main/java/com/ibm/fhir/model/path/util/FHIRPathUtil.java
prb112: No problem... just verbalizing the pattern

fhir-model/src/main/java/com/ibm/fhir/model/path/evaluator/FHIRPathEvaluator.java
JohnTimm: I just need to remove the tick if it is being used to escape an identifier. visitIdentifier has the same behavior.

fhir-model/src/main/java/com/ibm/fhir/model/path/evaluator/FHIRPathEvaluator.java
JohnTimm: Removed

PR #447, opened 2019-11-27
URI should default to complete match (not starts-with) #273

prb112:
- introduce specific behavior for the URI type parameter


fhir-persistence/src/test/java/com/ibm/fhir/persistence/search/test/AbstractSearchURITest.java
```diff
@@ -38,8 +38,10 @@ public void testSearchURI_uri() throws Exception {
         assertSearchReturnsSavedResource("uri", "http://hl7.org/fhir/DSTU2");
         assertSearchReturnsSavedResource("uri", "urn:uuid:53fefa32-1111-2222-3333-55ee120877b7");
         
-        // https://github.com/IBM/FHIR/issues/273
         // Matches are supposed to be precise (e.g. case, accent, and escape sensitive), but aren't
```
albertwang-ibm: should we remove "but aren't" in the comments?

prb112: Yes, you are 100% right.  Let me fix

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -248,12 +248,22 @@ protected JDBCOperator getOperator(Parameter queryParm) {
         JDBCOperator operator = null;
         Modifier modifier = queryParm.getModifier();
 
-        if (modifier != null) {
+        // In the case where a URI, we need specific behavior/manipulation
+        // so that URI defaults to EQ, unless... BELOW
+        if (Type.URI.compareTo(queryParm.getType()) == 0) {
+            if(modifier != null && Modifier.BELOW.compareTo(modifier) == 0) {
```
albertwang-ibm: according to the spec https://www.hl7.org/fhir/search.html, we should consider both "below" and "above":
"For uri: :below, :above indicate that instead of an exact match, either the search term left-matches the value, or vice-versa."
I guess "left-matches" means begin with, and "right-matches" means end-with ...

prb112: Above would be a totally different implementation, and require some re-coding

I think we can add the support doing the following: 

- If Modifier is above, switch to LIKE
- Find the domain of the URL using `/` 
     http://acme.org/fhir/ValueSet/123/_history/5
- Use a BIND value of `http://acme.org/`

Does this approach make sense?

albertwang-ibm: I'm not sure what's the real meaning of left-match and right-match here... @JohnTimm , what's your suggestion?

albertwang-ibm: BTW, seems the null check is not needed here...

prb112: It's not a left-right match with above. 
It's matching on the individual components

prb112: PR https://github.com/IBM/FHIR/pull/new/issue-448

albertwang-ibm: seem we can just remove modifier != null.

prb112: I'm sorry we cannot... when I did it throws a NPE

PR #452, opened 2019-12-02
clean up documentation to make tag visible #270

prb112:
clean up documentation to make tag visible #270
-rm  seemed to be a stray entry into the docker commands


albertwang-ibm:
@prb112 seems missing DCO

prb112:
> @prb112 seems missing DCO

thanks ... I forgot the problems with the web-ui.  I have just fixed and pushed

PR #454, opened 2019-12-02
Introduce fhir-path and fhir-profile modules

JohnTimm:


fhir-path/src/main/java/com/ibm/fhir/path/function/registry/FHIRPathFunctionRegistry.java
```diff
@@ -0,0 +1,170 @@
...
+import com.ibm.fhir.path.function.ToQuantityFunction;
+import com.ibm.fhir.path.function.ToStringFunction;
+import com.ibm.fhir.path.function.TodayFunction;
+import com.ibm.fhir.path.function.TypeFunction;
+import com.ibm.fhir.path.function.UnionFunction;
+import com.ibm.fhir.path.function.UpperFunction;
+
+public class FHIRPathFunctionRegistry {
+    private static final FHIRPathFunctionRegistry INSTANCE = new FHIRPathFunctionRegistry();
+    private Map<String, FHIRPathFunction> functionMap = new ConcurrentHashMap<>();
```
prb112: small thing... do you intend this to mutable? e.g. a downstream consumer could change this.

fhir-validation/pom.xml
```diff
@@ -18,6 +18,16 @@
...
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>fhir-path</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>fhir-path</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
prb112: is the second one intended for the same scope?

fhir-path/src/main/java/com/ibm/fhir/path/function/registry/FHIRPathFunctionRegistry.java
prb112: discussed with John - no direct access

JohnTimm: Added `final` keyword to FHIRPathFunctionRegistry class declaration.

fhir-validation/pom.xml
JohnTimm: Removed unnecessary dependency.

PR #457, opened 2019-12-02
more swagger updates

lmsurpre:
1. I found that the swagger definitions are missing the 'resourceType'
propery; after this change it should be there for all resource types.
  Unfortunately, I think we need to change our whole approach if we want
to indicate the value is 'fixed'. Because there are some places where
any Resource is allowed, I added this property to the base Resource type (as an enum).
However, Swagger/OpenAPI--or at least the
Liberty support for it--doesn't like it when a property is defined in
both halves of the *allOf* constraint...even if those two are compatible.
To address it, I think we should avoid the *allOf* construct and
just copy the duplicated Resource properties into each different
Resource subtype. For now, its just a TODO in the code.

2. Added an extra check to prevent the creation of a DomainResource
definition...only children of DomainResource should get an endpoint, not
DomainResource itself.

3. Introduced generateAllInOne for the FHIROpenApiGenerator. When using
an API Gateway like IBM API Connect, we found it most useful to generate
one swagger definition per resource type. However, I couldn't figure out
how to make Liberty work with multiple static files, and so instead I
created this new method. By default, both methods will be called from
Main and the 'all-in-one' will be written to a file named
"all-openapi.json". I did *not* create a corresponding method for the
Swagger (2.0) variant.

4. Used the new method to regenerate our openapi.json file thats
included in the fhir-openapi web project.

### DIFF
--- a/fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
+++ b/fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
@@ -18,9 +18,11 @@ import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.json.Json;
 import javax.json.JsonArray;
@@ -47,7 +49,6 @@ import com.ibm.fhir.model.resource.EvidenceVariable;
 import com.ibm.fhir.model.resource.GuidanceResponse;
 import com.ibm.fhir.model.resource.Library;
 import com.ibm.fhir.model.resource.Measure;
-
 import com.ibm.fhir.model.resource.MedicationDispense;
 import com.ibm.fhir.model.resource.MedicationKnowledge;
 import com.ibm.fhir.model.resource.MedicationRequest;
@@ -86,6 +87,7 @@ import com.ibm.fhir.model.type.SubstanceAmount;
 import com.ibm.fhir.model.type.TriggerDefinition;
 import com.ibm.fhir.model.type.UsageContext;
 import com.ibm.fhir.model.type.Uuid;
+import com.ibm.fhir.model.type.code.ResourceType;
 import com.ibm.fhir.model.util.FHIRUtil;
 import com.ibm.fhir.model.util.ModelSupport;
 import com.ibm.fhir.model.visitor.AbstractVisitable;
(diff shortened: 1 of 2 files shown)


fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
```diff
@@ -133,10 +135,136 @@ public static void main(String[] args) throws Exception {
...
+        generateBatchTransactionOpenApi(filter);
+    }
+    
+    private static void generateAllInOne(Filter filter) throws Exception {
+        JsonObjectBuilder swagger = factory.createObjectBuilder();
+        swagger.add("openapi", "3.0.0");
+
+        JsonObjectBuilder info = factory.createObjectBuilder();
+        info.add("title", "Simplified FHIR API");
+        info.add("description", "A simplified version of the FHIR API");
```
albertwang-ibm: Simplified FHIR API or "IBM FHIR Server API"? just for discussion.

lmsurpre: I'm fine with either.  I remember now that I had changed it (but just in this one spot) so as not to imply that the IBM FHIR Server itself can only support this simplified model.

That said, I agree we should either update it in all the spots across all swagger/openapi files, or just revert this change and go with "IBM FHIR Server API"

albertwang-ibm: yes, telling our server API is simplified doesn't sound great for me :)

lmsurpre: OK, I updated all the descriptions to match now.

fhir-swagger-generator/src/main/java/com/ibm/fhir/swagger/generator/FHIRSwaggerGenerator.java
```diff
@@ -259,25 +264,25 @@ private static void generateBatchTransactionSwagger() throws Exception, ClassNot
...
-        }
+        generateDefinition(Bundle.class, definitions);
 
         // generate definition for all inner classes inside the top level resources.
         for (String innerClassName : FHIROpenApiGenerator.getAllResourceInnerClasses()) {
             Class<?> parentClass = Class.forName(RESOURCEPACKAGENAME + "." + innerClassName.split("\\$")[0]);
-            if (CapabilityStatement.class.equals(parentClass)) {
+            if (Bundle.class.equals(parentClass)) {
                 Class<?> innerModelClass = Class.forName(RESOURCEPACKAGENAME + "." + innerClassName);
                 // System.out.println("Resource:  " + className);
```
albertwang-ibm: Is this commented line for debugging purposes only?

fhir-swagger-generator/src/main/java/com/ibm/fhir/swagger/generator/FHIRSwaggerGenerator.java
```diff
@@ -259,25 +264,25 @@ private static void generateBatchTransactionSwagger() throws Exception, ClassNot
...
                 generateDefinition(innerModelClass, definitions);
             }
         }
 
+        // generate definition for all the defined Types.
+        for (String typeClassName : FHIROpenApiGenerator.getAllTypesList()) {
+            Class<?> typeModelClass = Class.forName(TYPEPACKAGENAME + "." + typeClassName);
+            // System.out.println("Type:  " + className);
```
albertwang-ibm: Is this commented line for debugging purposes only?

albertwang-ibm: If so, maybe add a Boolean DEBUG as switch to turn off by default?

lmsurpre: i just removed them.  i had removed from FHIROpenApiGenerator before (per your previous PR review), but never removed them from FHIRSwaggerGenerator until now.

Issue #460, opened 2019-12-03
Expand all model tests


prb112:
Expand all model tests

PR #461, opened 2019-12-03
 Expand all model tests #460

prb112:
Expand all model tests #460 

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,6 +54,9 @@ jobs:
         )
         model_paths=(
           fhir-model
+          fhir-validation/src/main
+          fhir-path/src/main
+          fhir-profile/src/main
         )
         declare -a PROFILES_ARR
 


albertwang-ibm:
I didn't add the validator into R4ExampleTest.java for fhir-model project to avoid cycling dependency situation, because the ValidationProcessor is defined in fhir-validation project which depends on fhir-model project,  neither can we simply move ValidationProcessor into fhir-model projects due to that FhirValidator used by ValidationProcessor is actually defined in fhir-validation project.  Seems one option for us is to change the workflow to run all.txt index file for fhir-persistence-jdbc project whenever any project in either model-pathes or jdbc-pathes is changed...

prb112:
Sure. since this is @lmsurpre I'll leave with him to decide on the approach

lmsurpre:
should we have a separate test within fhir-validation which executes the validator against all the resource?

prb112:
> should we have a separate test within fhir-validation which executes the validator against all the resource?

Sure. I think that's valid.

PR #470, opened 2019-12-03
issue #460 - refactor R4ExamplesDriver to invoke tests as appropriate

lmsurpre:
1. Introduced Index.java in the fhir-examples module and moved the
"TestType" concept from fhir-model into this class as a list of
filenames with corresponding Enum values.

2. Renamed ExamplesUtil.reader to "resourceReader" and introduced a
separate helper for getting a Reader from an Index instance

3. Moved all fhir-example index files to the top level of the
fhir-examples module and updated the R4ExampleDriver logic to infer the
resource type from the filename's file extension

4. Moved System Property logic from R4ExamplesDriver to the classes
which call this driver, making it easier to set each one individually

5. Updated pom.xml files and our CI's build.yml so that fhir-model,
fhir-validation, fhir-search, and fhir-persistence-jdbc tests can all be
configured independently (and automatically based on the changes in the
pull request).

Also dropped PoolConnectionProvider log levels from INFO to FINE.


.github/workflows/build.yml
```diff
@@ -46,30 +46,48 @@ jobs:
...
+        model_paths=(
+          fhir-model
+        )
+        validation_paths=(
+          fhir-validation
+          fhir-path/src/main
+        )
+        search_paths=(
+          fhir-search
+        )
```
prb112: I'll close the PR I opened

fhir-examples/src/main/java/com/ibm/fhir/examples/Index.java
```diff
@@ -0,0 +1,62 @@
...
+     * All R4 spec examples
+     */
+    SPEC_XML("/spec-xml.txt"),
+
+    /**
+     * All IBM generated examples
+     */
+    IBM_XML("/ibm-xml.txt");
+
+
```
prb112: minor, but extra space.

fhir-examples/src/main/java/com/ibm/fhir/examples/Index.java
```diff
@@ -0,0 +1,62 @@
+package com.ibm.fhir.examples;
```
prb112: No copyright header

fhir-examples/src/main/java/com/ibm/fhir/examples/Index.java
```diff
@@ -0,0 +1,62 @@
...
+
+public enum Index {
+    /**
+     * Both R4 spec and IBM generated examples
+     */
+    ALL_JSON("/all-json.txt"),
```
prb112: I'm looking at the format. 
perhaps unifying the layout. 

[PROVIDER]-[TYPE]-[FORMAT].txt
provider =  IBM | SPEC | third-party ? 
type = all | minimal | performance ? 
format = xml | json 

just a thought...

fhir-parent/pom.xml
```diff
@@ -669,15 +672,31 @@
...
             <properties>
-                <!-- for the fhir-model project -->
-                <fhir-model.testType>ALL</fhir-model.testType>
+                <fhir-model.testType>ALL_JSON</fhir-model.testType>
+            </properties>
+        </profile>
+        <profile>
+            <id>validation-all-tests</id>
+            <properties>
+                <fhir-persistence-jdbc.testType>ALL_JSON</fhir-persistence-jdbc.testType>
```
prb112: should this be .index?

.github/workflows/build.yml
lmsurpre: good catch.  will update now.

fhir-examples/src/main/java/com/ibm/fhir/examples/Index.java
lmsurpre: i don't see it...

fhir-examples/src/main/java/com/ibm/fhir/examples/Index.java
lmsurpre: good catch. will update now.

fhir-examples/src/main/java/com/ibm/fhir/examples/Index.java
prb112: line 48/49 are empty lines

fhir-examples/src/main/java/com/ibm/fhir/examples/Index.java
lmsurpre: we used to have "TestType" and "Format" as different fields, but I think it just got in the way.  I think we'd like to be able to create a single index file with a mix of all 3 of those in it.
For now, I left the index files as-is and just named them to indicate what was in them, but now there is nothing stopping us from defining one with a mix of JSON and XML.  (and we already have ones with a mix of IBM and SPEC).
If we weren't files-system bound, we could just keep all the files in a flat structure and tag them as appropriate, but that definitely makes it hard on eclipse (and programs like it) when you open that directory...

fhir-parent/pom.xml
lmsurpre: yes, good catch

fhir-examples/src/main/java/com/ibm/fhir/examples/Index.java
prb112: no problem. just food for thought.

fhir-examples/src/main/java/com/ibm/fhir/examples/Index.java
lmsurpre: ah, extra line.  I left it on purpose to visually distinguish the enum values from the rest of the file...I thought I had seen it in one of our other enums somewhere, but definitely not attached to it.  let me know if you really want it squashed, otherwise i'll leave it.

.github/workflows/build.yml
```diff
@@ -46,30 +46,48 @@ jobs:
...
+        fi
+        if [ -n "$(git --no-pager diff --name-only ${BASE} ${GITHUB_SHA} -- ${jdbc_paths[@]})" ]; then
+          echo "Changes affect fhir-persistence-jdbc, running ALL"
+          PROFILES_ARR+=(jdbc-all-tests)
+        else
+          echo "No changes affect fhir-persistence-jdbc, running MINIMAL"
+        fi
 
         if [ ${#PROFILES[@]} -eq 0 ]; then
           # Convert the PROFILES_ARR array to a comma-delimited string
```
albertwang-ibm: I think we'd better also enable the ability of cross check, e.g, (1) add fhir-example to the loop, and whenever fhir-example changes, maybe we should run all.txt for validation, jdbc etc.  (2) whenever model-path change happens, maybe we should run all.txt for jdbc or validation(as we discussed, there is no validator in fhir-model project at present).

fhir-examples/src/main/java/com/ibm/fhir/examples/Index.java
```diff
@@ -0,0 +1,68 @@
...
+     * All R4 spec examples
+     */
+    SPEC_XML("/spec-xml.txt"),
+
+    /**
+     * All IBM generated examples
+     */
+    IBM_XML("/ibm-xml.txt");
+
+
```
albertwang-ibm: Can we merge the index files(for both json and xml) to one set instead of 2? currently, we have to double the efforts to make index changes to both.
e.g, with index content like:
       xxxxx-xxxxxx           (both json and xml)
       xxxxx-xxxxx.json    (json only) 
       xxxxxxx-xxxxxxx.xml (xml only)
       ...

Just for discussion as an potential enhancement, not a must.

lmsurpre: yes, this approach now makes that possible.  we can slide/dice the examples in any way we want...its as simple as just copy/pasting the lines from one to the other.  do we want to do that?

i think the format (json vs xml) doesn't really matter for any of the tests beyond the fhir-model test, and i don't want to make the MINIMAL index too large since we run that on every build.  Maybe we combine the `ibm` and `spec` index files, but leave MINIMAL and ALL separate?

albertwang-ibm: MINIMAL-json.txt + MINIMAL-xml.txt --> MINIMAL.txt
the same for other index files, because seems the file names in them are all the same, just file extension is different... 
and we just need to change the driver a little bit to process the merged index file... 
Anyway, this is just a enhancement suggestion, not really needed for this PR.

.github/workflows/build.yml
lmsurpre: on (1) I totally agree, I'll add fhir-example to all 4 paths
on (2), this PR adds the logic so that if anything under `fhir-validation` or `fhir-path/src/main` changes, then we'll run the validation test with the ALL_JSON index

fhir-examples/src/main/java/com/ibm/fhir/examples/Index.java
lmsurpre: the set isn't exactly the same.  there are some spec examples that have json and no xml and vice versa.

.github/workflows/build.yml
albertwang-ibm: Should we add fhir-model project to validation-paths? I mean e.g, if resource classes in it is updated or re-generated, should we trigger validation for all examples?

lmsurpre: sure, i can add that

albertwang-ibm: thanks, just try to make the matrix better.

Issue #403, opened 2019-11-13
Upgrade FHIR spec version to 4.0.1


JohnTimm:
HL7 recently released a technical correction to FHIR R4 v4.0.0. We need to download the latest artifacts (e.g. structure definitions, search parameters, etc.), regenerate the model code and run a regression to see what else needs to get updated.

- [x] Updated definitions in fhir-tools and fhir-registry
- [x] Regenerate fhir-model
- [x] Regenerate ibm examples
- [x] Overwrite spec examples in fhir-examples
- [x] Update FHIRPathEvaluator for %rootResource
- [x] other?

lmsurpre:
Turns out there was quite a bit of "other"...mostly related to a spec change where Resource.id is no longer an Id but instead has a type code of `http://hl7.org/fhirpath/System.String`

prb112:
Whole team reviewed this one

PR #478, opened 2019-12-04
issue #403 - Upgrade to FHIR 4.0.1

lmsurpre:
Added 4.0.1 definitions and updated CodeGenerator and model
* Resource.id is now a java.lang.String and not an Id (still with id format constraint, but no extensions allowed)
  * Updated PathAwareVisitor to track the path as we visit non-value elements of type java.lang.String (i.e. ones like `Resource.id` but not ones like `Resource.Meta.versionId.value`)
* Removed unneccessary escaping of keywords in generated constraint annotations 

Updated fhir-examples
* fixed generator issues and regenerated all ibm examples
* downloaded spec from https://www.hl7.org/fhir/downloads.html and overwrote the 4.0.0 examples
* updated and reorganized all index files

Updated Validator and FHIRPathEvaluator to support the new `%rootResource` external constant


fhir-model/src/main/java/com/ibm/fhir/model/resource/AppointmentResponse.java
```diff
@@ -625,7 +624,7 @@ public Builder actor(Reference actor) {
          * <p>This element is required.
          * 
          * @param participantStatus
-         *     accepted | declined | tentative | in-process | completed | needs-action | entered-in-error
```
lmsurpre: note:  this is a breaking change!

fhir-model/src/main/java/com/ibm/fhir/model/resource/CarePlan.java
```diff
@@ -994,7 +993,7 @@ public Builder partOf(Collection<Reference> partOf) {
          * <p>This element is required.
          * 
          * @param status
-         *     draft | active | suspended | completed | entered-in-error | cancelled | unknown
```
lmsurpre: this is a breaking change!

fhir-examples/src/main/resources/json/ibm/complete-absent/Account-1.json
```diff
@@ -1,13 +1,5 @@
 {
     "resourceType": "Account",
-    "_id": {
-        "extension": [
-            {
-                "url": "http://hl7.org/fhir/StructureDefinition/data-absent-reason",
```
prb112: I take it this is no longer used? it's absent = therefore it should be treated as absent?

prb112: Crazy question, what happens when you leave it?  If it'd be valid, maybe we should have a test with data like this?

lmsurpre: as of 4.0.1 that is invalid.  i did manually test this and it complains its an unrecognized element (which i think is the right behavior), but didn't introduce a negative test case for it.

Issue #438, opened 2019-11-21
_id should not be stored as pivoted name-value pair values


prb112:
__id should not be stored as pivoted name-value pair value_

The _id parameter/code are stored as first-class columns in the _logical_resources and _resources tables. Therefore, we should not store these attributes in the name-value pair pivoted _values tables, and search queries should be written to use the columns directly, which is more efficient (and logical).

**Steps**

1. Investigate Schema Status
2. Investigate Persistence Layer changes and Search changes
3. Implement change to support and use _id and _lastUpdated uniquely
4. Confirm common ids don't collide on the the Search (e.g. Practioner/1 doesn't collide with Patient/1)

Related to: Global Search Performance

Original Issue: _id and _lastUpdated should not be stored as pivoted name-value pair values #324 #255

prb112:
moved to review qa

Issue #483, opened 2019-12-05
Implement 'memberOf' function


JohnTimm:

prb112:
Tests pass and works for me.

PR #484, opened 2019-12-05
Issue #438 - initial version of 'memberOf' function

JohnTimm:


JohnTimm:
Commit message uses in correct issue number. Should be #483 (instead of #438).

Issue #261, opened 2019-10-15
whole-system search(search all and search across multiple resource types)  on the base endpoint


lmsurpre:
We currently implement whole-system search by handling GET requests to `[base]/_search` in `FHIRResource.searchAll()`.  However, that conflicts with the following from the [specification](https://www.hl7.org/fhir/http.html#vsearch):
```
...
It is also possible to search across multiple resource types:

  GET [base]?_type=Condition,Observation&[parameters]{&_format=[mime-type]}

This is a request to search on both Condition and Observation. In this case, the only parameters that can be used are those defined for both Condition and Observation (using SearchParameter.base - see Cross-resource Search Parameters), or the parameters defined for all resources. If a search lists types not listed in SearchParameter.base for any of the parameters, this is an error, and a server SHOULD return a 400 status. It is also possible to search on all types at once:

  GET [base]?[parameters]{&_format=[mime-type]}
```

Maybe we were confused by this paragraph at the [top of the page](https://www.hl7.org/fhir/http.html#styleguide):
```
This specification uses the underscore as a prefix to disambiguate reserved names from other names in 3 cases:

    To differentiate system wide history and search interactions from interactions on Resource Types
    To differentiate search, history and similar interactions from instances of a resource type
    To differentiate search parameters defined for all resources from those defined for specific resource types
```

Specifically, I think the first part there (To differentiate system wide history and search interactions) is misleading.

NOTE:  rather than removing support for the `[base]/_search` endpoint, we should support POST method searches on that endpoint.

albertwang-ibm:
After studying Fhir r4 search spec, I think we need to add:
(1) Multiple resource types search with GET using base url with _type parameter
(2) Multiple resource types search with POST using base url+_search with _type parameter.
(3) All resource types search with GET using base url.
(4) All resource types search with POST using base url+_search

prb112:
Closed as the PR is accepted

PR #489, opened 2019-12-06
issue #261 #494 whole system search enhancement and fix for self url of chained parameter search result

albertwang-ibm:
Summary of the changes:
(1) Changed search all GET path to base url.
(2) Added search all POST support.
(3) Added multiple resource types search support.
(4) Refactored ServerTestBase.
(5) Fixed self url of chained parameter search response.
(6) Added new error messages for search all tests.
(7) Modified and enhanced test client codes to support the new search paths.
(8) Added new integration test cases for the new multiple resource types search feature.


fhir-client/src/main/java/com/ibm/fhir/client/FHIRClient.java
```diff
@@ -344,7 +344,10 @@
      * @return a FHIRResponse that contains the results of the 'search-all' operation
      * @throws Exception
      */
-    FHIRResponse searchAll(FHIRParameters parameters, FHIRRequestHeader... headers) throws Exception;
+    FHIRResponse searchAllGet(FHIRParameters parameters, FHIRRequestHeader... headers) throws Exception;
+    
+    
+    FHIRResponse searchAllPost(FHIRParameters parameters, FHIRRequestHeader... headers) throws Exception;
```
prb112: I think it's worth documenting the reason that there are two interfaces.  some javadocs might help explain.

fhir-client/src/main/java/com/ibm/fhir/client/FHIRClient.java
```diff
@@ -344,7 +344,10 @@
      * @return a FHIRResponse that contains the results of the 'search-all' operation
```
prb112: might want to add that this is specific to get

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
```diff
@@ -205,6 +212,10 @@ protected SqlQueryData buildSystemLevelQuery(String selectRoot, String subSelect
         
         for(Map.Entry<String,Integer> resourceEntry : resourceNameMap.entrySet()) {
             String resourceTypeName =  resourceEntry.getKey();
+            // Only search the required resource types if any.
+            if (this.resourceTypes != null && !this.resourceTypes.contains(resourceTypeName)) {
```
prb112: what's the point of this code?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
```diff
@@ -50,6 +50,9 @@
     private static final Set<String> SKIP_WHERE = new HashSet<>(Arrays.asList("_id"));
     
     protected Class<?> resourceType;
+    
+    // Used for whole system search on multiple resource types.
+    private List<String> resourceTypes = null;
```
prb112: maybe set as a default Collections.emptyList();

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
prb112: if you set the emptyList then use !this.resourceTypes.isEmpty

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregatorFactory.java
```diff
@@ -41,6 +42,10 @@ else if (searchContext.hasSortParameters()) {
             qsa = new QuerySegmentAggregator(resourceType, offset, pageSize, parameterDao, resourceDao);
         }
         
+        if( Resource.class.equals(resourceType) && searchContext.getSearchResourceTypes()!= null) {
```
prb112: !isEmpty ? instead...

fhir-search/src/main/java/com/ibm/fhir/search/SearchConstants.java
```diff
@@ -89,6 +89,9 @@ private SearchConstants() {
     
     // _pretty
     public static final String PRETTY = "_pretty";
+    
+    // _type
+    public static final String RESTYPE = "_type";
```
prb112: Spell it out? 
RESOURCE_TYPE

fhir-search/src/main/java/com/ibm/fhir/search/uri/UriBuilder.java
```diff
@@ -108,6 +109,20 @@ public String toSearchSelfUri() throws URISyntaxException {
...
     }
 
+    private void appendResourceTypesParameter() {
+        if (context.getSearchResourceTypes() != null && !context.getSearchResourceTypes().isEmpty()) {
+            queryString.append(SearchConstants.AND_CHAR);
+            queryString.append(SearchConstants.RESTYPE);
+            queryString.append(SearchConstants.EQUALS_CHAR);
+            String delim = SearchConstants.EMPTY_QUERY_STRING;
+            for (String param : context.getSearchResourceTypes()) {
+                queryString.append(delim + param);
```
prb112: break into two appends 

append the DELIMITER char and then the param string 
it'll be slightly more efficient. 
you won't create a string, and then append the created string. 
you'll add a char, then a string.

prb112: Also you could use StringJoiner here.

fhir-search/src/main/java/com/ibm/fhir/search/uri/UriBuilder.java
```diff
@@ -207,16 +222,20 @@ private String serializeSearchParmToQueryString(Parameter param) {
             while (process) {
                 param   = param.getNextParameter();
                 process = param != null && param.isChained();
+                // A guard is added as the original code 'could' result in an NPE.
                 if (param != null) {
-                    // A guard is added as the original code 'could' result in an NPE.
-                    returnString.append(param.getCode());
+                    if (param.isChained()) {
+                        returnString.append(SearchConstants.CHAINED_PARAMETER_CHARACTER).append(param.getCode());
```
prb112: break this into two parts. 
returnString.append(SearchConstants.CHAINED_PARAMETER_CHARACTER) can go outside the if/else

and then you append(param.getCode) in the if
else you appendNormalParameter

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -83,7 +83,13 @@
             "Search parameter: '%s' must have resource type name modifier";
     private static final String INVALID_TARGET_TYPE_EXCEPTION = "Invalid target type for the Inclusion Parameter.";
     private static final String UNSUPPOTED_EXPR_NULL =
-            "An empty expression is found or the parameter type is unsupported [%s][%s]";;
+            "An empty expression is found or the parameter type is unsupported [%s][%s]";
+    
+    private static final String MODIFIYERRESOURCETYPE_NOT_ALLOWED_FOR_RESOURCETYPE =
```
prb112: I think Modifier is not allowed for search parameter of resource type would be sufficient.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -588,6 +594,43 @@ public static FHIRSearchContext parseQueryParameters(Class<?> resourceType,
...
+            // Because _include and _revinclude searches all require certain resource type modifier in 
+            // search parameter, so we just don't support it.
+            if (queryParameters.containsKey(SearchConstants.INCLUDE)
+                        || queryParameters.containsKey(SearchConstants.REVINCLUDE)) {
+                throw SearchExceptionUtil.buildNewInvalidSearchException(
+                        "system search not supported with _include or _revinclude.");
+            }
+            
+            if (queryParameters.containsKey(SearchConstants.RESTYPE)) {
+                for (String ResTypes : queryParameters.get(SearchConstants.RESTYPE)) {
```
prb112: lower camelcase 
resTypes

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -588,6 +594,43 @@ public static FHIRSearchContext parseQueryParameters(Class<?> resourceType,
...
+                    for (String resType : tmpResourceTypes) {
+                        if (ModelSupport.isResourceType(resType)) {
+                            resourceTypes.add(resType);
+                        } else {
+                            if (lenient) {
+                                log.log(Level.FINE, resType + " is not valid resource type!");
+                                continue;
+                            } else {
+                                throw SearchExceptionUtil.buildNewInvalidSearchException(
+                                        "_type search parameter has invilid resource type:" + resType);
```
prb112: it may be good to escape the resType on the way back out

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -588,6 +594,43 @@ public static FHIRSearchContext parseQueryParameters(Class<?> resourceType,
...
+                            }
+                        }
+                    }
+                }
+            } 
+        }
+        
+        queryParameters.remove(SearchConstants.RESTYPE);
```
prb112: How does the self uri look on the way back in the response body? is it still right?

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -620,14 +668,31 @@ public static FHIRSearchContext parseQueryParameters(Class<?> resourceType,
...
+                    if (isMultiResTypeSearch) {
+                      // Find the SearchParameter that will apply to all the resource types.
+                      for (String resType: resourceTypes) {
+                          Map<String, SearchParameter> resTypeSPs = getApplicableSearchParametersMap(resType);
+  
+                          // Get the search parameter from our filtered set of applicable SPs for this resource type.
+                          searchParameter = resTypeSPs.get(parameterName);
+                          if (searchParameter == null) {
+                              String msg =
+                                      "Search parameter '" + parameterName + "' for resource type '"
```
prb112: please change the references to parameterName to parameterCode. 
We pass in the code (it may map to name in most cases, however is not consistent with the change from dstu2 to r4.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -687,7 +752,7 @@ public static FHIRSearchContext parseQueryParameters(Class<?> resourceType,
                 throw SearchExceptionUtil.buildNewParseParameterException(name, e);
             }
         } // end for
-
+
```
prb112: extra spacing

fhir-server-test/src/test/java/com/ibm/fhir/server/test/FHIROperationTest.java
```diff
@@ -231,12 +231,6 @@ private Composition buildComposition(String practitionerId, String patientId, St
         return composition;
     }
 
-    private Condition buildCondition(String patientId, String fileName) throws Exception {
```
prb112: refactored into FHIRServerTestBase?

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchAllTest.java
```diff
@@ -154,7 +175,7 @@ public void testSearchAllUsingSecurity() throws Exception {
     public void testSearchAllUsingProfile() throws Exception {
         FHIRParameters parameters = new FHIRParameters();
         parameters.searchParam("_profile", "http://ibm.com/fhir/profile/Profile");
-        FHIRResponse response = client.searchAll(parameters);
+        FHIRResponse response = client.searchAllGet(parameters);
```
prb112: I'm looking at the signature some more.  is there a better option?  where we can have something like this? 

client.searchAll(parameters) ... goes to post. 
client.searchAll(parameters, boolean) 
true goes to post
false goes to get

fhir-client/src/main/java/com/ibm/fhir/client/FHIRClient.java
albertwang-ibm: yeah, this is align with FHIR spec. will add

fhir-client/src/main/java/com/ibm/fhir/client/FHIRClient.java
albertwang-ibm: go point, will add

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
albertwang-ibm: resourceTypes can be empty, if it's empty, then means not a resource type specific search

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
albertwang-ibm: I prefer null here, because we will simply pass in one if need instead of add to existing list.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregatorFactory.java
albertwang-ibm: align with previous

fhir-search/src/main/java/com/ibm/fhir/search/SearchConstants.java
albertwang-ibm: yeah, agree better to change, will do

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
prb112: right - however downstream you have two calls to check isNotNull and size 
it's purely preference.

fhir-search/src/main/java/com/ibm/fhir/search/uri/UriBuilder.java
albertwang-ibm: make sense to use append instead of + in this cases, will change

fhir-search/src/main/java/com/ibm/fhir/search/uri/UriBuilder.java
albertwang-ibm: make sense, will change.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
albertwang-ibm: I think we'd better make the error message more accurate.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
albertwang-ibm: good catch, will change

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
albertwang-ibm: yeah, it's correct, because I added it back when building the url.
It's removed here to avoid the unknown search parameter error when processing the parameters.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
albertwang-ibm: en.. this is interesting, I think all the extra spaces are already removed by eclipsse save action. ah. seems not for this workspace. will remove

fhir-server-test/src/test/java/com/ibm/fhir/server/test/FHIROperationTest.java
albertwang-ibm: yes

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchAllTest.java
albertwang-ibm: make sense, will change

fhir-client/src/main/java/com/ibm/fhir/client/FHIRClient.java
albertwang-ibm: done

fhir-client/src/main/java/com/ibm/fhir/client/FHIRClient.java
albertwang-ibm: done

fhir-search/src/main/java/com/ibm/fhir/search/SearchConstants.java
albertwang-ibm: done

fhir-search/src/main/java/com/ibm/fhir/search/uri/UriBuilder.java
albertwang-ibm: done

fhir-search/src/main/java/com/ibm/fhir/search/uri/UriBuilder.java
albertwang-ibm: done

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
albertwang-ibm: done

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
albertwang-ibm: done

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchAllTest.java
albertwang-ibm: done

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
albertwang-ibm: sound good, done

PR #492, opened 2019-12-09
Minor fix on line 08

kant:


prb112:
Thanks for your documentation fix @kant 

As we use the DCO check it's preferred that each contributor use: 

```
git commit --amend --signoff
git push --force-with-lease origin patch-1
```

Issue #259, opened 2019-10-15
Support "special" search param Location-near


lmsurpre:
In DSTU2 we did our own "special" Location search behavior and I think our schema still includes a Location search parameters table.

In STU3 or R4, they defined the "special" search parameter Location-near:
Search for locations where the location.position is near to, or within a specified distance of, the provided coordinates expressed as [latitude]\|[longitude]\|[distance]\|[units] (using the WGS84 datum, see notes). If the units are omitted, then kms should be assumed. If the distance is omitted, then the server can use its own discretion as to what distances should be considered near (and units are irrelevant) Servers may search using various techniques that might have differing accuracies, depending on implementation efficiency. Requires the near-distance parameter to be provided also | Location.position

We should either:
A. map this new search parameter onto the current functionality (and test test test); or
B. remove the current functionality if its broken / in disrepair

prb112:
Put in Review QA

prb112:
PR https://github.com/IBM/FHIR/pull/497/files

PR #497, opened 2019-12-10
Support "special" search param Location-near #259

prb112:
Support "special" search param Location-near #259

This PR is broken into 3 core elements: 
1 - fhir-search location enablement 
2 - fhir-persistence-jdbc location enablement
3 - extraction of parameter values

    - Add support for 'special' in parseQueryString with Prefix
    - Add near.unitTest.properties to turn on fhir-search and
    fhir-persistence tracing
    - Update the AbstractQueryBuilder to wrap location parsing errors in a
    FHIRPersistenceException and re-throw
    - Add configuration for FHIRConfiguration for switching bounding type
    - Update the LocationParmBehaviorUtil to use BoundingType configuration
    - Add JDBCSearchNearTest
    - Update Units-of-Measure
    - Update FHIR Search NearLocationHandler
    - Update Bounding interface with Box, Radius types
    - Extend tests to cover the new types, and confirm validity of the types
    - Introduce enumeration with BoundingType
    - Update the QueryBuilder signature
    - Update NearLocationHandler
    - Refactor Units to use units-of-measure definitions
            - StatuteLengths
            - ImperialUnits
            - Metric Units
            - Remove OtherUnits (Angstrom)
            - configure Units in lookup map in UOMManager
    - Refactor FHIRPersistenceUtil to put NEAR processing into fhir-search
    - AbstractQueryBuilder
            - Change the boundingbox to a list of boundingradius
            - Change to checking near (near-distance is deprecated) and add to a
    check to NearLocationHandler to support a generic check
            - Add Bounding, BoundingBox, BoundingRadius
            - Add Tests for Bounding types
    - Update QuerySegmentAggregator to use the single check for near
    - Update JDBCQueryBuilder to match the changes to AbstractQueryBuilder
    - Introduce LocationBehaviorUtil to more consistently model the Location
    signature query
    - Remove from ParameterDAOImplementation near-distance
    - Add SearchConstants.Type 'special'
    - Add JDBCSearchNearTest with 'near'
    - Remove unused call to continue in SearchUtil
    - Update the validation for the BoundingBox
    - Remove the guard for the 'special' type which blocks near processing
    - Add support for the extraction routine in the JDBC visitor code
    (Location.position)
    - Add logicalId to the ParameterVisitorBatchDAO location
    - Add 'special' to the DTO Parameter object


docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -1212,6 +1212,7 @@ This section contains reference information about each of the configuration prop
 |`fhirServer/audit/serviceProperties/geoCity`|string|The Geo City configure for CADF audit logging service.|
 |`fhirServer/audit/serviceProperties/geoState`|string|The Geo State configure for CADF audit logging service.|
 |`fhirServer/audit/serviceProperties/geoCounty`|string|The Geo Country configure for CADF audit logging service.|
+|`fhirServer/search/boundingAreaType`|boolean|True, the bounding area is a Radius, else the bounding area is a box.|
```
lmsurpre: maybe rename it to `fhirServer/search/useBoundingRadius` to make it clear what true/false mean

or, alternatively, make it a string and validate that the value is either `box` or `radius`.

fhir-model/src/main/java/com/ibm/fhir/model/visitor/Visitor.java
```diff
@@ -1347,4 +1348,14 @@
     void visit(java.lang.String elementName, Year value);
     void visit(java.lang.String elementName, YearMonth value);
     void visit(java.lang.String elementName, ZonedDateTime value);
+
+    /**
+     * @return the coordinates for the lat/long.
```
lmsurpre: this isn't right.  it returns true or false which controls whether child elements should be visited.

fhir-model/src/main/java/com/ibm/fhir/model/visitor/Visitor.java
```diff
@@ -1347,4 +1348,14 @@
     void visit(java.lang.String elementName, Year value);
     void visit(java.lang.String elementName, YearMonth value);
     void visit(java.lang.String elementName, ZonedDateTime value);
+
+    /**
+     * @return the coordinates for the lat/long. 
+     * 
+     * @param elementName
+     * @param elementIndex
+     * @param location
```
lmsurpre: this should match the name of the parameter

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dto/Parameter.java
```diff
@@ -220,6 +220,9 @@ public void visit(IParameterVisitor visitor) throws FHIRPersistenceException {
         case URI:
             visitor.stringValue(name, this.valueString, baseLevel);
             break;
+        case SPECIAL:
```
lmsurpre: i think there is one other parameter with type=special ...maybe we should verify that we have a 'near' search parameter here (just to be safe)

prb112: I checked this out using the search-parameters.json and the website.

prb112: only 1

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/type/LocationParmBehaviorUtil.java
```diff
@@ -0,0 +1,160 @@
...
+import com.ibm.fhir.search.location.bounding.Bounding;
+import com.ibm.fhir.search.location.bounding.BoundingBox;
+import com.ibm.fhir.search.location.bounding.BoundingRadius;
+
+/**
+ * Location Behavior Util generates SQL and loads the variables into bind
+ * variables.
+ * <br>
+ * The bounding areas at this point are already valid as the fhir-search
+ * component has processed the received point into a set of bounding boxes.
```
lmsurpre: is this right?  i see two buildLocationSearchQuery methods: one takes a List of Boundings (not boxes) and the other takes a `single` bounding box.

prb112: There are actually three methods
1 - coordinates the type of the boundingarea 
2 - for the radius 
3 - for the box

fhir-persistence-jdbc/src/test/java/com/ibm/fhir/persistence/jdbc/search/test/JDBCSearchNearTest.java
```diff
@@ -0,0 +1,296 @@
...
+        }
+
+        FHIRSearchContext ctx = SearchUtil.parseQueryParameters(Location.class, queryParms, true);
+        FHIRPersistenceContext persistenceContext = FHIRPersistenceContextFactory.createPersistenceContext(null, ctx);
+        MultiResourceResult<Resource> result = persistence.search(persistenceContext, Location.class);
+        return result;
+    }
+
+    @AfterClass
+    public void teardown() throws Exception {
```
lmsurpre: should we also delete the saved resource(s)?

lmsurpre: should we also delete the saved resource(s)?

prb112: sure ... that's no problem to add

fhir-persistence-jdbc/src/test/java/com/ibm/fhir/persistence/jdbc/search/test/JDBCSearchNearTest.java
```diff
@@ -0,0 +1,296 @@
...
+public class JDBCSearchNearTest {
+    private Properties testProps;
+
+    protected Location savedResource;
+
+    protected static FHIRPersistence persistence = null;
+
+    @BeforeClass
+    public void startup() throws Exception {
+        InputStream inputStream = JDBCSearchNearTest.class.getResourceAsStream("/near.unitTest.properties");
```
lmsurpre: does it need its own unitTest properties file?  why not just re-use `logging.unitTest.properties`?

prb112: I turn on com.ibm.fhir.search trace with this one.

prb112: also exists in fhir-persistence-jdbc

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dto/Parameter.java
lmsurpre: 2 listed at https://www.hl7.org/fhir/search.html#special
* near
* _filter

but yeah, if `_filter` doesn't appear in search-parameters.json then i guess we're good

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/type/LocationParmBehaviorUtil.java
lmsurpre: ok, i guess my point was that the "the fhir-search component has processed the received point into a set of bounding boxes" comment doesn't see to belong here (at the class level)...maybe move it to one of the methods or just delete it?

fhir-persistence-jdbc/src/test/java/com/ibm/fhir/persistence/jdbc/search/test/JDBCSearchNearTest.java
lmsurpre: yeah, i guess i was just thinking you could add a line for `com.ibm.fhir.search` in the existing `logging.unitTest.properties` file.  both will be INFO by default and you can selectively change the log level.  i don't see why we need a separate file for it, but i guess its not a big deal.

prb112: right.  just explaining my choice.  I can reconcile with the existing one

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dto/Parameter.java
prb112: it is not.

fhir-model/src/main/java/com/ibm/fhir/model/visitor/Visitor.java
prb112: fixed in code generator.  I modified this class by hand (as you can see above).  I have fixed this case.

fhir-model/src/main/java/com/ibm/fhir/model/visitor/Visitor.java
prb112: fixed in code generator.  I modified this class by hand (as you can see above).  I have fixed this case.

fhir-persistence-jdbc/src/test/java/com/ibm/fhir/persistence/jdbc/search/test/JDBCSearchNearTest.java
prb112: Reconciled.

fhir-persistence-jdbc/src/test/java/com/ibm/fhir/persistence/jdbc/search/test/JDBCSearchNearTest.java
prb112: done

docs/src/pages/guides/FHIRServerUsersGuide.md
prb112: Fixed

prb112:
```suggestion
|`fhirServer/search/useBoundingRadius`|boolean|True, the bounding area is a Radius, else the bounding area is a box.|
```

fhir-model/src/main/java/com/ibm/fhir/model/visitor/DefaultVisitor.java
```diff
@@ -29,8 +29,8 @@
  * <ul>
  * <li>preVisit methods to control whether a given Resource or Element gets visited
  * <li>visitStart methods to provide setup behavior prior to the visit
- * <li>supertype visit methods to perform some common action on all visited Resources and Elements
- * <li>subtype visit methods to perform unique behavior that varies by the type being visited
+ * <li>defaultAction methods to perform some common action on all visited Resources and Elements
+ * <li>specific visit methods to perform unique behavior that varies by the type being visited
```
lmsurpre: I forgot this class was generated and I updated this javadoc by hand.  Regenerating the model has overwritten these edits.  Mind updating the codeGenerator to preserve the `supertype`/`subtype` version?

prb112: got it.  updating the code generator

Issue #317, opened 2019-10-28
Support for search parameters of type Composite


lmsurpre:
**Describe the bug**
In DSTU2, there are only 2 or so `composite` SearchParameter types.  As noted in https://github.ibm.com/watson-health-fhir-server/fhir-docs/blob/master/Conformance.md
In DSTU2, the composite SearchParameters are disabled (removed from the `search-parameters.json`.  

Per the Spec https://www.hl7.org/fhir/valueset-search-param-type.html
`composite | Composite | A composite search parameter that combines a search on two values together.`

The SearchParameter drives a composite key from the parameter.  In this case, when running an Observation resource, the entire resource is dropped into a single output value using the expression `"expression" : "value.as(string)"`. 

**To Reproduce**
In `SearchUtil`, there is a shortcircuit of the value `public static <T> Map<SearchParameter, List<FHIRPathNode>> extractParameterValues(T resource) throws Exception {`  

- Disable the exception `!SearchParamType.COMPOSITE.equals(type)`.  
- Ensure that there is at least one value.as()
```
  {
    "fullUrl" : "http://hl7.org/fhir/SearchParameter/Questionnaire-context-type-quantity",
    "resource" : {
      "resourceType" : "SearchParameter",
      "id" : "Questionnaire-context-type-quantity",
      "extension" : [{
        "url" : "http://hl7.org/fhir/StructureDefinition/structuredefinition-standards-status",
        "valueCode" : "trial-use"
      }],
      "url" : "http://hl7.org/fhir/SearchParameter/Questionnaire-context-type-quantity",
      "version" : "4.0.0",
      "name" : "context-type-quantity",
      "status" : "draft",
      "experimental" : false,
      "date" : "2018-12-27T22:37:54+11:00",
      "publisher" : "Health Level Seven International (FHIR Infrastructure)",
      "contact" : [{
        "telecom" : [{
          "system" : "url",
          "value" : "http://hl7.org/fhir"
        }]
      },
      {
        "telecom" : [{
          "system" : "url",
          "value" : "http://www.hl7.org/Special/committees/fiwg/index.cfm"
        }]
      }],
      "description" : "A use context type and quantity- or range-based value assigned to the questionnaire",
      "code" : "context-type-quantity",
      "base" : ["Questionnaire"],
      "type" : "composite",
      "expression" : "Questionnaire.useContext",
      "xpathUsage" : "normal",
      "multipleOr" : false,
      "multipleAnd" : true,
[...]

PR #499, opened 2019-12-11
Issue #317 - Composite Search

lmsurpre:
1. Converted JDBC DTO Parameter into a "proper" visitor pattern. This
makes adding a CompositeParameter a lot easier and generally cleans
things up a bit.

2. Introduced CompositParameter which contains a list of component
parameters of any type.

3. Stubbed out Composite search tests, including:
    * extension-search-parameters.json with a single composite search parm
    * BasicComposite.json with a repeating element for testing composite
    * AbstractSearchCompositeTest stub for laying out the test logic
    * JDBCSearchCompositeTest for executing the tests on Derby

4. SearchUtil and ParametersUtil updates to include SearchParameter
canonical uris as an alternative key to the SearchParameter code
    * Also updated fhir-search tests to double the expected search
parameter counts accordingly

5. Implemented initial CompositeParameter extraction logic in
FHIRPersistenceJDBCImpl; needs further cleanup

6. Introduce support for db-generated identity columns in
fhir-database-utils. Instead of adding to columndefinition, I added it
to Table...similar to how primaryKey was implemented.

7. Added ROW_ID columns for all X_Y_VALUES tables and set to `GENERATED
BY DEFAULT` and made primary key to ensure uniqueness.

8. Created X_COMPOSITES table with foreign keys to the X_Y_VALUES
ROW_IDs. Also renamed FhirResourceGroup to FhirResourceTableGroup to
better reflect its purpose.

9. Added visit(CompositeParmVal) variant to
ExtractedParameterValueVisitor and implemented initial DTO logic in
ParameterVisitorBatchDAO; now the components are saved to the X_Y_VALUES
tables and the composite row is written using their generated ROW_IDs
(obtained from `preparedStatement.getGeneratedKeys()`)

10. Updated SearchUtil to parse composite parameters, both for regular and chained searches

11. Added processComposite method to AbstractQueryBuilder and JDBCQueryBuilder:
    * where clause gets generated in processCompositeParm
    * for "normal" requests, the query segment gets manipulated by QuerySegmentAggregator
    * for "chained" requests, the query segment is manipulated from JDBCQueryBuilder.appendInnerSelect

12. Expanded Composite search tests (and fixed related issues):
* at least one test for each search parameter table type we support (except `Location`)
* chained searches (where the composite parameter is at the end of a reference chain)
* support for the `missing` modifier on composite parameters
* 'or' behavior (e.g. `val1$val2,val1$val3`)

13. Updated Db2 stored proc to remove composite parameters. Also removed the outdated `addParameterArrayTypes.sql` stored proc and the associated array types which were still in our schema but no longer used.

14. Addressed issue #508 by adding a dependency from the resource tables to the global parameter value tables...now the global parameter value tables should always be added first and this should hopefully avoid the deadlock.


fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/common/CommonDatabaseAdapter.java
```diff
@@ -114,12 +116,12 @@ protected String buildColumns(List<ColumnBase> columns) {
      * @param tablespaceName
      * @return
      */
-    protected String buildCreateTableStatement(String schema, String name, List<ColumnBase> columns, PrimaryKeyDef pkDef, String tablespaceName) {
+    protected String buildCreateTableStatement(String schema, String name, List<ColumnBase> columns, PrimaryKeyDef pkDef, IdentityDef identity, String tablespaceName) {
         StringBuilder result = new StringBuilder();
         result.append("CREATE TABLE ");
         result.append(getQualifiedName(schema, name));
         result.append("(");
```
prb112: small change here.  
```suggestion
        result.append('(');
```

prb112: changed to character which is slightly smaller

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/Generated.java
```diff
@@ -0,0 +1,19 @@
...
+
+package com.ibm.fhir.database.utils.model;
+
+public enum Generated {
+    ALWAYS, BY_DEFAULT;
+    
+    @Override
+    public String toString() {
+        switch (this) {
+        case BY_DEFAULT: return "BY DEFAULT";
```
prb112: I realize this is a return, and naturally breaks.  maybe a comment? explaining the expected behavior ?

fhir-examples/src/main/resources/json/ibm/basic/BasicComposite.json
```diff
@@ -0,0 +1,144 @@
...
+        },
+        {
+          "url": "http://example.org/dateTime",
+          "valueDateTime": "2018-10-29T17:12:00-04:00"
+        },
+        {
+          "url": "http://example.org/time",
+          "valueTime": "17:12:00"
+        },
+
```
prb112:
```suggestion
```
Trying to remove the extra line

fhir-search/src/test/java/com/ibm/fhir/search/parameters/MultiTenantSearchParameterTest.java
```diff
@@ -53,7 +53,7 @@ public void testGetApplicableSearchParameters2() throws Exception {
         List<SearchParameter> result = SearchUtil.getApplicableSearchParameters("Medication");
         assertNotNull(result);
         printSearchParameters("testGetApplicableSearchParameters2", result);
-        assertEquals(23, result.size());
+        assertEquals(23 * 2, result.size());
```
prb112: `* 2, ` has a particular purpose.  Can we throw this in a Constant that explains it? 
SearchTest.NUMBER_2_DOES_X?

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -82,12 +85,12 @@
...
-    private static final String UNSUPPOTED_EXPR_NULL =
+    private static final String UNSUPPORTED_EXPR_NULL =
             "An empty expression is found or the parameter type is unsupported [%s][%s]";
-    
+
     private static final String MODIFIYERRESOURCETYPE_NOT_ALLOWED_FOR_RESOURCETYPE =
             "Modifier resource type [%s] is not allowed for search parameter [%s] of resource type [%s].";
-    
+
     private static final String Different_MODIFIYERRESOURCETYPES_FOUND_FOR_RESOURCETYPES =
```
prb112: While you are in here... can you refactor/rename to all upper case?

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersUtil.java
```diff
@@ -224,9 +224,9 @@ public static void init() {
                 if (parameter != null) {
 
                     /*
-                     * In R4, SearchParameter changes from a single Base resource to an array In prior releases, the
-                     * code transformed VirtualResources to Basic. As Base is an array, there are going to result in
-                     * potential collisions in the map. transformParameter and transformPath
+                     * In R4, SearchParameter changes from a single Base resource to an array.
+                     * In prior releases, the code transformed VirtualResources to Basic.
```
prb112:
```suggestion
```

prb112: Remove reference to VirtualResource

fhir-search/src/main/java/com/ibm/fhir/search/SearchConstants.java
```diff
@@ -140,21 +140,30 @@ private SearchConstants() {
...
-                    put(SearchConstants.Type.DATE, Arrays.asList(SearchConstants.Modifier.MISSING));
-                    put(SearchConstants.Type.QUANTITY, Arrays.asList(SearchConstants.Modifier.MISSING));
+                    put(SearchConstants.Type.STRING, Arrays.asList(Modifier.EXACT, Modifier.CONTAINS, Modifier.MISSING));
+                    put(SearchConstants.Type.REFERENCE, Arrays.asList(Modifier.TYPE, Modifier.MISSING));
+                    put(SearchConstants.Type.URI, Arrays.asList(Modifier.BELOW, Modifier.ABOVE, Modifier.MISSING));
+                    put(SearchConstants.Type.TOKEN, Arrays.asList(Modifier.BELOW, Modifier.NOT, Modifier.MISSING));
+                    put(SearchConstants.Type.NUMBER, Arrays.asList(Modifier.MISSING));
+                    put(SearchConstants.Type.DATE, Arrays.asList(Modifier.MISSING));
+                    put(SearchConstants.Type.QUANTITY, Arrays.asList(Modifier.MISSING));
+                    put(SearchConstants.Type.COMPOSITE, Arrays.asList(Modifier.MISSING));
```
prb112: this might be where a conflict comes up... with 'near'

fhir-persistence-schema/pom.xml
```diff
@@ -30,7 +30,6 @@
         <dependency>
             <groupId>com.ibm.db2</groupId>
             <artifactId>jcc</artifactId>
-            <scope>test</scope>
```
prb112: 👍 perfectly fine with this approach.

fhir-persistence-schema/README.md
```diff
@@ -83,39 +110,24 @@ Edit `wlp/usr/servers/fhir-server/config/TNT1/fhir-server-config.json` and add t
...
+1. Print the schema to files by executing the SchemaPrinter:
+    `java -cp ./fhir-database-utils.jar:fhir-persistence-schema.jar com.ibm.fhir.schema.app.SchemaPrinter --to-file`
 
-# Creating the database 
+2. Connect to your instance and execute each of the following:
+    - schema.sql:  `db2 -tvf schema.sql`
+    - grants.sql:  `db2 -tvf grants.sql`
+    - stored-procedures.sql:  `db2 -td@ -vf stored-procedures.sql`
 
-To create the database and database user, please use the following command.
```
prb112: please add this back.  I actually ran into an issue, and found this helpful (when setting up a local docker image).    Maybe add as an appendix?

prb112: nevermind (didn't link with the above)

fhir-persistence-jdbc/src/test/java/com/ibm/fhir/persistence/jdbc/test/util/QuantityParmBehaviorUtilTest.java
```diff
@@ -242,61 +241,18 @@ public int acquireCodeSystemId(String codeSystemName) throws FHIRPersistenceExce
...
+                // do nothing
             }
 
             @Override
             public void addParameterNamesCacheCandidate(String parameterName, Integer parameterId)
                     throws FHIRPersistenceException {
-
-            }
-
-            @Override
```
prb112: Nice to see the unused code go.

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirResourceTableGroup.java
```diff
@@ -511,6 +554,79 @@ public void addQuantityValues(List<IDatabaseObject> group, String prefix) {
...
+ALTER TABLE device_composites ADD CONSTRAINT fk_device_composites_rid  FOREIGN KEY (comp1_number)   REFERENCES device_number_values;
+ALTER TABLE device_composites ADD CONSTRAINT fk_device_composites_rid  FOREIGN KEY (comp1_date)     REFERENCES device_date_values;
+ALTER TABLE device_composites ADD CONSTRAINT fk_device_composites_rid  FOREIGN KEY (comp1_token)    REFERENCES device_token_values;
+ALTER TABLE device_composites ADD CONSTRAINT fk_device_composites_rid  FOREIGN KEY (comp1_quantity) REFERENCES device_quantity_values;
+ALTER TABLE device_composites ADD CONSTRAINT fk_device_composites_rid  FOREIGN KEY (comp1_latlng)   REFERENCES device_latlng_values;
+...
+ALTER TABLE device_quantity_values ADD CONSTRAINT fk_device_quantity_values_pn FOREIGN KEY (parameter_name_id) REFERENCES parameter_names ON DELETE CASCADE;
+ALTER TABLE device_quantity_values ADD CONSTRAINT fk_device_quantity_values_r  FOREIGN KEY (resource_id)       REFERENCES device_logical_resources ON DELETE CASCADE;
+     * </pre>
+     * @param prefix
```
prb112: missing the param group

fhir-search/src/main/java/com/ibm/fhir/search/parameters/QueryParameter.java
```diff
@@ -95,78 +93,73 @@ public String toString() {
...
             String valueSystem = value.getValueSystem();
             if (valueSystem != null) {
                 buffer.append("    valueSystem: " + valueSystem);
-                buffer.append(SearchConstants.NL);
+                buffer.append(NL);
             }
 
             String valueCode = value.getValueCode();
             if (valueCode != null) {
                 buffer.append("    valueCode: " + valueCode);
```
prb112: we should probably split this into two appends. 
assemble string1 and string 2 into string 3
append string 3

fhir-search/src/main/java/com/ibm/fhir/search/parameters/QueryParameterValue.java
```diff
@@ -112,6 +143,18 @@ public String toString() {
...
         outputBuilder(returnString, valueString);
         outputBuilder(returnString, valueDate);
+        
+        if (component != null && !component.isEmpty()) {
+            returnString.append("composite[");
+            // temporarily change the delimiter; NOT thread-safe
+            String oldDelim = delim;
+            delim = "";
+            for (QueryParameter componentParam : component) {
+                returnString.append(delim + componentParam);
```
prb112:
```suggestion
                returnString.append(delim).append(componentParam);
```

fhir-search/src/main/java/com/ibm/fhir/search/parameters/QueryParameterValue.java
```diff
@@ -112,6 +143,18 @@ public String toString() {
         outputBuilder(returnString, valueCode);
         outputBuilder(returnString, valueString);
         outputBuilder(returnString, valueDate);
+        
+        if (component != null && !component.isEmpty()) {
+            returnString.append("composite[");
+            // temporarily change the delimiter; NOT thread-safe
```
prb112: Why is this not threadsafe?  it's in the threadlocal?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -314,12 +315,75 @@ public JDBCOperator getPrefixOperator(ParameterValue queryParmValue) {
...
+                case TOKEN:     databaseQueryParm = this.processTokenParm(queryParm, tableAlias);
+                        break;
+                case NUMBER:    databaseQueryParm = this.processNumberParm(resourceType, queryParm, tableAlias);
+                        break;
+                case QUANTITY:  databaseQueryParm = this.processQuantityParm(resourceType, queryParm, tableAlias);
+                        break;
+                case URI:       databaseQueryParm = this.processUriParm(queryParm, tableAlias);
+                        break;
+                case COMPOSITE: databaseQueryParm = this.processCompositeParm(resourceType, queryParm, tableAlias);
+                        break;
```
prb112: I'm expecting after the merge, there will have to be case SPECIAL:

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -605,35 +669,44 @@ private void appendInnerSelect(StringBuilder whereClauseSegment, Parameter curre
...
-        case TOKEN:
-            whereClauseSegment.append(resourceTypeName + "_TOKEN_VALUES ");
-            break;
-        case REFERENCE:
-        case URI:
-        case STRING:
-        default:
-            whereClauseSegment.append(resourceTypeName + "_STR_VALUES ");
-            break;
+        whereClauseSegment.append(" FROM ")
```
prb112: I think FROM is also a JDBCConstant

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1124,6 +1197,79 @@ private SqlQueryData processQuantityParm(Class<?> resourceType, Parameter queryP
...
+            List<QueryParameter> components = compositeValue.getComponent();
+            
+            whereClauseSegment.append(valueSeparator);
+            
+            String componentSeparator = "";
+            for (int j = 0; j < components.size(); j++) {
+                try {
+                    QueryParameter component = components.get(j);
+                    SqlQueryData subQueryData = buildQueryParm(resourceType, component, tableAlias + "_p" + (j+1));
+                    whereClauseSegment.append(componentSeparator + subQueryData.getQueryString());
```
prb112: does componentSeparator ever change? 
also split into separate append operations.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1124,6 +1197,79 @@ private SqlQueryData processQuantityParm(Class<?> resourceType, Parameter queryP
...
+        for (QueryParameterValue compositeValue : queryParm.getValues()) {
+            List<QueryParameter> components = compositeValue.getComponent();
+            
+            whereClauseSegment.append(valueSeparator);
+            
+            String componentSeparator = "";
+            for (int j = 0; j < components.size(); j++) {
+                try {
+                    QueryParameter component = components.get(j);
+                    SqlQueryData subQueryData = buildQueryParm(resourceType, component, tableAlias + "_p" + (j+1));
```
prb112: the max for j+1 is 3 ? 
throw an illegal value? or prune at 3? or stop the for loop at 3?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
prb112: sorry... down the stream it's set.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1124,6 +1197,79 @@ private SqlQueryData processQuantityParm(Class<?> resourceType, Parameter queryP
...
+                    whereClauseSegment.append(componentSeparator + subQueryData.getQueryString());
+                    bindVariables.addAll(subQueryData.getBindVariables());
+                } catch (Exception e) {
+                    throw new FHIRPersistenceException("Error while creating subquery for parameter " + queryParm, e);
+                }
+                componentSeparator = AND;
```
prb112: I think we should apply this pattern to all of the CODE HERE.

fhir-examples/src/main/resources/json/ibm/basic/BasicComposite.json
lmsurpre: if JSON had comments, this would be a comment to indicate that the "primitive" types are over and now the "complex" types are starting...certainly not obvious from an empty line but still useful to me

fhir-persistence-schema/README.md
lmsurpre: yeah, i felt like "how to see what the schema Main app is doing" should go at the top, but still tried to be explicit that users can either:
* A. deploy the schema with the tool; or
* B. manually apply the DDL from the printed SQL files

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirResourceTableGroup.java
lmsurpre: it turns out the javadoc for this parameter was missing form every other method in this class as well...will add it everywhere

fhir-search/src/main/java/com/ibm/fhir/search/SearchConstants.java
lmsurpre: Actually there no conflict on this part since Location search doesn't currently support any modifiers.
When we get to #510, we'll need to add Type.SPECIAL to this list with a singleton of `Modifier.MISSING`.

fhir-search/src/main/java/com/ibm/fhir/search/parameters/QueryParameter.java
lmsurpre: Not sure how much this matters; the one thing I did is add a static import for `NL`.  And since I did that for one spot, I went through and removed the unneccessary class qualifier everywhere else...the rest was already here.

fhir-search/src/main/java/com/ibm/fhir/search/parameters/QueryParameterValue.java
lmsurpre: Because, for some reason, the `delim` is a member of the class and not just a a local var.
If two threads are somehow sharing the instance, they will fight over who "owns" the delim...leading to unpredictable toString behavior.

I think its ugly, but this was already the design and so instead of changing that I just wanted to flag it.

lmsurpre: That said, if the class is only ever used from a threadlocal, then we shouldn't really ever see an issue with this.  I don't know if thats the case or not.

lmsurpre: Actually, looking at this again, I don't know why I didn't just define my own local var.  I'll do that now.
I'll leave the original `delim` member alone, but I still think thats ugly.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
lmsurpre: yep

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
lmsurpre: yeah, really wasn't sure the best spot to put the max number of components.
at first, i was thinking it should go in the search layer, but the limitation is really a DB one...theoretically you could define a search parameter with more than 3 components but our implementation just won't support that

lmsurpre: FWIW we do have this check in `SearchUtil.parseCompositeQueryParameterValuesString`
```
            if (compTypes.size() != componentValueStrings.length) {
                throw new FHIRSearchException(String.format("Expected %d components but found %d in composite query value '%s'", 
                    compTypes.size(), componentValueStrings.length, v));
            }
```
That should ensure that a query string value can only contain the number of components included in the search parameter definition.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/common/CommonDatabaseAdapter.java
lmsurpre: done

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/Generated.java
lmsurpre: added some javadoc

fhir-search/src/test/java/com/ibm/fhir/search/parameters/MultiTenantSearchParameterTest.java
lmsurpre: Because we're now indexin the list by both the code and the canonical uri, the expected size is now 2 times the number of search parameters.
I don't even love these tests which just happen to know the number of expected parameters in the file...seems like it would be better to just compute it from the file...otherwise its for sure gonna break each time we add or remove a search parameter.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
lmsurpre: done

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersUtil.java
lmsurpre: done

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirResourceTableGroup.java
lmsurpre: done

fhir-search/src/main/java/com/ibm/fhir/search/parameters/QueryParameter.java
lmsurpre: didn't do this one, but can if its important to you

fhir-search/src/main/java/com/ibm/fhir/search/parameters/QueryParameterValue.java
lmsurpre: done

fhir-search/src/main/java/com/ibm/fhir/search/parameters/QueryParameterValue.java
lmsurpre: done

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
lmsurpre: OK, I introduced a constant `JDBCConstants.MAX_NUM_OF_COMPOSITE_COMPONENTS` with a value of 3 and added a guard in both `FHIRPersistenceJDBCImpl.extractSearchParameters` (extraction time) and `JDBCQueryBuilder.processCompositeParm` (query time).

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
lmsurpre: its hard to do anything consistent inside this class, but yeah I've grown fond of this approach for delimiting complex content (for simple strings you can just use `String.join` now)

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
lmsurpre: done

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
```diff
@@ -558,10 +566,10 @@ public boolean visit(java.lang.String elementName, int elementIndex, Timing timi
 
     @Override
     public boolean visit(java.lang.String elementName, int elementIndex, Location.Position position) {
-        Parameter p = new Parameter();
+        LocationParmVal p = new LocationParmVal();
         p.setName(searchParamCode);
         
-        // The following code ensures that the lat/lon is only added when there is a pair. 
+        // The following code ensures that the lat/long is only added when there is a pair.
```
prb112:
```suggestion
        // The following code ensures that the lat/lon is only added when there is a pair. 
```

prb112: this comment refers to the lat/lon that follow. 
long is a reserved word, so I wanted to avoid it in this comment

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
```diff
@@ -619,18 +627,18 @@ private void setDateValues(Parameter p, DateTime dateTime) {
...
      * @param end
      */
-    private void setDateValues(Parameter p, java.time.Instant start, java.time.Instant end) {
+    private void setDateValues(DateParmVal p, java.time.Instant start, java.time.Instant end) {
         if (start != null) {
             Timestamp startTime = Timestamp.from(start);
             p.setValueDateStart(startTime);
             p.setValueDate(startTime);
-            p.setTimeType(TimeType.UNKNOWN);
+//            p.setTimeType(TimeType.UNKNOWN);
```
prb112: should we remove this line? 
```suggestion
```

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
```diff
@@ -335,60 +336,103 @@ protected String buildWhereClause(String overrideType) {
...
-                        whereClause.append(tmpTableName).append(ON).append(tmpTableName).append(".LOGICAL_RESOURCE_ID = R.LOGICAL_RESOURCE_ID");
                     }
+                    whereClause.append(" WHERE ").append(whereClauseSegment).append(") ");
+                    String tmpTableName = overrideType + i;
+                    whereClause.append(tmpTableName).append(ON).append(tmpTableName).append(".LOGICAL_RESOURCE_ID = R.LOGICAL_RESOURCE_ID");
                 }
-            }
-        }
+            } // end if SKIP
+        } // end for
```
prb112: very helpful to understanding the code.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
lmsurpre: yes, good catch

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
lmsurpre: reverted

fhir-search/src/main/java/com/ibm/fhir/search/parameters/QueryParameter.java
prb112: no worries. this is a-ok

PR #509, opened 2019-12-16
Issue #483 - 'memberOf' function refactor

JohnTimm:

### DIFF
--- a/fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
+++ b/fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
@@ -13,30 +13,39 @@ import static com.ibm.fhir.path.util.FHIRPathUtil.getElementNode;
 import static com.ibm.fhir.path.util.FHIRPathUtil.getString;
 import static com.ibm.fhir.path.util.FHIRPathUtil.hasElementNode;
 import static com.ibm.fhir.path.util.FHIRPathUtil.hasStringValue;
+import static com.ibm.fhir.profile.ValueSetSupport.expand;
+import static com.ibm.fhir.profile.ValueSetSupport.getContains;
+import static com.ibm.fhir.profile.ValueSetSupport.getValueSet;
 
 import java.util.Collection;
+import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
-import com.ibm.fhir.model.resource.CodeSystem;
 import com.ibm.fhir.model.resource.ValueSet;
-import com.ibm.fhir.model.resource.ValueSet.Compose;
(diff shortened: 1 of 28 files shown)


fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
```diff
@@ -95,65 +107,88 @@ public int getMaxArity() {
...
-        values.add(concept.getCode().getValue());
-        for (CodeSystem.Concept c : concept.getConcept()) {
-            values.addAll(getValues(c));
+        for (FHIRPathNode child : node.children()) {
+            if ("system".equals(child.name()) && FHIRPathType.FHIR_URI.equals(node.type())) {
+                return child.asElementNode().element().as(Uri.class).getValue();
+            }
         }
-        return values;
+        return null;
```
lmsurpre: worth adding method doc to explain when it returns null and what it means?

fhir-path/src/test/java/com/ibm/fhir/path/test/MemberOfFunctionTest.java
```diff
@@ -0,0 +1,106 @@
...
+import org.testng.annotations.Test;
+
+import com.ibm.fhir.model.type.Code;
+import com.ibm.fhir.model.type.CodeableConcept;
+import com.ibm.fhir.model.type.Coding;
+import com.ibm.fhir.model.type.Uri;
+import com.ibm.fhir.path.FHIRPathNode;
+import com.ibm.fhir.path.evaluator.FHIRPathEvaluator;
+
+public class MemberOfFunctionTest {
```
lmsurpre: any javadoc worth adding here?  maybe high-level overview of the function (copy from operationDefinition?) with highlights for any "decisions" you made (any ambiguity in spec?).  for example, it might be a good spot to summarize the high-level algorithm for matching concepts in versioned systems...like whether we behave differently for systems where versionNeeded is true vs false (e.g. see https://chat.fhir.org/#narrow/stream/179202-terminology/topic/CodeSystem.2EversionNeeded)

fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
JohnTimm: Added some javadoc to the `MemberOfFunction.getSystem` method

fhir-path/src/test/java/com/ibm/fhir/path/test/MemberOfFunctionTest.java
JohnTimm: My plan is to create a README.md in the repo and/or on the GH pages site that explains the capability in more detail with some examples.

lmsurpre: FYI I intended for this comment to be on MemberOfFunction, not MemberOfFunctionTest

JohnTimm: Added some additional javadoc to `MemberOfFunction` class.

PR #520, opened 2019-12-19
Update Db2Adapter.createFhirSchemas to succeed when schemas exist

lmsurpre:
Without this, the step always fails if you've already created a schema
on this db because FHIR_ADMIN already exists. Now it will succeed but
print a WARNING to indicate it already existed.

1. introduced DuplicateNameException and a corresponding helper to
identify when schema creation fails due to the schema already existing

2. renamed DuplicateValueException to UniquenessViolationException to
avoid confusion with DuplicateNameException

3. updated Db2Adapter to catch the DuplicateNameException and WARN
instead of failing

Also added log statement at the end of tenant creation so that you don't
need to scroll to the very top of the output to find the generated key


prb112:
interesting, the fhir-audit changes are in the PR.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/api/IDatabaseTranslator.java
```diff
@@ -50,6 +50,14 @@
      */
     boolean isDuplicate(SQLException x);
     
+    /**
+     * Check the exception to see if it is reporting that THE NAME OF THE OBJECT TO BE CREATED 
+     * OR THE TARGET OF A RENAME STATEMENT IS IDENTICAL TO THE EXISTING NAME OF THE OBJECT TYPE
+     * @param x
+     * @return
+     */
+    boolean isAlreadyExists(SQLException x);
```
prb112: the name is a bit convoluted. 

isAlreadyExists  ... may be alternative 
alreadyExists
exists

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/api/UniquenessViolationException.java
```diff
@@ -9,16 +9,16 @@
 /**
  * Translation of a duplicate key or value constraint SQLException.
  */
-public class DuplicateValueException extends DataAccessException {
+public class UniquenessViolationException extends DataAccessException {
```
prb112: UniqueConstraintViolationException?

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/api/IDatabaseTranslator.java
lmsurpre: yeah, agree its a bit akward.  i went with `is` to match the other methods which are similar in nature.  its saying "is the SQLException an AlreadyExists one".

I had the same feeling about the existing `isDuplicate(SQLException x)` method name as well...

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/api/UniquenessViolationException.java
lmsurpre: I actual bounced between that and this one.  Will change to that.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/api/IDatabaseTranslator.java
prb112: I have no problemo with it.

lmsurpre:
> interesting, the fhir-audit changes are in the PR.

Sorry, that was just a bit lazy on my part.  I created this PR first, then did those changes on top of these.  Then, instead of cherry-picking it to a new one from master, I just opened that PR against this one.  Just disregard those changes since you already reviewed them (and keep in mind they aren't in master until we merge this one).

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/api/UniquenessViolationException.java
lmsurpre: done

Issue #528, opened 2019-12-20
All timestamps should be stored in DB as UTC


lmsurpre:
**Describe the bug**
We thought this was this case, but I found it is not.

**To Reproduce**
1. Run JDBCSearchDateTest.java
2. Look at the BASIC_RESOURCES table's LAST_UPDATED column

**Expected behavior**
The database value should be stored as UTC.

PR #529, opened 2019-12-21
issue #528 - update Conformance.md and tweak datetime precision

lmsurpre:
1. added a bunch more date and datetime tests to AbstractSearchDateTest
(and reorganized some ones that were there)

2. found and fixed one problematic case:

We were capturing the precision wrong for Datetime values which include
0 seconds.  Our algorithm relied on TemporalAccessor.toString but this
method actually excludes the seconds when they are 0. This was causing
us to store values like `2019-12-31T20:00:00Z` as the range
`[2019-12-31T20:00:00Z, 2019-12-31T20:00:59.999999Z)` instead of
`[2019-12-31T20:00:00Z, 2019-12-31T20:00:00.999999Z)`. I addressed it by
using the DateTime or Date PARSER_FORMATTER (depending on its type).

3. updated BasicDate.json to test with a more "interesting"
dateTime...one that strattles the edge of a year, month, and day.

### DIFF
--- a/docs/src/pages/Conformance.md
+++ b/docs/src/pages/Conformance.md
@@ -127,7 +127,7 @@ If not specified on a query string, the default prefix is `eq`.
 The IBM FHIR Server adheres to the specification with two minor exceptions:
 * The server supports search query values with fractional seconds; any search query value given with fractional seconds is treated as precise value, whereas search query values without fractional seconds are handled as an implicit range (e.g. `2000-04-30T23:59:00` becomes the range `[2000-04-30T23:59:00, 2000-04-30T23:59:01)`).
 * Dates and DateTimes (and query parameter values for date search parameters) which are expressed without timezones are handled as UTC values.
-  * This differs slightly from the specification which indicates that "Where both search parameters and resource element date times do not have time zones, the servers local time zone should be assumed". However, because both the element values AND search query parameters are handled in the same way, this difference will only matter when timezones are specified on one side (resource element or search query parameter) but not the other. For example, a query like `Patient?birthdate=2019-01-01` would match a resource with a value of `2019-01-01`, but would *not* match a resource with a value of `2019-01-01T20:00:00-04:00`.
+  * This differs slightly from the specification which indicates that "Where both search parameters and resource element date times do not have time zones, the servers local time zone should be assumed". However, because both the element values AND search query parameters are handled in the same way, this difference matters only when timezones are specified on one side (resource element or search query parameter) but not the other. For example, a query like `Patient?birthdate=2019-01-01` would match a resource with a value of `2019-01-01`, but would *not* match a resource with a value of `2019-01-01T20:00:00-04:00`.
 
 All search parameter values are stored in the database in UTC time in order to improve data portability.
 
@@ -138,7 +138,7 @@ Query parameter values without fractional seconds will be handled as an implicit
 * 2019-01-01T12:00:00.1Z
 * 2019-01-01T12:00:00.999999Z
 
-Query parameter values with fractional seconds will be handled with exact match semantics (ignoring precision). For example, a search like `Patient?birthdate=2019-01-01T12:00:00.1Z` would include resources with the following effectiveDateTime values:
+Query parameter values with fractional seconds is handled with exact match semantics (ignoring precision). For example, a search like `Patient?birthdate=2019-01-01T12:00:00.1Z` would include resources with the following effectiveDateTime values:
 * 2019-01-01T12:00:00.1Z
 * 2019-01-01T12:00:00.100Z
 * 2019-01-01T12:00:00.100000Z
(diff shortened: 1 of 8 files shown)


docs/src/pages/Conformance.md
```diff
@@ -124,16 +124,21 @@ The `eb` and `ap` prefixes are not supported for searches which target values of
...
+The IBM FHIR Server adheres to the specification with two minor exceptions:
+* The server supports search query values with fractional seconds; any search query value given with fractional seconds is treated as precise value, whereas search query values without fractional seconds are handled as an implicit range (e.g. `2000-04-30T23:59:00` becomes the range `[2000-04-30T23:59:00, 2000-04-30T23:59:01)`).
+* Dates and DateTimes (and query parameter values for date search parameters) which are expressed without timezones are handled as UTC values.
+  * This differs slightly from the specification which indicates that "Where both search parameters and resource element date times do not have time zones, the servers local time zone should be assumed". However, because both the element values AND search query parameters are handled in the same way, this difference will only matter when timezones are specified on one side (resource element or search query parameter) but not the other. For example, a query like `Patient?birthdate=2019-01-01` would match a resource with a value of `2019-01-01`, but would *not* match a resource with a value of `2019-01-01T20:00:00-04:00`.
```
prb112:
```suggestion
  * This differs slightly from the specification which indicates that "Where both search parameters and resource element date times do not have time zones, the servers local time zone should be assumed". However, because both the element values AND search query parameters are handled in the same way, this difference matters only when timezones are specified on one side (resource element or search query parameter) but not the other. For example, a query like `Patient?birthdate=2019-01-01` would match a resource with a value of `2019-01-01`, but would *not* match a resource with a value of `2019-01-01T20:00:00-04:00`.
```

docs/src/pages/Conformance.md
```diff
@@ -124,16 +124,21 @@ The `eb` and `ap` prefixes are not supported for searches which target values of
...
+
+The IBM FHIR Server stores up to 6 fractional seconds (microsecond granularity) for Instant and DateTime values.
 
 Query parameter values without fractional seconds will be handled as an implicit range. For example, a search like `Patient?date=2019-01-01T12:00:00Z` would include resources with the following effectiveDateTime values:
 * 2019-01-01T12:00:00Z
 * 2019-01-01T12:00:00.1Z
 * 2019-01-01T12:00:00.999999Z
 
-Query parameter values with fractional seconds will be handled with exact match semantics (ignoring precision). For example, a search like `Patient?date=2019-01-01T12:00:00.100Z` would include resources with the following effectiveDateTime values:
+Query parameter values with fractional seconds will be handled with exact match semantics (ignoring precision). For example, a search like `Patient?birthdate=2019-01-01T12:00:00.1Z` would include resources with the following effectiveDateTime values:
```
prb112:
```suggestion
Query parameter values with fractional seconds is handled with exact match semantics (ignoring precision). For example, a search like `Patient?birthdate=2019-01-01T12:00:00.1Z` would include resources with the following effectiveDateTime values:
```

fhir-persistence/src/test/java/com/ibm/fhir/persistence/search/test/AbstractPLSearchTest.java
```diff
@@ -72,7 +72,7 @@ public void createResources() throws Exception {
     @AfterClass
     public void removeSavedResourcesAndResetTenant() throws Exception {
         if (savedResource != null && persistence.isDeleteSupported()) {
-            persistence.delete(getDefaultPersistenceContext(), Basic.class, savedResource.getId());
+//            persistence.delete(getDefaultPersistenceContext(), Basic.class, savedResource.getId());
```
prb112: do you want to actually do this? 
```suggestion
            persistence.delete(getDefaultPersistenceContext(), Basic.class, savedResource.getId());
```

fhir-search/src/main/java/com/ibm/fhir/search/date/DateTimeHandler.java
```diff
@@ -172,12 +174,26 @@ public static Instant generateValue(TemporalAccessor value) {
...
      */
-    public static Instant generateUpperBound(TemporalAccessor value) {
-        return generateUpperBound(null, value, value.toString());
+    public static Instant generateUpperBound(DateTime value) {
+        // from https://www.hl7.org/fhir/datatypes.html#dateTime:
+        // "Seconds must be provided due to schema type constraints but may be zero-filled and may be ignored at receiver discretion."
+        
+        // use value.getValue().toString() instead if you want to interpret zero-filled seconds ("ss" = 00) 
+        // as a datetime with a precision of 1 minute (i.e. HH:mm rather than HH:mm:ss)
+        return generateUpperBound(null, value.getValue(), DateTime.PARSER_FORMATTER.format(value.getValue()));
```
prb112: 👍 sounds fine

PR #530, opened 2019-12-21
issue #528 - use server timezone to interpret dates

lmsurpre:
This PR builds on #529 and can be taken in whole or in part.

The spec says "Where both search parameters and resource element date
times do not have time zones, the servers local time zone should be
assumed" and so I wanted to see what that would look like.

Its not very clear what is "correct" if only one of those two excludes
time zones. It doesn't matter much on the parameter extraction side
because all datetimes with times in them are required to have a
timezone, but I think I like the idea of handling searches w/o timezone
info via local time, because usually if someone puts a date or datetime
without a timezone they are expecting local time and not UTC time.

In the process, I also found a discrepency between our EQ and NE
behavior. To rectify this, I removed the "exactly equals DATE_START"
omptimization because I think its not always right...for example if the
indexed value was actually a Period then we shold consider an exact
point to be equal to that Period (even if the start matches).

### DIFF
--- a/fhir-search/src/main/java/com/ibm/fhir/search/date/DateTimeHandler.java
+++ b/fhir-search/src/main/java/com/ibm/fhir/search/date/DateTimeHandler.java
@@ -13,7 +13,7 @@ import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.Year;
 import java.time.YearMonth;
-import java.time.ZoneOffset;
+import java.time.ZoneId;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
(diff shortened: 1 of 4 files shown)


fhir-search/src/main/java/com/ibm/fhir/search/date/DateTimeHandler.java
```diff
@@ -141,29 +142,30 @@ public static Instant generateValue(TemporalAccessor value, String originalStrin
      */
     public static Instant generateValue(TemporalAccessor value) {
         Instant response;
+        ZoneRules defaultOffsetRules = ZoneId.systemDefault().getRules();
```
prb112: good catch

Issue #535, opened 2019-12-23
SwaggerGenerator double-writes all search parameters


lmsurpre:
**Describe the bug**
I tried running the SwaggerGenerator from the 4.0.0 release and, unfortunately, it produces invalid Swagger.  OpenApiGenerator has the same issue.
![image](https://user-images.githubusercontent.com/5478798/71373612-d0ab8080-2585-11ea-9fe7-44c9b74c34ae.png)

**To Reproduce**
Execute OpenApiGenerator and validate the result.

**Expected behavior**
Search parameters should only be listed once per type.

**Additional context**
There was a late change leading up to 4.0.0 in which I introduced a second entry in the search parameters map for each search parameter.  In the past we only needed to look up the applicable search parameters by "code" but for Composite search we needed to look them up by "url" as well.
Obviously, I didn't know that would cause this knock-on effect to the Swagger/OpenAPI.

prb112:
interesting side effect.

prb112:
Reviewed, and closing after discussions in PR are complete

PR #537, opened 2019-12-23
issue #535 - Deduplicate the SearchParameter map

lmsurpre:
Previously, the parameter maps built by ParametersUtil had each
SearchParameter listed twice...once by "code" and once by "url".

This pull request introduces a new ParametersMap object that
encapsulates two separate maps and presents it as a single object.

Also removed the notion of "UNSUPPORTED_EXPRESSION" since 
4.0.1 artifacts now have valid FHIRPath and our FHIRPath engine
now supports `resolve()`.

### DIFF
--- a/fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersUtil.java
+++ b/fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersUtil.java
@@ -20,7 +20,6 @@ import java.util.logging.Logger;
 
 import com.ibm.fhir.model.format.Format;
 import com.ibm.fhir.model.parser.FHIRParser;
-import com.ibm.fhir.model.parser.exception.FHIRParserException;
 import com.ibm.fhir.model.resource.Bundle;
 import com.ibm.fhir.model.resource.SearchParameter;
 import com.ibm.fhir.model.type.code.ResourceType;
@@ -84,7 +83,6 @@ public final class ParametersUtil {
      * Loads the built-in search parameters and constructs . 
      * 
      * @return a map of ParametersMaps, keyed by resourceType
-     * @throws FHIRParserException 
      */
     private static Map<String, ParametersMap> loadBuiltIns() {
         Map<String, ParametersMap> map;
@@ -124,30 +122,29 @@ public final class ParametersUtil {
                 if (log.isLoggable(Level.FINE)) {
                     log.fine(String.format(MISSING_EXPRESSION, parameter.getCode().getValue()));
                 }
-                continue;
-            }
-
-            /*
-             * In R4, SearchParameter changes from a single Base resource to an array.
-             * As Base is an array, there are going be potential collisions in the map.
-             */
-            List<ResourceType> types = parameter.getBase();
-            for (ResourceType type : types) {
-                String base = type.getValue();
-
-                ParametersMap map = typeToParamMap.get(base);
-                if (map == null) {
-                    map = new ParametersMap();
-                    typeToParamMap.put(base, map);
+            } else {
+                /*
+                 * In R4, SearchParameter changes from a single Base resource to an array.
+                 * As Base is an array, there are going be potential collisions in the map.
+                 */
+                List<ResourceType> types = parameter.getBase();
+                for (ResourceType type : types) {
+                    String base = type.getValue();
+    
+                    ParametersMap map = typeToParamMap.get(base);
+                    if (map == null) {
+                        map = new ParametersMap();
+                        typeToParamMap.put(base, map);
+                    }
+    
+                    // check and warn if the parameter name and code do not agree.
+                    String code = parameter.getCode().getValue();
+                    String name = parameter.getName().getValue();
+                    checkAndWarnForIssueWithCodeAndName(code, name);
+    
+                    // add the map entry with keys for both the code and the url
+                    map.insert(code, parameter.getUrl().getValue(), parameter);
                 }
-
-                // check and warn if the parameter name and code do not agree.
-                String code = parameter.getCode().getValue();
-                String name = parameter.getName().getValue();
-                checkAndWarnForIssueWithCodeAndName(code, name);
-
-                // add the map entry with keys for both the code and the url
-                map.insert(code, parameter.getUrl().getValue(), parameter);
             }
         }
         
--- a/fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersUtil.java
+++ b/fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersUtil.java
@@ -6,36 +6,24 @@
 
 package com.ibm.fhir.search.parameters;
 
-import static com.ibm.fhir.model.type.String.string;
-
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileNotFoundException;
-import java.io.IOException;
 import java.io.InputStream;
 import java.io.PrintStream;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import java.util.stream.Collectors;
 
-import com.ibm.fhir.exception.FHIRException;
 import com.ibm.fhir.model.format.Format;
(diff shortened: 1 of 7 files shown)


fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersUtil.java
```diff
@@ -79,185 +64,95 @@
...
-
-            FHIRPathEvaluator evaluator = FHIRPathEvaluator.evaluator();
-            EvaluationContext evaluationContext = new EvaluationContext(bundle);
-            Collection<FHIRPathNode> result = evaluator.evaluate(evaluationContext, FHIR_PATH_BUNDLE_ENTRY);
-
-            List<SearchParameter> paramList = result.stream()
-                    .map(node -> node.asResourceNode().resource().as(SearchParameter.class))
-                    .collect(Collectors.toList());
-            for (SearchParameter parameter : paramList) {
+    public static Map<String, ParametersMap> buildSearchParametersMapFromBundle(Bundle bundle) {
```
prb112: any reason for the separation of bundle vs InputStream? curiosity not critique ... 
this is technically now a breaking change. since 
`Map<String, Map<String, SearchParameter>> populateSearchParameterMapFromStreamXML(InputStream stream) {` and the corresponding json method are now public in 4.0.0 

Yes, I highly doubt someone is using this exact method, however, we should carefully treat these method changes going forward.  Once again - discussion point not a critique

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersUtil.java
```diff
@@ -79,185 +64,95 @@
...
-     * resource type.
+     * Loads the built-in search parameters and constructs . 
+     * 
+     * @return a map of ParametersMaps, keyed by resourceType
+     * @throws FHIRParserException 
      */
-    public static Map<String, Map<String, SearchParameter>> getBuiltInSearchParameterMap() {
-        return builtInSearchParameters;
-    }
+    private static Map<String, ParametersMap> loadBuiltIns() {
```
prb112: If we're in a full refactoring mode, I'd suggest extrapolating the relevant signatures to an interface/factory, and the containing class/object into a interface so future changes are opaque to the user.

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersUtil.java
```diff
@@ -79,185 +64,95 @@
...
-        return result;
-    }
-
     /**
-     * This is a convenience function that simply returns the built-in (spec-defined) SearchParameters as a map keyed by
-     * resource type.
+     * Loads the built-in search parameters and constructs . 
+     * 
+     * @return a map of ParametersMaps, keyed by resourceType
+     * @throws FHIRParserException
```
prb112: hum... do you actually throw this?  note the signature is not showing it.

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersUtil.java
```diff
@@ -79,185 +64,95 @@
...
+            }
 
-                // Conditional Logging intentionally avoids forming of the String.
+            if (parameter.getExpression() == null || !parameter.getExpression().hasValue()) {
                 if (log.isLoggable(Level.FINE)) {
-                    log.fine(String.format(LOG_PARAMETERS, parameter.getCode().getValue()));
+                    log.fine(String.format(MISSING_EXPRESSION, parameter.getCode().getValue()));
                 }
+                continue;
+            }
```
prb112: I suggest explicitly adding an else block in the loop.  continues can be lost in a complicated code block.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -459,20 +465,27 @@ public static SearchParameter getSearchParameter(Class<?> resourceType, Canonica
...
         String tenantId = FHIRRequestContext.get().getTenantId();
-
+        SearchParameter result = null;
+        
         // First try to find the search parameter within the specified tenant's map.
-        SearchParameter result =
-                getSearchParameterInternal(getTenantOrDefaultSPMap(tenantId), resourceType, uri.getValue(), false);
+        Map<String, ParametersMap> spMaps = getTenantOrDefaultSPMap(tenantId);
+        if (spMaps != null && !spMaps.isEmpty()) {
+            ParametersMap parametersMap = spMaps.get(resourceType);
```
prb112: 473 and 481 have nearly identical blocks.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -417,21 +414,30 @@ public static SearchParameter getSearchParameter(Class<?> resourceType, String n
 
     /**
      * @param resourceType
-     * @param name
+     * @param code
      * @return
      * @throws Exception
      */
-    public static SearchParameter getSearchParameter(String resourceType, String name) throws Exception {
+    public static SearchParameter getSearchParameter(String resourceType, String code) throws Exception {
```
prb112: good catch

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersUtil.java
lmsurpre: Yeah, not sure how careful to be about preserving backwards compatibility on the public methods...this is one reason i was pushing to better define internal vs external api.

In this particular case, I just thought it was a little cleaner to the map building logic from the reading of the Bundle from disk.  It means we don't need to deal with the JSON vs XML variants or deal with "what if the passed file isn't a Bundle" or what-have-you...basically just a tighter API.

Originally, I had a variant that took a list of SearchParameters instead of a Bundle (since thats all we really need as input), but I ended up removing that since we're already expecting to have a Bundle in the only two places calling this method right now.

Anyway, I'm not dead-set on this particular change (InputStream -> Bundle), so let me know if you'd like to see it reverted in favor of backwards compat.

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersUtil.java
lmsurpre: The diff on this one isn't great (maybe due to the large scope of changes).  We already had a private method called loadBuildIn.  I also preserved a method similar to getBuiltInSearchParameterMap but renamed it to buildSearchParametersMapFromBundle.

As you can see, I didn't really attempt to preserve consistency with the past...just went with whatever felt right to me.  I'm happy to try preserving as much of the former API as desired if thats what we want, although the big change here is that the return types will have the new ParametersMap instead of the old `Map<String, SearchParameter>`.

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersUtil.java
prb112: No change needed at all.  Just thinking ahead.

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersUtil.java
prb112: No worries.  just a point to consider.  No action required.

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersUtil.java
lmsurpre: good catch, will remove

lmsurpre: done

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersUtil.java
lmsurpre: I generally prefer the early exit (or continue in this case) for the exception cases...I don't like long else blocks which lead to extra level of indentation and I forget what was the if/else conditions above.  Just a difference in style preference I suppose, but in this case I went with the suggestion.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
lmsurpre: I wasn't sure if factoring these couple lines into a separate method would be worth it, but I did it and I agree it cleans up the method a bit.

Issue #536, opened 2019-12-23
Include examples in the generated Swagger/OpenAPI


lmsurpre:
**Is your feature request related to a problem? Please describe.**
Tools like IBM API Connect attempt to generate examples, but this can be very difficult with a schema of this size/complexity.  For example, because all elements contain extensions and extensions contain extensions, the schema is infinite and the tools need to inteligently select when "enough is enough".

**Describe the solution you'd like**
We're sitting on a whole pile of example in the `fhir-examples` project, so can we include some of those in the schema?

**Additional context**
Tools like IBM API Connect attempt to generate examples, but this can be very difficult with a schema of this size/complexity.  If we provide our own examples, will they just use those?

prb112:
Reviewed and QAd.  Closing.

PR #538, opened 2019-12-24
issue #536 - introduce examples to the Swagger/OpenAPI

lmsurpre:
By default, we'll now use the first `complete-mock` example for each
applicable resourceType.

In addition, I flipped this to "minimal" and used that to generate a new
version of the "all-in-one" in our `fhir-openapi` project.

### DIFF
--- a/fhir-swagger-generator/pom.xml
+++ b/fhir-swagger-generator/pom.xml
@@ -13,6 +13,12 @@
     <dependencies>
         <dependency>
             <groupId>${project.groupId}</groupId>
+            <artifactId>fhir-examples</artifactId>
+            <version>${project.version}</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
             <artifactId>fhir-core</artifactId>
             <version>${project.version}</version>
         </dependency>
--- a/fhir-model/src/main/java/com/ibm/fhir/model/resource/Resource.java
+++ b/fhir-model/src/main/java/com/ibm/fhir/model/resource/Resource.java
@@ -93,6 +93,10 @@ public abstract class Resource extends AbstractVisitable {
         return resourceType.isInstance(this);
     }
 
+    /**
+     * @throws ClassCastException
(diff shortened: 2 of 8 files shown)


fhir-swagger-generator/pom.xml
```diff
@@ -11,6 +11,12 @@
     <artifactId>fhir-swagger-generator</artifactId>
 
     <dependencies>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>fhir-examples</artifactId>
+            <version>${project.version}</version>
+            <scope>compile</scope>
```
prb112: these are not picked up in any other project? if so, we may want to add an exclusion. 

Double checking with this comment.

fhir-openapi/src/main/webapp/META-INF/openapi.json
```diff
@@ -3,7 +3,7 @@
     "openapi": "3.0.0",
     "info": {
         "title": "Simplified FHIR API",
-        "description": "A simplified version of the FHIR API",
+        "description": "A simplified version of the HL7 FHIR API",
         "version": "4.0.0"
```
prb112: Should the title match our IBM FHIR Server? or update to 4.0.1 to match the spec?

lmsurpre: IMHO this is mostly independent of our implementation...we don't have our own "flavor" of FHIR.  However, due to complexities with representing FHIR in Swagger/OpenAPI, this is not a representation of the *full* FHIR API.  This is why I settled on "A simplified version of the HL7 FHIR API".

However, I do agree that we should update it to version 4.0.0.

Also, I've been meaning to add a README for this module so that others can have some clue about what it does...I'll try adding that tonight and so please hold off on merging this one for now.

fhir-swagger-generator/pom.xml
lmsurpre: Nothing else should depend on this one...its basically an independent project for making some swagger / openapi definitions (similar to fhir-tools, except that it depends on fhir-model).

PR #539, opened 2019-12-24
Fix Site Automation

prb112:

### DIFF
--- a/.github/workflows/site.yml
+++ b/.github/workflows/site.yml
@@ -64,8 +64,8 @@ jobs:
             echo "cleaning up the prior files on the branch" 
             if [ ! -f .nojekyll ]
             then 
-            touch .nojekyll
-            rm -f _config.yml
+                touch .nojekyll
+                rm -f _config.yml
             fi
 
             rm -f *.js webpack.stats.json styles-*.js styles-*.js.map webpack-runtime-*.js.map webpack-runtime-*.js manifest.webmanifest component---*.js* app-*.js*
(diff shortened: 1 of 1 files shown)


.github/workflows/site.yml
```diff
@@ -71,18 +71,18 @@ jobs:
...
-      env: 
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        GITHUB_REPOSITORY: ${{ secrets.GITHUB_REPOSITORY }}
-        GITHUB_ACTOR: ${{ secrets.GITHUB_ACTOR }}
-        CI: true
+    env: 
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      GITHUB_REPOSITORY: ${{ secrets.GITHUB_REPOSITORY }}
+      GITHUB_ACTOR: ${{ secrets.GITHUB_ACTOR }}
+      CI: true
```
lmsurpre: I don't think this change should be needed.  `env` is supported at both the job_id level AND the step level.
If we really want it at the job_id level, I'd like to see it moved above the steps.

.github/workflows/site.yml
```diff
@@ -3,86 +3,91 @@ name: Docs and Site
...
-        if [ ! -f .nojekyll ]
-        then 
-          touch .nojekyll
-          rm -f _config.yml
-        fi
+            echo "cleaning up the prior files on the branch" 
+            if [ ! -f .nojekyll ]
+            then 
+            touch .nojekyll
+            rm -f _config.yml
```
lmsurpre: maybe indent this to make it clear its within the then?

prb112: done

Issue #545, opened 2019-12-26
Add Lint for the Javadocs


prb112:
Add Lint for the Javadocs

prb112:
PR https://github.com/IBM/FHIR/pull/547

prb112:
adds javadoc:javadoc (limits the amount of duplicate work)

PR #547, opened 2019-12-26
Add Lint for the Javadocs #545

prb112:
``` sh
       JAVADOC_GOAL=""
        if [ "${{ matrix.java }}" != "openjdk8" ]
        then 
          export JAVA_HOME=/opt/hostedtoolcache/AdoptOpenJDK/1.0.0-releases-${{ matrix.java }}-hotspot-normal-latest/x64/
          JAVADOC_GOAL=" -Pvalidate-javadoc javadoc:javadoc "
        fi
        mvn -B ${JAVADOC_GOAL} install --file fhir-parent/pom.xml -DskipTests --no-transfer-progress -X
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -121,7 +121,7 @@ jobs:
       # As the e2e tests are the fastest, opting to put javadoc:javadoc in fhir-parent
       # The profile avoids UML diagram creation
       run: |
-        find / -iname javadoc
+        export JAVA_HOME=/opt/hostedtoolcache/AdoptOpenJDK/1.0.0-releases-${{ matrix.java }}-hotspot-normal-latest/x64/
         mvn -B -Pvalidate-javadoc javadoc:javadoc install --file fhir-parent/pom.xml -DskipTests --no-transfer-progress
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -118,7 +118,16 @@ jobs:
     - name: Build samples
       run: mvn -B install --file fhir-examples/pom.xml --no-transfer-progress
     - name: Build parent without tests
-      run: mvn -B install --file fhir-parent/pom.xml -DskipTests --no-transfer-progress
+      # As the e2e tests are the fastest, opting to put javadoc:javadoc in fhir-parent
+      # The profile avoids UML diagram creation
```
lmsurpre: I wonder if it should just be its own separate step.  Or maybe even its own job...

prb112: I thought about the cognitive complexity of where you would need to go for checks.

prb112: I'm not sure if it should be it's own step.  It seems like a waste of resources. 
... 
And at the same time, I'm not sure it's a bad idea an atomic flag to a possible failure.

PR #549, opened 2019-12-28
build and post latest javadocs

prb112:

### DIFF
--- a/fhir-parent/pom.xml
+++ b/fhir-parent/pom.xml
@@ -230,7 +230,7 @@
                 <scope>test</scope>
             </dependency>
             <dependency>
-                <!--  required for the Tyrus client to work on Java 11 -->
+                <!-- required for the Tyrus client to work on Java 11 -->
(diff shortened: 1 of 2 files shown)


fhir-parent/pom.xml
```diff
@@ -498,23 +498,47 @@
...
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-project-info-reports-plugin</artifactId>
-                    <version>2.9</version>
+                    <version>3.0.0</version>
                 </plugin>
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-javadoc-plugin</artifactId>
-                    <version>3.1.1</version>
+                    <version>3.0.1</version>
```
lmsurpre: Sorry for the late comment; can we leave a comment on here to explain why we shouldn't upgrade to 3.1

prb112: Let me put this in my latest PR

Issue #160, opened 2019-09-27
Support conditional read operation


albertwang-ibm:
We should give client the option of performing a conditional read. This should save bandwidth and reduce latency for frequently accessed resources.

From https://www.hl7.org/FHIR/DSTU2/http.html#cread:

>   Clients may use the If-Modified-Since, or If-None-Match HTTP header on a read request. If so, they MUST accept either a 304 Not Modified as a valid status code on the response (which means that the content is unchanged since that date) or full content (either the content has not changed, or the server does not support conditional request).

>   Servers can return 304 Not Modified where content is unchanged since the If-Modified-Since date-time or the If-None-Match ETag specified or they can return the full content as normal. This optimisation is relevant in reducing bandwidth for caching purposes and servers are encouraged but not required to support this. If servers don't support conditional read, they just return the full content.

https://touchstone.aegis.net/touchstone/scriptexecution?exec=201909271137469410753433&qn=/FHIR4-0-0-Advanced/Patient/01-read/fhir-api-r4-patient-read-04-conditional-found-notmodified#WAITING_OPER

albertwang-ibm:
This could be low priority for us before we don't support this in our capabilitystatement, and touchstone knows that we don't support this by checking our capabilitystatement.

albertwang-ibm:
verified in touchstone test, so close this issue

PR #556, opened 2020-01-06
issue #160 conditional read support

albertwang-ibm:
1. Added full support for conditional read:
(1) Support If-Modified-Since
      Support RFC_1123, RFC_850, ANSIC and touchstone date time formats.
(2) Support If-None-Match
      Support ETag value with or without " (and W/)
       e.g:  1, "1", W/1, W/"1" (the first format is used by TouchStone)
2. Added statement  of conditional read full support to server capability.
3. Added server integrations tests for conditional read.
4. Fixed the date time format of Last-Modified response header.
5. Updated last-modified response header test codes accordingly.


fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -600,6 +618,30 @@ public Response conditionalDelete(@PathParam("type") String type) throws Excepti
             log.exiting(this.getClass().getName(), "delete(String,String)");
         }
     }
+    
+    
+    private long parseIfModifiedSince() {
```
prb112: We can probably use DateTimeHandler from Search's pattern and use a parseBest
Snippet

```
private static final DateTimeFormatter DATE_TIME_PARSER_FORMATTER =
            new DateTimeFormatterBuilder()
                    .appendPattern("yyyy")
                    .optionalStart()
                        .appendLiteral('-')
                        .appendPattern("MM")
                        .optionalStart()
                            .appendLiteral('-')
                            .appendPattern("dd")
                            .optionalStart()
                                .appendLiteral("T")
                                .optionalStart()
                                .appendPattern("HH")
                                .optionalStart()
                                    .appendLiteral(':')
                                    .appendPattern("mm")
                                    .optionalStart()
                                        .appendLiteral(':')
                                        .appendPattern("ss")
                                        .optionalStart()
                                            .appendFraction(ChronoField.MICRO_OF_SECOND, 0, 6, true)
                                        .optionalEnd()
                                    .optionalEnd()
                                .optionalEnd()
                                .optionalStart()
                                    .appendPattern("XXX")
                                .optionalEnd()
                            .optionalEnd()
                            .optionalEnd()
                        .optionalEnd()
                    .optionalEnd()
                    .toFormatter();
    // @formatter:on
```

then one call to 

```
return DATE_TIME_PARSER_FORMATTER.withResolverStyle(ResolverStyle.SMART).parseBest(value,
                    ZonedDateTime::from, LocalDateTime::from, LocalDate::from, YearMonth::from, Year::from);
```

albertwang-ibm: en... seems the new pattern is much more complicated then using SimpleDateFormat, but let me try to change to use it ... :)

albertwang-ibm: just noticed that it's mainly used for formater with optional elements, but in my case, we need 2 different formaters with the required different elements. so I would prefer to keep current implementation.

prb112: I don't think you need two formatters.  I think breaking this discussion out and talking with John may be best

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -162,6 +172,17 @@
     private static final String LOCAL_REF_PREFIX = "urn:";
     private static final String HEADERNAME_IF_NONE_EXIST = "If-None-Exist";
     private static final String HEADERNAME_PREFER = "Prefer";
+    private static final String HEADERNAME_IF_MODIFIED_SINCE = "If-Modified-Since";
+    private static final String HEADERNAME_IF_NONE_MATCH = "If-None-Match";
+
+    public static final DateTimeFormatter PARSER_FORMATTER = new DateTimeFormatterBuilder()
```
prb112: Maybe add a comment indicating what it supports? and format?

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -600,6 +621,28 @@ public Response conditionalDelete(@PathParam("type") String type) throws Excepti
...
     }
+    
+    
+    private long parseIfModifiedSince() {
+        long modifiedSince = -1;
+        try {
+            // Handle RFC_1123 and RFC_850 formats first. 
+            // e.g "Sun, 06 Nov 1994 08:49:37 GMT", "Sunday, 06-Nov-94 08:49:37 GMT", "Sunday, 06-Nov-1994 08:49:37 GMT"
+            // If 2 digits year is used, then means 1940 to 2039.
+            modifiedSince = httpServletRequest.getDateHeader(HEADERNAME_IF_MODIFIED_SINCE);
```
prb112: so it can be a long? or timestamp?

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -610,9 +653,50 @@ public Response read(@PathParam("type") String type, @PathParam("id") String id)
...
+            
+            String ifNoneMatch = httpHeaders.getHeaderString(HEADERNAME_IF_NONE_MATCH);
+            
             Resource resource = doRead(type, id, true, false, null, null, queryParameters);
-            ResponseBuilder response = Response.ok().entity(resource);
-            response = addHeaders(response, resource);
+            int version2Match = -1;
+            // Support ETag value with or without " (and W/)
+            // e.g:  1, "1", W/1, W/"1" (the first format is used by TouchStone)
+            if (ifNoneMatch != null && !ifNoneMatch.replaceAll("\"", "").replaceAll("W/", "").trim().isEmpty()) {
```
prb112: I think this part should be done one time, not twice as on 664/666

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -610,9 +653,50 @@ public Response read(@PathParam("type") String type, @PathParam("id") String id)
...
+            Instant modifiedTime2Compare = null;
+            if (modifiedSince > 0 ) {
+                modifiedTime2Compare = Instant.ofEpochMilli(modifiedSince);
+            }
+            
+            boolean isModifed = true;
+            // check if-not-match first
+            if (version2Match != -1) {
+                if (version2Match == Integer.parseInt(resource.getMeta().getVersionId().getValue())) {
+                    isModifed = false;
```
prb112: slight spelling ... Modified

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
albertwang-ibm: long, the getDateHeader function returns EpochMilli. will add comment to clarify this.

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
albertwang-ibm: good point, will add comment

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
albertwang-ibm: ha, good catch!

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
albertwang-ibm: I thought about this also, and because this requires another if block, so, I didn't. make sense to parse only once, will change.

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
JohnTimm: I'm not sure that this pattern is exactly what we want. You have two optional segments and, technically, they could both be present and it would still work.

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
albertwang-ibm: done

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
albertwang-ibm: done

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
albertwang-ibm: done

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
albertwang-ibm: done

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
albertwang-ibm: we want to support 2 different patterns with only one formatter, this is the only way so far which works in my tests. I don't need they work together, just need it works if any of them presents.  :( .   PARSER_FORMATTER.parse(httpHeaders.getHeaderString(HEADERNAME_IF_MODIFIED_SINCE), LocalDateTime::from).     and I require to parse as LocalDateTime, let me add test codes to find out what happens if both present, I would guess it will throw exception.

albertwang-ibm: added more test codes to cover negative cases.

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright IBM Corp. 2016,2019
+ * (C) Copyright IBM Corp. 2020
```
lmsurpre: this was not right.  should have been 2016,2020

fhir-client/src/main/java/com/ibm/fhir/client/impl/FHIRResponseImpl.java
```diff
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright IBM Corp. 2016,2019
+ * (C) Copyright IBM Corp. 2020
```
lmsurpre: not right.  should be 2016,2020

Issue #572, opened 2020-01-10
Source and Javadoc Jars are missing from 4.0.0 release artifactory


prb112:
**Describe the bug**
Source and Javadoc Jars are missing from 4.0.0 release artifactory

**To Reproduce**
1. Check Artifactory https://bintray.com/beta/#/ibm-watson-health/ibm-fhir-server-releases/fhir-model?tab=files or https://dl.bintray.com/ibm-watson-health/ibm-fhir-server-releases/com/ibm/fhir/fhir-operation/4.0.0/
2. List the files and see source and javadoc are missing

**Expected behavior**
Source and Javadoc should be there

prb112:
PR https://github.com/IBM/FHIR/pull/576

PR #576, opened 2020-01-10
Source and Javadoc Jars are missing from 4.0.0 release artifactory #572

prb112:


prb112:
> I sort of expected this to be in our travis config, but I'm OK with the separate script. LGTM.

absolutely, I just made the script, so we 'could' fixup if we need to.

Issue #280, opened 2019-10-17
FHIR Search - Remove need for Value Types Metadata


prb112:
1. remove valutypes metadata from all places that have it (code and docs)
2. remove DATE_VALUE column (use schema migration logic)
3. write up "best practice" inside the SearchParameter extensions guide.

prb112:
PR https://github.com/IBM/FHIR/pull/593

prb112:
PR https://github.com/IBM/FHIR/pull/593 is updated per review/discussion with Lee

PR #593, opened 2020-01-15
issue #280

prb112:


prb112:
Please hold on the review, I'm tracking down where the above error is found

docs/src/pages/guides/FHIRSearchConfiguration.md
```diff
@@ -100,34 +98,15 @@ The `fhir-search` module requires that the [expression](https://www.hl7.org/fhir
...
-        "name": "favorite-color",
-        "targetClasses": ["String"]
-    }]
-}
-```
-
-#### 1.1.3 Test
-When creating a SearchParameter extension, it is recommended that the FHIRPath expression is checked for validity. Many examples exist in the module `fhir-search`. 
+#### 1.1.2 Test
+When creating a SearchParameter extension, it is recommended that the FHIRPath expression is checked for validity. Many examples exist in the module `fhir-search`. Further, optimizations to the custom SearchParameter extensions may add explict casts to the intended type and activate the specific behavior `refSequence.start-end.as(FHIRQuantity)`. An entry may look like `"expression": "Basic.extension.where(url='http://example.org/decimal').value.as(Decimal)"`.
```
lmsurpre: This line is confusing to me:
> Further, optimizations to the custom SearchParameter extensions may add explict casts to the intended type and activate the specific behavior `refSequence.start-end.as(FHIRQuantity)`.

lmsurpre: I also think we can be a little more clear that the server will skip any search parameters that cannot be mapped to the specified search parameter type.  And probably link to https://www.hl7.org/fhir/R4/search.html#table as the documentation for the allowed mappings.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -1236,7 +1236,7 @@ private ExtractedParameterValue processPrimitiveValue(FHIRPathSystemValue system
             DateParmVal p = new DateParmVal();
             TemporalAccessor v = systemValue.asTemporalValue().temporal();
             java.time.Instant inst = DateTimeHandler.generateValue(v);
-            p.setValueDate(DateTimeHandler.generateTimestamp(inst));
```
lmsurpre: Its interesting to see that we DID have a case where we'd populate the Value and not the Start or End.  Good catch.

lmsurpre: But actually, I think we should set both the Start *and* End values in this case.  Agree?

docs/src/pages/guides/FHIRSearchConfiguration.md
prb112: sure.  let's discuss post your first pass of the review

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -925,13 +908,10 @@ private SqlQueryData processDateParm(Class<?> resourceType, QueryParameter query
...
-        boolean isDateRangeSearch = isDateRangeSearch(resourceType, queryParm);
-
-        if (log.isLoggable(Level.FINE)) {
-            log.fine("isDateSearch=" + isDateSearch + "  isDateRangeSearch=" + isDateRangeSearch);
-        }
-        if (!isDateSearch && !isDateRangeSearch) {
+        // Check to see if this is of a type is date. 
+        String code = queryParm.getCode();
+        SearchParamType type = SearchUtil.getSearchParameter(resourceType, code).getType();
+        if (!SearchParamType.DATE.getValue().equals(type.getValue())) {
```
lmsurpre: Is this needed?  How would we get here if the SearchParamType isn't date?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/SortedQuerySegmentAggregator.java
```diff
@@ -129,6 +122,15 @@ public SqlQueryData buildQuery() throws Exception {
...
             buildFromClause(sqlSortQuery, resourceType.getSimpleName());
 
+            // Gather up all bind variables from the query segments
+            // An important step here is to add _id and _lastUpdated
+            List<Object> allBindVariables = new ArrayList<>();
+            allBindVariables.addAll(idsObjects);
+            allBindVariables.addAll(lastUpdatedObjects);
+            for (SqlQueryData querySegment : this.querySegments) {
+                allBindVariables.addAll(querySegment.getBindVariables());
+            }
```
lmsurpre: this looks like its more related to that issue we were getting when _lastUpdate and _id weren't getting handled right.  i guess there still some case where we're failing to use those properly?  maybe we need to get this fix into 4.0.1?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/SortedQuerySegmentAggregator.java
```diff
@@ -254,7 +256,7 @@ private String buildAggregateExpression(SortParameter sortParm, int sortParmInde
             attributeNames.add(STR_VALUE);
             break;
         case DATE:
-            attributeNames.add(DATE_VALUE);
+            attributeNames.add(DATE_START);
```
lmsurpre: Is it ok to use DATE_START and no DATE_END?  I'm not sure if I really understand how this getValueAttributeNames is used...

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/type/NumberParmBehaviorUtil.java
```diff
@@ -269,14 +284,7 @@ public static int calculateSignificantFigures(BigDecimal original) {
...
             throws FHIRPersistenceException {
-        boolean isIntegerSearch = false;
-        try {
-            isIntegerSearch = ValueTypesFactory.getValueTypesProcessor().isIntegerSearch(resourceType, queryParm);
-        } catch (FHIRSearchException e) {
-            log.log(Level.INFO, "Caught exception while checking the value types for parameter '"
-                    + queryParm.getCode() + "'; continuing...", e);
-            // do nothing
-        }
+        boolean isIntegerSearch = isIntegerSearch(resourceType, queryParm);
```
lmsurpre: Are you sure we need to treat integer search as "special"?

It seems like a lot of code just to fail something.  I don't think supporting "eb" and "sa" prefixes for integers would cause anyone any issue.

lmsurpre: The spec says `sa (starts-after) and eb (ends-before) are not used with integer values but are used for decimals.` which is why we had that check in the first place.  But to me, thats pretty weak.  I recommend removing.

prb112: FYI - it's a pretty drastic change.  would love to sync back to explain.  I'll save for post 4.0.1

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/SortedQuerySegmentAggregator.java
prb112: Yes, in this case it's for a consistent SORT order.  If you add DATE_END as well, it'll lead to very irregular sorts.

docs/src/pages/guides/FHIRSearchConfiguration.md
prb112: since we are now removing the integer part, it's no longer needed above

prb112: I've modified the paragaph

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
prb112: Yes, updated to set the same end time - as this is actually an Instant (at this point in the code)

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/SortedQuerySegmentAggregator.java
prb112: already in 4.0.1 :)

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
prb112: Good point, I'm not sure where this originated. It's definitely not needed.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/type/NumberParmBehaviorUtil.java
prb112: removed, and changed per discussion with Lee

docs/src/pages/guides/FHIRSearchConfiguration.md
lmsurpre: > Further, optimizations to the custom SearchParameter extensions may add explict casts to the intended type and activate the specific behavior `refSequence.start-end.as(FHIRQuantity)`. An entry may look like `"expression": "Basic.extension.where(url='http://example.org/decimal').value.as(Decimal)"`.

Still not clear to me if we're recommending them to use `as` or saying the server might do it automatically?  Its the phrase "optimizations to the custom SearchParameter extensions may add explict casts" messing me up.
Maybe change that to:
> If a search parameter expression extracts an element with a data type that is incompatible with the declared search parameter type, the server will skip the value and log a message. For choice elements, like `Extension.value`, its recommended to restrict the expression to values of the desired type by using the `as` function. For example, select only Decimal values from the `http://example.org/demical` extension, use and expressions like `Basic.extension.where(url='http://example.org/decimal').value.as(Decimal)`.

docs/src/pages/guides/FHIRSearchConfiguration.md
```diff
@@ -100,34 +98,15 @@ The `fhir-search` module requires that the [expression](https://www.hl7.org/fhir
...
-        "name": "favorite-color",
-        "targetClasses": ["String"]
-    }]
-}
-```
-
-#### 1.1.3 Test
-When creating a SearchParameter extension, it is recommended that the FHIRPath expression is checked for validity. Many examples exist in the module `fhir-search`. 
+#### 1.1.2 Test
+When creating a SearchParameter extension, it is recommended that the FHIRPath expression is checked for validity. Many examples exist in the module `fhir-search`. An entry may look like `"expression": "Basic.extension.where(url='http://example.org/decimal').value`.
```
prb112:
```suggestion
When creating the SearchParameter FHIRPath expression, one should test the FHIRPath expression and test the search parameter. 

If a search parameter expression extracts an element with a data type that is incompatible with the declared search parameter type, the server skips the value and logs a message. For choice elements, like Extension.value, its recommended to restrict the expression to values of the desired type by using the `as` function. For example, to select only Decimal values from the http://example.org/demical extension, use an expressions like Basic.extension.where(url='http://example.org/decimal').value.as(Decimal).
```

docs/src/pages/guides/FHIRSearchConfiguration.md
```diff
@@ -100,34 +98,15 @@ The `fhir-search` module requires that the [expression](https://www.hl7.org/fhir
...
-        "resourceType": "Patient",
-        "name": "favorite-color",
-        "targetClasses": ["String"]
-    }]
-}
-```
-
-#### 1.1.3 Test
-When creating a SearchParameter extension, it is recommended that the FHIRPath expression is checked for validity. Many examples exist in the module `fhir-search`. 
+#### 1.1.2 Test
```
prb112:
```suggestion
#### 1.1.2 Recommendations
```

Issue #608, opened 2020-01-20
Run XML tests whenever fhir-model updates


lmsurpre:
Currently, we only run the JSON tests in our pipelines.

prb112:
Confirmed

PR #610, opened 2020-01-20
issue #608 - introduce the ALL index and use it for fhir-model changes

lmsurpre:
Also refactored the Index class to take a list of paths so that we don't
need to maintain the same rows in 3 different files.
Maintaining backwards compat was a little ugly, but the index files
shouldn't be too big and so combining them into a single string
shouldn't be too bad.


prb112:
image

prb112:
Raised to 72%

Issue #112, opened 2019-09-20
Bulk Export/Import: $export-status Operation support for Delete


albertwang-ibm:
Bulk Export/Import $export-status operation support for Delete
The operation should stop the JavaBatch Job. 

**Input**
- After a JavaBatch job is submitted, there is a response which includes the ID. 
- Validate the Tenant Id for the Job
- Call the $export-status operation and validate the tenant

**Output** 
400 - Bad Request  
404 - Job Execution not found  
403 - forbidden access to job (beyond normal 403)  
500 - Internal error   

**Steps**
- Use the Java BatchRuntime to access and stop the Job. 

Alternative, URL to stop the above job is:
   https://localhost:9443/ibm/api/batch/jobexecutions/1?action=stop

Related to Issue #55

prb112:
I'd really like to be assigned this work in sprint planning.

prb112:
PR https://github.com/IBM/FHIR/pull/633

Issue #55, opened 2019-09-13
Refine $import, $export and $export-status operations


prb112:
Refine $import, $export and $export-status operations

- confirm the multi-tenancy
- resolve `TODO`
- refactor to support `DELETE` operations
- push batch code down to the library for bulkdata and pull into the Operation 

Reference - https://javaee.github.io/tutorial/batch-processing006.html#BCGCAHCB

```
 <dependency>
            <groupId>javax.batch</groupId>
            <artifactId>javax.batch-api</artifactId>
            <version>1.0.1</version>
        </dependency>


 JobOperator jobOperator = BatchRuntime.getJobOperator();
        jobOperator.start("FhirBulkExportChunkJob", toProperties(properties));


 private Properties toProperties(Map<String,String> props){
        Properties properties = new Properties();
        for(Entry<String, String> entry : props.entrySet()) {
            properties.put(entry.getKey(), entry.getValue());
        }
        
        return properties;
    }
```

prb112:
per #67 and #69 - update the code.

prb112:
linked to  Custom operation should support more types being passed from URI #75

prb112:
Linked the two together in ZenHub

prb112:
Unfortunately, JobOperator was not an option.

prb112:
I still refactored the code.

prb112:
PR https://github.com/IBM/FHIR/pull/633

PR #633, opened 2020-01-25
Bulk Export/Import: $export-status Operation support for Delete #112

prb112:
- Clean up BulkData Constants
- Update BulkData Operations
- Refactor the BulkDataClient
- Also implemented, Refine $import, $export and $export-status
operations #55


prb112:
Note, I adjusted the points taken on the PR.  (3 in total)

fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/cache/BulkDataTenantSpecificCache.java
```diff
@@ -17,13 +16,11 @@
 import com.ibm.fhir.operation.bulkdata.config.BulkDataConfigUtil;
 
 /**
- * This class implements a cache of BulkData configuration organized by tenantId. Each object stored in the cache will be a
+ * This class implements a cache of BulkData configuration organized by tenantId. Each object stored in the cache will
+ * be a
  * two-level map of SearchParameters organized first by resource type, then by search parameter name.
```
albertwang-ibm: SearchParameters or jobParameters?

prb112: funny - we never caught this in the initial implementation. I'll fix. It's actually suppose to say a map of properties related to the bulkdata export.

albertwang-ibm:
The refactoring and the new job delete codes look great for me!!
But this does remind that we need a mechanism to protect the job instance id in the polling location url, otherwise, user will be able to poll and get the exported url of any other user, or delete the request submitted by others.  seems a simple solution is to encrypt the job instance Id, and because the javabatch job id is not reused(always increase if non in-memory db) in server side,  so user will not be able to use his/her old job info to get or deleted other's export.

prb112:
> The refactoring and the new job delete codes look great for me!!
> But this does remind that we need a mechanism to protect the job instance id in the polling location url, otherwise, user will be able to poll and get the exported url of any other user, or delete the request submitted by others. seems a simple solution is to encrypt the job instance Id, and because the javabatch job id is not reused(always increase if non in-memory db) in server side, so user will not be able to use his/her old job info to get or deleted other's export.

Let's sync up on Monday.  I have a good place to put it, and I can explain what I do in verifyTenant.

fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/cache/BulkDataTenantSpecificCache.java
```diff
@@ -17,13 +16,11 @@
 import com.ibm.fhir.operation.bulkdata.config.BulkDataConfigUtil;
 
 /**
- * This class implements a cache of BulkData configuration organized by tenantId. Each object stored in the cache will be a
+ * This class implements a cache of BulkData configuration organized by tenantId. Each object stored in the cache will
```
prb112:
```suggestion
 * This class implements a cache of BulkData configuration organized by tenantId. Each object stored in the cache is a 
```

fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/cache/BulkDataTenantSpecificCache.java
```diff
@@ -17,13 +16,11 @@
...
+ * be a
```
prb112:
```suggestion
```

fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/cache/BulkDataTenantSpecificCache.java
prb112:
```suggestion
 * a map of properties related to bulkdata operations. 
```

Issue #636, opened 2020-01-25
Simplify javabatch client codes and configuration with JobOperator and BatchRuntime


albertwang-ibm:
**Task**: Instead of using REST API to start/monitor/stop the job, we can use **JobOperator** and **BatchRuntime**.
**Reference**: page 54 of http://www-03.ibm.com/support/techdocs/atsmastr.nsf/002573f7000ac64286256c71006d2e0a/f34b9c5a6e29bc7886257e7c004c93eb/$FILE/WP102544%20-%20WLB%20Step-by-Step%20Implementation%20Guide.002.pdf/WP102544%20-%20WLB%20Step-by-Step%20Implementation%20Guide.pdf

**Sample codes**:

import javax.batch.operations.JobOperator; 
import javax.batch.runtime.BatchRuntime; 
import javax.batch.runtime.JobExecution; 
import javax.batch.runtime.JobInstance;

Start a job:
JobOperator jobOperator = BatchRuntime.getJobOperator(); 
long execId = jobOperator.start("job-id", JobParameters); 
JobInstance jobInstance = jobOperator.getJobInstance(execId);
JobExecution jobExecution = jobOperator.getJobExecution(execId);

Status of a job:
JobOperator jobOperator = BatchRuntime.getJobOperator();
JobInstance jobInstance = jobOperator.getJobInstance(execId); 
List<JobExecution> jobExecutions = jobOperator.getJobExecutions(jobInstance);
...

Stop a job:
BatchRuntime.getJobOperator().stop(execId);
 
Restart a job:
newExecId = BatchRuntime.getJobOperator().restart(execId, JobParameters);

prb112:
@albertwang-ibm I looked into this and tried this approach.  It does not work.  It automatically starts the job on the local operator.

albertwang-ibm:
per Paul's experiments and discussed in the scrum, because if using these APIs, the job will be started inside fhir-server itself, this is not what we want to see, so we will not use these APIs.

PR #647, opened 2020-01-28
issue #636 encrypt the internal Javabatch job id used in bulkdata apis

albertwang-ibm:

### DIFF
--- a/fhir-operation-bulkdata/pom.xml
+++ b/fhir-operation-bulkdata/pom.xml
@@ -34,5 +34,10 @@
             <artifactId>jsonassert</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>fhir-server</artifactId>
+            <version>${project.version}</version>
+        </dependency>
     </dependencies>
 </project>
--- a/fhir-operation-bulkdata/pom.xml
+++ b/fhir-operation-bulkdata/pom.xml
@@ -34,10 +34,5 @@
             <artifactId>jsonassert</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>${project.groupId}</groupId>
-            <artifactId>fhir-server</artifactId>
-            <version>${project.version}</version>
-        </dependency>
     </dependencies>
 </project>
--- a/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/BulkDataConstants.java
+++ b/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/BulkDataConstants.java
@@ -12,6 +12,7 @@ import java.util.List;
 import javax.crypto.spec.SecretKeySpec;
 
 import com.ibm.fhir.operation.bulkdata.config.BulkDataConfigUtil;
+import com.ibm.fhir.server.helper.FHIRServerUtils;
 
 /**
  * Constants for BulkData operations
@@ -35,7 +36,9 @@ public class BulkDataConstants {
     public static final String PARAM_GROUP_ID = "groupId";
     public static final String PARAM_JOB = "job";
 
-    public static final SecretKeySpec JOB_ENCRYPTION_KEY = BulkDataConfigUtil.getJobIdEncryptionKey();
+    // Encryption key used for JavaBatch Job ID
+    public static final SecretKeySpec BATCHJOBID_ENCRYPTION_KEY =
+            BulkDataConfigUtil.getBatchJobIdEncryptionKey(FHIRServerUtils.getJNDIValue("config/bulkDataBatchJobIdEncryptionKey", null));
 
     // Status
     public static final List<String> SUCCESS_STATUS = Collections.unmodifiableList(Arrays.asList("COMPLETED"));
--- a/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/BulkDataConstants.java
+++ b/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/BulkDataConstants.java
@@ -11,8 +11,9 @@ import java.util.List;
 
 import javax.crypto.spec.SecretKeySpec;
 
+import com.ibm.fhir.config.FHIRConfigHelper;
+import com.ibm.fhir.config.FHIRConfiguration;
 import com.ibm.fhir.operation.bulkdata.config.BulkDataConfigUtil;
-import com.ibm.fhir.server.helper.FHIRServerUtils;
 
 /**
  * Constants for BulkData operations
@@ -38,7 +39,7 @@ public class BulkDataConstants {
 
     // Encryption key used for JavaBatch Job ID
     public static final SecretKeySpec BATCHJOBID_ENCRYPTION_KEY =
-            BulkDataConfigUtil.getBatchJobIdEncryptionKey(FHIRServerUtils.getJNDIValue("config/bulkDataBatchJobIdEncryptionKey", null));
+            BulkDataConfigUtil.getBatchJobIdEncryptionKey(FHIRConfigHelper.getStringProperty(FHIRConfiguration.PROPERTY_BULKDATA_BATCHJOBID_ENCRYPTION_KEY, null));
 
     // Status
     public static final List<String> SUCCESS_STATUS = Collections.unmodifiableList(Arrays.asList("COMPLETED"));
--- a/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/BulkDataConstants.java
+++ b/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/BulkDataConstants.java
@@ -9,6 +9,10 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
+import javax.crypto.spec.SecretKeySpec;
+
+import com.ibm.fhir.operation.bulkdata.config.BulkDataConfigUtil;
+
 /**
  * Constants for BulkData operations
  */
@@ -30,6 +34,8 @@ public class BulkDataConstants {
     public static final String PARAM_TYPE_FILTER = "_typeFilter";
     public static final String PARAM_GROUP_ID = "groupId";
     public static final String PARAM_JOB = "job";
+
+    public static final SecretKeySpec JOB_ENCRYPTION_KEY = BulkDataConfigUtil.getJobIdEncryptionKey();
 
     // Status
     public static final List<String> SUCCESS_STATUS = Collections.unmodifiableList(Arrays.asList("COMPLETED"));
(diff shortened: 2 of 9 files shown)


fhir-operation-bulkdata/pom.xml
```diff
@@ -34,5 +34,10 @@
             <artifactId>jsonassert</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>fhir-server</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
prb112: I think this should be removed.

fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/BulkDataConstants.java
```diff
@@ -31,6 +36,10 @@
     public static final String PARAM_GROUP_ID = "groupId";
     public static final String PARAM_JOB = "job";
 
+    // Encryption key used for JavaBatch Job ID
+    public static final SecretKeySpec BATCHJOBID_ENCRYPTION_KEY =
+            BulkDataConfigUtil.getBatchJobIdEncryptionKey(FHIRServerUtils.getJNDIValue("config/bulkDataBatchJobIdEncryptionKey", null));
```
prb112: Pull this value from the tenant configuration for bulkdata or pull from the property group.

fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataUtil.java
```diff
@@ -284,4 +287,37 @@ public static String checkAndValidateJob(Parameters parameters) throws FHIROpera
...
+        if (key == null) {
+            return strToEncrypt;
+        } else {
+            try
+            {
+                Cipher cp = Cipher.getInstance("AES/ECB/PKCS5Padding");
+                cp.init(Cipher.ENCRYPT_MODE, key);
+                return java.net.URLEncoder.encode(Base64.getEncoder().encodeToString(cp.doFinal(strToEncrypt.getBytes("UTF-8"))), StandardCharsets.UTF_8.name());
+            } catch (Exception e) {
+                return strToEncrypt;
```
prb112: so it will default through to the normal job id?

prb112: I think we should remove '+' and '=' before URL encoding

fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataUtil.java
```diff
@@ -284,4 +287,37 @@ public static String checkAndValidateJob(Parameters parameters) throws FHIROpera
 
         throw new FHIROperationException("no job identifier is passed");
     }
+
+
+    public static String encryptBatchJobId(String strToEncrypt, SecretKeySpec key) {
```
prb112: worthy of a comment

fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataUtil.java
```diff
@@ -284,4 +287,37 @@ public static String checkAndValidateJob(Parameters parameters) throws FHIROpera
...
+                Cipher cp = Cipher.getInstance("AES/ECB/PKCS5Padding");
+                cp.init(Cipher.ENCRYPT_MODE, key);
+                return java.net.URLEncoder.encode(Base64.getEncoder().encodeToString(cp.doFinal(strToEncrypt.getBytes("UTF-8"))), StandardCharsets.UTF_8.name());
+            } catch (Exception e) {
+                return strToEncrypt;
+            }
+        }
+    }
+
+    public static String decryptBatchJobId(String strToDecrypt, SecretKeySpec key) {
```
prb112: Worthy of a comment

fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataUtil.java
```diff
@@ -284,4 +287,37 @@ public static String checkAndValidateJob(Parameters parameters) throws FHIROpera
...
+        // Decrypt to get the batch job id
+        if (key == null) {
+            return strToDecrypt;
+        } else {
+            try
+            {
+                Cipher cp = Cipher.getInstance("AES/ECB/PKCS5PADDING");
```
prb112: is this the lightest weight cipher?  it doesn't need to be strong.

prb112: Where is the URL decode? shouldn't this be symmetric, or are you relying on the Liberty runtime to previously decode

prb112: maybe add a comment

fhir-server/liberty-config/server.xml
```diff
@@ -199,4 +199,7 @@
         </application-bnd>
     </webApplication>
     <include optional="true" location="${server.config.dir}/batchDs.xml"/>
+    
+    <!-- The key used for encryption of the internal JavaBatch Job instance/execution id which is used in bulkdata APIs  -->
+    <jndiEntry jndiName="config/bulkdatajobencryptionkey" value="change-password"/>
```
prb112: I think this should go into the properties file separate from JNDI

fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/BulkDataConstants.java
albertwang-ibm: I don't think this need to be tenant specific. so I didn't put it into bulkdata.json or any of then per tenant config file. I considered server.xml or server.env, and thought server.xml is the better place to put this tenant independent config.

fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataUtil.java
albertwang-ibm: yes, if the encryption key is not configured in server.xml, then the id will not be encrypted.
en, if remove, how can we bring back the the '+' and '=' before decryption later when receiving the pull request ...

fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataUtil.java
albertwang-ibm: let me add, I thought it's straight forward, so I didn't.

fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataUtil.java
albertwang-ibm: let me add one, I thought it's straight forward, so I didn't.

fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataUtil.java
albertwang-ibm: yes, this is lightest weight cipher, and I didn't add salt to it to make the config simpler, otherwise, we will need config for both the key and salt.
yes, liberty run time or the rest framework did the url decoding automatically, I added decoding in this function before, and found it didn't work, and then through debugging, I found it's already decoded, double decode the encrypted job id caused error, so I removed the url decode from this function.  
Yes, will add comment for this.

fhir-server/liberty-config/server.xml
albertwang-ibm: I didn't put it into any of the tenant config files because I want the same key to be used for all the tenants. I considered either server.xml or server.env, and I thought server.xml seems a better choice for me.

fhir-operation-bulkdata/pom.xml
albertwang-ibm: this is added for being able to use the util function there to read config from server.xml

albertwang-ibm: changed to use fhir server config.

fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/BulkDataConstants.java
albertwang-ibm: changed to use fhir server config.

fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataUtil.java
albertwang-ibm: changed to use fhir server config.

fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataUtil.java
albertwang-ibm: done

fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataUtil.java
albertwang-ibm: done

fhir-server/liberty-config/server.xml
albertwang-ibm: changed to use fhir server config.

fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataUtil.java
prb112: great confirmation, I think adding the comment is the only part left?

prb112:
Added a comment on the unresolved one.  I think it just helps in the event of someone scanning for weak ciphers.

fhir-operation-bulkdata/pom.xml
prb112: pom.xml needs to be updated - remove fhir-server

fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataUtil.java
albertwang-ibm: added commenats

fhir-operation-bulkdata/pom.xml
albertwang-ibm: it's removed from the pom

Issue #656, opened 2020-01-30
Error running https://<host>:<port>/fhir-server/api/v4/ right after install


festewart:
**Describe the bug**
After installing and before writing anything to FHIR, curl -k -u fhiruser https://localhost:9447/fhir-server/api/v4/  returns error outcome and errors logged to messages.log

**To Reproduce**
1. Install R4.0.1 using defaults (no data loaded)
2. curl -k -u fhiruser https://localhost:9447/fhir-server/api/v4/
Enter host password for user 'fhiruser'
Output is:
{"resourceType":"OperationOutcome","id":"c0-a8-30-21-58f8457b-3823-48ae-9916-859993ce8f33","issue":[{"severity":"fatal","code":"processing","details":{"text":"FHIRPersistenceDataAccessException: Server error: failure retrieving count"}}]}
3. Error in messages.log is:
[1/30/20 6:57:38:595 CST] 00000079 com.ibm.fhir.server.resources.FHIRResource                   W Caught exception while processing 'search' request.
com.ibm.fhir.persistence.jdbc.exception.FHIRPersistenceDataAccessException: Server error: failure retrieving count  [probeId=c0-a8-30-21-58f8457b-3823-48ae-9916-859993ce8f33]
    at com.ibm.fhir.persistence.jdbc.dao.impl.FHIRDbDAOImpl.runCountQuery(FHIRDbDAOImpl.java:454)
    at com.ibm.fhir.persistence.jdbc.dao.impl.ResourceDAOImpl.searchCount(ResourceDAOImpl.java:291)
    at com.ibm.fhir.persistence.jdbc.impl.FHIRPersistenceJDBCImpl.search(FHIRPersistenceJDBCImpl.java:429)
    at com.ibm.fhir.server.resources.FHIRResource.doSearch(FHIRResource.java:2146)
    at com.ibm.fhir.server.resources.FHIRResource.doSearchAll(FHIRResource.java:855)
    at com.ibm.fhir.server.resources.FHIRResource.searchAllGet(FHIRResource.java:871)

**Expected behavior**
No error - maybe something saying nothing was found

prb112:
PR https://github.com/IBM/FHIR/pull/658

prb112:
Fix is merged

PR #658, opened 2020-01-30
System-Count when no data in system #656

prb112:

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
@@ -240,6 +240,12 @@ public class QuerySegmentAggregator {
 
         // Processes through EACH register parameter extracting the integer value
         Map<String, Integer> resourceNameMap = resourceDao.readAllResourceTypeNames();
+        if (resourceNameMap.size() == 0) {
+            // Special condition where we have no registered resources and therefore no data.
+            // This is only used in COUNT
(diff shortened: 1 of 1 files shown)


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
```diff
@@ -240,6 +240,12 @@ protected SqlQueryData buildSystemLevelQuery(String selectRoot, String subSelect
 
         // Processes through EACH register parameter extracting the integer value
         Map<String, Integer> resourceNameMap = resourceDao.readAllResourceTypeNames();
+        if (resourceNameMap.size() == 0) {
+            // Special condition where we have no registered resources and therefore no data.
+            // This is only used in COUNT
+            queryString.append("SELECT LR.LOGICAL_RESOURCE_ID AS CNT FROM LOGICAL_RESOURCES LR");
```
lmsurpre: If we know we have no data, should it just be `SELECT 0` or some such?

prb112: since it's a nested SELECT SELECT it ends up being a pain in the butt to do VALUES() since we don't really want to return a value.

lmsurpre: yeah, i almost suggested defining and throwing a checked exception so that we don't even attempt the query, but I guess the above is ok.  marking it approved.

PR #666, opened 2020-02-02
issue #679 - FHIRPath Patch

lmsurpre:
Added add, delete, replace, insert, and move FHIRPath helpers with the meat of the FHIRPath Patch logic. For now, these methods require both the FHIRPath AND the elementName which they are interacting with.

Added FHIRPathPatch classes and added content-negotiated `patch` JAXRS methods in FHIRResource

Added FHIRPathSpecTest for executing the spec-provided
FHIRPath Patch test cases. I needed to tweak the test file to work
around our strict interpretation of `ele-1`
which makes empty resources invalid.
I also removed a single fhirpath patch operation from the
`Full Resource` testcase...we currently cannot handle replace of
DomainResource.text.div

Added logic to throw UnsupportedOperationException for patches
with nested part values and added corresponding skip logic from the test

Added skip logic for 2 other test cases, both with `mode=forwards`
* `Delete Nested Primitive #2`: this test expects that the server will
automatically remove a parent elements when the last element from within
one is removed. We don't do that.
* `Reorder List #4`: on this one, I disagree with the expected output

Finally, I added the FHIR mimetypes to the patchFormat element of our
CapabilityStatement to indicate that we now support FHIRPath patch.

### DIFF
--- a/fhir-config/src/main/java/com/ibm/fhir/config/FHIRConfiguration.java
+++ b/fhir-config/src/main/java/com/ibm/fhir/config/FHIRConfiguration.java
@@ -89,7 +89,7 @@ public class FHIRConfiguration {
      * This is our in-memory cache of PropertyGroup's keyed by tenant-id.
      */
     private TenantSpecificPropertyGroupCache configCache = new TenantSpecificPropertyGroupCache();
-    
+
     /**
      * This method is used to configure an explicit top-level directory where FHIR Server configuration
      * information is expected to reside.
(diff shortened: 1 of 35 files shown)


lmsurpre:
> For naming prefix consistency, I think that FHIRPatchType should be FHIRPathPatchType and the various operations should be named FHIRPathPatchReplace, FHIRPathPatchAdd, etc.

done

JohnTimm:
LGTM

PR #685, opened 2020-02-06
Refactored FHIRPathPatch into builder pattern

lmsurpre:
Also:
1. added `toParameters()` to set up use from the client
2. removed hacky logic for retrieving the element name from a path
    It turns out that our FHIRPathNode object already exposes the element
    name from a selected node.
    This works for all cases other than inserting an element at position 0
    of an empty list, and its not clear from the spec whether that should be
    supported or not. Now we throw UnsupportedOperationException in that
    case.
3. reduced the visibility of all the patch visitors, added null
    checks, and improved documentation of both the visitors and select
    methods in the FHIRPath model.
4. fixed major issue caused by use of parent identity in the fhirpathpatch visitors

### DIFF
--- a/fhir-path/src/main/java/com/ibm/fhir/path/patch/FHIRPathPatch.java
+++ b/fhir-path/src/main/java/com/ibm/fhir/path/patch/FHIRPathPatch.java
@@ -7,7 +7,6 @@
 package com.ibm.fhir.path.patch;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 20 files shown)


fhir-path/src/main/java/com/ibm/fhir/path/patch/FHIRPathPatch.java
```diff
@@ -17,33 +16,189 @@
...
      * @throws IllegalArgumentException if the Parameters object does not satisfy the requirements of a FHIRPath Patch
      */
-    public FHIRPathPatch(Parameters params) {
+    public static FHIRPathPatch from(Parameters params) {
         Objects.requireNonNull(params);
-        operations = new ArrayList<FHIRPathPatchOperation>(2);
+        Builder builder = FHIRPathPatch.builder();
+        
         for (Parameter param : params.getParameter()) {
             if (!"operation".equals(param.getName().getValue())) {
```
bfelaco: Is there a constant for this defined somewhere?

fhir-path/src/main/java/com/ibm/fhir/path/patch/FHIRPathPatch.java
```diff
@@ -17,33 +16,189 @@
...
+         * Build the {@link FHIRPathPatch}
+         * 
+         * @return
+         *     An immutable object of type {@link FHIRPathPatch}
+         */
+        public FHIRPathPatch build() {
+            return new FHIRPathPatch(this);
+        }
+
+        protected Builder from(FHIRPathPatch patch) {
```
bfelaco: Is the intention of making this an instance method to support inter-mixing other patches with manually added operations like this?

```java
     FHIRPathPatch.builder()
          .add(...)
          .move(...)
          .from(someExistingPatch)
          .from(someOtherPatch)
          .build();
```

Or were you only thinking of the simple case of just wanting to mutate a single existing immutable `FHIRPathPatch`?

If you want to allow/support the former, then you should make this method public so anyone can do it. If not, then I'd suggest changing this to an additional private ctor which can only be used by the `toBuilder()` method of the outer class.

lmsurpre: > s the intention of making this an instance method to support inter-mixing other patches with manually added operations like this?

No, although I did consider something like that for supporting combining of disparate patches into a single patch.  I don't think it would be too hard to add, but since presently it just naively applies them one-at-a-time, I don't think it makes much difference whether you combine into a single patch or just execute them one-after-another.

> Or were you only thinking of the simple case of just wanting to mutate a single existing immutable FHIRPathPatch?

this

I just followed the precedent set in our generated model.  For example, see
https://github.com/IBM/FHIR/blob/master/fhir-model/src/main/java/com/ibm/fhir/model/resource/Account.java#L1381

I think an overloaded [protected] constructor is another option, but I think it just comes down to preference.  @JohnTimm any thoughts on this?

fhir-path/src/main/java/com/ibm/fhir/path/patch/FHIRPathPatch.java
lmsurpre: yeah, good catch
```suggestion
            if (!FHIRPathPatchOperation.OPERATION.equals(param.getName().getValue())) {
```

fhir-path/src/main/java/com/ibm/fhir/path/patch/FHIRPathPatch.java
lmsurpre: Looking at it some more, I guess it wouldn't hurt to just flip this to public...then it could serve both use cases.  I can't think of a downside...

lmsurpre: done

lmsurpre:
> Can we call call the new method `merge` instead of `from`?

To me merge implies its combining them in some clever way, perhaps removing redundancies, or similar whereas its really just adding all the operations from that other patch.  I brainstormed on a couple names but couldn't come up with anything I really liked.  John agree that `from` is ok since it has javadoc that more accurately describes what it does.

Issue #601, opened 2020-01-17
Validate required bindings from the base spec


lmsurpre:
For code values with a required binding, we generate Code subtypes in the model and invalid values cannot be parsed.
However, for CodeableConcept values with required bindings in the base spec, we don't validate the concepts at all.

Now that we have valueset membership operational, I think we should start using this to validate valueset membership for required bindings in the base spec.

To reproduce:
Create or validate a resource that has an element of type CodeableConcept and a ValueSet binding of Required.
Note that the server happily accepts any value for this element, even if there is no coding from the required ValueSet.

This should fail validation:
```
{
  "resourceType": "Condition",
  "id": "example",
  "clinicalStatus": {
    "coding": [
      {
        "system": "http://example.com",
        "code": "fake-code"
      }
    ]
  },
  "subject": {
    "reference": "Patient/example"
  }
}
```

This should pass validation:
```
{
  "resourceType": "Condition",
  "id": "example",
  "clinicalStatus": {
    "coding": [
      {
        "system": "http://example.com",
        "code": "fake-code"
      },
      {
        "system": "http://terminology.hl7.org/CodeSystem/condition-clinical",
        "code": "active"
      }
    ]
  },
  "subject": {
    "reference": "Patient/example"
  }
}
```

PR #696, opened 2020-02-11
Issue #601 - add validation for required bindings on codeable concepts

JohnTimm:


fhir-model/src/main/java/com/ibm/fhir/model/util/ValidationSupport.java
```diff
@@ -355,4 +357,37 @@ public static void prohibited(Element element, String elementName) {
...
+        if (codeableConcept != null && !codeableConcept.getCoding().isEmpty() && hasCodingWithSystemAndCodeValues(codeableConcept)) {
+            List<String> codeList = Arrays.asList(codes);
+            for (Coding coding : codeableConcept.getCoding()) {
+                if (hasSystemAndCodeValues(coding) && 
+                        system.equals(coding.getSystem().getValue()) && 
+                        codeList.contains(coding.getCode().getValue())) {
+                    return;
+                }
+            }
+            throw new IllegalStateException(String.format("Element: '%s' must contain a valid code from system: '%s'", elementName, system));
```
lmsurpre: isn't it more about containing a valid value from a given valueset?
anyway, i guess all the valuesets in question here are "identity" valuesets for a given codesystem, so probably its fine.

JohnTimm: Instances are concerned with systems but from an error message perspective, I could see how that could be improved.

JohnTimm: Reversed course and put value set in error message instead of code system.

PR #697, opened 2020-02-11
issue #601 - update generated resources with valid codeableconcepts

lmsurpre:


lmsurpre:
pushed the fix directly to #696

Issue #668, opened 2020-02-03
Build pipeline failure due to maven timeout issue


albertwang-ibm:
**Problem:** Execution create-archive of goal org.apache.maven.plugins:maven-assembly-plugin:3.1.0:single failed: Plugin org.apache.maven.plugins:maven-assembly-plugin:3.1.0 or one of its dependencies could not be resolved: Failed to collect dependencies at org.apache.maven.plugins:maven-assembly-plugin:jar:3.1.0 -> org.apache.maven.shared:maven-artifact-transfer:jar:0.9.0: Failed to read artifact descriptor for org.apache.maven.shared:maven-artifact-transfer:jar:0.9.0: Could not transfer artifact org.apache.maven.shared:maven-artifact-transfer:pom:0.9.0 from/to central (https://repo.maven.apache.org/maven2): Connect to repo.maven.apache.org:443 [repo.maven.apache.org/151.101.184.215] failed: Connection timed out

**Possible solution:**
(1) add -Dmaven.wagon.http.pool=false` OR `-Dhttp.keepAlive=false` as flags to the maven tasks.
(2) upgrade to maven 3.2 or after.

**Reference:** 
https://developercommunity.visualstudio.com/content/problem/292284/maven-hosted-agent-connection-timeout.html

albertwang-ibm:
per discussion, we'd like to try the the first solution of adding flag.

lmsurpre:
The linked issues from https://developercommunity.visualstudio.com/content/problem/292284/maven-hosted-agent-connection-timeout.html are all marked as resolved (as of 3.3.x).

I double-checked our pipeline and we're already using the latest maven:
`Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)`

Something not right...

lmsurpre:
I tried adding both `-Dmaven.wagon.http.pool=false` and `-Dhttp.keepAlive=false` but it broker the build.

I updated it to -Dmaven.wagon.httpconnectionManager.ttlSeconds=240` and it works, but its hard to confirm whether it addresses the issue or not.  If we hit the issue again, please reopen this one.

albertwang-ibm:
seems it works so far with multiple builds of my PR have passed, so close this issue for now.

lmsurpre:
I just hit this one:
```
[...]

PR #703, opened 2020-02-13
issue #668 - add http flags to try avoiding timeouts

lmsurpre:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,6 +51,7 @@ jobs:
       env:
         BASE: origin/${{ github['base_ref'] }}
       run: |
+        mvn -version
         model_paths=(
           fhir-model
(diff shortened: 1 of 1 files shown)


albertwang-ibm:
if your new setting still doesn't work, maybe we can try -Dhttp.keepAlive=false only...

lmsurpre:
I will leave #668 in "under review" for now...if you see the timeout again, please speak up!

albertwang-ibm:
a wild guess is that: if the builds for different jdk/os are parallel, and the same maven target repo is used for all builds, maybe this can cause contention happening, maybe more than 1 build tried to push the same jar to the repo, and then only the first can succeed. so, if the builds are parallel now and the same repo is used, maybe we can try to make the builds in sequence as an experiment ...

Issue #711, opened 2020-02-18
Look up a model class from its typeName


lmsurpre:
**Is your feature request related to a problem? Please describe.**
We have a declarative mapping and I'd like to store just the type name for mapping targets (rather than fully qualified name). To do it, we need a way to lookup the target class from just the type name.

**Describe the solution you'd like**
Add ModelSupport.getTypeClass(String typeName) which performs the lookup...an exact logical reverse of 
`ModelSupport.getTypeName(Class<?> clazz)`

**Describe alternatives you've considered**
The alternative would be to use something like this wherever we need it:
```
Class.forName("com.ibm.fhir.model.type." + name.substring(0, 1).toUpperCase() + name.substring(1));
```

PR #712, opened 2020-02-18
Issue #711 - implement ModelSupport.getDataType(...) method

JohnTimm:


fhir-model/src/main/java/com/ibm/fhir/model/util/ModelSupport.java
```diff
@@ -889,4 +898,11 @@ public static String getSystem(Code code) {
         }
         return null;
     }
+
+    /**
+     * @return the data type class associated with {@code typeName) parameter if exists, otherwise null
```
lmsurpre:
```suggestion
     * @return the data type class associated with {@code typeName} parameter if exists, otherwise null
```

JohnTimm: Fixed Javadoc

Issue #723, opened 2020-02-24
Warning messages when importing from COS


albertwang-ibm:
When importing r4_AllergyIntolerance.ndjson from fhir-performance COS bucket, got a lot warning messages like this following:

Not all bytes were read from the S3ObjectInputStream, aborting HTTP connection. This is likely an error and may result in sub-optimal behavior. Request only the bytes you need via a ranged GET or drain the input stream after use.

PR #725, opened 2020-02-25
 issue #723 #728 upgrade COS client and fix Base64BinaryTest 

albertwang-ibm:
For issue #723, According to: 
(1) https://stackoverflow.com/questions/45624175/amazons3-getting-warning-s3abortableinputstreamnot-all-bytes-were-read-from-t, we should call s3ObjectInputStream.abort() to prevent the warning.
(2) https://github.com/aws/aws-sdk-java/issues/1657, I found the old COS client was not with the correct version of aws s3 client codes, so upgrade COS client from 2.1.0 to 2.6.1 (the most current version now.) and the httpclient is also upgraded accordingly.

For issue #728:
(1) Moved FHIRModelConfig.setCheckReferences(false) to FhirServerTestBase.
(2) changed the type of "who" in testCreateParameters of Base64BinaryTest to "Patient".


fhir-bulkimportexport-webapp/src/main/java/com/ibm/fhir/bulkcommon/BulkDataUtils.java
```diff
@@ -157,8 +158,11 @@ public static int readFhirResourceFromObjectStore(AmazonS3 cosClient, String buc
            int numOfLinesToSkip, List<Resource> fhirResources) {
         int exported;
         S3Object item = cosClient.getObject(new GetObjectRequest(bucketName, itemName));
-        try (BufferedReader resReader = new BufferedReader(new InputStreamReader(item.getObjectContent()))) {
+        try (S3ObjectInputStream s3InStream = item.getObjectContent();
+             BufferedReader resReader = new BufferedReader(new InputStreamReader(s3InStream))) {
             exported = getFhirResourceFromBufferReader(resReader, numOfLinesToSkip, fhirResources);
+            // Notify s3 client to abort and prevent the server from keeping on sending data.
+            s3InStream.abort();
```
prb112: Is abort for exception handling logic? if so, why is it in the main `try`?

fhir-parent/pom.xml
```diff
@@ -285,7 +285,7 @@
             <dependency>
                 <groupId>com.ibm.cos</groupId>
                 <artifactId>ibm-cos-java-sdk</artifactId>
-                <version>2.1.0</version>
+                <version>2.6.1</version>
```
prb112: FYI, I looked at the dependency trees: 
 +- com.ibm.cos:ibm-cos-java-sdk:jar:2.1.0:compile
[INFO] |  +- com.ibm.cos:ibm-cos-java-sdk-s3:jar:2.1.0:compile
[INFO] |  |  +- (com.ibm.cos:ibm-cos-java-sdk-kms:jar:2.1.0:compile - omitted for duplicate)
[INFO] |  |  \- (com.ibm.cos:ibm-cos-java-sdk-core:jar:2.1.0:compile - omitted for duplicate)
[INFO] |  +- com.ibm.cos:ibm-cos-java-sdk-kms:jar:2.1.0:compile
[INFO] |  |  \- (com.ibm.cos:ibm-cos-java-sdk-core:jar:2.1.0:compile - omitted for duplicate)
[INFO] |  \- com.ibm.cos:ibm-cos-java-sdk-core:jar:2.1.0:compile
[INFO] |     +- (commons-logging:commons-logging:jar:1.1.3:compile - omitted for conflict with 1.2)
[INFO] |     +- (org.apache.httpcomponents:httpclient:jar:4.5.6:compile - version managed from 4.5.2; omitted for duplicate)
[INFO] |     +- software.amazon.ion:ion-java:jar:1.0.2:compile
[INFO] |     +- com.fasterxml.jackson.core:jackson-databind:jar:2.6.6:compile
[INFO] |     |  +- com.fasterxml.jackson.core:jackson-annotations:jar:2.6.0:compile
[INFO] |     |  \- com.fasterxml.jackson.core:jackson-core:jar:2.6.6:compile
[INFO] |     +- com.fasterxml.jackson.dataformat:jackson-dataformat-cbor:jar:2.6.6:compile
[INFO] |     |  \- (com.fasterxml.jackson.core:jackson-core:jar:2.6.6:compile - omitted for duplicate)
[INFO] |     +- (org.apache.httpcomponents:httpasyncclient:jar:4.1.3:compile - omitted for conflict with 4.1.4)
[INFO] |     \- joda-time:joda-time:jar:2.8.1:compile

+- com.ibm.cos:ibm-cos-java-sdk:jar:2.6.1:compile
[INFO] |  +- com.ibm.cos:ibm-cos-java-sdk-s3:jar:2.6.1:compile
[INFO] |  |  +- (com.ibm.cos:ibm-cos-java-sdk-kms:jar:2.6.1:compile - omitted for duplicate)
[INFO] |  |  \- (com.ibm.cos:ibm-cos-java-sdk-core:jar:2.6.1:compile - omitted for duplicate)
[INFO] |  +- com.ibm.cos:ibm-cos-java-sdk-kms:jar:2.6.1:compile
[INFO] |  |  \- (com.ibm.cos:ibm-cos-java-sdk-core:jar:2.6.1:compile - omitted for duplicate)
[INFO] |  \- com.ibm.cos:ibm-cos-java-sdk-core:jar:2.6.1:compile
[INFO] |     +- (commons-logging:commons-logging:jar:1.1.3:compile - omitted for conflict with 1.2)
[INFO] |     +- (org.apache.httpcomponents:httpclient:jar:4.5.6:compile - version managed from 4.5.9; omitted for duplicate)
[INFO] |     +- software.amazon.ion:ion-java:jar:1.2.0:compile
[INFO] |     +- com.fasterxml.jackson.core:jackson-databind:jar:2.10.2:compile
[INFO] |     |  +- com.fasterxml.jackson.core:jackson-annotations:jar:2.10.2:compile
[INFO] |     |  \- com.fasterxml.jackson.core:jackson-core:jar:2.10.2:compile
[INFO] |     +- com.fasterxml.jackson.dataformat:jackson-dataformat-cbor:jar:2.10.0:compile
[INFO] |     |  \- (com.fasterxml.jackson.core:jackson-core:jar:2.10.0:compile - omitted for conflict with 2.10.2)
[INFO] |     \- joda-time:joda-time:jar:2.8.2:compile

prb112: reasonably the same and up to date.

fhir-bulkimportexport-webapp/src/main/java/com/ibm/fhir/bulkcommon/BulkDataUtils.java
albertwang-ibm: it's not for exception handling logic, if exception happens, the exported will be 0, and the job will be finished with error message in the log, so there will be no issue.   This is for the normal flow of the chunk job, e.g, in first chunk, the job opens the objectstream, read 1000 lines and then abort the stream, and then writes to fhir server, persistence 1000 to checkpoint as the tobeskipped lines, and then in the following second chunk, the job opens the objectstream again, skip 1000 lines first, then read 1000 lines and then abort the stream, and then writes to fhir server, and persistence 2000 to checkpoint as the tobeskipped lines, repeat the above steps till all the lines are processed.      Note: In current implementation,  I didn't provide option to reuse the inputstream/reader across all the chunk operations, because I thought the to-be-imported file can be very huge, and this can cause the connection to be occupied by the import job for too long time which in theory may cause issue for the server if too many concurrent huge files are imported at the same time.  But I did have opened issue #726 to provide option(e.g, controlled by a switch) to reuse the inputstream/reader across all chunk operations if the job is not paused or stopped and continued.

fhir-parent/pom.xml
albertwang-ibm: thanks for the check!

fhir-bulkimportexport-webapp/src/main/java/com/ibm/fhir/bulkcommon/BulkDataUtils.java
prb112: I get it! this makes sense.

Issue #194, opened 2019-10-04
Refactor persistence layer to return OperationOutcome instead of throwing


lmsurpre:
In issue #115 we added support for the persistence layer to return the success status and an optional OperationOutcome (instead of or in addition to the result of the interaction).  Now the task is to use this new feature to chip away at the number of places the Persistence Layer is throwing exceptions to pass state.

lmsurpre:
I'm not sure how far to take this one.  Some potential options:
* current:  only return the Response object for "successful" requests, otherwise continue to throw . (in which case we could perhaps remove the "success/isSuccess()" field/getter that I had added in #115)
* middle-ground:  only throw exceptions on Server errors, not client errors
* all the way:  remove throws FHIRPersistenceException form the interface, and force it to always return a Response object

I'm gonna take a shot at the "middle-ground" option....

prb112:
comment on 'all-the-way' it makes jax-rs more of a requirement when using the persistence layer separate from the rest layer.

Issue #17, opened 2019-09-04
Missing exception in logs after addHeader, audit, or txn.rollback error


lmsurpre:
**Describe the bug**
Under certain circumstances, the server will return a 500 error and there will be no exception details in the server logs.

**To Reproduce**
1. Implement a fake AuditLogger and configure the FHIR Server to use it.
2. Have it throw an exception when you call the logX methods
3. Issue a request and note the 500 error response
4. Check the logs and note there is no stacktrace

**Expected behavior**
All 500 errors should result in an exception in the server logs.

**Additional context**
We always had this issue, but since we weren't doing much of anything in the finally block, we never hit it.  We recently moved audit logging to the finally block at the end of the interactions and this has widened

I suggest we move the log statement to the REST interaction methods (create, update, read, etc. instead of the doCreate, doUpdate, doRead variants).

lmsurpre:
While performing some related cleanup/refactoring, I noticed that I broke the test on accident, by following the spec instead of the current behavior.
from https://www.hl7.org/fhir/http.html#concurrency

If the version id given in the If-Match header does not match, the server returns a 412 Precondition Failed status code instead of updating the resource.

so i think it was a bug in our current impl and I'm HOPING fixing that doesn't break our touchstone results.

PR #731, opened 2020-02-26
issues #747 and #748 - batch/transaction processing edits and error handling improvement

lmsurpre:
Updated bundle processing such that:
    1. 'batch' requests return with 200 OK when one or more of the entries
    fails (check each entry for its response status).
    2. 'transaction' requests return a single OperationOutcome in case of
    failure.  I normalize whatever the normal status code would be into
    either 400 (client error) or 500 (server error).

Also, I removed FHIRHttpException.  Instead of setting the HTTP Status code whenever we
throw, we will set just the OperationOutcome.issue.type and map those to
the desired HTTP status codes in a single spot (currently
IssueTypeToHttpStatusMapper).
This should help prepare us better for both #17 (due to fewer spots to
add logging) and #194 (replacing thrown exceptions with returned
OperationOutcome) in the future.


fhir-server/src/main/java/com/ibm/fhir/server/util/IssueTypeToHttpStatusMapper.java
```diff
@@ -36,15 +43,20 @@
             for (Issue issue : issues) {
                 IssueType code = issue.getCode();
                 if (code != null && code.getValue() != null) {
-                    return issueTypeToResponseCode(IssueType.ValueSet.from(code.getValue()));
+                    ValueSet issueType = IssueType.ValueSet.from(code.getValue());
+                    // Special case for IssueType CONFLICT which can be either an HTTP 409 (Conflict) or HTTP 412 (Precondition failed)
+                    if (issueType == IssueType.ValueSet.CONFLICT &&
+                            FHIRUtil.getExtensionStringValue(code, EXTENSION_URL_HTTP_FAILED_PRECONDITION) != null) {
```
prb112: is this method used anywhere else? is it in FHIRUtil because you think it will be? it might be more containable to add it to this class.

fhir-server/src/main/java/com/ibm/fhir/server/util/IssueTypeToHttpStatusMapper.java
```diff
@@ -36,15 +43,20 @@
...
+                    return issueTypeToResponseCode(issueType);
                 }
             }
         }
-        
         return Status.INTERNAL_SERVER_ERROR;
     }
     
-    public static Status issueTypeToResponseCode(IssueType.ValueSet value) {
+    private static Status issueTypeToResponseCode(IssueType.ValueSet value) {
```
prb112: any reason to 'lock' it down?

fhir-server-test/src/test/java/com/ibm/fhir/server/test/ServerSpecTest.java
```diff
@@ -299,7 +299,7 @@ public void testUpdatePatientVersionAwareError3() throws Exception {
                         .request()
                         .header("If-Match", ifMatchValue)
                         .put(entity, Response.class);
-        assertResponse(response, Response.Status.CONFLICT.getStatusCode());
+        assertResponse(response, Response.Status.PRECONDITION_FAILED.getStatusCode());
```
prb112: Confirming... this changed due to the IssueType mapping

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -249,23 +249,21 @@ public FHIRResource() throws Exception {
...
             checkInitComplete();
 
-            status = Response.Status.OK;
-            RestAuditLogger.logMetadata(httpServletRequest, startTime, new Date(), status);
+            CapabilityStatement capabilityStatement = getCapabilityStatement();
+            RestAuditLogger.logMetadata(httpServletRequest, startTime, new Date(), Response.Status.OK);
 
-            return Response.ok().entity(getCapabilityStatement()).build();
-        } catch (FHIRHttpException e) {
+            return Response.ok().entity(capabilityStatement).build();
```
prb112: do you need the capabilityStatement instance? immediately responding with it seems preferred

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -673,9 +645,8 @@ public Response delete(@PathParam("type") String type, @PathParam("id") String i
                 response = addHeaders(response, ior.getResource());
             }
             return response.build();
-        } catch (FHIRHttpException e) {
-            return exceptionResponse(e);
         } catch (FHIRPersistenceResourceNotFoundException e) {
+            // Overwrite the exception response status because we want NOT_FOUND to be success for delete
```
prb112: very helpful comment

fhir-server/src/main/java/com/ibm/fhir/server/util/IssueTypeToHttpStatusMapper.java
lmsurpre: FHIRUtil already had `getExtensionStringValue(Resource, String)` so I basically just added a sister one for Elements.  I think both could be modified to be more generic, but at least it won't throw a ClassCastException any more when you specify an extension url with a non-string value ;-)

fhir-server/src/main/java/com/ibm/fhir/server/util/IssueTypeToHttpStatusMapper.java
lmsurpre: I added special logic into the other method (which calls this one).  If you called this one directly (e.g. by accident) then you'd miss out on that new logic.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/ServerSpecTest.java
lmsurpre: actually, I broke the test on accident, by following the spec instead of the current behavior. 
 from https://www.hl7.org/fhir/http.html#concurrency
> If the version id given in the If-Match header does not match, the server returns a 412 Precondition Failed status code instead of updating the resource.

so i think it was a bug in our current impl and I'm HOPING fixing that doesn't break our touchstone results.

lmsurpre: P.S. let me know if you think i should open a separate issue for it

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
lmsurpre: I forget why I split it out...maybe I thought I was going to use it but then never did.  I can revert that part of the change and call it inline if you prefer.

lmsurpre: In general, I'd love to start caching this, maybe on a per-tenant basis....our calls to `metadata` take a long time and it shouldn't change unless someone updates the config...

prb112: no need to revert, I wanted to confirm the intent.  I didn't know if there was a side effect

fhir-server-test/src/test/java/com/ibm/fhir/server/test/ServerSpecTest.java
prb112: small... small... suggestion, add the comment to the issue, these are harder to find in the PRs. 

if it's in the issue, at least we can search on it (and I'm suggesting adding to the issue related to this PR only)

fhir-server/src/main/java/com/ibm/fhir/server/util/IssueTypeToHttpStatusMapper.java
```diff
@@ -36,15 +43,20 @@
             for (Issue issue : issues) {
                 IssueType code = issue.getCode();
                 if (code != null && code.getValue() != null) {
-                    return issueTypeToResponseCode(IssueType.ValueSet.from(code.getValue()));
+                    ValueSet issueType = IssueType.ValueSet.from(code.getValue());
```
JohnTimm: Should this use the new `getValueAsEnumConstant` method?

lmsurpre: done

fhir-model/src/main/java/com/ibm/fhir/model/util/FHIRUtil.java
```diff
@@ -433,7 +438,8 @@ public static String getExtensionStringValue(Resource resource, String extension
             if (DomainResource.class.isAssignableFrom(resource.getClass())) {
                 DomainResource dr = (DomainResource) resource;
                 for (Extension ext : dr.getExtension()) {
```
prb112: Essentially 460 is duplicate - just a different origin.  Perhaps refactor into an intermediate method.

fhir-server/src/main/java/com/ibm/fhir/server/exception/FHIRRestBundledRequestException.java
```diff
@@ -6,14 +6,10 @@
 
 package com.ibm.fhir.server.exception;
 
-import java.util.Collection;
-
-import javax.ws.rs.core.Response.Status;
-
+import com.ibm.fhir.exception.FHIROperationException;
```
prb112: Please update copyright

fhir-server/src/main/java/com/ibm/fhir/server/util/IssueTypeToHttpStatusMapper.java
```diff
@@ -14,8 +14,20 @@
 import com.ibm.fhir.model.resource.OperationOutcome;
 import com.ibm.fhir.model.resource.OperationOutcome.Issue;
 import com.ibm.fhir.model.type.code.IssueType;
+import com.ibm.fhir.model.util.FHIRUtil;
```
prb112: please update copyright

fhir-model/src/main/java/com/ibm/fhir/model/util/FHIRUtil.java
lmsurpre: done

Issue #734, opened 2020-02-26
Add cache support for the capabilites request


lmsurpre:
**Is your feature request related to a problem? Please describe.**
Calls to the /metadata take a long time and it shouldn't change unless someone updates the config.  At the very least, we should make sure we're adding reasonable cache-control headers.

**Describe the solution you'd like**
I'd love to start caching this, maybe on a per-tenant basis....

**Describe alternatives you've considered**
Make it more static.  For example, IGs sometimes ship with their own CapabilityStatement resources.  How will we expose those?

lmsurpre:
potentially related to #661

JohnTimm:
One thing we need to consider on this one is how to dynamically support definitional resources that contribute to the capabilities statement (i.e. profiles that are POST or PUT through the endpoint). I think, per Lee's suggestion, we would need to go to a time-based cached.

prb112:
PR https://github.com/IBM/FHIR/pull/1078

I went with a time based cache (since the date in the generated capabilities statement is using ZonedDateTime.now I check to see if the current ZonedDateTime.now is within 1 hr)

JohnTimm:
https://guava.dev/releases/17.0/api/docs/com/google/common/cache/CacheBuilder.html

PR #757, opened 2020-03-04
REST layer refactoring/cleanup

lmsurpre:
1. added originalRequestUri to FHIRRequestContext
2. set it from FHIRRestServletFilter
3. use it from FHIRResource
4. remove unused `queryString` parm from SearchUtil.parseQueryParameters

### DIFF
--- a/fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright IBM Corp. 2016,2019
+ * (C) Copyright IBM Corp. 2016, 2020
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -7,25 +7,27 @@
 package com.ibm.fhir.server.filter.rest;
 
 import java.io.IOException;
+import java.net.URI;
 import java.security.Principal;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-import javax.servlet.Filter;
 import javax.servlet.FilterChain;
 import javax.servlet.FilterConfig;
 import javax.servlet.ServletException;
 import javax.servlet.ServletRequest;
-import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpFilter;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.UriBuilder;
 
 import org.owasp.encoder.Encode;
 
 import com.ibm.fhir.config.FHIRConfigHelper;
 import com.ibm.fhir.config.FHIRConfiguration;
 import com.ibm.fhir.config.FHIRRequestContext;
+import com.ibm.fhir.config.PropertyGroup;
 import com.ibm.fhir.core.HTTPHandlingPreference;
 import com.ibm.fhir.core.HTTPReturnPreference;
 import com.ibm.fhir.exception.FHIRException;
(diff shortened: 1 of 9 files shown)


fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -3560,10 +3561,9 @@ private String serializeOperationOutcome(OperationOutcome oo) {
         }
     }
 
-    private synchronized CapabilityStatement getCapabilityStatement() throws Exception {
+    private synchronized CapabilityStatement getCapabilityStatement() throws FHIROperationException {
```
prb112: After our earlier discussion, this is only called once? or does it synchronize across multiple requests, I could see this being slow unintentionally.

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
```diff
@@ -39,54 +41,54 @@
 /**
  * This class is a servlet filter which is registered with the REST API's servlet. The main purpose of the class is to
  * log entry/exit information and elapsed time for each REST API request processed by the server.
- * 
- * @author padams
  */
-public class FHIRRestServletFilter implements Filter {
+public class FHIRRestServletFilter extends HttpFilter {
+    private static final long serialVersionUID = 1L;
```
prb112: are you actually serializing this class?  best to generate it not use 1l.

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
```diff
@@ -100,56 +102,33 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
-                }
-            }
-            FHIRRequestContext.get().setReturnPreference(returnPref);
+            HTTPReturnPreference returnPref = computeReturnPref(request, handlingPref);
+            context.setReturnPreference(returnPref);
 
             // Pass the request through to the next filter in the chain.
             chain.doFilter(request, response);
-        } catch (FHIRException e) {
+        } catch (Exception e) {
```
prb112: Totally agree with this change!

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
```diff
@@ -39,54 +41,54 @@
...
  */
-public class FHIRRestServletFilter implements Filter {
+public class FHIRRestServletFilter extends HttpFilter {
+    private static final long serialVersionUID = 1L;
+
     private static final Logger log = Logger.getLogger(FHIRRestServletFilter.class.getName());
 
     private static String tenantIdHeaderName = null;
     private static String datastoreIdHeaderName = null;
+    private static String originalRequestUriHeaderName = null;
```
prb112: I thought Filters were stateless? ... so the reason I mention this... 
if the configuration changes, a server restart needs to be called to re-activate the init filter

prb112: maybe document that behavior?

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
lmsurpre: I don't think this comment is really related to this change.  It was (and still is) called on each `metadata` request and thats what we have #734 for

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
lmsurpre: I am not, it was just to make eclipse not complain about missing it (apparently HttpFilter is serializable whereas Filter was not)

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
prb112: yeah ... I'm just wondering if we're injecting a pause with the synchronized when two threads are trying to generate it

prb112: No need to resolve here

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
lmsurpre: added to section 5.1.3 of the userguide

lmsurpre:
discussed with paul and we think 1L is fine for now...no immediate plans to serialize it

Issue #686, opened 2020-02-07
Add javadoc to FHIRPath model


lmsurpre:
I added description for two methods on FHIRPathNode, everything else needs work.

lmsurpre:
Will get published to site as part of 4.1.0 release process.

PR #761, opened 2020-03-05
Issue #686 - added Javadoc to FHIRPath model

JohnTimm:


fhir-path/src/main/java/com/ibm/fhir/path/FHIRPathAbstractNode.java
```diff
@@ -28,56 +28,92 @@ protected FHIRPathAbstractNode(Builder builder) {
         children = Collections.unmodifiableCollection(builder.children);
     }
     
+    /**
+     * {@inheritDoc}
+     */
```
lmsurpre: I think `@inheritDoc` is just if you want to combine some new text with the parent text.  If you're not adding anything, just leave it off.  But maybe double-check me :-)

JohnTimm: Removed unnecessary `{@inheritDoc}` usage.

fhir-path/src/main/java/com/ibm/fhir/path/evaluator/FHIRPathEvaluator.java
```diff
@@ -118,29 +146,95 @@ public EvaluationContext getEvaluationContext() {
...
     
     /**
+     * Evaluate a FHIRPathExpression using an existing EvaluationContext against a collection of FHIRPath nodes
+     * 
      * @param evaluationContext
+     *     the evaluation context
      * @param expr
+     *     the FHIRPath expression to evaluate
      * @param initialContext
-     * @return a potentially-empty non-null collection of FHIRPathNodes
```
lmsurpre: is is worth keeping the "potentially-empty non-null" part?

fhir-path/src/main/java/com/ibm/fhir/path/evaluator/FHIRPathEvaluator.java
```diff
@@ -1262,22 +1369,58 @@ private EvaluationContext(FHIRPathTree tree) {
...
+         *     the name
+         * @param nodes
+         *     the collection of FHIRPath nodes
+         */
         public void setExternalConstant(String name, Collection<FHIRPathNode> nodes) {
             externalConstantMap.put(name, nodes);
         }
         
+        /**
+         * Unset an external constant using a name and FHIRPath node
```
lmsurpre: how does this use a FHIRPath node?

fhir-path/src/main/java/com/ibm/fhir/path/evaluator/FHIRPathEvaluator.java
JohnTimm: not in my opinion.

fhir-path/src/main/java/com/ibm/fhir/path/evaluator/FHIRPathEvaluator.java
```diff
@@ -1297,6 +1440,14 @@ public void unsetExternalConstant(String name) {
             }
         }
         
+        /**
+         * Indicates whether this EvaluationContext has an external constant with the given name
+         * 
+         * @param name
+         *     the name
+         * @return
+         *     true if this EvaluationContext has an external constant with the given name if exists, otherwise false
```
lmsurpre:
```suggestion
         *     true if this EvaluationContext has an external constant with the given name, otherwise false
```

fhir-path/src/main/java/com/ibm/fhir/path/evaluator/FHIRPathEvaluator.java
JohnTimm: It doesn't

fhir-path/src/main/java/com/ibm/fhir/path/evaluator/FHIRPathEvaluator.java
```diff
@@ -90,20 +93,45 @@
...
+     *     the {@link Resource} or {@link Element}
      * @param expr
-     * @return a potentially-empty non-null collection of FHIRPathNodes
+     *     the FHIRPath expression to evaluate
+     * @return
+     *     the result of evaluation as a collection of FHIRPath nodes
+     * @throws NullPointerException
+     *     if any of the parameters are null    
      * @throws FHIRPathException
-     * @throws NullPointerException if any of the passed arguments are null
```
lmsurpre: does it not throw NPE any more?

fhir-path/src/main/java/com/ibm/fhir/path/FHIRPathTypeInfoNode.java
```diff
@@ -44,23 +69,35 @@ private Builder(FHIRPathType type, TypeInfo typeInfo) {
...
+    /**
+     * The compareTo operation is not supported for this FHIRPathTypeInfoNode
+     */
     @Override
     public int compareTo(FHIRPathNode o) {
         throw new UnsupportedOperationException();
     }
     
+    /**
+     * Indicates whether the {@link TypeInfo} wrapped by this FHIRPathTypeInfoNode is equal the parameter
```
lmsurpre:
```suggestion
     * Indicates whether the {@link TypeInfo} wrapped by this FHIRPathTypeInfoNode is equal to the parameter
```

fhir-path/src/main/java/com/ibm/fhir/path/evaluator/FHIRPathEvaluator.java
lmsurpre: I originally added that because, as a user of this method, I wanted a guarantee that it will never be null, and a reminder that it might come back empty.  Should a consumer assume those things based on it not saying it can be null?

fhir-path/src/main/java/com/ibm/fhir/path/evaluator/FHIRPathEvaluator.java
lmsurpre:
```suggestion
         * Unset an external constant
```

fhir-path/src/main/java/com/ibm/fhir/path/evaluator/FHIRPathEvaluator.java
```diff
@@ -1262,22 +1369,58 @@ private EvaluationContext(FHIRPathTree tree) {
...
+         *     the name
+         * @param node
+         *     the FHIRPath node
+         */
         public void setExternalConstant(String name, FHIRPathNode node) {
             externalConstantMap.put(name, singleton(node));
         }
         
+        /**
+         * Set an external constant using a name and a collection of FHIRPath node
```
lmsurpre: how is it set?  to each node in the collection?

fhir-path/src/main/java/com/ibm/fhir/path/evaluator/FHIRPathEvaluator.java
JohnTimm: FHIRPath has no notion of null. A null result would be an empty result

fhir-path/src/main/java/com/ibm/fhir/path/evaluator/FHIRPathEvaluator.java
JohnTimm: Yes, via Objects.requireNonNull

fhir-path/src/main/java/com/ibm/fhir/path/evaluator/FHIRPathEvaluator.java
JohnTimm: Added `non-null, potentially empty` language back to the Javadoc

fhir-path/src/main/java/com/ibm/fhir/path/evaluator/FHIRPathEvaluator.java
JohnTimm: Updated comment

fhir-path/src/main/java/com/ibm/fhir/path/evaluator/FHIRPathEvaluator.java
JohnTimm: Updated the comment

fhir-path/src/main/java/com/ibm/fhir/path/evaluator/FHIRPathEvaluator.java
JohnTimm: The collection is the value of the constant. If the constant is a single value, then the collection would be a singleton.

fhir-path/src/main/java/com/ibm/fhir/path/FHIRPathTypeInfoNode.java
JohnTimm: Updated comment

PR #778, opened 2020-03-12
Issue 700 - set reference types based on constraints

lmsurpre:
1. Introduced `com.ibm.fhir.model.annotation.ReferenceTarget` for
capturing the allowed target types for each Reference type in the model

2. Updated CodeGenerator to include these new annotations and also add
the allowed reference types to the javadoc of the Builder setters

3. Added support for introspecting this data in ModelSupport...similar
to what we did for allowed Choice types in the past

4. Updated ExamplesGenerator to set the `type` of each Reference based on
the allowed type and prefix the reference value itself with the same
type.  While in there, I also updated it to use ModelSupport for
choiceElement information and removed the previous approach of using a
pre-computed properties file (choiceElements.properties).

5. Regenerated the IBM fhir-examples

6. Set `checkReferenceTypes` to true and removed the setFalse workaround
for all our tests that were skipping reference type checking due to the invalid examples

7. Updated example indices as needed.
With reference type checking enabled, we found two new issues with the spec examples, one in `medicationrequest0301` and one in `devicemetric-example` (XML and JSON)


lmsurpre:
> I think we should use class objects instead of class simple names for the values in @ReferenceTarget as it is consistent with our other annotations (e.g. Choice).

I did consider this, but I went with Strings for a couple reasons:
* In the case of choice elements, you actually want the class because you need to create an instance of one in order to populate the field.  However, for Reference elements, you actually just want the String value because its either part of the string Reference.reference or a code string in Reference.type.
* The source for this data actually comes from `targetProfile` which can reference any StructureDefinition.  In the base spec, its always going to be a Resource type (and so we'll have a corresponding class), but I think using a string here allows for the possibility of including a full URI in the future.

Issue #781, opened 2020-03-13
Update Observation is slow (under certain circumstances)


lmsurpre:
Create performance looks pretty good to me, but the update performance is looking ugly in our test env (millions of observations).

Update is going to take longer because, with the current design, we do lots of extra work:
1. read before update to distinguish creates from updates and check versions
2. delete all existing search parameters for this logical resource before inserting the new ones

However, my initial analysis points to a possible degradation due to the support I added for Composite search parameter types (which cross multiple tables).  Needs further analysis.

PR #785, opened 2020-03-13
issue #781 - skip constraint checking for composite foreign keys

lmsurpre:
We found that the query plan was doing a full index scan of the
Observation_COMPOSITES table on each delete to each of the
Observation_X_VALUES tables. This is totally unneccessary because we
always delete the composite rows first before we delete the
corresponding X_VALUES rows. Telling Db2 not to enforce this particular
constraint should speed things up nicely.

I also found and fixed an issue where we were only generating the FK 
constraint for component 3 (comp3) of the composites tables. We added 
the FK for each component (all 3), but since we never changed the name 
it was overwriting it instead of adding it for each set of columns (i.e. components).

Also fixed javadoc in FhirResourceTableGroup by:
1. fixing copy-paste issue in addComposites javadoc
2. remove "ON DELETE CASCADE" from everywhere --its simply not used
3. added "NOT ENFORCED" for the addComposites sample FKs

### DIFF
--- a/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/api/IDatabaseAdapter.java
+++ b/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/api/IDatabaseAdapter.java
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright IBM Corp. 2019
+ * (C) Copyright IBM Corp. 2019, 2020
  *
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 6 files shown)


lmsurpre:
I think we just need to nail down the migration story and this should be good to go.

prb112:
verified the query plan changed
```
DELETE FROM FHIRDATA2.OBSERVATION_STR_VALUES SV 
WHERE SV.LOGICAL_RESOURCE_ID = ?;
```

**Prior Query Plan**
image

**New Query Plan**
image

PR #788, opened 2020-03-16
Issue #833 - prototype migration and add automated migration test

lmsurpre:
Add support for arbitrary database migration logic.

This proof-of-concept extends our fhir-database-utils project with the
ability to define migration steps for tables within the schema and should
extend to other object "types" as well.

It also illustrates the usage in two places:
1. drop the indices that were using the DATE_VALUE column
2. drop the ENFORCED FK constraints in all X_COMPOSITES tables (and add the new unenforced ones)


fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/common/ReorgTable.java
```diff
@@ -0,0 +1,48 @@
...
+        DataDefinitionUtil.assertValidName(schemaName);
+        DataDefinitionUtil.assertValidName(tableName);
+        this.schemaName = schemaName;
+        this.tableName = tableName;
+    }
+
+    @Override
+    public void run(IDatabaseTranslator translator, Connection c) {
+        final String qname = DataDefinitionUtil.getQualifiedName(schemaName, tableName);
+        final String ddl = "CALL SYSPROC.ADMIN_CMD ('REORG TABLE " + qname + "')";
```
prb112: I think you need to run this... https://dba.stackexchange.com/a/30233/202822
I think this should be a second run.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/NopObject.java
```diff
@@ -33,6 +33,11 @@ public void apply(IDatabaseAdapter target) {
         // We're NOP so we do nothing on purpose
     }
 
+    @Override
+    public void apply(Integer priorVersion, IDatabaseAdapter target) {
```
prb112: might as well push this into the interface as a default implementation.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/Tablespace.java
```diff
@@ -44,6 +44,17 @@ public void apply(IDatabaseAdapter target) {
         }
     }
 
+    @Override
+    public void apply(Integer priorVersion, IDatabaseAdapter target) {
```
prb112: don't do anything here. we shouldn't make a new tablespace, or alter here.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/BaseObject.java
```diff
@@ -31,28 +31,34 @@
...
+
     // The version number of the application schema this object applies to
     private final int version;
-    
+
+    // Steps to perform before updating from a previous version of this object
+    protected final List<Migration> preSteps;
+
+    // Steps to perform after updating from a previous version of this object
+    protected final List<Migration> postSteps;
```
prb112: I've been reading through the steps included.  I think we should add Migration to the VERSION_HISTORY tracking table and DatabaseObjectType. 

Without these steps, it'd be impossible to resume the specific migration in a deterministic way.  For instance, alter table would not work again, and fail.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/common/DropForeignKeyConstraint.java
```diff
@@ -0,0 +1,55 @@
...
+    private final String schemaName;
+    private final String tableName;
+    private final List<String> constraintNames;
+
+    /**
+     * Public constructor
+     * @param schemaName
+     * @param tableName
+     */
+    public DropForeignKeyConstraint(String schemaName, String tableName, String... constraintName) {
```
prb112: I like this approach 1..* instead of one at a time.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/common/DropForeignKeyConstraint.java
```diff
@@ -0,0 +1,55 @@
+/*
+ * (C) Copyright IBM Corp. 2019
```
prb112: 2020

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/api/AllVersionHistoryService.java
```diff
@@ -22,4 +22,10 @@ public boolean applies(String objectSchema, String objectType, String objectName
         return true;
     }
 
+    @Override
```
prb112: 2020

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/api/IVersionHistoryService.java
```diff
@@ -7,7 +7,7 @@
 package com.ibm.fhir.database.utils.api;
```
prb112: 2020

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/common/DropColumn.java
```diff
@@ -0,0 +1,55 @@
+/*
+ * (C) Copyright IBM Corp. 2019
```
prb112:
```suggestion
 * (C) Copyright IBM Corp. 2019, 2020
```

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/common/DropColumn.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+import com.ibm.fhir.database.utils.api.IDatabaseStatement;
+import com.ibm.fhir.database.utils.api.IDatabaseTranslator;
+
+/**
+ * Drop columns from the schema.table
+ */
+public class DropColumn implements IDatabaseStatement {
```
prb112: I'm leaning towards NOT dropping columns... 

1 - we always leave a rollback point with the original data, and we can MOVE the data at any point, or re-index. 
2 - it should be relatively transparent. 


Thoughts?

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/IDatabaseObject.java
```diff
@@ -16,7 +16,7 @@
 import com.ibm.fhir.task.api.ITaskGroup;
 
 /**
- * Defines the Database Object's expected methods. 
+ * Defines the Database Object's expected methods.
```
prb112: 2020

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/IDatabaseObject.java
```diff
@@ -25,21 +25,29 @@
...
      */
     public void apply(IDatabaseAdapter target);
-    
+
+    /**
+     * Apply migration logic to bring the target database to the current level of this object
+     * @param priorVersion
+     * @param target the database target
+     */
+    public void apply(Integer priorVersion, IDatabaseAdapter target);
```
prb112: I'd say add a NO OP default here.  It'll be fairly repeatable for StoredProcedures or other NO OP content such as Tablespaces.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/Migration.java
```diff
@@ -0,0 +1,17 @@
...
+
+import java.util.List;
+
+import com.ibm.fhir.database.utils.api.IDatabaseStatement;
+
+/**
+ * A database migration step
+ */
+public interface Migration {
+    public List<IDatabaseStatement> migrateFrom(Integer priorVersion);
```
prb112: I think documenting the method. 

Starting from the prior version, the Migration implements as a set of steps that are sequentially executed. 

Also the set of IDatabaseStatements, maybe they should indicate if each step is idempotent. 

Also a minor comment, to be consistent in this code base, shouldn't it be IMigration.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/ObjectGroup.java
```diff
@@ -22,10 +22,10 @@
 
     // the list of objects in our group
     private final List<IDatabaseObject> group = new ArrayList<>();
-    
+
```
prb112: 2020

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/PhysicalDataModel.java
```diff
@@ -32,27 +32,27 @@
  */
 public class PhysicalDataModel implements IDataModel {
     private static final Logger logger = Logger.getLogger(PhysicalDataModel.class.getName());
-    
+
```
prb112: 2020

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/ProcedureDef.java
```diff
@@ -12,14 +12,13 @@
 import com.ibm.fhir.database.utils.api.IDatabaseAdapter;
 
 /**
- * The definition of a stored procedure, whose content is provided by a Supplier<String>
- * function
+ * The definition of a stored procedure, whose content is provided by a Supplier<String> function
```
prb112: 2020

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/ProcedureDef.java
```diff
@@ -44,17 +40,23 @@ public void apply(IDatabaseAdapter target) {
...
 
-    /* (non-Javadoc)
-     * @see com.ibm.fhir.database.utils.model.IDatabaseObject#drop(com.ibm.fhir.database.utils.api.IDatabaseAdapter)
-     */
+    @Override
+    public void apply(Integer priorVersion, IDatabaseAdapter target) {
+        // Serialize the execution of the procedure, to try and avoid the
+        // horrible deadlocks we keep getting
+        synchronized(target) {
+            // TODO migration logic here?
```
prb112: Yes, pre-migration logic makes sense here. 
As does post-migration logic.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/ProcedureDef.java
```diff
@@ -44,17 +40,23 @@ public void apply(IDatabaseAdapter target) {
         }
     }
 
-    /* (non-Javadoc)
-     * @see com.ibm.fhir.database.utils.model.IDatabaseObject#drop(com.ibm.fhir.database.utils.api.IDatabaseAdapter)
-     */
+    @Override
+    public void apply(Integer priorVersion, IDatabaseAdapter target) {
+        // Serialize the execution of the procedure, to try and avoid the
```
prb112: how does synchronized help with deadlocks here? it's synchronized in java code? I realize you are not the original person to author it... but if you know...

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/Reorg.java
```diff
@@ -1,14 +0,0 @@
...
- *
- * SPDX-License-Identifier: Apache-2.0
- */
-
-package com.ibm.fhir.database.utils.model;
-
-/**
- * Representing the SQL function
- */
-public class Reorg {
```
prb112: I actually agree with Reorg.  
I don't think it should be Reorg Table
You can reorg indices as well and other database objects.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/ProcedureDef.java
```diff
@@ -44,17 +40,23 @@ public void apply(IDatabaseAdapter target) {
...
     }
-    
-       @Override
-       protected void grantGroupPrivileges(IDatabaseAdapter target, Set<Privilege> group, String toUser) {
-            target.grantProcedurePrivileges(getSchemaName(), getObjectName(), group, toUser);
-        }
 
+    @Override
+    protected void grantGroupPrivileges(IDatabaseAdapter target, Set<Privilege> group, String toUser) {
+        target.grantProcedurePrivileges(getSchemaName(), getObjectName(), group, toUser);
```
prb112: we should leave a note here to make it intentional that grants should be carefully evaluated to see if there is an update required in the Migration step, such as a change to the operational user.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/common/ReorgTable.java
prb112: I think we should pass in DatabaseObject and choose the object name to reorg or inherit off of the Reorg object (see Reorg below)

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/RowArrayType.java
```diff
@@ -39,6 +39,11 @@ public void apply(IDatabaseAdapter target) {
         target.createArrType(getSchemaName(), getObjectName(), rowTypeName, arraySize);
     }
 
+    @Override
+    public void apply(Integer priorVersion, IDatabaseAdapter target) {
```
prb112: We should only create this one time. I actually don't even think this is in our schema today.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/RowType.java
```diff
@@ -29,22 +29,27 @@ public void apply(IDatabaseAdapter target) {
         target.createRowType(getSchemaName(), getObjectName(), columns);
     }
 
+    @Override
+    public void apply(Integer priorVersion, IDatabaseAdapter target) {
+        target.createRowType(getSchemaName(), getObjectName(), columns);
```
prb112: similar comment to RowType.  we don't actually use it right now.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/Sequence.java
```diff
@@ -32,11 +32,16 @@ public void apply(IDatabaseAdapter target) {
         target.createSequence(getSchemaName(), getObjectName(), this.cache);
     }
 
+    @Override
+    public void apply(Integer priorVersion, IDatabaseAdapter target) {
+        target.createSequence(getSchemaName(), getObjectName(), this.cache);
```
prb112: I firmly believe we should have a NoOp here.  we should not create a sequence again.  only one time.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/RowType.java
```diff
@@ -18,7 +18,7 @@
  */
 public class RowType extends BaseObject {
     private final List<ColumnBase> columns = new ArrayList<>();
-    
+
```
prb112: 2020

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/RowArrayType.java
```diff
@@ -39,6 +39,11 @@ public void apply(IDatabaseAdapter target) {
         target.createArrType(getSchemaName(), getObjectName(), rowTypeName, arraySize);
```
prb112: 2020

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/Sequence.java
```diff
@@ -32,11 +32,16 @@ public void apply(IDatabaseAdapter target) {
         target.createSequence(getSchemaName(), getObjectName(), this.cache);
     }
 
+    @Override
```
prb112: 2020

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/SessionVariableDef.java
```diff
@@ -12,35 +12,31 @@
...
     public void apply(IDatabaseAdapter target) {
         target.createIntVariable(getSchemaName(), getObjectName());
     }
 
-    /* (non-Javadoc)
-     * @see com.ibm.fhir.database.utils.model.IDatabaseObject#drop(com.ibm.fhir.database.utils.api.IDatabaseAdapter)
-     */
+    @Override
+    public void apply(Integer priorVersion, IDatabaseAdapter target) {
+        target.createIntVariable(getSchemaName(), getObjectName());
```
prb112: we should not have logic here to apply a new version.  Only the first time. it's like a sequence, very integral to the overall operation.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/SessionVariableDef.java
```diff
@@ -12,35 +12,31 @@
...
-     */
     @Override
     public void apply(IDatabaseAdapter target) {
         target.createIntVariable(getSchemaName(), getObjectName());
     }
 
-    /* (non-Javadoc)
-     * @see com.ibm.fhir.database.utils.model.IDatabaseObject#drop(com.ibm.fhir.database.utils.api.IDatabaseAdapter)
-     */
+    @Override
```
prb112: 2020

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/Tablespace.java
```diff
@@ -18,8 +18,8 @@
 /**
  * Used to create and drop tablespaces within a database
  */
-public class Tablespace extends DatabaseObject {    
-    
+public class Tablespace extends DatabaseObject {
```
prb112: 2020

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/VersionHistoryService.java
```diff
@@ -24,10 +24,10 @@
 
     // The name of the admin schema we are working with
     private final String adminSchemaName;
-    
+
```
prb112: 2020

fhir-persistence-schema/src/test/java/com/ibm/fhir/schema/control/VersionHistoryServiceTest.java
```diff
@@ -23,24 +23,17 @@
     // history information collected during schema application
     private final ConcurrentHashMap<String, Integer> newHistory = new ConcurrentHashMap<>();
 
-    /* (non-Javadoc)
```
prb112: 2020

fhir-persistence-schema/src/test/java/com/ibm/fhir/schema/control/VersionHistoryServiceTest.java
```diff
@@ -64,4 +57,12 @@ public void addTestHistory(String objectSchema, String objectType, String object
     public static String makeKey(String objectSchema, String objectType, String objectName) {
         return objectSchema + ":" + objectType + ":" + objectName;
     }
+
+    /**
+     * Always returns null
+     */
+    @Override
+    public Integer getVersion(String objectSchema, String objectType, String objectName) {
+        return null;
```
prb112: Is this really testing VersionHistory service? shouldn't it return -1 or something to confirm?

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/Table.java
```diff
@@ -137,6 +137,17 @@ public void apply(IDatabaseAdapter target) {
...
     }
 
+    @Override
+    public void apply(Integer priorVersion, IDatabaseAdapter target) {
+        for (Migration step : preSteps) {
+            step.migrateFrom(priorVersion).stream().forEachOrdered(target::runStatement);
+        }
+        apply(target);
+        for (Migration step : postSteps) {
+            step.migrateFrom(priorVersion).stream().forEachOrdered(target::runStatement);
```
prb112: if you apply this (say 2 steps)
step 1 alter table add column succeeds
step 2 move content (say these are in pre-steps) 
what happens to the above. 

is it idempotent? is it in a deterministic state?  right now I don't think we could successfully migrate beyond one step one version in a deterministic way. 

I think we have to track the migration steps in VERSION_HISTORY and filter the successful steps.

prb112: maybe even log out that the step is being skipped.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/VersionedSchemaObject.java
```diff
@@ -6,6 +6,9 @@
 
 package com.ibm.fhir.database.utils.model;
 
+import java.util.Arrays;
```
prb112: 2020

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/VersionedSchemaObject.java
```diff
@@ -6,6 +6,9 @@
...
+import java.util.List;
+
```
prb112: The Title of the Java class makes me realize that we don't VERSION Indices.  I think we should (we have altered the signatures of a few of them without renaming them - drop-add)

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirResourceTableGroup.java
```diff
@@ -397,6 +400,14 @@ public void addDateValues(List<IDatabaseObject> group, String prefix) {
                 .setTablespace(fhirTablespace)
                 .addPrivileges(resourceTablePrivileges)
                 .enableAccessControl(this.sessionVariable)
+                .addPostStep(priorVersion -> {
```
prb112: I think the addPostStep should be backed by an ArrayList (I know we talked about a queue as well.  As long as all the statements are deterministic).

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirResourceTableGroup.java
```diff
@@ -397,6 +400,14 @@ public void addDateValues(List<IDatabaseObject> group, String prefix) {
...
+                    List<IDatabaseStatement> statements = new ArrayList<>();
+                    if (priorVersion == 1) {
+                        statements.add(new DropColumn(schemaName, tableName, "DATE_VALUE"));
```
prb112: I don't think we should drop this column.  I think we leave it for a minimum number of versions.

prb112: If we decide to keep it we should use a constant for DATE_VALUE_DROPPED_COLUMN something to clearly indicate don't use it.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/common/ReorgTable.java
lmsurpre: that answer is from 2012, i wonder if its still the "best practice".  it even mentions that the first runstats is more for older version...

lmsurpre:
```RUNSTATS
REORG
RUNSTATS
REBIND packages
FLUSH PACKAGE CACHE DYNAMIC```
i wonder how long all that takes?
at least with db2 on cloud, it should be running runstats automatically, although maybe its still "best practice" to run it before and after the reorg...will add it for now

Comment (prb112) [reply fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/common/ReorgTable.java]: on the dataset we had it took 10s of minutes in total

File: fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/common/DropColumn.java (hunk 10 above)
Comment (lmsurpre) [reply fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/common/DropColumn.java]: I'm torn on it.  I'd prefer not to create it going forward since its never used.  But I don't like the idea of some deployments having it and others not.  Hence the idea to drop it.  I think we could add a step to copy the data from that column into the LOW and HIGH columns if it makes us feel better.
However, I'd prefer to just tell the 2 people that installed version 4.0.0 that they need to be aware of it and just be done.

File: fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/common/ReorgTable.java (hunk 1 above)
Comment (lmsurpre) [reply fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/common/ReorgTable.java]: I added the requested logic in `Db2Adapter.runStatement`...but I only did the RUNSTATS sandwhich and not the rebind/flush for now.
Then, later I decided that maybe the whole REORG statement is db2-specific, so I renamed the class to Db2ReorgTable (so I guess I could have had the logic here all-along...).

File: fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/derby/DerbyAdapter.java [hunk 38]
```diff
@@ -174,12 +181,12 @@ public boolean doesTableExist(String schemaName, String tableName) {
...
-    */
+        /* CREATE SEQUENCE fhir_sequence
+         *     AS BIGINT
+         *     START WITH 1
+         *     CACHE 1000
+         *     NO CYCLE;
+        */
         // Derby doesn't support CACHE
         final String sname = DataDefinitionUtil.getQualifiedName(schemaName, sequenceName);
         final String ddl = "CREATE SEQUENCE " + sname + " AS BIGINT START WITH 1 NO CYCLE";
```
prb112: I didn't change this to 1000, should we change to 1000 here?

prb112: to align with db2

fhir-persistence-schema/pom.xml
```diff
@@ -45,24 +45,35 @@
 
     <build>
         <plugins>
+            <!-- <plugin>
```
prb112: how about we use the ant plugin here. 
it'd be platform agnostic.  

the commandline args below don't necessarily work on Windows

fhir-persistence-schema/pom.xml
```diff
@@ -45,24 +45,35 @@
...
                 <configuration>
                     <filesets>
-                        <fileset>
-                            <directory>derby</directory>
-                            <followSymlinks>false</followSymlinks>
-                            <useDefaultExcludes>true</useDefaultExcludes>
-                            <includes>
-                                <include>**/*</include>
-                            </includes>
-                        </fileset>
```
prb112: are we removing it on each build?

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
```diff
@@ -36,13 +36,17 @@
 import com.ibm.fhir.database.utils.api.TenantStatus;
 import com.ibm.fhir.database.utils.common.DataDefinitionUtil;
 import com.ibm.fhir.database.utils.common.JdbcConnectionProvider;
+import com.ibm.fhir.database.utils.common.JdbcPropertyAdapter;
```
prb112: 2020

fhir-persistence-schema/src/test/java/com/ibm/fhir/schema/derby/OldFhirSchemaGenerator.java
```diff
@@ -0,0 +1,790 @@
...
+import com.ibm.fhir.database.utils.model.SessionVariableDef;
+import com.ibm.fhir.database.utils.model.Table;
+import com.ibm.fhir.database.utils.model.Tablespace;
+import com.ibm.fhir.model.type.code.FHIRResourceType;
+import com.ibm.fhir.schema.control.FhirSchemaConstants;
+import com.ibm.fhir.schema.control.Replacer;
+import com.ibm.fhir.schema.control.SchemaGeneratorUtil;
+
+/**
+ * Encapsulates the generation of the FHIR schema artifacts from IBM FHIR Server version 4.0.1
```
prb112: add a comment on this class purpose

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/PopulateResourceTypes.java
```diff
@@ -50,7 +50,7 @@ public void run(IDatabaseTranslator translator, Connection c) {
         try (Statement s = c.createStatement(); PreparedStatement batch = c.prepareStatement(stmtResourceTypeInsert)) {
             s.execute(stmtVariable);
```
prb112: does the set variable work for derby?

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/NopObject.java
lmsurpre: didn't end up doing this one but can if you think its important.
instead, i opted to add "ThrowUnsupportedOperationException" in cases where we're trying to update objects that we shouldn't (i.e. tablespace and sequence).

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/Tablespace.java
lmsurpre: now it throws if you try doing an update:
```
        if (priorVersion != null && priorVersion > 0) {
            throw new UnsupportedOperationException("Modifying tablespaces is not supported");
        }
        apply(target);
```

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/BaseObject.java
lmsurpre: The updates should happen together in a transaction and should either fail or succeed together.
The migration steps themselves should have the logic of what versions they apply to and what versions they don't.
With that said, I'm not fundamentally opposed to adding Migrations to the VERSION_HISTORY if you think its important.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/common/DropForeignKeyConstraint.java
lmsurpre: done

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/common/DropColumn.java
lmsurpre: We discussed as a team and decided to NOT drop the old column.
We *do* drop the indices that were using the column though.

Also, we will *not* be adding the column going forward, so users on version 4.0.1 that upgrade to 4.1.0 or higher will have this extra column sitting around.
Also, we do NOT copy this older column's data into the DATE_START or DATE_END columns, so any old resources that had date values indexed should get updated to use the new approach.

fhir-persistence-schema/src/test/java/com/ibm/fhir/schema/derby/OldFhirResourceTableGroup.java
```diff
@@ -0,0 +1,706 @@
...
+import com.ibm.fhir.database.utils.model.IDatabaseObject;
+import com.ibm.fhir.database.utils.model.ObjectGroup;
+import com.ibm.fhir.database.utils.model.PhysicalDataModel;
+import com.ibm.fhir.database.utils.model.SessionVariableDef;
+import com.ibm.fhir.database.utils.model.Table;
+import com.ibm.fhir.database.utils.model.Tablespace;
+import com.ibm.fhir.schema.control.FhirSchemaTags;
+
+/**
+ * Utility to create all the tables associated with a particular resource type
```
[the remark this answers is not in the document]
lmsurpre: done

fhir-persistence-schema/src/test/java/com/ibm/fhir/schema/derby/OldFhirSchemaGenerator.java
lmsurpre: done

fhir-persistence-schema/pom.xml
lmsurpre: i'm open to it.  i maybe should have removed this commented out step since it doesn't work for creating the 4.0.1 schema, but left it here as a placeholder for that approach.  maybe just a comment above it to say to consider/test on windows?

fhir-persistence-schema/pom.xml
lmsurpre: this "derby" directory no longer exists because i moved the derby instance into the target dir a while back.
now that its in the target dir, its cleaned by `mvn clean` by default.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/derby/DerbyAdapter.java
lmsurpre: no opinion on this one...i forget why you said it was better to start at 1000 even for db2

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/PopulateResourceTypes.java
lmsurpre: Ah, I expect not.  Main doesn't even try it because it skips `allocateTenant` when dbType == DERBY.

lmsurpre: How should we handle for now?  Maybe we just add an `@implNote` to say it only works for multi-tenant schemas?

Or should we overload the constructor and let the class's user tell us if its for a multi-tenant schema or not?

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/IDatabaseObject.java
lmsurpre: didn't end up doing this one but can if you think its important.
instead, i opted to add "ThrowUnsupportedOperationException" in cases where we're trying to update objects that we shouldn't (i.e. tablespace and sequence).

fhir-persistence-schema/src/test/java/com/ibm/fhir/schema/control/VersionHistoryServiceTest.java
lmsurpre: We use the real VersionHistoryService for anywhere that needs a true version (like in the new DerbyMigrationTest).
Perhaps we could return a -1 (or a 0) here to avoid having the null checks on the `apply` methods, but I went with null since we were already using Integer arguments (and not ints).

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/Migration.java
lmsurpre: added this javadoc:
```
    /**
     * Migrate from some previous version to the current version of this database object
     *
     * @param priorVersion
     *            the version being migrated from
     * @return a list of statements to be executed sequentially to bring a database object from the prior version to the
     *         current version
     * @implSpec this method should only be executed when the current version > priorVersion and so the steps need not be
     *         idempotent
     */
```

> Also a minor comment, to be consistent in this code base, shouldn't it be IMigration.
I'm pretending I didn't see this part ;-). @JohnTimm

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/VersionedSchemaObject.java
lmsurpre: should we open it as a future task or you'd like to see that change as part of this PR?

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirResourceTableGroup.java
lmsurpre: addPostStep has been replaced by addMigration, but the comment still applies.
the actual list is instantiated in the VersionedSchemaObject constructor:
```
    public VersionedSchemaObject(String schemaName, String objectName) {
        DataDefinitionUtil.assertValidNames(schemaName, objectName);
        this.schemaName = schemaName;
        this.objectName = objectName;
        this.migrations = new ArrayList<>();
    }
```

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirResourceTableGroup.java
lmsurpre: We decided not to add migration logic which drops the column, but did we remove it from v2 of the table.  I added the following comment to try making it clear:
```
                        // Note: version 1 of this table had a DATE_VALUE column.
                        // We chose not to DROP it here, but it is not used going forward.
```

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/ProcedureDef.java
lmsurpre: pre and post-migration logic has been replaced by a single list of "migrations" for now.
if we end up needing pre and post-migration logic for procedures, we'll need to revisit that later.

for now, i'm just logging a warning if we get to `ProcedureDef.apply(priorVersion, target)` with migration steps, because instead it just calls apply (which invokes `create or replace`)

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/ProcedureDef.java
lmsurpre: i guess the idea is just to allow a single thread to call `createOrReplaceProcedure` at a time (for a given target at least)...i wonder if we should do something similar for adding foreign key constraints...

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/Reorg.java
lmsurpre: well the javadoc and the arguments all say table... how about `Db2Reorg`?
does Reorg apply to other dbs besides Db2?

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/ProcedureDef.java
lmsurpre: not sure why this code is showing as changed...all I did is fix the indentation.

do you think having that comment here makes sense, or is there a spot that makes more sense?
maybe at the class level on ProcedureDef?

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/RowArrayType.java
lmsurpre: yeah, i think this was something we used in our old (DSTU2) schema.  it got dropped when we moved to a dynamic statement because I guess those don't support array types or something (Robin knows the details).
for now, I've just added the same logic that I added for other things that shouldn't get updated:
```
        if (priorVersion != null && priorVersion > 0 && this.version > priorVersion) {
            throw new UnsupportedOperationException("Upgrading row array types is not supported");
        }
```

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/RowType.java
lmsurpre: same resolution as RowArrayType

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/ProcedureDef.java
prb112: that's a good point... now we thought it out, let's monitor it and make it consistent if we need to

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/Reorg.java
prb112: That's a good question.  MSSQL uses REORGANIZE https://docs.microsoft.com/en-us/sql/relational-databases/indexes/reorganize-and-rebuild-indexes?view=sql-server-ver15

prb112: I'm OK resolving.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/VersionedSchemaObject.java
prb112: Let's open a future task.

fhir-persistence-schema/pom.xml
prb112: or drop into a platform profile? or make a future todo

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/PopulateResourceTypes.java
prb112: @Impl note works for me

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirResourceTableGroup.java
lmsurpre: Now that I added the DROP back in, I used this suggestion to use a constant named `DATE_VALUE_DROPPED_COLUMN`

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/PopulateResourceTypes.java
lmsurpre: I added the `@implNote` to the class's javadoc...maybe i should have done it on the method too, but lets call it good for now

fhir-persistence-schema/pom.xml
lmsurpre: yeah, lets revisit this when we get to migration post 4.1.0

Issue #811, opened 2020-03-18
Prepopulate ResourceType Names in Tenant Allocation


prb112:
Pre-populate ResourceType Names in Tenant Allocation

The RESOURCE NAMES are added to the tenant's partition on each tenant as each unique FHIR Resource is added to the data store. 

Proposal is to load the Parameter's values

prb112:
The proposal is in QA

prb112:
I investigated using a CRC32 checksum to ensure a consistent output for the ResourceType.  The CRC creates a large LONG number.  The large number doesn't make sense, so I've inlined a resource_types.properties file with a corresponding test to ensure correctness.

PR #813, opened 2020-03-18
Prepopulate ResourceType Names in Tenant Allocation #811

prb112:


fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
```diff
@@ -816,6 +817,23 @@ protected void allocateTenant() {
         // Fill any static data tables (which are also partitioned by tenant)
         populateStaticTables(gen, tenantKey);
 
+        // Prepopulate the Resource Type Tables
+        try (ITransaction tx = TransactionFactory.openTransaction(connectionPool)) {
+            try {
+                Connection c = connectionPool.getConnection();
```
albertwang-ibm: can we use one try here?

lmsurpre:
I've always wondered same thing!  Why are we building up this static list dynamically?!

I think it would be even better if we could ensure that the RESOURCE_NAMES table is the same for each tenant.  That is, each tenant should have the same ids for each resource type (instead of letting the db assign them using a SEQUENCE like it does now).

prb112:
I'll also look at updating the Java code to optimize (instead of retrieving everytime)

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/Db2PopulateResourceTypes.java
```diff
@@ -0,0 +1,54 @@
...
+                try (CallableStatement cStmt = c.prepareCall(callableStr);) {
+                    cStmt.setString(1, resourceType);
+                    cStmt.registerOutParameter(2, Types.INTEGER);
+                    cStmt.execute();
+                }
+            }
+        } catch (SQLException x) {
+            throw translator.translate(x);
+        }
+
```
albertwang-ibm: better to remove the empty lines.

prb112: will do.

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
prb112: it's the pattern that Robin follows through the code. thus the transaction factory, and then the nested tries

albertwang-ibm: not sure if we need to close the connection or put it back to the connection pool ...

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/Db2PopulateResourceTypes.java
prb112: fixed in my local commit

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
prb112: We do not need to put it back in the pool... from what I've reviewed in the code, however, you are correct that we need to do autocloseable.

prb112: fixed

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/Db2PopulateResourceTypes.java
```diff
@@ -0,0 +1,127 @@
...
+                // Check Error Codes. 
+                int[] codes = batch.executeBatch();
+                int errorCodes = 0;
+                for (int code : codes) {
+                    if (code != 0) {
+                        errorCodes++;
+                    }
+                }
+                if (errorCodes > 0) {
+                    LOGGER.warning("at least one of the Resource Types are not populated [" + errorCodes + "]");
```
lmsurpre: maybe it should be `Level.SEVERE` and rollback?  Is there a case where we'd see this and want to keep going?

prb112: You are right - fixing

prb112:
Fixed

Issue #841, opened 2020-03-24
Reduce the Default Pool Size for the Schema Generator


prb112:
Reduce the Default Pool Size

PR #842, opened 2020-03-24
Reduce the Default Pool Size for the Schema Generator #841

prb112:

### DIFF
--- a/fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirSchemaConstants.java
+++ b/fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirSchemaConstants.java
@@ -13,7 +13,7 @@ public class FhirSchemaConstants {
 
     // A lower pool size is selected as default to limit the likelihood of contention on the DBMS.
     // Standard connection/thread pool size
-    public static final int DEFAULT_POOL_SIZE = 8;
+    public static final int DEFAULT_POOL_SIZE = 1;
     
     // Size of string columns in the search tables. DSTU2 was 511
     public static final int MAX_SEARCH_STRING_BYTES = 1024;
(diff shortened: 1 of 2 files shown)


fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirSchemaConstants.java
```diff
@@ -11,8 +11,9 @@
  */
 public class FhirSchemaConstants {
 
+    // A lower pool size is selected as default to limit the likelihood of contention on the DBMS.
     // Standard connection/thread pool size
-    public static final int DEFAULT_POOL_SIZE = 80;
```
albertwang-ibm: As I remember, when updateschema, we can only set the poolsize to 1, but we don't have to set it for the other operations. not sure if the other operations are using the pool or not, if they are using, will this change make the other operations slower?

prb112: you can always chose to up to 80. 
it may make them sequential, and avoid the resource contention from a higher number. 

I personally want lower than 8, but 8 is what Lee and I thought made sense.

albertwang-ibm: my point is that this change could make us have to set pool size for every other operations, the current setting only makes us have to set it for maybe only one operation ... but anyway, I'm fine with the change. we can always set the pool size as parameter if need to.

albertwang-ibm: or if the pool is only used by updateschema, then maybe we should make it even smaller... :) double checking the codes now ...

prb112: it's essentially used in the PoolConnectionProvider (underneath)

Issue #846, opened 2020-03-24
--add-tenant-key missing in docs


prb112:
--add-tenant-key missing in docs

PR #847, opened 2020-03-24
--add-tenant-key missing in docs #846

prb112:

### DIFF
--- a/fhir-persistence-schema/README.md
+++ b/fhir-persistence-schema/README.md
@@ -141,6 +141,23 @@ Edit `wlp/usr/servers/fhir-server/config/default/fhir-server-config.json` and ad
 --tenant-key "<the-base64-tenant-key>"
 ```
 
+### Add a Key to Existing Tenant
+To add a tenant key for an existing tenant, replace FHIRDATA with your client schema, and TENANT_NAME with your tenant's name. 
+
+```
+--prop-file db2.properties
+--schema-name FHIRDATA
+--add-tenant-key TENANT_NAME
+```
+
+**Example Output**
+```
+2020-03-24 13:54:36.387 00000001    INFO .common.JdbcConnectionProvider Opening connection to database: jdbc:db2://localhost:50000/FHIRDB
+2020-03-24 13:54:37.012 00000001    INFO   com.ibm.fhir.schema.app.Main New tenant key: TNT1 [key=LogFbM06+PLS1cur/NOTREALg=]
+2020-03-24 13:54:37.014 00000001    INFO   com.ibm.fhir.schema.app.Main Processing took:   0.637 s
+2020-03-24 13:54:37.015 00000001    INFO   com.ibm.fhir.schema.app.Main SCHEMA CHANGE: OK
+```
+
 ### Update the stored procedures for FHIRDATA (and not FHIR_ADMIN)
 
 ```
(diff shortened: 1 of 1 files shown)


fhir-persistence-schema/README.md
```diff
@@ -141,6 +141,23 @@ Edit `wlp/usr/servers/fhir-server/config/default/fhir-server-config.json` and ad
 --tenant-key "<the-base64-tenant-key>"
 ```
 
+### Add a Key to Existing Tenant
+To add a tenant key for an existing tenant, replace FHIRDATA with your client schema, and TENANT_NAME with your tenant's name. 
+
+```
+--prop-file db2.properties
+--schema-name FHIRDATA
+--add-tenant-key TENANT_NAME
```
lmsurpre: The other snippets in this README use `default` as the tenant name.
I think we should either do the same here (or update everywhere to TENANT_NAME).

fhir-persistence-schema/README.md
```diff
@@ -141,6 +141,23 @@ Edit `wlp/usr/servers/fhir-server/config/default/fhir-server-config.json` and ad
 --tenant-key "<the-base64-tenant-key>"
 ```
 
+### Add a Key to Existing Tenant
+To add a tenant key for an existing tenant, replace FHIRDATA with your client schema, and TENANT_NAME with your tenant's name. 
+
```
albertwang-ibm: do we have info about when is needed to add a key to existing tenant? and how to delete the old tenant key if need etc.

lmsurpre: I actually had the same thought.  I don't think there is a function to remove a prior key, but I think there should be.

prb112: I can add a comment on this.

prb112: image
We only have allocate and drop and add

prb112: thoughts?

albertwang-ibm: allocateTenant will add the tenant and also create a tenant key. 
does dropTenant drop all tenant keys? I guess it should. 
did Robin add the add key for existing tenant feature for security reason? e.g, the tenant key is expose, so need a new key for the tenant ... if so, I think, we do need a command to remove the key ...  and I guess, all the tenant keys of a tenant can work to access the tenant, so if don't delete the old key could bring security issue. ...

albertwang-ibm: or it could be that, Robin added this feature just in case the old tenant key is missed/forgotten, but we still want to access that tenant, so this make it handy for us to just create a new key for the tenant and then use it to access the tenant ...

prb112: dropTenant does not remove a key.  it just change the tenant status and removes the partition.  It was for key rotation (except there are NO delete-key features)

albertwang-ibm: ha, I see. then I think if possible, we need:
(1) add removing all the tenant keys function in dropTenant, otherwise they are just useless garbage data after the tenant is dropped.
(2) add remove key command, use tenant name and key as parameter.
And as I remember, the tenant key itself is not stored in DB, but the hash with a salt of the key is stored in db instead, so we will need to remove the hash from the table instead...

albertwang-ibm: but anyway, I'm OK with this document change.

prb112: it's best to leave it.  I think it's good audit data, and ensures the id is not re-used.

prb112: I think remove key should be added, that's fairly easy to use.  If agreed, we can create an issue to track

Issue #853, opened 2020-03-26
Issues with DerbyMaster and derby.properties design


lmsurpre:
**Describe the bug**
`fhir-persistence-schema` project has `com.ibm.fhir.database.utils.derby.DerbyMaster` which provides an abstraction for an AutoCloseable embedded derby instance thats used by both tests and "bootstrapping" code (when persistence type is derby and bootstrap is set to true--which is the default config).

Currently, whenever instantiated from a test, it logs the following warning:
`WARNING: Failed to create derby.properties file!`

Given the current design, this will always happen unless you happen to have a derby dir under the current working directory of the launched application.
And even then, derby will not use this derby.properties file unless its launched from that particular directory (like in the network derby case).

**To Reproduce**
Execute any test that invokes DerbyMaster

**Expected behavior**
DerbyMaster should have a more flexible design and we should have a better approach for providing derby.properties.

**Additional context**
Additionally, the current design of DerbyMaster doesn't seem compatible with our VersionHistoryService and, while adding a migration test for #833 I found that the DEBUG property doesn't work when launched as a test because derby only checks the working directory for the derby.properties file.

albertwang-ibm:
yes, the adding of the properties file for derby is for the performance issue that we found in pipeline integration test, and there is always derby directory there, so it worked well for integration tests.

PR #862, opened 2020-03-27
issue #853 use system properties instead of derby.properties

albertwang-ibm:
Changed to use system properties, tested and found this works well with all derby embedded related unit tests and server integration tests.


fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/derby/DerbyServerPropertiesMgr.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.util.Properties;
+
+
+/**
+ * Server properties for embedded derby which is used in unit tests and server integration tests,
+ * equal to setting in derby.properties.
+ */
+public class DerbyServerPropertiesMgr {
+    public static void setServerProperties (boolean isDebug) {
```
prb112:
```suggestion
    public static void setServerProperties(boolean isDebug) {
```

fhir-server/src/main/java/com/ibm/fhir/server/listener/FHIRServletContextListener.java
```diff
@@ -140,8 +141,13 @@ public void contextInitialized(ServletContextEvent event) {
      */
     private void bootstrapDerbyDatabases(PropertyGroup fhirConfig) throws Exception {
         Boolean performDbBootstrap = fhirConfig.getBooleanProperty(PROPERTY_JDBC_BOOTSTRAP_DB, Boolean.FALSE);
+        // Controls if we run derby in debugging mode which enables more logs.
+        boolean DEBUG = false;
```
prb112: it's not really a constant at this point. let's just pass in the boolean value and a comment on line 149....

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/derby/DerbyServerPropertiesMgr.java
```diff
@@ -0,0 +1,29 @@
...
+        Properties sysProperties = System.getProperties();
```
prb112: does Derby automatically write this out when it's set on System.getProperties?

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/derby/DerbyMaster.java
```diff
@@ -56,22 +51,7 @@
     public DerbyMaster(String database) {
         this.database = database;
 
-        try (PrintWriter out = new PrintWriter(DERBY_PROPERTIES)){
-            Class.forName(DERBY_TRANSLATOR.getDriverClassName());
```
prb112: I take it the translator already does the class name lookup?  otherwise if it doesn't I think this should be done on line 53. it's the general jdbc impl class loading pattern

fhir-server/src/main/java/com/ibm/fhir/server/listener/FHIRServletContextListener.java
albertwang-ibm: agreed, thanks!

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/derby/DerbyServerPropertiesMgr.java
albertwang-ibm: no, it doesn't.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/derby/DerbyMaster.java
albertwang-ibm: yeah, it's interesting that removing this has been working well. checking if this is really needed here ....

albertwang-ibm: did some search, JDBC 4.0 (and after) drivers that are found in your class path are automatically loaded, only JDBC drivers older than 4.0 have to be loaded by Class.forName.
I will keep it just in case JDBC older drivers are/(will be) still used...

fhir-server/src/main/java/com/ibm/fhir/server/listener/FHIRServletContextListener.java
albertwang-ibm: done

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/derby/DerbyMaster.java
albertwang-ibm: added it back with comment even thought I believe it's not really needed now

fhir-server/src/main/java/com/ibm/fhir/server/listener/FHIRServletContextListener.java
```diff
@@ -142,6 +143,10 @@ private void bootstrapDerbyDatabases(PropertyGroup fhirConfig) throws Exception
         Boolean performDbBootstrap = fhirConfig.getBooleanProperty(PROPERTY_JDBC_BOOTSTRAP_DB, Boolean.FALSE);
         if (performDbBootstrap) {
             log.info("Performing Derby database bootstrapping...");
+
+            // Start derby with debug off by default.
+            DerbyServerPropertiesMgr.setServerProperties(false);
```
lmsurpre: Is this actually needed?  Isn't "false" the default?

Or I guess you add this here just to make it easy to turn on should we want to turn it on?

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/derby/DerbyServerPropertiesMgr.java
prb112: How does this write out? ... 

`try (PrintWriter out = new PrintWriter(DERBY_PROPERTIES)){ `

Is very different from System.getProperties

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/derby/DerbyMaster.java
prb112: it's best to maintain a consistent pattern - it's in FhirDbDaoImpl too

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/derby/DerbyServerPropertiesMgr.java
albertwang-ibm: yes, you are correct. I had used the perperties file because I did find this way at that time, now I found this way is much easier and don't have to associate with any directory.

prb112: Excellent.  thanks

Issue #664, opened 2020-01-31
conditionalDelete/multiple support


albertwang-ibm:
In current conditionDelete implementation, if multiple resources are found, then server returns 412 to fail the deletion request because the server only supports conditionalDelete/single at present.

This issue is opened for enhancement of the server to support conditionalDelete/multiple.

PR #869, opened 2020-03-30
issue #664 #884 conditional delete multiple match support and config for disabling bulk export public access

albertwang-ibm:
(1)This PR enabled fhir server to support conditional delete with multiple matches.
passed all tests including the 2 touchstone fhir-api-r4-patient-delete-03-conditional-multiple test cases which failed before.

https://touchstone.aegis.net/touchstone/scriptexecution?exec=202003301558230438592632&qn=/FHIR4-0-1-Advanced/Patient/05-delete/fhir-api-r4-patient-delete-03-conditional-multiple#WAITING_OPER
(2) This PR also added server configure "fhirserver.bulkdata.isExportPublic" to allow user to disable public access to the exported files in COS/S3.

### DIFF
--- a/docs/src/pages/guides/FHIRServerUsersGuide.md
+++ b/docs/src/pages/guides/FHIRServerUsersGuide.md
@@ -1422,7 +1422,7 @@ must restart the server for that change to take effect.
 |`fhirServer/core/defaultHandling`|Y|Y|
 |`fhirServer/core/allowClientHandlingPref`|Y|Y|
 |`fhirServer/core/checkReferenceTypes`|N|N|
-|`fhirServer/core/conditionalDeleteMaxNumber`|N|N|
+|`fhirServer/core/conditionalDeleteMaxNumber`|Y|Y|
 |`fhirServer/searchParameterFilter`|Y|Y|
 |`fhirServer/notifications/common/includeResourceTypes`|N|N|
 |`fhirServer/notifications/websocket/enabled`|Y|Y|
--- a/docs/src/pages/guides/FHIRServerUsersGuide.md
+++ b/docs/src/pages/guides/FHIRServerUsersGuide.md
@@ -1358,7 +1358,7 @@ This section contains reference information about each of the configuration prop
 |`fhirServer/bulkdata/jobParameters/credential.ibm`| If use IBM credential, "Y" or "N" |
 |`fhirServer/bulkdata/jobParameters/cos.api.key`|string|API key for accessing IBM COS |
 |`fhirServer/bulkdata/jobParameters/cos.srvinst.id`|string|Service instance Id for accessing IBM COS |
-|`fhirServer/bulkdata/implementation_type`|string|"cos" (Any Amazon S3-compatible object stores including IBM COS) or "dummy" which matches the desired bulk operation implementation type |
+|`fhirServer/bulkdata/implementation_type`|string|Use "cos" for any S3-compatible object store |
 |`fhirServer/bulkdata/batch-uri`|string|The URL to access the FHIR server hosting the batch web application |
 |`fhirServer/bulkdata/batch-user`|string|User for submitting JavaBatch job |
 |`fhirServer/bulkdata/batch-user-password`|string|Password for above batch user |
--- a/docs/src/pages/guides/FHIRServerUsersGuide.md
+++ b/docs/src/pages/guides/FHIRServerUsersGuide.md
@@ -1143,7 +1143,30 @@ The *fhir-bulkimportexport-webapp* module is a wrapper for the whole BulkData we
     </webApplication>
 ```
 
-BulkData web application writes the exported FHIR resources to an IBM Cloud Object Storage (COS) or Amazon S3 bucket as configured in the per-tenant server configuration under fhirServer/bulkdata. 
+BulkData web application writes the exported FHIR resources to an IBM Cloud Object Storage (COS) or any Amazon S3-Compatible bucket as configured in the per-tenant server configuration under fhirServer/bulkdata. Following is an example configuration for bulkdata, please refer to section 5 for the detailed description of these properties:
+
+```
+        "bulkdata": {
+            "bulkDataBatchJobIdEncryptionKey": "change-password",
+            "applicationName": "fhir-bulkimportexport-webapp",
+            "moduleName": "fhir-bulkimportexport.war",
+            "jobParameters": {
+                "cos.bucket.name": "fhir-r4-connectathon",
+                "cos.location": "us",
+                "cos.endpointurl": "fake",
+                "cos.credential.ibm": "Y",
+                "cos.api.key": "fake",
+                "cos.srvinst.id": "fake"
+            },
+            "implementation_type": "cos",
+            "batch-uri": "https://localhost:9443/ibm/api/batch/jobinstances",
+            "batch-user": "fhiradmin",
+            "batch-user-password": "change-password",
+            "batch-truststore": "resources/security/fhirTruststore.jks",
+            "batch-truststore-password": "change-password",
+            "isExportPublic": true
+        }
+```
 
 To use Amazon S3 bucket for exporting, please set cos.credential.ibm to "N", set cos.api.key to S3 access key, and set cos.srvinst.id to S3 secret key. The following is a sample path to the exported ndjson file, the full path can be found in the response to the polling location request after the export request (please refer to the FHIR BulkDataAccess spec for details).  
 
@@ -1335,7 +1358,7 @@ This section contains reference information about each of the configuration prop
 |`fhirServer/bulkdata/jobParameters/credential.ibm`| If use IBM credential, "Y" or "N" |
 |`fhirServer/bulkdata/jobParameters/cos.api.key`|string|API key for accessing IBM COS |
 |`fhirServer/bulkdata/jobParameters/cos.srvinst.id`|string|Service instance Id for accessing IBM COS |
-|`fhirServer/bulkdata/implementation_type`|string|"cos" or "dummy" which matches the desired bulk operation implementation type |
+|`fhirServer/bulkdata/implementation_type`|string|"cos" (Any Amazon S3-compatible object stores including IBM COS) or "dummy" which matches the desired bulk operation implementation type |
 |`fhirServer/bulkdata/batch-uri`|string|The URL to access the FHIR server hosting the batch web application |
 |`fhirServer/bulkdata/batch-user`|string|User for submitting JavaBatch job |
 |`fhirServer/bulkdata/batch-user-password`|string|Password for above batch user |
--- a/docs/src/pages/guides/FHIRServerUsersGuide.md
+++ b/docs/src/pages/guides/FHIRServerUsersGuide.md
@@ -1143,50 +1143,7 @@ The *fhir-bulkimportexport-webapp* module is a wrapper for the whole BulkData we
     </webApplication>
 ```
 
-BulkData web application writes the exported FHIR resources to an IBM Cloud Object Storage (COS) or Amazon S3 bucket, as configured in the per-tenant bulkdata.json configuration file. The bulkdata.json file is stored in the tenant configuration directory of each fhir-server instance. The following is a bulkdata.json which is configured to export the FHIR resources into fhir-bulkdata-sample bucket of IBM COS:
-
-```json
-{
-  "applicationName": "fhir-bulkimportexport-webapp",
-  "moduleName": "fhir-bulkimportexport.war",
-  "jobParameters": {
-    "cos.bucket.name": "fhir-bulkdata-sample",
-    "cos.location": "us",
-    "cos.endpointurl": "https://s3.us-south.cloud-object-storage.appdomain.cloud",
-    "cos.credential.ibm": "Y",
-    "cos.api.key": "xxxxxxxxxxxxxxxxxxxxxxx",
-    "cos.srvinst.id": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
-  },
-  "implementation_type": "cos",
-  "batch-uri": "https://localhost:9443/ibm/api/batch/jobinstances",
-  "batch-user": "fhiradmin",
-  "batch-user-password": "xxxxxxxxxxxx",
-  "batch-truststore" : "resources/security/fhirTruststore.jks",
-  "batch-truststore-password" : "xxxxxxxxxxxx",
-  "serverHostname" : "localhost:9443",
-  "contextRoot" : "/fhir-server/api/v4"
-}
-```
-The following configured parameters are:
-
-|Parameter Name   | Description|
-|--------------| ------------|
-|`applicationName`| fixed value, always set to fhir-bulkimportexport-webapp |
-|`moduleName`| fixed value, always set to fhir-bulkimportexport.war |
-|`jobParameters.cos.bucket.name`| object store bucket name |
-|`jobParameters.cos.location`| object store location |
-|`jobParameters.cos.endpointurl`| object store end point url |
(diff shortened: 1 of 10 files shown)


fhir-server-test/src/test/java/com/ibm/fhir/server/test/BundleTest.java
```diff
@@ -2209,8 +2209,7 @@ public void testBatchConditionalDeletes() throws Exception {
         assertResponseBundle(responseBundle, BundleType.BATCH_RESPONSE, 4);
         assertGoodGetResponse(responseBundle.getEntry().get(0), Status.OK.getStatusCode(), HTTPReturnPreference.MINIMAL);
         assertGoodGetResponse(responseBundle.getEntry().get(1), Status.NO_CONTENT.getStatusCode(), HTTPReturnPreference.MINIMAL);
-        assertBadResponse(responseBundle.getEntry().get(2), Status.PRECONDITION_FAILED.getStatusCode(),
-                "returned multiple matches");
+        assertGoodGetResponse(responseBundle.getEntry().get(2), Status.NO_CONTENT.getStatusCode(), HTTPReturnPreference.MINIMAL);
```
albertwang-ibm: this is the only real change in this file, all the others are just formats like removing empty spaces at the line ends.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/BundleTest.java
```diff
@@ -2209,8 +2209,22 @@ public void testBatchConditionalDeletes() throws Exception {
         assertResponseBundle(responseBundle, BundleType.BATCH_RESPONSE, 4);
         assertGoodGetResponse(responseBundle.getEntry().get(0), Status.OK.getStatusCode(), HTTPReturnPreference.MINIMAL);
         assertGoodGetResponse(responseBundle.getEntry().get(1), Status.NO_CONTENT.getStatusCode(), HTTPReturnPreference.MINIMAL);
-        assertBadResponse(responseBundle.getEntry().get(2), Status.PRECONDITION_FAILED.getStatusCode(),
-                "returned multiple matches");
+
+        // A search that results in multiple matches:
+        // (1) if matches > FHIRConstants.FHIR_CONDITIONAL_DELETE_MAX_NUMBER_DEFAULT, then result in a 400 status code.
```
lmsurpre, line 2214: should it be a 412 to be more consistent with the fhir spec?

lmsurpre: > Multiple matches: A server may choose to delete all the matching resources, or it may choose to return a 412 Precondition Failed error indicating the client's criteria were not selective enough. A server indicates whether it can delete multiple resources in its Capability Statement (.rest.resource.conditionalDelete)

fhir-server-test/src/test/java/com/ibm/fhir/server/test/DeleteTest.java
```diff
@@ -430,17 +435,31 @@ public void testConditionalDeleteResource() throws Exception {
         } else {
             assertResponse(response.getResponse(), Response.Status.METHOD_NOT_ALLOWED.getStatusCode());
         }
-        
-        // A search that results in multiple matches should result in a 412 status code.
+
+        // A search that results in multiple matches:
+        // (1) if matches > FHIRConstants.FHIR_CONDITIONAL_DELETE_MAX_NUMBER_DEFAULT, then result in a 400 status code.
```
lmsurpre, line 440: should it be a 412 to be more consistent with the fhir spec?

docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -1420,6 +1422,7 @@ must restart the server for that change to take effect.
 |`fhirServer/core/defaultHandling`|Y|Y|
 |`fhirServer/core/allowClientHandlingPref`|Y|Y|
 |`fhirServer/core/checkReferenceTypes`|N|N|
+|`fhirServer/core/conditionalDeleteMaxNumber`|N|N|
```
lmsurpre, line 1425: because you use FHIRConfigHelper to look up the value of the config property, I think this means it *is* client-specific and dynamic
```suggestion
|`fhirServer/core/conditionalDeleteMaxNumber`|Y|Y|
```

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -499,26 +503,28 @@ public FHIRRestOperationResponse doDelete(String type, String id, String searchQ
             // resource to be deleted. Otherwise, we'll use the id value to identify the resource
             // to be deleted.
             Resource resourceToDelete = null;
+            Bundle responseBundle = null;
+
             if (searchQueryString != null) {
+                int searchPageSize = FHIRConfigHelper.getIntProperty(FHIRConfiguration.PROPERTY_CONDITIONAL_DELETE_MAX_NUMBER, FHIRConstants.FHIR_CONDITIONAL_DELETE_MAX_NUMBER_DEFAULT);
```
lmsurpre, line 509: because you use FHIRConfigHelper to look it up, I think this means it *is* client-specific

fhir-server-test/src/test/java/com/ibm/fhir/server/test/BundleTest.java
albertwang-ibm: I choose not to use 412 just because I thought 412 means we don't support multiple matches at all. but we does support. anyway, I'm ok to change to use 412 instead.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/DeleteTest.java
albertwang-ibm: yes, I'm Ok to change to use 412.

docs/src/pages/guides/FHIRServerUsersGuide.md
albertwang-ibm: good catch, actually I was not sure about this.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
albertwang-ibm: yes, I agree. just was a little bit confused about the document.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/BundleTest.java
albertwang-ibm: done

fhir-server-test/src/test/java/com/ibm/fhir/server/test/DeleteTest.java
albertwang-ibm: done

docs/src/pages/guides/FHIRServerUsersGuide.md
albertwang-ibm: done

fhir-bulkimportexport-webapp/src/main/java/com/ibm/fhir/bulkexport/system/ChunkWriter.java
```diff
@@ -154,11 +158,18 @@ private void pushFhirJsonsToCos(InputStream in, int dataLength) throws Exception
             if (cosBucketPathPrefix != null && cosBucketPathPrefix.trim().length() > 0) {
                 itemName = cosBucketPathPrefix + "/" + ResourceTypes.get(chunkData.getIndexOfCurrentResourceType())
                             + "_" + chunkData.getPartNum() + ".ndjson";
+                if (isExportPublic) {
+                    // Set expiration time to 2 hours(7200 seconds).
+                    metadata.setExpirationTime(Date.from(Instant.now().plusSeconds(7200)));
```
lmsurpre, line 163: Is this the part that you mentioned only works on Amazon S3 and not COS?  Or that's something else?

docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -1368,6 +1327,22 @@ This section contains reference information about each of the configuration prop
...
 |`fhirServer/search/useBoundingRadius`|boolean|True, the bounding area is a Radius, else the bounding area is a box.|
+|`fhirServer/bulkdata/applicationName`| string|Fixed value, always set to fhir-bulkimportexport-webapp |
+|`fhirServer/bulkdata/moduleName`|string| Fixed value, always set to fhir-bulkimportexport.war |
+|`fhirServer/bulkdata/jobParameters/cos.bucket.name`|string|Object store bucket name |
+|`fhirServer/bulkdata/jobParameters/cos.location`|Object store location |
+|`fhirServer/bulkdata/jobParameters/cos.endpointurl`| Object store end point url |
+|`fhirServer/bulkdata/jobParameters/credential.ibm`| If use IBM credential, "Y" or "N" |
+|`fhirServer/bulkdata/jobParameters/cos.api.key`|string|API key for accessing IBM COS |
+|`fhirServer/bulkdata/jobParameters/cos.srvinst.id`|string|Service instance Id for accessing IBM COS |
+|`fhirServer/bulkdata/implementation_type`|string|"cos" or "dummy" which matches the desired bulk operation implementation type |
```
lmsurpre, line 1338: does "cos" cover all S3-compatible object stores or just the IBM Cloud one?

docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -1143,50 +1143,7 @@ The *fhir-bulkimportexport-webapp* module is a wrapper for the whole BulkData we
...
-|`jobParameters.cos.api.key`| api key for accessing IBM COS |
-|`jobParameters.cos.srvinst.id`| service instance Id for accessing IBM COS |
-|`implementation_type`| cos or dummy which matches the desired bulk operation implementation type |
-|`batch-uri`| the URL to access the FHIR server hosting the batch web application |
-|`batch-user`| user for submitting JavaBatch job |
-|`batch-user-password`| password for above batch user |
-|`batch-truststore`| trust store for JavaBatch job submission |
-|`batch-truststore-password`| password for above trust store |
-|`serverHostname`| host name part of the server generated polling location url |
-|`contextRoot`| context root part of the server generated polling location url |
```
lmsurpre, line 1189: I agree its best to move these to the reference table in section 5; good call

docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -1143,50 +1143,7 @@ The *fhir-bulkimportexport-webapp* module is a wrapper for the whole BulkData we
...
-  "implementation_type": "cos",
-  "batch-uri": "https://localhost:9443/ibm/api/batch/jobinstances",
-  "batch-user": "fhiradmin",
-  "batch-user-password": "xxxxxxxxxxxx",
-  "batch-truststore" : "resources/security/fhirTruststore.jks",
-  "batch-truststore-password" : "xxxxxxxxxxxx",
-  "serverHostname" : "localhost:9443",
-  "contextRoot" : "/fhir-server/api/v4"
-}
-```
```
lmsurpre, line 1169: I agree this content is outdated; now the config data exists in fhir-server-config.json instead of a separate bulkdata.json; however, I think maybe we should leave a sample config here just to illustrate the properties (and refer to section 5 for the full list of properties).

fhir-bulkimportexport-webapp/src/main/java/com/ibm/fhir/bulkexport/system/ChunkWriter.java
albertwang-ibm: yes, correct. IBM COS doesn't support object level expiration time, it only supports bucket level with policy setting which is well documented in IBM COS document, but IBM COS doesn't complaint if you set it even though it doesn't take effect at all. 

I have moved the codes into a separated if (isExportPublic) to avoid confusing - the old codes worked for S3 without any problem which I verified using my Amazon acount and also verified it didn't work in IBM COS;  I think this is because the meta object was really used only when cosClient.putObject is called later, I moved it just for the purpose of avoiding confusing of why do we change the meta after the request is created with the meta as a parameter.  and still, it works for S3 only.

docs/src/pages/guides/FHIRServerUsersGuide.md
albertwang-ibm: I think the answer is all, let me change the description a little bit. 
IBM COS client SDK is forked from Amazon S3 SDK.

docs/src/pages/guides/FHIRServerUsersGuide.md
albertwang-ibm: good point, let me add one

albertwang-ibm: done

docs/src/pages/guides/FHIRServerUsersGuide.md
albertwang-ibm: done

fhir-bulkimportexport-webapp/src/main/java/com/ibm/fhir/bulkexport/system/ChunkWriter.java
lmsurpre: Thanks for the explanation...mind adding a quick comment in the code to indicate that IBM COS doesn't honor this (but also doesn't fail on it)?

docs/src/pages/guides/FHIRServerUsersGuide.md
lmsurpre: I think we should say this in the "description" column of this row. Something like this:
> Use "cos" for any S3-compatible object store

docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -1399,6 +1375,7 @@ This section contains reference information about each of the configuration prop
 |`fhirServer/audit/serviceProperties/geoCity`|Dallas|
 |`fhirServer/audit/serviceProperties/geoState`|TX|
 |`fhirServer/audit/serviceProperties/geoCounty`|US|
+|`fhirServer/bulkdata/isExportPublic`|true|
```
lmsurpre, line 1401: Should we add the other bulkdata properties here too?  Do any of them have a "default"?

fhir-bulkimportexport-webapp/src/main/java/com/ibm/fhir/bulkexport/system/ChunkWriter.java
albertwang-ibm: sure, good point!

docs/src/pages/guides/FHIRServerUsersGuide.md
albertwang-ibm: The others are either with fixed values, or bucket configuration or password or encryption key, so I didn't add them here.

docs/src/pages/guides/FHIRServerUsersGuide.md
albertwang-ibm: en, interesting, seems my change doesn't show up here. I changed to "  "cos" (Any Amazon S3-compatible object stores including IBM COS) or "dummy" which matches the desired bulk operation implementation type ".     but I do think your description is better, let me change to it.

albertwang-ibm: done

fhir-bulkimportexport-webapp/src/main/java/com/ibm/fhir/bulkexport/system/ChunkWriter.java
albertwang-ibm: done

Issue #823, opened 2020-03-20
Add "which operations are supported" to Conformance and CapabilityStatement


lmsurpre:
We need to advertise which operations we support in two places:

1. Update https://ibm.github.io/FHIR/Conformance with a version of the information we have at https://github.com/ibm/fhir#extended-operations

2. Update `Capabilities.java` with support for introspecting the OperationRegistry and advertise the supported "extended operations", including implementation-defined operations.

lmsurpre:
In #796 we added info about what profiles we support to the CapabilityStatement and ValidationGuide.  As part of this item, lets add this info to the Conformance.md doc too.

PR #881, opened 2020-04-01
issue #823 - Add extended operations to the CapabilityStatement

lmsurpre:
Also added helper methods for constructing a Canonical from a uri and an optional version and/or fragment.

and some miscelaneous javadoc changes


fhir-model/src/main/java/com/ibm/fhir/model/type/Canonical.java
```diff
@@ -33,6 +33,25 @@ public static Canonical of(java.lang.String value) {
...
 
+    public static Canonical of(java.lang.String uri, java.lang.String version) {
+        StringBuilder value = new StringBuilder(uri);
+        if (version != null && !version.isEmpty()) {
+            value.append("|" + version);
+        }
+        return Canonical.builder().value(value.toString()).build();
+    }
+
+    public static Canonical of(java.lang.String uri, java.lang.String version, java.lang.String fragment) {
```
JohnTimm, line 44: If you are going to use StringBuilder then use it for everything:

value.append("|").append(version);

don't mix and match string builder and the concatenation operator. this will give you the most benefit of using something like string builder

lmsurpre: done

Issue #838, opened 2020-03-23
More metrics for BulkImport JavaBatch Job


albertwang-ibm:
The enhancements to make:
(1) Calculate and log processing rate after a certain number of resources are ingested, e.g, calculate and log processing rate after every 1000 resources (or 50 chunks each with 20 resources) are ingested.
(2) Adding processing times for reading from data source and writing to target DB.

albertwang-ibm:
please notice that the in-fly rate calculation will be per partition instead of aggregated, because each partition process its own resources separated,  aggregation is done only at the end of the job.

PR #894, opened 2020-04-03
WIP: issue #838 Add more metrics for bulkImport java Batch Job

albertwang-ibm:
Added metrics for each resource type:
  (1) In-fly importing rate.
  (2) Processing times used for reading from data source and writing to target DB.
  (3) Total resource size and average resource size.
  (4) Validation time.

### DIFF
--- a/fhir-bulkimportexport-webapp/src/main/java/com/ibm/fhir/bulkcommon/BulkDataUtils.java
+++ b/fhir-bulkimportexport-webapp/src/main/java/com/ibm/fhir/bulkcommon/BulkDataUtils.java
@@ -7,6 +7,7 @@
 package com.ibm.fhir.bulkcommon;
 
 import java.io.BufferedReader;
+import java.io.File;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.StringReader;
@@ -24,6 +25,8 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+
+import javax.net.ssl.HttpsURLConnection;
 
 import com.ibm.cloud.objectstorage.ClientConfiguration;
 import com.ibm.cloud.objectstorage.SDKGlobalConfiguration;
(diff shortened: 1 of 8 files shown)


fhir-bulkimportexport-webapp/src/main/java/com/ibm/fhir/bulkcommon/BulkDataUtils.java
```diff
@@ -288,6 +297,11 @@ public static int readFhirResourceFromLocalFile(String filePath, int numOfLinesT
     }
 
 
+    public static long getLocallFileSize(String filePath) throws Exception {
```
lmsurpre, line 300:
```suggestion
    public static long getLocalFileSize(String filePath) throws Exception {
```

fhir-bulkimportexport-webapp/src/main/java/com/ibm/fhir/bulkimport/ChunkReader.java
```diff
@@ -167,12 +152,46 @@ public Object readItem() throws Exception {
...
+                break;
+            case FILE:
+                importFileSize = BulkDataUtils.getLocallFileSize(importPartitionWorkitem);
+                break;
+            case AWSS3:
+            case IBMCOS:
+                importFileSize = BulkDataUtils.getCosFileSize(cosClient, cosBucketName, importPartitionWorkitem);
+                break;
+            default:
+                throw new Exception ("Doesn't support data source storage type!");
```
lmsurpre, line 187:
```suggestion
                throw new Exception("Doesn't support data source storage type!");
```

lmsurpre: I generally
1. use IllegalStateException in default switch statements that come from enums; and
2. print the name or value of the enum value that caused it in the message

However, neither one very important given that we should never get here unless someone has added a new value to the enum without updating the places its used from.

fhir-bulkimportexport-webapp/src/main/java/com/ibm/fhir/bulkcommon/BulkDataUtils.java
albertwang-ibm: done

fhir-bulkimportexport-webapp/src/main/java/com/ibm/fhir/bulkimport/ChunkReader.java
albertwang-ibm: done

PR #897, opened 2020-04-06
issue #823 - documented extended operations in Conformance.md

lmsurpre:

### DIFF
--- a/docs/src/pages/Conformance.md
+++ b/docs/src/pages/Conformance.md
@@ -35,31 +35,31 @@ Alternatively, for operations with only simple input parameters (i.e. no complex
 #### System operations
 System operations are invoked at `[base]/$[operation]`
 
-|Operation|Definition|Short Description|Notes|
-|---------|----------|-----------------|-----|
-| $convert | https://hl7.org/fhir/R4/resource-operation-convert.html | Takes a resource in one form and returns it in another | Converts between JSON and XML but *not* between FHIR versions |
-| $export | https://hl7.org/fhir/uv/bulkdata/STU1/OperationDefinition-export.html | Export data from the server | exports to an S3-compatible data store; see the (user guide)[https://ibm.github.io/FHIR/guides/FHIRServerUsersGuide#4101-bulk-data-export] for config info |
-| $healthcheck | https://github.com/IBM/FHIR/blob/master/fhir-operation-healthcheck/src/main/resources/healthcheck.json | Check the health of the server | Checks for a valid connection to the database |
+|Operation|Short Description|Notes|
+|---------|-----------------|-----|
+| [$convert](https://hl7.org/fhir/R4/resource-operation-convert.html) | Takes a resource in one form and returns it in another | Converts between JSON and XML but *not* between FHIR versions |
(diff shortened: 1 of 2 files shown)


fhir-config/src/main/java/com/ibm/fhir/config/FHIRRequestContext.java
```diff
@@ -56,28 +56,29 @@ public FHIRRequestContext initialValue() {
...
     };
-    
+
     public FHIRRequestContext() {
         this.requestUniqueId = UUID.randomUUID().toString();
     }
-    
+
     public FHIRRequestContext(String tenantId) throws FHIRException {
+        this();
```
prb112, line 65: only change here?

docs/src/pages/Conformance.md
```diff
@@ -21,6 +21,41 @@ The HL7 FHIR specification is more than just a data format. It defines an [HTTP
...
+Operations can invoked via HTTP POST by passing a Parameters in the body or, for simple operations, via HTTP GET by passing the parameters in the URL.
+
+#### System operations
+System operations are invoked at `[base]/$[operation]`
+
+|Operation|Definition|Short Description|Notes|
+|---------|----------|-----------------|-----|
+| $convert | https://hl7.org/fhir/R4/resource-operation-convert.html | Takes a resource in one form and returns it in another | Converts between JSON and XML but *not* between FHIR versions |
+| $export | https://hl7.org/fhir/uv/bulkdata/STU1/OperationDefinition-export.html | Export data from a FHIR server whether or not it is associated with a patient ||
+| $healthcheck | https://github.com/IBM/FHIR/blob/master/fhir-operation-healthcheck/src/main/resources/healthcheck.json | Check the health of the server | Checks for a valid connection to the database |
```
prb112, line 37: as this is posting to our site, should add (an as of version column?) since the site is agnostic of versions.

fhir-config/src/main/java/com/ibm/fhir/config/FHIRRequestContext.java
lmsurpre: yes

lmsurpre: and totally unrelated to the issue; forgot i had made this minor change and just didn't want to lose it

docs/src/pages/Conformance.md
lmsurpre: I feel like this table is already going to be too wide for the site...

lmsurpre: and I left out `$import` since its not fully baked yet

lmsurpre: I wonder if we need a better approach for versioning the site in general...

prb112: I'd suggest linking the operation rather than putting the link following. 
*versioning is a fair point* we'd have to think that through.

lmsurpre: originally i had a column for whether its spec-defined or custom, but I took that out since it was sort of redundant with the url definition (hl7.org vs github.com).  still, i think you're right about turning those URLs into link targets to save on space...will do that now

lmsurpre: done

PR #834, opened 2020-03-23
WIP: Update Dockerfile to align with openliberty #814

prb112:
Updated `Dockerfile` to agree with changes to the Dockerfile-ol 

```
~$ curl -k -i -u 'fhiruser:change-password' 'https://localhost:9443/fhir-server/api/v4/$healthcheck'
HTTP/1.1 200 OK
Content-Type: application/fhir+json
Date: Mon, 23 Mar 2020 18:22:39 GMT
Content-Language: en-US
Content-Length: 123

{"resourceType":"OperationOutcome","issue":[{"severity":"information","code":"informational","details":{"text":"All OK"}}]}
```

Signed-off-by: Paul Bastide <pbastide@us.ibm.com>

PR #904, opened 2020-04-07
Issue 837 and 649 - double-down on the UBI-based OpenLiberty image

lmsurpre:
Subsumes #834 

1. Updated fhir-install/Dockerfile to use the UBI-based open-liberty image from https://hub.docker.com/r/openliberty/open-liberty

2. Moved fhir-install/docker to build to make it clear this is just for our CI process.
Previously, the CI pipeline used a mix of scripts from `build` and `fhir-install`.
Now there is a single Dockerfile in fhir-install and the variants used by our CI 
pipeline are nicely contained under the build directory.

3. Added dockerfile-maven-plugin to fhir-install; now it builds the
`ibmcom/ibm-fhir-server` docker image as part of the "install" phase and
pushes it to a registry as part of the "deploy" phase (config pending)

4. Removed build/docker/liberty/Dockerfile, moved the config and userlib dirs 
under a child directory named "volumes", and bind mounted those volumes 
from the fhir-server service defined in docker-compose.yml

5. Removed copy-dependencies.sh; now `pre-integration-test.sh` just
copies the dependencies directly to the target ($SIT) directory, so that
only the dockerized integration scripts actually use the build/docker
directory

6. Added docker-push.sh to the release workflow; this uses a username
and password that come from github secrets (currently my username and a
corresponding apikey)


build/.gitignore
```diff
@@ -1 +1,6 @@
-logs/
```
prb112, line 1: logs was added for release.sh - please add back

build/db2-migration-test-docker.sh
```diff
@@ -7,13 +7,13 @@
 set -ex
 
 echo "Preparing environment for fhir-server integration tests..."
-if [[ -z "${WORKSPACE}" ]]; then
-    echo "ERROR: WORKSPACE environment variable not set!"
-    exit 2
-fi
+
+# The full path to the directory of this script, no matter where its called from
```
prb112, line 11: I think what I built for persistence-automation for postgres, could play well here, and we should probably reconcile early as I think you have a `pattern` that I've just generalized.

build/docker/db2/Dockerfile
```diff
@@ -6,8 +6,7 @@
 
 FROM ibmcom/db2
 
-# Create a non-admin user for the FHIR server
-# This non-admin user will be used by fhir server to access db2
+# Create a non-admin user for the FHIR server to use to access db2
```
prb112, line 9: There has to be at least one link / reference to this file in an md

build/.gitignore
lmsurpre: done

build/db2-migration-test-docker.sh
lmsurpre: I find that having this run the same no matter where its invoked from simplifies the script, so I settled on this pattern based on https://stackoverflow.com/a/246128/161022

Previously, these scripts required us to set a WORKSPACE env variable, but this got me very confused when I cloned a second copy of our repo and tried running the scripts and couldn't understand why it wasn't using the right stuff.  Better to have it know its in a build dir under the project root and always execute relative to that.

build/docker/db2/Dockerfile
lmsurpre: Not sure I follow this comment...

build/docker/copy-schema-jar.sh
```diff
@@ -0,0 +1,22 @@
...
+###############################################################################
+# (C) Copyright IBM Corp. 2016, 2020
+#
+# SPDX-License-Identifier: Apache-2.0
+###############################################################################
+set -e
+
+if [[ -z "${WORKSPACE}" ]]; then
+    echo "ERROR: WORKSPACE environment variable not set!"
+    exit 2
```
prb112, line 11: interesting.. this is considered a reserved exit code
http://www.tldp.org/LDP/abs/html/exitcodes.html

just something to think about next time we're in the code

build/docker/db2/Dockerfile
prb112: sorry - I did a quick search in the docs, and we should probably make sure (since we move this file) that there are no documents that need updating

build/docker/copy-schema-jar.sh
lmsurpre: i was wondering why it exits with a code of 2!
lets just change to 1 and be done?

lmsurpre:
```suggestion
    exit 1
```

build/gen-code.sh
```diff
@@ -5,12 +5,16 @@
...
+WORKSPACE="$( dirname "${DIR}" )"
+
+# Install the latest fhir-tools
+mvn clean install -f ${WORKSPACE}/fhir-tools/pom.xml
 
 if [[ $? -eq 0 ]]
-then 
-    mvn com.ibm.fhir:fhir-tools:generate-model -f ../fhir-model/pom.xml
+then
+    mvn com.ibm.fhir:fhir-tools:generate-model -f ${WORKSPACE}/fhir-model/pom.xml
```
prb112, line 19: good fix

build/docker/copy-schema-jar.sh
lmsurpre: done

build/docker/db2/Dockerfile
lmsurpre: good call.  i just found one:
```
fhir-install/README-DB2.md:docker build -t fhirserverdb2 . -f Dockerfile-db2 --squash
```
will fix

lmsurpre: ok, i basically ended up re-writing this readme and added it to the build dir

PR #906, opened 2020-04-08
WIP: Persistence Integration Workflow

prb112:


albertwang-ibm:
For PostgreSql, we have to set max_prepared_transactions in postgresql.conf (usually under /usr/local/var/postgres), the default value is 0 which mean disabled. the suggested value is equal to the maximum connection number.  so, suggest  max_prepared_transactions = 64 for now.

Issue #910, opened 2020-04-08
Expand return preference handling to include validation warnings


lmsurpre:
Relates to #194 and #888

PR #915, opened 2020-04-09
WIP: issue #910 - introduce fhir-context and move the context classes there

lmsurpre:
* FHIRRequestContext (formerly in fhir-config)
* FHIRPagingContext (formerly in fhir-core)
* FHIRHistoryContext (formerly in fhir-persistence)

For now, I kept FHIRSearchContext in fhir-search and FHIRPersistenceContext in fhir-persistence.

As a result of the change, I had to remove the tenantId lookup from
FHIRConfigurationHelper since it no longer has access to the context
classes.  Instead, I just added a string "tenantId" parameter to all the
get methods and do the tenantId lookup before calling the methods.

PROBLEM:  I found that fhir-persistence-proxy depends on FHIRRequestContext and so this approach WONT WORK unless we move this new `fhir-context` module (and thus `fhir-model` as well) out of the webapp and into the shared lib.


fhir-context/src/main/java/com/ibm/fhir/context/FHIRRequestContext.java
```diff
@@ -218,4 +228,51 @@ public void setOriginalRequestUri(String originalRequestUri) {
...
+
+    /**
+     * Associate a list of supplemental issues with the current request
+     *
+     * @implSpec issues will be appended in-order to the resulting OperationOutcome
+     * @implNote these issues are supplemental and therefor it is NOT recommended to use this method
+     *           to communicate issues of SEVERITY greater than warning (i.e. do not be use for issues
+     *           of severity IssueSeverity.ValueSet.ERROR or IssueSeverity.ValueSet.FATAL)
+     */
+    public void addIssue(List<OperationOutcome.Issue> issues) {
```
lmsurpre, line 275:
```suggestion
    public void addIssues(List<OperationOutcome.Issue> issues) {
```

lmsurpre:
Subsumed by #919

Issue #918, opened 2020-04-09
DocumentReference content returned on read is not the same as was saved.


mjdisc:
I did a POST for creating a DocumentReference. The input looks like this:
``` json
{
	"resourceType": "DocumentReference",
	"subject": {
		"reference": "Patient/MJD-1"
	},
	"type": {
		"coding": [{
			"system": "http://loinc.org",
			"version": "2.61",
			"code": "68607-1"
		}]
	},
	"status": "current",
	"docStatus": "final",
	"content": [{
		"attachment": {
			"contentType": "text/plain",
			"data": "ABCDEF"
		}
	}]
}
```

That create ran successfully. I then did a read of that DocumentReference. I got this back:

``` json
{
    "resourceType": "DocumentReference",
    "id": "6a7efa25-40fc-4a39-a529-bcbeba624d05",
    "meta": {
        "versionId": "1",
        "lastUpdated": "2020-04-09T19:13:59.472Z"
    },
    "status": "current",
    "docStatus": "final",
    "type": {
        "coding": [
            {
                "system": "http://loinc.org",
                "version": "2.61",
                "code": "68607-1"
            }
        ]
    },
    "subject": {
        "reference": "Patient/MJD-1"
    },
    "content": [
        {
            "attachment": {
                "contentType": "text/plain",
                "data": "ABCDEA=="
            }
        }
    ]
}
```

Notice that content.attachment.data is not the same as what was originally passed in on the POST. Although ABCDEF is just junk test data that is not Base64 encoded, it is nonetheless valid Base64 data (I think). But not getting back the original input on a read is problematic.

I ran this test on: https://us-south.wh-fhir.dev.cloud.ibm.com/wh-fhir-dev/api/v4

prb112:
@mjdisc I believe the behavior is correct (it's just treating the original content as [Base64Binary](https://www.hl7.org/fhir/datatypes.html#attachment))

```
data |   | 0..1 | base64Binary | Data inline, base64ed
-- | -- | -- | -- | --
```

JohnTimm:
Per the Base64 encoding spec:
https://tools.ietf.org/html/rfc4648#page-5

```
   Special processing is performed if fewer than 24 bits are available
   at the end of the data being encoded.  A full encoding quantum is
   always completed at the end of a quantity.  When fewer than 24 input
   bits are available in an input group, bits with value zero are added
   (on the right) to form an integral number of 6-bit groups.  Padding
[...]

PR #920, opened 2020-04-10
Issue #918 - added validation for incorrect padding

JohnTimm:

### DIFF
--- a/fhir-model/src/test/java/com/ibm/fhir/model/test/Base64BinaryTest.java
+++ b/fhir-model/src/test/java/com/ibm/fhir/model/test/Base64BinaryTest.java
@@ -1,0 +1,33 @@
+/*
+ * (C) Copyright IBM Corp. 2020
+ *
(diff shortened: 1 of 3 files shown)


fhir-model/src/test/java/com/ibm/fhir/model/test/Base64BinaryTest.java
```diff
@@ -0,0 +1,33 @@
...
+    public void testBase64BinaryInvalid() throws Exception {
+        try {
+            Base64Binary.builder().value("ABCDEF").build();
+            fail();
+        } catch (Exception e) {
+            Assert.assertTrue(e instanceof IllegalArgumentException);
+            Assert.assertEquals(e.getMessage(), "Invalid base64 encoded string: incorrect padding");
+        }
+    }
+}
```
lmsurpre, line 33: there's also a way to just tell TestNG that you expect it to fail with an IllegalArgumentException, but this is fine

PR #922, opened 2020-04-10
issue #941 - force the use of TLSv1.2

lmsurpre:
As recommended at https://www.hl7.org/fhir/security.html#http


lmsurpre:
I confirmed it still works if you enabled HTTP (though its still not recommended)

PR #961, opened 2020-04-21
NATS notification support

ccorley:
Adds support for NATS for publishing of patient records.  Supports NATS clusters with and without TLS.  Provides a sample NATS cluster for a provided manual integration test.


ccorley:
Creating a new repo for a cleaner PR.  Closing.

Issue #930, opened 2020-04-14
NPE in Delete when adding an audit log entry


JohnTimm:
[4/14/20, 18:35:24:686 EDT] 000000d3 com.ibm.fhir.server.resources.Delete                         E An error occurred while writing the audit log message.
java.lang.NullPointerException
	at com.ibm.fhir.server.util.RestAuditLogger.populateAuditLogEntry(RestAuditLogger.java:493)
	at com.ibm.fhir.server.util.RestAuditLogger.logDelete(RestAuditLogger.java:178)
	at com.ibm.fhir.server.resources.Delete.delete(Delete.java:90)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at com.ibm.ws.jaxrs20.server.LibertyJaxRsServerFactoryBean.performInvocation(LibertyJaxRsServerFactoryBean.java:656)
	at com.ibm.ws.jaxrs20.server.LibertyJaxRsInvoker.performInvocation(LibertyJaxRsInvoker.java:160)
	at org.apache.cxf.service.invoker.AbstractInvoker.invoke(AbstractInvoker.java:96)
	at com.ibm.ws.jaxrs20.server.LibertyJaxRsInvoker.invoke(LibertyJaxRsInvoker.java:273)
	at org.apache.cxf.jaxrs.JAXRSInvoker.invoke(JAXRSInvoker.java:205)
	at com.ibm.ws.jaxrs20.server.LibertyJaxRsInvoker.invoke(LibertyJaxRsInvoker.java:444)
	at org.apache.cxf.jaxrs.JAXRSInvoker.invoke(JAXRSInvoker.java:111)
	at org.apache.cxf.interceptor.ServiceInvokerInterceptor$1.run(ServiceInvokerInterceptor.java:61)
	at org.apache.cxf.interceptor.ServiceInvokerInterceptor.handleMessage(ServiceInvokerInterceptor.java:99)
	at org.apache.cxf.phase.PhaseInterceptorChain.doIntercept(PhaseInterceptorChain.java:308)
[...]

PR #974, opened 2020-04-22
issue #930 - ensure status is never null from delete AND return 200 OK on success

lmsurpre:
While investigating the issue, I found that `delete` and `conditionalDelete` were handled subtely different.
Now the response status is set in FHIRRestHelper.doDelete() and these two variants share a common buildResponse helper.

Additionally, we were inconsistent in the response status being returned.
Previously we were returning 204 (No Content) for some deletes and 200 (OK) status for others.
Now we will always return a 200 (OK) for a successful delete and the response will always contain an OperationOutcome:
* For conditional deletes with no matches, we return a warning
* For the deletion of a previously deleted resource, we return a warning
* For the deletion of a non-existent resource, we return a warning
* For any other successful deletion (conditional or not), we return an informational message with the number of resources deleted (and their logical ids)

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
@@ -567,75 +567,77 @@ public class FHIRPersistenceJDBCImpl implements FHIRPersistence, FHIRPersistence
         try {
             existingResourceDTO = this.getResourceDao().read(logicalId, resourceType.getSimpleName());
 
-            if (existingResourceDTO != null) {
-                if (existingResourceDTO.isDeleted()) {
-                    existingResource = this.convertResourceDTO(existingResourceDTO, resourceType, null);
-                    resourceBuilder = existingResource.toBuilder();
-
-                    SingleResourceResult<T> result = new SingleResourceResult.Builder<T>()
-                            .success(true)
-                            .resource(existingResource)
-                            .build();
-
-                    return result;
-                }
-                else {
-                    existingResource = this.convertResourceDTO(existingResourceDTO, resourceType, null);
-
-                    // Resources are immutable, so we need a new builder to update it (since R4)
-                    resourceBuilder = existingResource.toBuilder();
-
-                    int newVersionNumber = existingResourceDTO.getVersionId() + 1;
-                    Instant lastUpdated = Instant.now(ZoneOffset.UTC);
-
-                    // Update the soft-delete resource to reflect the new version and lastUpdated values.
-                    Meta meta = existingResource.getMeta();
-                    Meta.Builder metaBuilder = meta == null ? Meta.builder() : meta.toBuilder();
-                    metaBuilder.versionId(Id.of(Integer.toString(newVersionNumber)));
-                    metaBuilder.lastUpdated(lastUpdated);
-                    resourceBuilder.meta(metaBuilder.build());
-
-
-                    @SuppressWarnings("unchecked")
-                    T updatedResource = (T) resourceBuilder.build();
-
-                    // Create a new Resource DTO instance to represent the deleted version.
-                    com.ibm.fhir.persistence.jdbc.dto.Resource resourceDTO = new com.ibm.fhir.persistence.jdbc.dto.Resource();
-                    resourceDTO.setLogicalId(logicalId);
-                    resourceDTO.setVersionId(newVersionNumber);
-
-                    // Serialize and compress the Resource
-                    GZIPOutputStream zipStream = new GZIPOutputStream(stream);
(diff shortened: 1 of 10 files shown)


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -567,75 +567,77 @@ private ResourceDAO getResourceDao() {
...
-                    resourceBuilder = existingResource.toBuilder();
-
-                    SingleResourceResult<T> result = new SingleResourceResult.Builder<T>()
-                            .success(true)
-                            .resource(existingResource)
-                            .build();
+            if (existingResourceDTO == null) {
+                throw new FHIRPersistenceResourceNotFoundException("resource does not exist: " + 
+                        resourceType.getSimpleName() + ":" + logicalId);
+            }
```
lmsurpre, line 573: the changes here are mostly just for style:  by moving the exception cases to the top, i reduced the main code block by two levels of indentation

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -567,75 +567,77 @@ private ResourceDAO getResourceDao() {
...
-                    existingResource = this.convertResourceDTO(existingResourceDTO, resourceType, null);
-                    resourceBuilder = existingResource.toBuilder();
-
-                    SingleResourceResult<T> result = new SingleResourceResult.Builder<T>()
-                            .success(true)
-                            .resource(existingResource)
-                            .build();
+            if (existingResourceDTO == null) {
+                throw new FHIRPersistenceResourceNotFoundException("resource does not exist: " + 
+                        resourceType.getSimpleName() + ":" + logicalId);
```
albertwang-ibm, line 572: 2 ":" seems a little bit strange, can we change to 
resourceType.getSimpleName() + ":" + logicalId" + "does not exist." 

or use the same style as the other warning info of the following codes.

albertwang-ibm:
this should be able to pass touchstone tests, touch stone only verifies: if 200 then should have body, if 204, then no body.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
prb112: Adding to Albert's comments... we should bracket the values in the message.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -893,8 +902,7 @@ private ResourceDAO getResourceDao() {
      * @throws IOException
      */
     protected List<Resource> buildSortedFhirResources(FHIRPersistenceContext context, Class<? extends Resource> resourceType, List<Long> sortedIdList,
-                                                      List<String> elements)
-                            throws FHIRException, FHIRPersistenceException, IOException {
+            List<String> elements) throws FHIRException, FHIRPersistenceException, IOException {
         final String METHOD_NAME = "buildFhirResource";
```
prb112, line 906: wrong method name

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
lmsurpre: My PR didn't actually change the error message from before, but I agree with these comments and will address now.

lmsurpre: I think I'll just change it to `"resource does not exist: " + resourceType.getSimpleName() + "/" + logicalId)`
I usually like to bracket the values in `'` but in this case the values can't contain spaces

lmsurpre, line 572:
```suggestion
                        resourceType.getSimpleName() + "/" + logicalId);
```

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
lmsurpre:
```suggestion
        final String METHOD_NAME = "buildSortedFhirResources";
```

Issue #977, opened 2020-04-23
remove derby from bulkdata.xml


albertwang-ibm:
Because liberty uses derby as default job repo(without any config needed), so don't need to config derby as default javabatch job repo.

PR #978, opened 2020-04-23
issue #977 remove derby config from bulkdata.xml

albertwang-ibm:

### DIFF
--- a/fhir-server/liberty-config/configDropins/defaults/bulkdata.xml
+++ b/fhir-server/liberty-config/configDropins/defaults/bulkdata.xml
@@ -30,7 +30,9 @@
         </application-bnd>
     </webApplication>
 
-    <!-- use IBM Cloud DB2 for job repository 
+    <!-- By default, in-memory derby is used for job repository, no configuration is needed. -->
+    
+    <!-- Use IBM Cloud DB2 for job repository 
     <dataSource id="fhirbatchDS" jndiName="jdbc/fhirbatchDB">
         <jdbcDriver libraryRef="fhirSharedLib"/>
         <properties.db2.jcc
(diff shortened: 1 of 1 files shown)


fhir-server/liberty-config/configDropins/defaults/bulkdata.xml
```diff
@@ -41,15 +44,21 @@
...
             serverName="localhost"
             user="javabatch"
-            password="change-password" /> -->
-        <properties.derby.embedded createDatabase="create" databaseName="derby/fhirDB"/>
+            password="change-password" />
     </dataSource>
+    -->
 
     <batchPersistence jobStoreRef="BatchDatabaseStore"/>
     <databaseStore dataSourceRef="fhirbatchDS" id="BatchDatabaseStore" schema="JBATCH" tablePrefix=""/>
```
lmsurpre, line 64: so it doesn't complain that `fhirbatchDS` doesn't exist by default?

albertwang-ibm: yes, it doesn't complaint on these 2 lines, so I kept them here, and so only need to define the datasource in the overrides

albertwang-ibm: and yes, I verified that derby is used in this case. the job id is always 1 after each restart of the server.

lmsurpre: OK, I havn't quite got around to test bulk export locally, but I'll take your word for it :-)

albertwang-ibm: thanks! :)

PR #998, opened 2020-04-28
Move to @RolesAllowed security, update server config, and introduce jwtRP.xml

lmsurpre:
1. Move webapp security info into `RolesAllowed` annotations on a per-JAXRS-resource basis and make the `metadata` and `.well-known/smart-configuration` endpoints open by default

2. Introduce jwtRP.xml as a new configDropin alternative to oauthResourceServer which
can be used to introspect jwt tokens that are passed to the server for
introspecting scopes and making authorization decisions

3. Re-introduce `singleSignonEnabled="false"` on the
default server.xml and override this setting within the
`oidcProvider.xml` dropin

### DIFF
--- a/fhir-server-webapp/src/main/webapp/WEB-INF/web.xml
+++ b/fhir-server-webapp/src/main/webapp/WEB-INF/web.xml
@@ -18,26 +18,6 @@
         <url-pattern>/*</url-pattern>
     </servlet-mapping>
 
-    <!-- Security definitions for the web app -->
-    <security-role>
-        <role-name>FHIRUsers</role-name>
-    </security-role>
-    <security-constraint>
-        <web-resource-collection>
-            <web-resource-name>FHIR REST API</web-resource-name>
-            <url-pattern>/*</url-pattern>
-            <http-method>GET</http-method>
-            <http-method>POST</http-method>
-            <http-method>PUT</http-method>
-            <http-method>PATCH</http-method>
-            <http-method>DELETE</http-method>
-            <http-method>HEAD</http-method>
-            <http-method>TRACE</http-method>
-        </web-resource-collection>
-        <auth-constraint>
-            <role-name>FHIRUsers</role-name>
-        </auth-constraint>
-    </security-constraint>
     <login-config>
         <auth-method>CLIENT-CERT</auth-method>
     </login-config>
--- a/fhir-server/liberty-config/configDropins/disabled/jwtRP.xml
+++ b/fhir-server/liberty-config/configDropins/disabled/jwtRP.xml
@@ -1,0 +1,32 @@
+<server description="fhir-server">
+
+    <!-- Enable features -->
+    <featureManager>
+        <!-- mpJwt-1.1 is already enabled in the default server.xml, but it doesn't hurt to repeat it here -->
+        <feature>mpJwt-1.1</feature>
+    </featureManager>
+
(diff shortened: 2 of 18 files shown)


fhir-server-webapp/src/main/webapp/WEB-INF/web.xml
```diff
@@ -18,30 +18,6 @@
...
-        <web-resource-collection>
-            <web-resource-name>FHIR REST API</web-resource-name>
-            <url-pattern>/*</url-pattern>
-            <http-method>GET</http-method>
-            <http-method>POST</http-method>
-            <http-method>PUT</http-method>
-            <http-method>PATCH</http-method>
-            <http-method>DELETE</http-method>
-            <http-method>HEAD</http-method>
-            <http-method>TRACE</http-method>
```
prb112, line 35: This is pretty explicit whitelisting.  What happens with websockets? What happens with OPTIONS?  are there any behavior changes?

fhir-server/liberty-config/configDropins/disabled/jwtRP.xml
```diff
@@ -0,0 +1,32 @@
...
+    <mpJwt id="jwtConsumer"
+           keyName="libertyop"
+           userNameAttribute="sub"
+           groupNameAttribute="groupIds"
+           issuer="https://localhost:9443/oauth2/endpoint/oauth2-provider,https://host.docker.internal:9443/oauth2/endpoint/oauth2-provider"
+           authFilterRef="filter"/>
+
+    <authFilter id="filter">
+        <requestUrl urlPattern="/fhir-server" />
+        <requestUrl matchType="notContain" urlPattern="/fhir-server/api/v4/metadata" />
```
prb112, line 29: posted a message about implications for opening up... is this cached?

lmsurpre: we have https://github.com/IBM/FHIR/issues/734 for caching of metadata

it doesn't connect to the database, but it does read the config (both the tenant config and the default config), and programatically builds up the capabilitystatement, which can take some.

lmsurpre:
> LGTM with two questions...
> 
> 1 - does metadata make any api calls?
> 2 - does this behave successfully without tenant specified? Is there a integration test update?

1. no, it just reads the config, constructs the CapabilityStatement, and returns it.  in our cloud deploy, we store the config in a mounted COS bucket, so it *might* result in an api call to COS in that case (but there's already caching for that)
2. if tenant isn't specified, it just reads the default config.  i assume we already have a test for that and I manually invoked it (with and without the jwt config).   that said, the quest is a good one for deployments where we determine the tenant from the jwt token (like our current cloud deployment)...because if you request the metadata endpoint without the token that has the tenant info, you can't get back the tenant-specific config

fhir-server-webapp/src/main/webapp/WEB-INF/web.xml
lmsurpre: Websockets seems to work the same (with the default config at least).  OPTIONS also worked the same (always open), but had some interesting behavior when I added the mpJwt config:  it still returned HTTP 200 and listed the allowed methods, but it also included a `WWW-Authenticate` header with value `Bearer realm="MP-JWT", error="invalid_token"` when I invoked it without a jwt.

Issue #1007, opened 2020-04-30
ResourceTypesCache minor enhancement 


albertwang-ibm:
Currently, when a new resourceType is found, it's not updated to the static ResourceTypesCache directly, instead there is a local Map for the resourceType candidates, and all these candidates are updated to the resourceTypeCache only when the current transaction is committed, this does make sense for the purpose of avoiding the shared ResourceTypesCache to be updated too frequently.

But in bulkdata import case, this mean before the first batch (currently 20 FHIR resources) is committed, we always have to query database to find the resource type id for the to-be-imported resources even if it's already in the local Map. 

The small enhancement is to query ResourceTypesCache first, if no found, then query the local MAP, this will help to avoid at least (batch size -1) query to the database.

PR #1008, opened 2020-04-30
issue #1007 minor enhancement to resourceTypesCache

albertwang-ibm:
The small enhancement is to query ResourceTypesCache first, if no found, then query the local MAP, this will help to avoid at least (batch size -1) query to the database.

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
@@ -460,6 +460,12 @@ public class ResourceDAOImpl extends FHIRDbDAOImpl implements ResourceDAO {
         log.exiting(CLASSNAME, METHODNAME);
 
     }
+    
+    protected  Integer getResourceTypeIdFromCandidatorsCache(String resourceType) {
+        // Get resourceTypeId from local newResourceTypeIds in case this id is already in newResourceTypeIds
+        // but has not been updated to ResourceTypesCache yet.
+        return this.newResourceTypeIds.get(resourceType);
+    }
 
     @Override
     public Resource insert(Resource resource, List<ExtractedParameterValue> parameters, ParameterDAO parameterDao)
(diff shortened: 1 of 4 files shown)


prb112:
how do you warm up the cache?

albertwang-ibm:
> how do you warm up the cache?

I don't really want to change the current cache warmup or loading mechanism, I think it's good enough to fetch the resourceType from DB table when it's used the first time in the same VM, I think this is good for distributing the cache warmup evenly, I just make a little enhancement to allow the codes to get the resource type ID from the 2nd Cache - the local Map, this prevents the codes from having to query DB and get the resource type ID again because the 2nd Cache is not pushed to the main Cache before the transaction is committed.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
```diff
@@ -460,6 +460,19 @@ public void addResourceTypeCacheCandidate(String resourceType, Integer resourceT
         log.exiting(CLASSNAME, METHODNAME);
 
     }
+    
+    protected  Integer getResourceTypeIdFromCaches(String resourceType) {
+        Integer resourceTypeId;
+        // Get resourceTypeId from ResourceTypesCache first.
+        resourceTypeId = ResourceTypesCache.getResourceTypeId(resourceType);
+        // If no found, then get resourceTypeId from local newResourceTypeIds in case this id is already in newResourceTypeIds
```
prb112, line 468:
```suggestion
        // Get resourceTypeId from ResourceTypesCache first.
        Integer resourceTypeId= ResourceTypesCache.getResourceTypeId(resourceType);
        // If not found, then get resourceTypeId from local newResourceTypeIds in case this id is already in newResourceTypeIds
```
honestly, it should never hit newResourceTypeIds with the change in PR1010, but seems reasonable. 
small edits above

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/postgresql/PostgreSqlParameterNamesDAO.java
```diff
@@ -42,18 +42,7 @@ public int readOrAddParameterNameId(String parameterName) throws FHIRPersistence
...
-
-                    if (result == null) {
-                        // would be extremely weird, but good to protect against anyway
-                        throw new IllegalStateException("No parameter id returned after duplicate found!");
-                    }
-                }
-                else {
-                    throw new FHIRPersistenceDataAccessException("parameterName=" + parameterName, e);
-                }
+                throw new FHIRPersistenceDataAccessException("parameterName=" + parameterName, e);
```
lmsurpre, line 45: Maybe a bit more description in the error message?   Like `"Error while writing parameterName '" + parameterName + "'"` ?

Issue #1011, opened 2020-05-01
FHIRPath exists function should return false if the criteria parameter evaluates to false


JohnTimm:
The current implementation of the FHIRPath exists function checks for empty when the criteria parameter is specified. However, since exists(criteria) is shorthand for where(criteria).exists(), if criteria evaluates to false, then the exists(criteria) should also return false.

PR #1012, opened 2020-05-01
Issue #1011 - fixed bug in exists function

JohnTimm:

### DIFF
--- a/fhir-path/src/main/java/com/ibm/fhir/path/evaluator/FHIRPathEvaluator.java
+++ b/fhir-path/src/main/java/com/ibm/fhir/path/evaluator/FHIRPathEvaluator.java
@@ -367,8 +367,10 @@ public class FHIRPathEvaluator {
             if (arguments.size() < 0 || arguments.size() > 1) {
                 throw unexpectedNumberOfArguments(arguments.size(), "exists");
             }
-            Collection<FHIRPathNode> nodes = arguments.isEmpty() ? getCurrentContext() : visit(arguments.get(0));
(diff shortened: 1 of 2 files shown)


fhir-path/src/main/java/com/ibm/fhir/path/evaluator/FHIRPathEvaluator.java
```diff
@@ -367,8 +367,10 @@ private void reset() {
             if (arguments.size() < 0 || arguments.size() > 1) {
                 throw unexpectedNumberOfArguments(arguments.size(), "exists");
             }
-            Collection<FHIRPathNode> nodes = arguments.isEmpty() ? getCurrentContext() : visit(arguments.get(0));
-            return !nodes.isEmpty() ? SINGLETON_TRUE : SINGLETON_FALSE;
+            if (arguments.isEmpty()) {
+                return !getCurrentContext().isEmpty() ? SINGLETON_TRUE : SINGLETON_FALSE;
+            }
+            return isTrue(visit(arguments.get(0))) ? SINGLETON_TRUE : SINGLETON_FALSE;
```
lmsurpre, line 373: So this is consistent with the `where` function?
What is the result if the argument is non-empty and non-boolean?

JohnTimm: it will return false if the argument evaluates to a non-empty / non-boolean

lmsurpre: i confirmed with john that this matches the `where` function, so this looks good to me

Issue #1005, opened 2020-04-30
Package Implementation Guide (IG) jars as userlib jars


JohnTimm:
Currently the IG jars are packaged with the Web application. Packaging them as userlib jars will allow FHIR server users to remove or replace the jars in their respective deployments.

PR #1013, opened 2020-05-01
issue #1005 - move ig jars to userlib dir

lmsurpre:
Instead of packaging them in the fhir-server-webapp war, now we place
them in a userlib dir and the server will discover them from there.

I also performed some path simplification on the install scripts.
We used to create the install files at `/fhir/server/usr/*` but, since
we no longer package the fhir-client or the cli utilities in the
distribution, there's really no need to have these 3 levels.

Finally, I moved the docs (which have a single README which now points
at our website) into the fhir-server directory.

### DIFF
--- a/docs/src/pages/guides/FHIRServerUsersGuide.md
+++ b/docs/src/pages/guides/FHIRServerUsersGuide.md
@@ -1,4 +1,4 @@
-<WLP_HOME>---
+---
 layout: post
 title:  IBM FHIR Server User's Guide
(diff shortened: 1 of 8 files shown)


docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -1,4 +1,4 @@
----
+<WLP_HOME>---
```
lmsurpre, line 1: weird
```suggestion
---
```

Issue #1018, opened 2020-05-04
Improve the Exception Case for Db2 Configured Multitenancy


prb112:
Improve the Exception Case for Db2 Configured Multitenancy

Currently, if the tenantKey is missing, the code is shortcircuited.  The expectation is that the code raises an error in the following cases: 
- tenantKey does not exist and default handling of Db2 multitenancy (on)
- tenantKey does exists and default handling of Db2 multitenancy (off)

Related to #929

PR #1019, opened 2020-05-04
Improve the Exception Case for Db2 Configured Multitenancy #1018

prb112:

### DIFF
--- a/fhir-config/src/main/java/com/ibm/fhir/config/FHIRRequestContext.java
+++ b/fhir-config/src/main/java/com/ibm/fhir/config/FHIRRequestContext.java
@@ -33,6 +33,8 @@ public class FHIRRequestContext {
     private String tenantId;
     private String tenantKey;
     private String dataStoreId;
+    // By default, setting to False.
+    private Boolean dataStoreMultiTenant = Boolean.FALSE;
     private String requestUniqueId;
     private String originalRequestUri;
     private Map<String, List<String>> httpHeaders;
@@ -90,15 +92,15 @@ public class FHIRRequestContext {
 
     /**
      * Setter for the tenant key
-     * @param b64
+     * 
+     * @param base64
      * @throws FHIRException if the given value is not a valid Base64 string
      */
-    public void setTenantKey(String b64) throws FHIRException {
+    public void setTenantKey(String base64) throws FHIRException {
         try {
-            Base64.getDecoder().decode(b64);
-            this.tenantKey = b64;
-        }
-        catch (IllegalArgumentException x) {
+            Base64.getDecoder().decode(base64);
+            this.tenantKey = base64;
+        } catch (IllegalArgumentException x) {
             // Tenant key is a secret, so don't include it in any error message
             throw new FHIRException("Invalid tenantKey.");
         }
@@ -217,5 +219,23 @@ public class FHIRRequestContext {
      */
     public void setHttpHeaders(Map<String, List<String>> httpHeaders) {
         this.httpHeaders = httpHeaders;
+    }
+
+    /**
+     * set the datastore multitenant feature status
+     * 
+     * @param enabled
+     */
+    public void setDataStoreMultiTenant(Boolean enabled) {
+        dataStoreMultiTenant = enabled;
+    }
+
+    /**
+     * get the status of the multitenant feature.
+     * 
+     * @return
+     */
+    public Boolean isDataStoreMultiTenant() {
+        return dataStoreMultiTenant;
     }
 }
--- a/fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
@@ -134,6 +134,15 @@ public class FHIRRestServletFilter extends HttpFilter {
                 if (tenantKey != null) {
                     context.setTenantKey(tenantKey);
                 }
+
+                String type = dsPG.getStringProperty("type", null);
+                if (type != null) {
+                    // Based on the default for the database type, the code.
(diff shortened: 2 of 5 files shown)


fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
```diff
@@ -134,6 +134,15 @@ public void doFilter(HttpServletRequest request, HttpServletResponse response, F
                 if (tenantKey != null) {
                     context.setTenantKey(tenantKey);
                 }
+
+                String type = dsPG.getStringProperty("type", null);
+                if (type != null) {
+                    // Based on the default for the database type, the code.
+                    Boolean enabled =
+                            FHIRConfigHelper.getBooleanProperty("multitenant",
```
lmsurpre, line 142: should this be `String dsPropertyName = FHIRConfiguration.PROPERTY_DATASOURCES + "/" + dsId + "/multitenant";`?

fhir-config/src/main/java/com/ibm/fhir/config/FHIRRequestContext.java
```diff
@@ -33,6 +33,8 @@
     private String tenantId;
     private String tenantKey;
     private String dataStoreId;
+    // By default, setting to False.
+    private Boolean dataStoreMultiTenant = Boolean.FALSE;
```
lmsurpre, line 37: trying to decide if we want this here...I guess is the idea is its similar to the tenantKey so put it in the same spot?

that one made sense to me because it was already here, but I'm not sure if we want to keep adding stuff to the requestcontext.  can we keep it isolated to just the FHIRDbDAOImpl?

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
lmsurpre: where `dsId` is the "datastoreId" which can be looked up from the request context

fhir-config/src/main/java/com/ibm/fhir/config/FHIRRequestContext.java
prb112: I think it makes sense here. it maintains state, and keeps the DAO somewhat stateless

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
prb112: right. this is the pattern from the XA Datasource

fhir-config/src/main/java/com/ibm/fhir/config/FHIRRequestContext.java
lmsurpre: I guess I was thinking we could look up the config property where its needed without storing it as state anywhere.  It didn't seem to be used from anywhere other than FHIRDbDAOImpl.

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
prb112: resolved

fhir-config/src/main/java/com/ibm/fhir/config/FHIRRequestContext.java
prb112: Right, but you end up looking up the property group again, and processing again.  Wouldn't it be good to pick it off one time?

lmsurpre: We do have caching for the config and I think its consistent with other places that read the config whenever we need a config value.
That said, if we're calling it over and over then maybe you're right...do we call getConnection() a lot of times during a typical create/update request?

prb112: While that is true, the config is still navigated down the tree to the property, versus a point lookup. 

In general it's called in multiple spots, but if it's cached in memory (resourceType, parameterName et cetra it shouldn't hit that point), but it is referenced in a lot of places with many different paths.

lmsurpre: OK, I think its probably fine either way.  Would like to hear from @JohnTimm and @albertwang-ibm in terms of preference, but I think I'm good with whatever the group decides.

lmsurpre:
> I would think this could cause confusing, what if a user set multitenant to true for derby or postgresql? what should they expect?

IMHO, they should expect an error because those databases don't support that feature.
Its not so different from bootstrap=true which currently only works for Derby.

fhir-config/src/main/java/com/ibm/fhir/config/FHIRRequestContext.java
JohnTimm: I don’t have a strong opinion either way but if `tenantId` and `dsId` are already in the `FHIRRequestContext` then adding a derived field for convenience doesn’t bother me.

albertwang-ibm: I would agree with Paul's current codes.

lmsurpre: OK, marking it resolved then.  Thanks for weighing in.

prb112:
> I would think this could cause confusing, what if a user set multitenant to true for derby or postgresql? what should they expect?

I explain in some is used to trigger custom logic for multitenant datastore support.  I added to the readme.

Issue #1024, opened 2020-05-05
Duplicate Entry on Execution of Batch with Build


prb112:
Duplicate Entry on Execution of Batch with Build

prb112:
PR https://github.com/IBM/FHIR/pull/1025

PR #1025, opened 2020-05-05
Duplicate Entry on Execution of Batch with Build #1024

prb112:

### DIFF
--- a/fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/derby/DerbyFhirDatabase.java
+++ b/fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/derby/DerbyFhirDatabase.java
@@ -7,6 +7,8 @@
 package com.ibm.fhir.schema.derby;
 
 import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.Properties;
 import java.util.logging.Logger;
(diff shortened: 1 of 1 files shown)


albertwang-ibm:
MULTITENANT_FEATURE_ENABLED is a little bit conflict with what I'm doing to enable storeprocedure for postgresql -- postgresql db is not multi-tenant, but we do need update-procedure support for postgresql in the schema tool...

prb112:
> MULTITENANT_FEATURE_ENABLED is a little bit conflict with what I'm doing to enable storeprocedure for postgresql -- postgresql db is not multi-tenant, but we do need update-procedure support for postgresql in the schema tool...

I get that it is confusing. If we need to update the name in the code we can. 
The point is that it looks up the feature supporting a multitenant datastore.

albertwang-ibm:
> > MULTITENANT_FEATURE_ENABLED is a little bit conflict with what I'm doing to enable storeprocedure for postgresql -- postgresql db is not multi-tenant, but we do need update-procedure support for postgresql in the schema tool...
> 
> I get that it is confusing. If we need to update the name in the code we can.
> The point is that it looks up the feature supporting a multitenant datastore.

yes, my comment is for the schema tool only and I may have to change it. this PR looks great for me.

prb112:
> > > MULTITENANT_FEATURE_ENABLED is a little bit conflict with what I'm doing to enable storeprocedure for postgresql -- postgresql db is not multi-tenant, but we do need update-procedure support for postgresql in the schema tool...
> > 
> > 
> > I get that it is confusing. If we need to update the name in the code we can.
> > The point is that it looks up the feature supporting a multitenant datastore.
> 
> yes, my comment is for the schema tool only and I may have to change it. this PR looks great for me.

If you change it, I'd simply suggest changing the name, and not the functionality.
It's intended to shortcircuit multiple tenants in the same datastore.  If you want to add stored procedures for Postgres, then we add a specific gate for that feature.

PR #1027, opened 2020-05-05
Fix Build Issue with Size of Upload

prb112:

### DIFF
--- a/build/release/diagnostics.sh
+++ b/build/release/diagnostics.sh
@@ -8,7 +8,7 @@ set -eu -o pipefail
 # SPDX-License-Identifier: Apache-2.0
 ###############################################################################
 
-# gather diagnostics 
+# gather diagnostics and relevant files
 
 # Store the current directory to reset to
 pushd $(pwd) > /dev/null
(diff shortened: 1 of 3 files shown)


build/release/diagnostics.sh
```diff
@@ -33,6 +33,6 @@ diagnostic_details > diagnostic_details.log
 
 # To include the distribution, include ' -or -iname fhir-server-distribution.zip'
 # or '-or -iname surefire-reports -or -iname jacoco.exec'
-tar -czf release-test-coverage-and-logs.tgz $(find . -iname diagnostic_details.log -or -iname jacoco-aggregate -or -iname release-commit-details.txt)
+tar -czf release-and-coverage-and-logs.tgz $(find . -iname diagnostic_details.log -or -iname jacoco-aggregate -or -iname release-commit-details.txt -or -iname fhir-cli.zip -or -iname fhir-validation-distribution.zip -or -iname fhir-server-distribution.zip -or -iname fhir-persistence-proxy-distribution.zip)
```
lmsurpre, line 36: might it be better to have these as separate artifacts?

prb112: agreed to do as a separate PR

Issue #1072, opened 2020-05-11
Provide supplemental issues when an instance includes extensions we cannot validate


lmsurpre:
**Is your feature request related to a problem? Please describe.**
I noticed that pdex-plan-net references some of the extensions from the vhdir implementation guide.  Yet, when I validated the instances, it never complained about these extensions, even when the structure was off.  It turns out, the current constraintGenerator skips extension constraints when it cannot find their definition.

**Describe the solution you'd like**
1. Always generate constraints for extensions referenced from a profile, regardless of whether you have access to it.

2. But that constraint is just the cardinality part of it, remove the "conformsTo" part of it.

3. In the validator, always validate extension structures if we can resolve the url.  Else include informational issue.

If its referenced by a profile but we don't have that extension's structure definition: emit a supplemental warning. (And log a warning server-side)

If its NOT referenced in the profile, just a random extension:  still validate it if we have the url, but if we don't then emit an info message (not a warning)

**Describe alternatives you've considered**
Currently, everything happens in the ConstraintGenerator and it works for profiles but doesn't cover extension used outside those profiles.

lmsurpre:
seems to be working

PR #1076, opened 2020-05-12
Issue #1072 - move extension validation (conformsTo) to FHIRValidator

JohnTimm:


JohnTimm:
> I like that unknown extensions are handled similarly to unknown profiles, but I'd prefer it to be Informational vs Warning and I'd prefer them to say "is unknown to the server" instead of "is not supported". Still, this LGTM.

The "not supported" wording matches the issue type (NOT_SUPPORTED) and the wording for both profiles and value sets.

lmsurpre:
> The "not supported" wording matches the issue type (NOT_SUPPORTED) and the wording for both profiles and value sets.

Yep, I don't love it there either.  Because what does "supported" even mean.

JohnTimm:
> Yep, I don't love it there either. Because what does "supported" even mean.
> In the case of the issue type, we're constrainted to the valueset, but for all the messages we control that and can be more precise with our language.

Not supported as in not supported for the intended operation. I don't think the language is any more or less precise than "unknown to the server".

PR #1080, opened 2020-05-13
Add referenced vhdir codesystems, valuesets, and structuredefinitions

lmsurpre:
These resources are reference transitively from the vhdir extensions.

Previously, I included only the extensions themselves. Now, I include
all conformance resources needed to validate those extensions (as well
as the valuesets and codesystems that are provided with binding strength
"example"). The new StructureDefinitions might reference yet more
resources, but this set is sufficient for the purposes of validating
DaVinci PlanNet resources which reference these extensions.

Note:  The `StructureDefinition-vhdir-organization.json` in VHDir Ballot
2 was not valid, so I replaced just that one with the 'latest' from
master, which is still labeled as version `0.2.0`.


JohnTimm:
Can you move this to its own module?

lmsurpre:
> Can you move this to its own module?
I'd love to do this in the near future.  The main reason I packaged the VHDir stuff inside plan-net for now is because there is no suitable version of VHDir to use for FHIR R4...we'd need to package the "continuous build" version from http://build.fhir.org/ig/HL7/VhDir/index.html and I'd rather not advertise support for that.

PR #1081, opened 2020-05-13
Update release.sh to push validation package

prb112:


prb112:
I did a local build, and the classifier wasn't in the zip name, so I opted for the fhir-cli.zip I also checked version 4.1.0 which had it without the classifier.

Issue #1087, opened 2020-05-13
FhirPersistenceDataAccessException on search with multiple _lastUpdated search parameters


mjdisc:
A Condition search with multiple _lastUpdated search parameters fails with a FhirPersistenceDataAccessException .

I'm able to reproduce this on FHIR 4.1.1 and FHIR 4.2.2. 
Here is the 4.2.2 URL I used: 
https://us-south.wh-fhir.dev.cloud.ibm.com/wh-fhir-dev/api/v4/Condition?_lastUpdated=ge2018-09-01T00:00:00Z&_lastUpdated=le2020-05-01T00:00:00Z&_count=100&_tag=http://ibm.com/watson-health-wpd/pdm/data-type%7Cprimary

On a 4.1.1 system being used in PDM test, I have access to the logs and found this:
5/13/20 11:30:54:695 CDT] 00000783 com.ibm.fhir.persistence.jdbc.dao.impl.FHIRDbDAOImpl         E Failure retrieving count.
SQL=SELECT COUNT(R.RESOURCE_ID)  
FROM Condition_LOGICAL_RESOURCES LR JOIN 
	( SELECT * FROM Condition_RESOURCES IR WHERE (LAST_UPDATED <= ?)) 
	  AND ((LAST_UPDATED >= ?)) R ON R.LOGICAL_RESOURCE_ID=LR.LOGICAL_RESOURCE_ID AND R.RESOURCE_ID = LR.CURRENT_RESOURCE_ID AND R.IS_DELETED <> 'Y'  
	  JOIN (SELECT DISTINCT LOGICAL_RESOURCE_ID FROM Condition_TOKEN_VALUES  WHERE (PARAMETER_NAME_ID=162 AND ((TOKEN_VALUE = ? AND CODE_SYSTEM_ID = ?)))) 
	  Condition0 ON Condition0.LOGICAL_RESOURCE_ID = R.LOGICAL_RESOURCE_ID JOIN (SELECT DISTINCT LOGICAL_RESOURCE_ID FROM Condition_TOKEN_VALUES  WHERE 
	  (PARAMETER_NAME_ID=162 AND ((TOKEN_VALUE = ? AND CODE_SYSTEM_ID = ?)))) Condition1 ON Condition1.LOGICAL_RESOURCE_ID = R.LOGICAL_RESOURCE_ID
[...]

PR #1089, opened 2020-05-14
Resolve multiple _lastUpdated search parameters #1087

prb112:


JohnTimm:
Does this change the behavior when we get a parameter like this:

param1=value1,value2,value3

we need to differentiate between these two cases:
 
param1=value1,value2,value3’

and:

param1=value1&param1=value2

lmsurpre:
Can we add a test-case that reproduces the issue and then show its been resolved?

prb112:
Sure

prb112:
> Does this change the behavior when we get a parameter like this:
> 
> param1=value1,value2,value3
> 
> we need to differentiate between these two cases:
> 
> param1=value1,value2,value3 (needs to be OR)
> 
> and:
> 
> param1=value1&param1=value2 (needs to be AND or not allowed)

No, it doesn't change that behavior

prb112:
snuck in a readme per a chat with Steve/John

PR #1108, opened 2020-05-17
Distribution

prb112:


JohnTimm:
Any idea what the DB access exception is about?

prb112:
> Any idea what the DB access exception is about?

Yeah - it means the db is not yet active. We've been seeing it a few times in the last week.

JohnTimm:
> > Any idea what the DB access exception is about?
> 
> Yeah - it means the db is not yet active. We've been seeing it a few times in the last week.

I’ve tried restarting it about 3-4 times with no luck.

Issue #1107, opened 2020-05-15
CARIN BB Organization that doesn’t have both NPI and TAX identifiers fail


prb112:
CARIN BB Organization that doesn’t have both NPI and TAX identifiers fail

1. in`fhir-examples` profiles-fhir-ig-carin-bb-json.txt, edit `VALIDATION json/profiles/fhir-ig-carin-bb/Organization-Org45.json` to `OK json/profiles/fhir-ig-carin-bb/Organization-Org45.json`
1. Run `fhir-ig-carin-bb` `ExamplesValidationTest` 
1. The following errors are reported:

```
-- tokens xml/profiles/fhir-ig-carin-bb/Organization-Org45.xml
{
    "severity": "error",
    "code": "invariant",
    "details": {
        "text": "generated-carin-bb-organization-2: Constraint violation: identifier.exists() implies (identifier.where((type.where(coding.where(code = 'TAX').exists()).exists())).exists())"
    },
    "expression": [
        "Organization"
    ]
}
```

```
-- tokens json/profiles/fhir-ig-carin-bb/Organization-Org45.json
{
    "severity": "error",
    "code": "invariant",
    "details": {
        "text": "generated-carin-bb-organization-2: Constraint violation: identifier.exists() implies (identifier.where((type.where(coding.where(code = 'TAX').exists()).exists())).exists())"
    },
    "expression": [
        "Organization"
    ]
}
```

Expected to pass these tests with OK

PR #1115, opened 2020-05-18
Issue #1107 - add support for Identifier pattern value constraints

JohnTimm:

### DIFF
--- a/fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
+++ b/fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
@@ -31,7 +31,6 @@ import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-import com.ibm.fhir.model.resource.CodeSystem;
 import com.ibm.fhir.model.resource.ValueSet;
 import com.ibm.fhir.model.resource.ValueSet.Expansion;
 import com.ibm.fhir.model.resource.ValueSet.Expansion.Contains;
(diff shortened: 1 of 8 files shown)


fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
```diff
@@ -150,7 +148,7 @@ public int getMaxArity() {
 
     private boolean contains(Map<String, Set<String>> codeSetMap, Coding coding) {
         String system = (coding.getSystem() != null) ? coding.getSystem().getValue() : null;
-        String version = (coding.getVersion() != null) ? coding.getVersion().getValue() : FHIRRegistry.getInstance().getLatestVersion(system, CodeSystem.class);
+        String version = (coding.getVersion() != null) ? coding.getVersion().getValue() : null;
```
lmsurpre, line 151: is this change related?

fhir-profile/src/main/java/com/ibm/fhir/profile/ConstraintGenerator.java
```diff
@@ -304,13 +313,38 @@ private String generatePatternValueConstraint(ElementDefinition elementDefinitio
...
+                    sb.append(" <= 1");
+                } else {
+                    sb.append(" = 1");
+                }
+            } else {
+                sb.append(" >= ").append(min).append(" and ").append(identifier).append(".where(system = '").append(system).append("').count() <= ").append(max);
+            }
+
+            if (!node.children.isEmpty()) {
+                sb.append(" and (").append(identifier).append(".where(system = '").append(system).append("').exists()").append(" implies (").append(identifier).append(".where(system = '").append(system).append("' and ").append(generate(node.children)).append(")))");
```
lmsurpre, line 346: any way to format this to make it a little simpler to see whats going on?

lmsurpre: or maybe just a comment to indicate what this and clause is doing (at a high level)?

JohnTimm: The way the constraint is generated, there isn't much in the way of indentation that will do it justice. Looking at the actual generated constraint is more helpful.

fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
JohnTimm: Only tangentially related. There were warnings associated with a bug in the CARIN BB IG because we were defaulting an unknown version to Latest Version was causing superfluous warnings to be generated. This change makes the code a bit more tolerant.

Issue #1110, opened 2020-05-18
Include Bulk and Oauth tables in the grant action of fhir-persistence-schema


lmsurpre:
**Describe the bug**
While preping for Connectathon we found that bulk export was not working as expected.  It only worked after Paul ran a grant all command on the user id that was being used in the bulk data config.

Similarly, while setting up OAuth using Db2 on Cloud, we found that we needed to use bluadmin in order for the dynamic client registration to work.

In the case of the bulk tables, that might be especially difficult because the Liberty framework creates these tables and not us...needs investigation.

**To Reproduce**
First, deploy the server to Db2 on Cloud, following https://ibm.github.io/FHIR/guides/DB2OnCloudSetup

Issue 1: Bulk Data
1. Configure the bulkdata.xml configDropin to use a non-admin user account
2. Attempt to invoke a bulk import or bulk export job

Issue 2: OAuth
1. Add `oidcProvider.xml` from configDropins/disabled to configDropins/overrides
2. Configure db2 as the OAuthDataSource
3. Attempt to register a client using a userid in the clientAdministrator group

**Expected behavior**
Bulk Data:  the job is successfully submitted
OAuth:  the client is successfully registered

Issue #946, opened 2020-04-20
Add postgresql support for grant command of the schema tool


albertwang-ibm:
Investigate if grant command is needed for postgresql database, if need, then enhance the schema tool to support grant for postgresql database.

prb112:
I've implemented this in Sprint 12

PR #1123, opened 2020-05-19
Improve Bulk and OAuth Support #1110

prb112:
- Add Support for SQL Type SmallInt
- Add Support for multi-column unique constraints
- Improve Foreign Key Constraints for self-referenced foreign keys
(JavaBatch) and update BaseObject to work with FK
- Improve Foreign Key Constraints for Alternative Targets (columns don't
match)
- Add Restrictive Grants for OAuth Tables
- Add Restrictive Grants for Java Batch Tables
- Add Feature Flag for Grants in Schema / Main
- Added SchemaPrinter support for OAuth and JBatch
- Tuned logging for DerbyAdapter (tablespace is now log level fine)
- Tuned logging for ParametersUtil ( changed warning name-code match to
log level fine)
- Tuned DerbyAdapter warnings

Indirectly fixed...
- Add postgresql support for grant command of the schema tool #946


fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/JavaBatchSchemaGenerator.java
```diff
@@ -0,0 +1,450 @@
...
+        Table parameter = Table.builder(schemaName, JOBPARAMETER_TABLE)
+                .addVarcharColumn(NAME, 255, NULL) // NAME VARCHAR(255)
+                .addVarcharColumn(VALUE, 4096, NULL) // VALUE VARCHAR(255) (updated to 4096)
+                .addBigIntColumn(FK_JOBEXECID, NOT_NULL) // FK_JOBEXECID BIGINT
+                .addForeignKeyConstraintAltTarget("JBPRMETERFKJBXECID", schemaName, JOBEXECUTION_TABLE, JOBEXECID, FK_JOBEXECID)
+                .addIndex("JP_FKJOBEXECID_IX", FK_JOBEXECID)
+                .addPrivilege(FhirSchemaConstants.FHIR_BATCH_GRANT_GROUP, Privilege.SELECT)
+                .addPrivilege(FhirSchemaConstants.FHIR_BATCH_GRANT_GROUP, Privilege.INSERT)
+                .addPrivilege(FhirSchemaConstants.FHIR_BATCH_GRANT_GROUP, Privilege.DELETE)
+                .addPrivilege(FhirSchemaConstants.FHIR_BATCH_GRANT_GROUP, Privilege.UPDATE)
```
lmsurpre, line 141: maybe create these up front like we do in FHIRSchemaGenerator.  then, replaced these four lines with:
```
.addPrivileges(List<GroupPrivilege>)
```

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/common/CommonDatabaseAdapter.java
```diff
@@ -318,20 +318,46 @@ public void createForeignKeyConstraint(String constraintName, String schemaName,
         ddl.append(DataDefinitionUtil.join(cols));
         ddl.append(") REFERENCES ");
         ddl.append(targetName);
+        if(targetColumnName != null && !targetColumnName.isEmpty()) {
```
prb112, line 321: Fix spacing

prb112: Objects.isNull

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/derby/DerbyAdapter.java
```diff
@@ -153,12 +149,17 @@ public void dropProcedure(String schemaName, String procedureName) {
...
+
+    @Override
+    public void createTablespace(String tablespaceName, int extentSizeKB) {
+        logger.fine("Create tablespace not supported in Derby");
     }
 
     @Override
     public void dropTablespace(String tablespaceName) {
-        warnOnce(MessageKey.TABLESPACE, "Drop tablespace not supported in Derby");
+        logger.fine("Create tablespace not supported in Derby");
```
prb112, line 162: Drop / Fix

lmsurpre:
Now we need to add this new schema to the VersionHistory stuff as well...

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/DerbyBootstrapper.java
```diff
@@ -111,7 +113,7 @@ public static void bootstrap(Connection connection, String adminSchemaName, Stri
         // Current version history for the database. This is used by applyWithHistory
         // to determine which updates to apply and to record the new changes as they
         // are applied
-        VersionHistoryService vhs = new VersionHistoryService(adminSchemaName, dataSchemaName, OAUTH_SCHEMANAME);
+        VersionHistoryService vhs = new VersionHistoryService(adminSchemaName, dataSchemaName, OAUTH_SCHEMANAME, BATCH_SCHEMANAME);
```
albertwang-ibm, line 116: cool so, we support or are going to support upgrading of OAUTH and BATCH tables.

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirSchemaConstants.java
```diff
@@ -10,10 +10,9 @@
  * Constants related to Schema creation and updating. 
  */
 public class FhirSchemaConstants {
-
     // A lower pool size is selected as default to limit the likelihood of contention on the DBMS.
     // Standard connection/thread pool size
-    public static final int DEFAULT_POOL_SIZE = 1;
+    public static final int DEFAULT_POOL_SIZE = 10;
```
prb112, line 15: 2

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
```diff
@@ -575,6 +581,12 @@ protected void updateSchema() {
             oauthSchemaGenerator.buildOAuthSchema(pdm);
         }
 
+        // Build/update the Liberty JBatch related tables
```
albertwang-ibm, line 584: Do we assume the javabatch tables are in the same database as fhir data but with different schema?

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
```diff
@@ -798,7 +814,7 @@ protected void process() {
      * @param groupName
      */
     protected void grantPrivileges(String groupName) {
-        if (!MULTITENANT_FEATURE_ENABLED.contains(dbType)) {
+        if (!PRIVILEGES_FEATURE_ENABLED.contains(dbType)) {
```
albertwang-ibm, line 817: if enable this command for postgresql, can we still use the Db2Adapter?

fhir-server/liberty-config/configDropins/defaults/bulkdata.xml
```diff
@@ -61,5 +61,5 @@
     -->
 
     <batchPersistence jobStoreRef="BatchDatabaseStore"/>
-    <databaseStore dataSourceRef="fhirbatchDS" id="BatchDatabaseStore" schema="JBATCH" tablePrefix=""/>
+    <databaseStore dataSourceRef="fhirbatchDS" id="BatchDatabaseStore" schema="FHIR_BATCH" tablePrefix="" createTables="false"/>
```
albertwang-ibm, line 64: better to use FHIR_JBATCH, because JBATCH is the default schema of javabatch.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/derby/DerbyAdapter.java
prb112: Fixed

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/common/CommonDatabaseAdapter.java
prb112: Fixed

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/DerbyBootstrapper.java
prb112: Absolutely.

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
prb112: that's right.  Same database, same schema.

If necessary, I can add a separate command to deploy the Batch tables

prb112: Thoughts?

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
prb112: Thank you for seeing this, I'll fix

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
albertwang-ibm: so, drop schema will keep the batch tables but only remove the fhirdata tables ... 
I think because we support different databases in the same fhir server, e.g, in CDT env, we configured performance## tenants using fhir-db2-perf1 database, and the others use databases of the other DB2 instances.   so, maybe it's better to use different job database for javabatch, because it's actually shared/used for export/import from/to all the other databases...  in CDT env, we used the same database as fhirdata, I think maybe this is not the best approach.  In my local, I always configure separated database for javabatch job repo...

albertwang-ibm: I think we'd better to separate the job repo.  but this is not a block for 4.2.2. if necessary, we can just create a issue for future enhancement.

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
prb112: I've fixed this.

fhir-server/liberty-config/configDropins/defaults/bulkdata.xml
prb112: does it make a difference?

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
prb112: I've added --batch-only so you can deploy to a different db.  
It'll still install in the original main db, I don't think this is bad.

At least it can go into a separate db if needed (without the overhead of the rest of the tables)

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirSchemaConstants.java
prb112: Went to 1

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/JavaBatchSchemaGenerator.java
prb112: done

fhir-server/liberty-config/configDropins/defaults/bulkdata.xml
prb112: Converted to JBatch

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
```diff
@@ -291,6 +301,9 @@ protected void parseArgs(String[] args) {
             case "--dry-run":
                 this.dryRun = Boolean.TRUE;
                 break;
+            case "--batch-db-only":
+                this.batchOnly = Boolean.TRUE;
+                break;
```
lmsurpre, line 306: whats the intent of this one?  why not just do it lock-step with the update-schema like we're doing for the OAuth schema?

lmsurpre: if it really needs to be separate, maybe we should split it into three actions?
* `--update-schema`
* `--update-batch-schema`
* `--update-oauth-schema`

Like have those set the `createFhirSchema`, `createOauthSchema`, and `createJavaBatchSchema` flags respectively, then base the update on those flags alone...

lmsurpre: when i did the oauth schema, i was going to do something like that until i decided that the separate actions weren't really needed and so i just did the oauth ones as part of the normal create and update schema actions

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
lmsurpre: haha, just read this one and at least i understand the motivation for the `--batch-only` flag now

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
albertwang-ibm: After discuss with @lmsurpre , we agree that both oauth and batch has the same issue. and we'd better allow to use different databases for batch and oauth which is/are shared by the whole service. so the wanted changes are: admin always created with each database; and admin keep info for schemas in the same database; not matter they are fhirbath, oauth or fhirdata; and fhirbatch, oauth or fhirdata schema can be updated seperately.

prb112: I'm not going to drop it - I'm just going to change to match Lee's request. 
I'd rather do this all at once, it's an easyish change

albertwang-ibm: sound good! thanks!

fhir-config/src/main/java/com/ibm/fhir/config/FHIRRequestContext.java
```diff
@@ -94,6 +95,7 @@ public void setTenantId(String tenantId) throws FHIRException {
      * @param base64
      * @throws FHIRException if the given value is not a valid Base64 string
      */
+    @Deprecated
```
albertwang-ibm, line 98: maybe better to show the reason like it's not needed anymore or not useful anymore because it's handled in fhirdaoimp..

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/api/IConnectionProvider.java
```diff
@@ -57,4 +56,11 @@
      * @param key
      */
     public void describe(String prefix, StringBuilder cfg, String key);
-}
+
+    /**
+     * used to control threading size.
+     */
+    public default int getPoolSize() {
+        return -1;
```
albertwang-ibm, line 64: does -1 mean no limit?

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/postgresql/PostgreSqlAdapter.java
```diff
@@ -302,4 +304,33 @@ public void createIndex(String schemaName, String tableName, String indexName, S
         String ddl = DataDefinitionUtil.createIndex(schemaName, tableName, indexName, indexColumns, false);
         runStatement(ddl);
     }
-}
+
+    @Override
+    public boolean checkCompatibility(String adminSchema) {
+        final String statement = "SELECT 1";
+        boolean result = false;
+        try (Connection c = connectionProvider.getConnection(); PreparedStatement stmt = c.prepareStatement(statement)) {
```
albertwang-ibm, line 312: does this always return true?
the comments for --checkCompatibility "checks feature compatibility" seems a little bit confusing for me...
I didn't notice this before, could you help to tell what's this used for?

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
lmsurpre: Paul added separate actions for
```
    --update-schema-fhir SCHEMANAME
    --update-schema-batch SCHEMANAME
    --update-schema-oauth SCHEMANAME
```
while preserving the existing behavior of `--update-schema`.

I think the updated design manages to walk the fine line of backwards-compatibility while still adding the requested functionality.

fhir-persistence-schema/docs/tests/Db2.md
```diff
@@ -0,0 +1,217 @@
...
+
+# Test 13 - Allocate Tenant - FHIRDATA
+--prop-file fhir-persistence-schema/db2.properties
+--schema-name FHIRDATA
+--allocate-tenant default
+2020-05-20 13:26:22.393 00000001    INFO   com.ibm.fhir.schema.app.Main tenantId [1] is being pre-populated with lookup table data.
+2020-05-20 13:26:22.606 00000001    INFO   com.ibm.fhir.schema.app.Main Finished prepopulating the resource type and search parameter code/name tables tables
+2020-05-20 13:26:22.649 00000001    INFO   com.ibm.fhir.schema.app.Main Allocated tenant: default [key=DOd7xmwhkx2FhJMyPLwaHeU8PAUWcpGj7VjYgMf2hy0=] with Id = 1
+2020-05-20 13:26:22.651 00000001    INFO   com.ibm.fhir.schema.app.Main Processing took: 270.022 s
+2020-05-20 13:26:22.651 00000001    INFO   com.ibm.fhir.schema.app.Main SCHEMA CHANGE: OK
```
albertwang-ibm, line 217: postgresql.md is empty, is it the same as db2.md? if so, maybe should not name as db2.md ...

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/derby/DerbyFhirDatabase.java
```diff
@@ -40,6 +40,7 @@
     private static final String SCHEMA_NAME = "FHIRDATA";
     private static final String ADMIN_SCHEMA_NAME = "FHIR_ADMIN";
     private static final String OAUTH_SCHEMANAME = "FHIR_OAUTH";
+    private static final String BATCH_SCHEMANAME = "FHIR_BATCH";
```
albertwang-ibm, line 43: better to change to FHIR_JBATCH

fhir-config/src/main/java/com/ibm/fhir/config/FHIRRequestContext.java
```diff
@@ -94,6 +95,7 @@ public void setTenantId(String tenantId) throws FHIRException {
      * @param base64
      * @throws FHIRException if the given value is not a valid Base64 string
```
lmsurpre, line 96:
```suggestion
     * @throws FHIRException if the given value is not a valid Base64 string
     * @deprecated tenantKey should be looked up on-the-fly from the fhir server configuration (e.g. from 
     *                        FHIRDbDAOImpl.getConnection() in fhir-persistence-jdbc)
```

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/derby/DerbyFhirDatabase.java
albertwang-ibm: and are those constants common for derby, db2 etc.

fhir-config/src/main/java/com/ibm/fhir/config/FHIRRequestContext.java
lmsurpre: I agree and so I just left a suggestion that I believe would address the comment.  Should we add a similar javadoc on the `getTenantKey()` method as well?

fhir-server/src/main/java/com/ibm/fhir/server/listener/FHIRServletContextListener.java
```diff
@@ -224,6 +224,17 @@ private void bootstrapDerbyDatabases(PropertyGroup fhirConfig) throws Exception
                 log.info("No '" + datasourceJndiName + "' dataSource found; skipping OAuth client table bootstrapping");
             }
 
+            try {
+                // Check the batch database, if the batch database configuration is there, and available.
+                ds = (DataSource) ctxt.lookup("jdbc/fhirbatchDB");
```
albertwang-ibm, line 229: what if we already configured either DB2 or postgresql as job repo? will this still try to bootstrap derby job repo ...

albertwang-ibm: and the same question for the OAUTH bootstraping...

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/ProcedureDef.java
```diff
@@ -72,7 +72,7 @@ protected void grantGroupPrivileges(IDatabaseAdapter target, Set<Privilege> grou
         // Only apply DB Type specific store procedures.
         if (driveClassName.contains(this.getDbType().value())) {
             // Remove the postgresql tag "_pg" from the end of the object name and create the stored procedure.
-            target.grantProcedurePrivileges(getSchemaName(), getObjectName(), group, toUser);
+            target.grantProcedureAndFunctionPrivileges(getSchemaName(), getObjectName(), group, toUser);
```
lmsurpre, line 75:
```suggestion
            target.grantProcedureAndFunctionPrivileges(getSchemaName(), getObjectName().replace("_pg",""), group, toUser);
```

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/ProcedureDef.java
```diff
@@ -72,7 +72,7 @@ protected void grantGroupPrivileges(IDatabaseAdapter target, Set<Privilege> grou
         // Only apply DB Type specific store procedures.
         if (driveClassName.contains(this.getDbType().value())) {
             // Remove the postgresql tag "_pg" from the end of the object name and create the stored procedure.
```
lmsurpre, line 74:
```suggestion
            // Remove the postgresql tag "_pg" from the end of the object name and execute the grants for the stored procedure.
```

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/api/IConnectionProvider.java
prb112: it means there is now pool and to signal the downstream code to change.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/postgresql/PostgreSqlAdapter.java
prb112: it's used to check the connectivity to the database. 
returns an exception if there is an issue.

fhir-persistence-schema/docs/tests/Db2.md
prb112: intentional (I hadn't yet finished the tests)
it'll be uploaded, thanks for noticing.

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/derby/DerbyFhirDatabase.java
prb112: thanks for seeing this... I am trying to use JBATCH (good find)

prb112: Fixed in next PR.

fhir-config/src/main/java/com/ibm/fhir/config/FHIRRequestContext.java
prb112: I'm going to leave this here to pick up shortly

fhir-server/src/main/java/com/ibm/fhir/server/listener/FHIRServletContextListener.java
prb112: only if you have enabled boostrap as true.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/ProcedureDef.java
prb112: I'll change this... thanks

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/postgresql/PostgreSqlAdapter.java
albertwang-ibm: ha, interesting,  the function name seems a little bit confusing, I though it's compatible with something ...

fhir-server/src/main/java/com/ibm/fhir/server/listener/FHIRServletContextListener.java
albertwang-ibm: but our default setting is true, maybe better to check if the configured datasource is derby before bootstrap it ...

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirSchemaConstants.java
```diff
@@ -29,6 +28,7 @@
 
     // Group of privilege grants used for FHIRUSER access
     public static final String FHIR_USER_GRANT_GROUP = "fhiruser";
+    public static final String FHIR_BATCH_GRANT_GROUP = "fhirbatch";
```
lmsurpre, line 31: is it ever used anywhere?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/DerbyBootstrapper.java
```diff
@@ -172,7 +174,7 @@ public static void bootstrapOauthDb(DataSource ds) throws Exception {
                 // to determine which updates to apply and to record the new changes as they
                 // are applied
                 VersionHistoryService vhs =
-                        new VersionHistoryService(ADMIN_SCHEMANAME, OAUTH_SCHEMANAME, OAUTH_SCHEMANAME);
+                        new VersionHistoryService(ADMIN_SCHEMANAME, OAUTH_SCHEMANAME);
```
lmsurpre, line 177: good catch.  i wonder how that second OAUTH_SCHEMANAME has slipped in there...

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirSchemaConstants.java
prb112: no but it should be! great find

prb112: Fixed, it's now used

fhir-server/src/main/java/com/ibm/fhir/server/listener/FHIRServletContextListener.java
prb112: I follow you. I am just testing my locallized fix (which is in boostrapBatchDb(ds)

fhir-config/src/main/java/com/ibm/fhir/config/FHIRRequestContext.java
prb112: Picked up the suggestion and applied... I'm going to apply to getTenantKey as well

prb112: Applied in next commit

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/ProcedureDef.java
prb112: Fixed in code

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/ProcedureDef.java
prb112: Changed around and cleaned up

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/DerbyBootstrapper.java
```diff
@@ -197,31 +199,48 @@ public static void bootstrapOauthDb(DataSource ds) throws Exception {
...
-                CreateVersionHistory.createTableIfNeeded(ADMIN_SCHEMANAME, adapter);
-
-                // Current version history for the database. This is used by applyWithHistory
-                // to determine which updates to apply and to record the new changes as they
-                // are applied
-                VersionHistoryService vhs =
-                        new VersionHistoryService(ADMIN_SCHEMANAME, BATCH_SCHEMANAME);
-                vhs.setTarget(adapter);
-                vhs.init();
+            try (Statement stmt = c.createStatement();
```
albertwang-ibm, line 205: I thought you may want to check the datasource configuration in server.xml, but this also works. thanks!

prb112: yes, that was one approach, I considered

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirSchemaConstants.java
lmsurpre: It makes me wonder if we should have a FHIR_OAUTH_GRANT_GROUP as well, but I think this PR already has enough in it.
We'll need to remember to update the diagrams with this new role (e.g. in the fhir-persistence-schema/docs and update the guides to indicate that the batch user must be in the fhirbatch role for this feature to work).
Personally, I might have saved that for a separate PR, but I'm OK with including it here.

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
```diff
@@ -219,7 +219,7 @@ protected void buildCommonModel(PhysicalDataModel pdm, boolean fhirSchema, boole
             gen.buildSchema(pdm);
             switch (dbType) {
             case DB2:
-                gen.buildDatabaseSpecificArtifactsPostgres(pdm);
+                gen.buildDatabaseSpecificArtifactsDb2(pdm);
```
albertwang-ibm, line 222: I thought to limit the DB differences in the datautil/model layer, so I had added dbtype for the procedures, and then check dbtype before apply the objects.

albertwang-ibm: I totally agree with this change, just wanted to share my thoughts. :)

prb112: it worked, I changed it and thought I could get away with doing what you did.  It didn't work so I had to split it out at this layer.

PR #1145, opened 2020-05-22
Update DeleteTest to be more self-contained

lmsurpre:
Previously, this test was over-zealously deleting Observations it didn't
create.
Now it should be much more self-contained and more consistent: it always
tests both multi-delete scenarios (result <= max & result > max).


fhir-server-test/src/test/java/com/ibm/fhir/server/test/DeleteTest.java
```diff
@@ -398,20 +379,14 @@ public void testSearch3() throws Exception {
         assertEquals(5, searchResults.getEntry().size());
     }
 
-    @Test
-    public void testConditionalDeleteResource() throws Exception {
-        if (!deleteSupported) {
-            return;
-        }
-
+    public void testConditionalDeleteNoResource() throws Exception {
```
prb112, line 383: @Test annotation is gone?

Issue #1169, opened 2020-05-29
fhir-persistence-schema GRANT must consider "GRANT_GROUP"


lmsurpre:
**Describe the bug**
When attempting to use the 4.2.2 fhir-persistence-schema `--grant-to` action, it doesn't seem to apply the grants to the newly created FHIR_JBATCH tables.

**To Reproduce**
1. deploy the schema
2. run the grant-to

**Expected behavior**
It runs grant statements so that the user/role that was passed can now work with these tables

PR #1171, opened 2020-05-29
fhir-persistence-schema GRANT must consider resource #1169

prb112:

### DIFF
--- a/fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
+++ b/fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
@@ -351,7 +351,7 @@ public class Main {
         vhs.init();
 
         // Use the version history service to determine if this table existed before we run `applyWithHistory`
-        boolean newDb = vhs.getVersion(schemaName, DatabaseObjectType.TABLE.name(), "PARAMETER_NAMES") == null || 
+        boolean newDb = vhs.getVersion(schemaName, DatabaseObjectType.TABLE.name(), "PARAMETER_NAMES") == null ||
                 vhs.getVersion(schemaName, DatabaseObjectType.TABLE.name(), "PARAMETER_NAMES") == 0;
 
         applyModel(pdm, adapter, collector, vhs);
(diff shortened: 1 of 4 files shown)


fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
```diff
@@ -497,7 +495,9 @@ protected void grantPrivileges(String groupName) {
         final IDatabaseAdapter adapter = getDbAdapter(dbType, connectionPool);
         try (ITransaction tx = TransactionFactory.openTransaction(connectionPool)) {
             try {
-                pdm.applyGrants(adapter, groupName, grantTo);
+                pdm.applyGrants(adapter, FhirSchemaConstants.FHIR_USER_GRANT_GROUP, grantTo);
+                pdm.applyGrants(adapter, FhirSchemaConstants.FHIR_OAUTH_GRANT_GROUP, grantTo);
+                pdm.applyGrants(adapter, FhirSchemaConstants.FHIR_BATCH_GRANT_GROUP, grantTo);
```
albertwang-ibm, line 500: so, all the db objects will be processed 3 times - once for each group?  even thought maybe only some of them belong to certain group ...

lmsurpre: well, if I understand it right, `buildCommonModel` will only add the "target" objects to the pdm, so its not necessarilly "all" the db objects.

prb112: it's not necessarily all, but yes, it'll potentially process three times, applying only where the group matches

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirSchemaGenerator.java
```diff
@@ -83,13 +83,6 @@
     private static final String ADD_RESOURCE_TYPE = "ADD_RESOURCE_TYPE";
     private static final String ADD_ANY_RESOURCE = "ADD_ANY_RESOURCE";
 
-    // Tags used to control how we manage privilege grants
```
prb112, line 86: @punktilious below

lmsurpre:
Robin helped me confirm that the GRANT_GROUP is never actually persisted, so I think that resolves my one concern

PR #1204, opened 2020-06-08
Fix for Site Build Issue

prb112:
Fix for Site Build Issue


JohnTimm:
@prb112 is there a way we can get this site to show up on a mobile browser as well?

PR #1207, opened 2020-06-09
Issue 1136 - refactor project layout

lmsurpre:
1. merged fhir-operation into fhir-server and moved fhir-operation projects into common sub-dir

2. moved ig projects into common sub-dir

3. moved notification projects into common sub-dir


prb112:
Does this need to be change https://github.com/IBM/FHIR/blob/master/.github/workflows/build.yml#L55 ? 

This needs to change *just change the for loop conditions*
https://github.com/IBM/FHIR/blob/master/build/release/release.sh

prb112:
Also the error wasn't jumping out at me (why the build failed)

lmsurpre:
> Does this need to be change https://github.com/IBM/FHIR/blob/master/.github/workflows/build.yml#L55 ?

I think not.  I didn't do the persistence change (yet).

> This needs to change just change the for loop conditions
https://github.com/IBM/FHIR/blob/master/build/release/release.sh

I assume you mean around this line: `for PROJ in `find . -type d -maxdepth 1 | grep -v '.git' | grep -v 'build' | grep -v '/docs' | sed 's|.\/||g' | grep -v '\.' ` ?
Think I could just change maxdepth to 2?

prb112:
only comment is more of a question.. I saw the pom.xml change for fhir-validation with the fhir-core exclusion? is this new?

Issue #1230, opened 2020-06-15
NPE exception when attempting to validate against a profile that does not contain a snapshot


michaelwschroeder:
When attempting to validate against a profile that does not contain a snapshot, we get a fatal error from FHIR due to a null pointer exception in the FHIR code. Would be good to get a meaningful error message. 

stacktrace:
```
java.lang.NullPointerException
	at com.ibm.fhir.profile.ProfileSupport.getKeys(ProfileSupport.java:241)
	at com.ibm.fhir.profile.ProfileSupport.computeConstraints(ProfileSupport.java:66)
	at com.ibm.fhir.profile.ProfileSupport.lambda$getConstraints$1(ProfileSupport.java:216)
	at com.ibm.fhir.profile.ProfileSupport$$Lambda$147/0000000000000000.apply(Unknown Source)
	at java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1671)
	at com.ibm.fhir.profile.ProfileSupport.getConstraints(ProfileSupport.java:216)
	at com.ibm.fhir.profile.ProfileSupport.getConstraints(ProfileSupport.java:185)
	at com.ibm.fhir.validation.FHIRValidator$ValidatingNodeVisitor.validate(FHIRValidator.java:300)
	at com.ibm.fhir.validation.FHIRValidator$ValidatingNodeVisitor.doVisit(FHIRValidator.java:259)
	at com.ibm.fhir.path.visitor.FHIRPathDefaultNodeVisitor.visit(FHIRPathDefaultNodeVisitor.java:81)
	at com.ibm.fhir.path.FHIRPathResourceNode.accept(FHIRPathResourceNode.java:176)
	at com.ibm.fhir.validation.FHIRValidator$ValidatingNodeVisitor.validate(FHIRValidator.java:244)
	at com.ibm.fhir.validation.FHIRValidator$ValidatingNodeVisitor.access$100(FHIRValidator.java:230)
	at com.ibm.fhir.validation.FHIRValidator.validate(FHIRValidator.java:172)
	... 66 more 
```

prb112:
PR https://github.com/IBM/FHIR/pull/1235

prb112:
Fix is in 4.3.0

PR #1235, opened 2020-06-16
NPE exception when attempting to validate against a profile that does

prb112:
not contain a snapshot #1230

- Added a Snapshot/Elements check to Profile Support


JohnTimm:
What about simply putting: `requireNonNull(structureDefinition.getSnapshot(), "StructureDefinition.snapshot element is required")` as the first line in the requisite methods? I think there are maybe 3 methods impacted. That would be a 3 line change and it would be consistent with other places that we guard against null using `Objects.requireNonNull`

prb112:
Updated per review from John

Issue #1237, opened 2020-06-17
tenant-specific Resource and DomainResource search params only apply to specific types


lmsurpre:
**Describe the bug**
Under certain circumstances, tenant-specific extension search parameters defined for base "Resource" or "DomainResource" are not properly applied to all resource types on the server.

Update:  it turns out that the `typeToParamMap` being passed to `ParametersUtil.assignInheritedToAll(typeToParamMap)` for extension-search-parameters will only include resource types with extension-search-parameters defined.
That means that the following loop will only add these Resource/DomainResource search parameters to resource types that are explicitly listed as the "base" for at least one SearchParameter in the same extension-search-parameters bundle:
```
for (Entry<String, ParametersMap> entry : searchParameterMap.entrySet()) {
```

**To Reproduce**
1. Define a Search parameter (e.g. 'favorite-color') with a base of `["Resource"]`
2. Attempt to search a concrete type like Patient by this search parameter 

Note the response includes error text like the following:
> Search parameter 'favorite-color' for resource type 'Patient' was not found.

**Expected behavior**
A search parameter with a base of Resource should apply to all resource types.
A search parameter with a base of DomainResource should apply to all domain resource types.

**Additional context**
On the create path, we call `SearchUtil.getSearchParameters(String resourceType)`.

On the query path, we call `SearchUtil.getApplicableSearchParameters(String resourceType)`.

From what I can tell, these two methods do basically the same thing.  They both:
1. load the built-in parameters from `ParametersUtil.getBuiltInSearchParametersMap(String resourceType)`
2. filter them via `SearchUtil.filterSearchParameters()`
3. load the tenant-specific parameters via `getUserDefinedSearchParameters(String resourceType)` (which eventually calls `ParametersUtil.buildSearchParametersMapFromBundle`); and
4. add the tenant parameters to the applicable built-in ones.

This complication made the root cause of this issue harder to find; I think we should remove one of these two as part of the fix.

**Workaround**
list all resource types in the base field instead of Resource:
```
"base": ["Account",
      "ActivityDefinition",
      "AdverseEvent",
      "AllergyIntolerance",
      "Appointment",
      "AppointmentResponse",
      "AuditEvent",
      "Basic",
      "Binary",
      "BiologicallyDerivedProduct",
      "BodyStructure",
      "Bundle",
      "CapabilityStatement",
      "CarePlan",
      "CareTeam",
      "CatalogEntry",
      "ChargeItem",
      "ChargeItemDefinition",
[...]

PR #1240, opened 2020-06-17
issue #1237 - stop adding duplicate SearchParameters with base Resource

lmsurpre:
I found that we were adding the Resource-level SearchParameters in two
different places and this was causing some inconsistent behavior.
In this option, I stop adding the Resource search-parameters at the end
of getFilteredBuiltinSearchParameters.

This effectively reverts us to the searchfilter behavior we had in our
DSTU2 implementation; if a filter is present for a given type, then the
common parameters must be explicitly listed as allowed search parameters
as well.

Also: removes redundant getSearchParameterByCodeIfPresent in favor of
`getApplicableSearchParametersMap(resourceType).get(code)`

### DIFF
--- a/fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
+++ b/fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
@@ -195,12 +195,6 @@ public class SearchUtil {
             result.addAll(filterSearchParameters(filterRules, resourceType, spMap.values()));
         }
 
-        // Retrieve the SPs associated with the "Resource" resource type and filter per the filter rules.
-        spMap = spBuiltin.get(SearchConstants.RESOURCE_RESOURCE);
-        if (spMap != null && !spMap.isEmpty()) {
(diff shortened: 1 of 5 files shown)


fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -923,7 +884,12 @@ protected static boolean isAllowed(Type type, Modifier modifier) {
         Map<String, SearchParameter> result = new HashMap<>();
         List<SearchParameter> list = getApplicableSearchParameters(resourceType);
         for (SearchParameter sp : list) {
-            result.put(sp.getCode().getValue(), sp);
+            String code = sp.getCode().getValue();
+            if (result.containsKey(code)) {
+                log.info("SearchParametersMap for resource '" + resourceType + "' "
```
prb112, line 889: Info or Warn?

lmsurpre:
closed in favor of #1239

PR #1249, opened 2020-06-19
removed problematic example from indices

lmsurpre:
This example is valid to the core spec but invalid to the US Core IG and
we're not equipped to handle examples like this just yet.


punktilious:
LGTM

Issue #1257, opened 2020-06-20
touschstone test error - search parameter name is not unique


albertwang-ibm:
https://touchstone.aegis.net/touchstone/scriptexecution?exec=202006192132136000589822&qn=/FHIR4-0-1-Advanced/Capabilities/fhir-api-r4-capabilities-01-metadata#WAITING_OPER

(1) errors:
Validation of response body against profile 'http://hl7.org/fhir/StructureDefinition/CapabilityStatement' by FHIR specification's Validation Engine produced the following results:

    ERROR: Search parameter names must be unique in the context of a resource. [searchParam.select(name).isDistinct()]. Location: CapabilityStatement.rest[0].resource[41] (line 5905, col 18).

    ERROR: Search parameter names must be unique in the context of a resource. [searchParam.select(name).isDistinct()]. Location: CapabilityStatement.rest[0].resource[44] (line 6355, col 18).

    ERROR: Search parameter names must be unique in the context of a resource. [searchParam.select(name).isDistinct()]. Location: CapabilityStatement.rest[0].resource[103] (line 14709, col 18).

    ERROR: Search parameter names must be unique in the context of a resource. [searchParam.select(name).isDistinct()]. Location: CapabilityStatement.rest[0].resource[120] (line 17560, col 18).

(2) warning
missing etag

albertwang-ibm:
DocumentReference, Encounter, Patient and Resource all have 2 "_id" in searchParam.

PR #1258, opened 2020-06-20
Issue #1257 fix duplicated search parameters of Capstatement

albertwang-ibm:
Changes:
(1) Use Set to hold the search parameters instead of list.
(2) Avoid search parameters of "Resource" itself to be added twice.

### DIFF
--- a/fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
+++ b/fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
@@ -196,11 +196,9 @@ public class SearchUtil {
         }
 
         // Retrieve the SPs associated with the "Resource" resource type and filter per the filter rules.
-        if (resourceType.contentEquals(SearchConstants.RESOURCE_RESOURCE)) {
-            spMap = spBuiltin.get(SearchConstants.RESOURCE_RESOURCE);
-            if (spMap != null && !spMap.isEmpty()) {
-                resultSet.addAll(filterSearchParameters(filterRules, SearchConstants.RESOURCE_RESOURCE, spMap.values()));
-            }
+        spMap = spBuiltin.get(SearchConstants.RESOURCE_RESOURCE);
+        if (spMap != null && !spMap.isEmpty()) {
+            resultSet.addAll(filterSearchParameters(filterRules, SearchConstants.RESOURCE_RESOURCE, spMap.values()));
         }
 
         List<SearchParameter> result = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -192,15 +192,19 @@ public static void init() {
...
         if (spMap != null && !spMap.isEmpty()) {
-            result.addAll(filterSearchParameters(filterRules, resourceType, spMap.values()));
+            resultSet.addAll(filterSearchParameters(filterRules, resourceType, spMap.values()));
         }
 
         // Retrieve the SPs associated with the "Resource" resource type and filter per the filter rules.
-        spMap = spBuiltin.get(SearchConstants.RESOURCE_RESOURCE);
-        if (spMap != null && !spMap.isEmpty()) {
-            result.addAll(filterSearchParameters(filterRules, SearchConstants.RESOURCE_RESOURCE, spMap.values()));
+        if (resourceType.contentEquals(SearchConstants.RESOURCE_RESOURCE)) {
```
lmsurpre, line 199: This line breaks the contract of the method.
```Returns a filtered list of built-in SearchParameters associated with the
     * specified resource type and those
     * associated with the "Resource" resource type.```

File: fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java [hunk 2]
```diff
@@ -192,15 +192,19 @@ public static void init() {
...
+            spMap = spBuiltin.get(SearchConstants.RESOURCE_RESOURCE);
+            if (spMap != null && !spMap.isEmpty()) {
+                resultSet.addAll(filterSearchParameters(filterRules, SearchConstants.RESOURCE_RESOURCE, spMap.values()));
+            }
         }
```
lmsurpre, line 204: I'm gonna merge this change to hopefully avoid loading up my workspace :-)
```suggestion
        spMap = spBuiltin.get(SearchConstants.RESOURCE_RESOURCE);
        if (spMap != null && !spMap.isEmpty()) {
            resultSet.addAll(filterSearchParameters(filterRules, SearchConstants.RESOURCE_RESOURCE, spMap.values()));
        }
```

lmsurpre:
Even if we fix the build, this approach of using a Set instead of a List won't fix the issue because the real issue is that US Core (and other IGs) introduce search parameters that really do conflict with search parameters defined in the base spec.  That is, they have the same code for the same resource, but are otherwise different (e.g. different id, different name, different expression, etc).

Issue #1319, opened 2020-07-07
Failing Conditional Update within Transaction: 'Bundle.Entry.resource must contain an id field for a PUT operation.'


chgl:
**Describe the bug**
When sending a transaction containing a conditional update operation where `resource.id` is not specified, the operation fails with `Bundle.Entry.resource must contain an id field for a PUT operation.` despite this not being a requirement for conditional updates.

**To Reproduce**
1. Go to 'https://www.hl7.org/fhir/bundle-transaction.json.html'
2. Copy the transaction JSON
3. Send it to the FHIR Server via curl

The (simplified) failing entry in the sample transaction is:

```json
{
  "resourceType": "Bundle",
  "id": "bundle-transaction",
  "type": "transaction",
  "entry": [
    {
      "fullUrl": "urn:uuid:74891afc-ed52-42a2-bcd7-f13d9b60f096",
      "resource": {
        "resourceType": "Patient",
        "identifier": [
          {
            "system": "http:/example.org/fhir/ids",
            "value": "456456"
          }
        ],
        "birthDate": "1974-12-25"
      },
      "request": {
        "method": "PUT",
        "url": "Patient?identifier=http:/example.org/fhir/ids|456456"
      }
    }
  ]
}
```

**Expected behavior**
Instead of returning an error, the server should accept the transaction, check if a Patient resource with the given `identifier` (search expression) is found. If yes, it should update the contents of the resource with data provided in the transaction. If no resource with the given `identifier` is found, a new resource should be generated with the provided contents. Ie following the specification here: https://www.hl7.org/fhir/http.html#cond-update

> The resource MAY contain an id element, but does not need to (this is one of the few cases where a resource exists without an id element). If an id is provided, the server SHALL ignore it - see documentation for the update interaction.

**Additional context**

If a resource id is specified but changed between two `POST`s, while the search expression stays the same, it fails with `Input resource 'id' attribute must match the id of the search result resource.`

https://www.hl7.org/fhir/http.html#cond-update says

> Unlike this rest of this page, the conditional create, update, patch and delete are trial use until further experience is gained with their use. Their status will be reviewed in a future version of FHIR.

So I understand if this is something you are currently not tackling. A workaround for some cases might be to use update-as-create.

prb112:
Fix is in 4.3.1

chgl:
@prb112 awesome! Thanks for quick fix, @JohnTimm

PR #1320, opened 2020-07-07
Issue #1319 - conditional update where new resource has no id

JohnTimm:

### DIFF
--- a/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
@@ -338,7 +338,7 @@ public class FHIRRestHelper implements FHIRResourceHelpers {
 
                     // If the id of the input resource is different from the id of the search result,
                     // then throw exception.
-                    if (newResource.getId() != null && id != null && !newResource.getId().equalsIgnoreCase(id)) {
+                    if (newResource.getId() != null && id != null && !newResource.getId().equals(id)) {
                         String msg = "Input resource 'id' attribute must match the id of the search result resource.";
                         throw buildRestException(msg, IssueType.VALUE);
                     }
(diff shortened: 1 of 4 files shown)


fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -338,11 +338,11 @@ private FHIRRestOperationResponse doPatchOrUpdate(String type, String id, FHIRPa
 
                     // If the id of the input resource is different from the id of the search result,
                     // then throw exception.
-                    if (newResource.getId() != null && newResource.getId() != null
-                            && !newResource.getId().equalsIgnoreCase(id)) {
+                    if (newResource.getId() != null && id != null && !newResource.getId().equalsIgnoreCase(id)) {
```
punktilious, line 341: do we consistently ignore case when working with id values?

JohnTimm: In this particular case, I left it the way it was because I assume it was put there for good reason. We can certainly verify that equalsIgnoreCase is the right behavior there.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -1324,11 +1324,13 @@ private void methodValidation(HTTPVerb method, Resource resource) throws FHIRPer
                         "Bundle.Entry.resource is required for BundleEntry with PUT method.";
                 throw buildRestException(msg, IssueType.INVALID);
             }
+            /*
```
punktilious, line 1327: Did you intend to leave this commented out? Should probably remove it.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
JohnTimm: From here:
http://hl7.org/fhir/http.html#general
```
Note: All URLs (and ids that form part of the URL) defined by this specification are case sensitive.
```

JohnTimm: I will update the PR.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
JohnTimm: Removed commented out code.

PR #1324, opened 2020-07-08
Issue 1122 remove tenantKey from request context and support tenantKey without requiring fhir-server-config files

punktilious:
This change permits the R4 examples JDBC Main to run against DB2 databases without requiring a fhir-server-config.json file to be defined with the tenantKey (which in this case is provided as a command-line argument). This change follows on from a previous change which moved the tenantKey out of the FHIRRequestContext.

This change also now removes the tenantKey property from the FHIRRequestContext, which had previously been marked as deprecated.

### DIFF
--- a/fhir-persistence-jdbc/src/test/java/com/ibm/fhir/persistence/jdbc/test/spec/R4JDBCExamplesProcessor.java
+++ b/fhir-persistence-jdbc/src/test/java/com/ibm/fhir/persistence/jdbc/test/spec/R4JDBCExamplesProcessor.java
@@ -6,13 +6,23 @@
 
 package com.ibm.fhir.persistence.jdbc.test.spec;
 
+import java.io.ByteArrayInputStream;
+import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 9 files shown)


fhir-persistence-jdbc/src/test/java/com/ibm/fhir/persistence/jdbc/test/spec/R4JDBCExamplesProcessor.java
```diff
@@ -12,6 +12,7 @@
 import java.util.Properties;
 import java.util.function.Supplier;
 
+import com.ibm.fhir.config.FHIRConfigProvider;
 import com.ibm.fhir.config.FHIRRequestContext;
```
prb112, line 16: Please fix the copyright

punktilious: fixed

PR #1338, opened 2020-07-17
Duplicate Condition in Resolve

prb112:


fhir-path/src/main/java/com/ibm/fhir/path/function/ResolveFunction.java
```diff
@@ -122,7 +122,6 @@ private String resolveInternalFragmentReference(FHIRPathTree tree, String refere
                     DomainResource domainResource = (DomainResource) resource;
                     for (Resource contained : domainResource.getContained()) {
                         if (contained.getId() != null &&
-                                contained.getId() != null &&
```
punktilious, line 125: perhaps it was supposed to be contained != null &&  contained.getId() != null

JohnTimm: It is impossible for `contained` to be null in this context. Likely this was supposed to be `contained.getId() != null && id != null` but it is also impossible for `id` to be null in this context.

PR #1342, opened 2020-07-21
Update to new version of jcip

prb112:


prb112:
Waiting on feedback from team

PR #1359, opened 2020-07-24
Clarify what "API-stable" means

lmsurpre:


JohnTimm:
Maybe we should name the column "Java API stability"

PR #1364, opened 2020-07-27
issue 1362 Use timestamp prefix for identity strings to reduce random reads during ingestion

punktilious:
Closes 1362.

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
@@ -113,9 +113,11 @@ import com.ibm.fhir.persistence.jdbc.exception.FHIRPersistenceFKVException;
 import com.ibm.fhir.persistence.jdbc.util.CodeSystemsCache;
 import com.ibm.fhir.persistence.jdbc.util.JDBCParameterBuildingVisitor;
 import com.ibm.fhir.persistence.jdbc.util.JDBCQueryBuilder;
+import com.ibm.fhir.persistence.jdbc.util.LogicalIdentityProvider;
 import com.ibm.fhir.persistence.jdbc.util.ParameterNamesCache;
 import com.ibm.fhir.persistence.jdbc.util.ResourceTypesCache;
 import com.ibm.fhir.persistence.jdbc.util.SqlQueryData;
+import com.ibm.fhir.persistence.jdbc.util.TimestampPrefixedUUID;
 import com.ibm.fhir.persistence.util.FHIRPersistenceUtil;
 import com.ibm.fhir.search.SearchConstants;
 import com.ibm.fhir.search.SummaryValueSet;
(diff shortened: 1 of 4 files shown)


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/TimestampPrefixedUUID.java
```diff
@@ -0,0 +1,42 @@
...
+        // hex is sufficient, although a custom encoding using the full character set
+        // supported by FHIR identifiers would be a little more compact (== smaller indexes).
+        // Do not use Base64.
+        String prefix = Long.toHexString(millis);
+        
+        UUID uuid = UUID.randomUUID();
+        
+        StringBuilder result = new StringBuilder();
+        result.append(prefix);
+        result.append("-"); // redundant, but more visually appealing.
```
prb112, line 38: Does this impact any regular expressions? or potentially any fhir path? 

Does it impact FHIR Uuid type? 
`private static final Pattern PATTERN = Pattern.compile("urn:uuid:[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}");`.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -278,7 +283,7 @@ protected Action buildActionChain() {
             // system-generated value. For the update-or-create scenario, see update().
             // Default version is 1 for a brand new FHIR Resource.
             int newVersionNumber = 1;
-            logicalId = UUID.randomUUID().toString();
+            logicalId = logicalIdentityProvider.createNewIdentityValue();
             if (log.isLoggable(Level.FINE)) {
                 log.fine("Creating new FHIR Resource of type '" + resource.getClass().getSimpleName() + "'");
```
prb112, line 288: For ResourceDAOImpl, Also what about line 518? `stmt.setString(6, UUID.randomUUID().toString());`

prb112: Also DerbyResourceDao on 117?

punktilious: Yikes...no idea why we'd be manipulating data values in the DAO...good catch...will check

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/TimestampPrefixedUUID.java
lmsurpre: Resource.id is a system string with a hinted type of https://www.hl7.org/fhir/datatypes.html#id (and not UUID), so as long as we only use it for that, then I think we're OK on this one.

punktilious: It shouldn't, because identifiers are constrained by a different pattern according to the spec.

prb112: Works for me

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
punktilious: That's the sourceKey which was previously used to ensure idempotence in replication scenarios. It is no longer used and can be removed. Will create a new issue. It is benign here.

Issue #1380, opened 2020-07-30
extension-search-parameter.json using wildcard RESOURCE do not get extracted


prb112:
**Describe the bug**
extension-search-parameter.json using wildcard RESOURCE do not get extracted

**To Reproduce**
1. Create an extension-search-parameter with a base of RESOURCE
2. This is the wildcard for all RESOURCES.  
3.  It does not extract the value of any resource

**Expected behavior**
A value is extracted

**Additional context**
Customer reported

PR #1381, opened 2020-07-30
extension-search-parameter.json using wildcard RESOURCE do not get extracted #1380

prb112:


punktilious:
Is there a unit-test or SIT to catch this?

prb112:
> Is there a unit-test or SIT to catch this?

No perfectly valid question.  Give me a few to whip a unit test up

PR #1394, opened 2020-07-31
Fix javadoc and parameter name and add warning to FHIRUtil.getRandomKey

lmsurpre:

### DIFF
--- a/fhir-model/src/main/java/com/ibm/fhir/model/util/FHIRUtil.java
+++ b/fhir-model/src/main/java/com/ibm/fhir/model/util/FHIRUtil.java
@@ -79,6 +79,7 @@ public class FHIRUtil {
     public static final com.ibm.fhir.model.type.String STRING_DATA_ABSENT_REASON_UNKNOWN = com.ibm.fhir.model.type.String.builder()
             .extension(DATA_ABSENT_REASON_UNKNOWN)
             .build();
+    @Deprecated
     private static final SecureRandom RANDOM = new SecureRandom();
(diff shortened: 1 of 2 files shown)


fhir-model/src/main/java/com/ibm/fhir/model/util/FHIRUtil.java
```diff
@@ -570,17 +570,19 @@ public static boolean isFailure(IssueSeverity severity) {
...
         KeyGenerator keyGen;
         try {
-            keyGen = KeyGenerator.getInstance(key);
+            keyGen = KeyGenerator.getInstance(algorithm);
             keyGen.init(256);
             return Base64.getEncoder().encodeToString(keyGen.generateKey().getEncoded());
         } catch (NoSuchAlgorithmException e) {
+            log.warning("Algorithm '" + algorithm + "' is not supported; using SecureRandom instead");
             byte[] buffer = new byte[32];
             RANDOM.setSeed(System.currentTimeMillis());
```
punktilious, line 587: I know it's not related to this change, but why are we seeding here? Potential OWASP because currentTimeMillis isn't sufficiently random.

lmsurpre: I was wondering the same thing, which is why i added the warning...i'm not even sure it should fallback to SecureRandom

prb112: I don't think this should be in FHIRUtil -- it should go into bulkdata, and be hidden away.

Issue #693, opened 2020-02-11
Validate IETF BCP47 language codes


lmsurpre:
The valueset is defined at https://www.hl7.org/fhir/valueset-languages.html

This is the "max valueset" for anywhere referencing human languages in the base FHIR spec:
```
Max ValueSet: Resource.language (code / Preferred)
Max ValueSet: RelatedPerson.communication.language (CodeableConcept / Preferred)
Max ValueSet: Practitioner.communication (CodeableConcept / Preferred)
Max ValueSet: ExplanationOfBenefit.processNote.language (CodeableConcept / Preferred)
Max ValueSet: ValueSet.compose.include.concept.designation.language (code / Preferred)
Max ValueSet: CodeSystem.concept.designation.language (code / Preferred)
Max ValueSet: ClaimResponse.processNote.language (CodeableConcept / Preferred)
Max ValueSet: HealthcareService.communication (CodeableConcept / Preferred)
Max ValueSet: Patient.communication.language (CodeableConcept / Preferred)
Max ValueSet: Attachment.language (code / Preferred)
```

So, per #601, we should verify that these values are valid when we validate the corresponding resources.

lmsurpre:
The actually required valueset is https://www.hl7.org/fhir/valueset-all-languages.html

JohnTimm:
Per the description here: http://hl7.org/fhir/valueset-all-languages.html

The value set contains an infinite number of members. I think what we are looking to do is validate the syntax. Basic syntax validation can be accomplished with vanilla Java as follows:
```
public void validateLanguage(String language) {
    new Locale.Builder().setLanguageTag(language);
}
```
This method will throw a runtime exception (`IllformedLocaleException`) if the syntax is not valid.

lmsurpre:
The PRs cover core-spec "max-bindings" and also required or max-binding valuesets bound via a profile.  The only remaining work is to add generic CodeSystem-level validate-code and that function should be handled as part of #1217

lmsurpre:
I tried validating the following resource type as was surprised when it passed without issue:
```
{
  "resourceType": "Patient",
  "id": "{{patientid}}",
  "language": "xxxxxxxx",
  "meta": {
    "tag": [{
    	"code": "test"
    }]
  },
  "active": true
}
```

I guess we are only checking bcp47 well-formedness and not doing any checks on the IANA registry.
In this case, I would expect *at least* a warning that the value I have provided is not in the preferred valueset for the bound element.

[...]

PR #1414, opened 2020-08-14
issue #693 - memberOf handles maxValueSet and all-languages value set

tbieste:

This change adds validation of the maxValueSet defined for a binding, and validation of language codes against BCP-47 when all-languages is the value set.

### DIFF
--- a/fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
+++ b/fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
@@ -147,7 +147,7 @@ public class MemberOfFunction extends FHIRPathAbstractFunction {
      * @return true or false
      */
     private boolean isSyntaxBased(ValueSet valueSet) {
-        String valueSetUrl = valueSet.getUrl() != null ? valueSet.getUrl().getValue().split("\\|")[0] : null;
+        String valueSetUrl = valueSet.getUrl() != null ? valueSet.getUrl().getValue() : null;
         return ALL_LANG_VALUE_SET_URL.equals(valueSetUrl);
     }
     
@@ -196,7 +196,7 @@ public class MemberOfFunction extends FHIRPathAbstractFunction {
      * @return true if validation was successful, otherwise false
      */
     private boolean validateCodeAgainstSyntaxBasedValuedSet(ValueSet valueSet, Code code, Coding coding, CodeableConcept codeableConcept, EvaluationContext evaluationContext, FHIRPathElementNode elementNode, String strength) {
-        String valueSetUrl = valueSet.getUrl() != null ? valueSet.getUrl().getValue().split("\\|")[0] : null;
+        String valueSetUrl = valueSet.getUrl() != null ? valueSet.getUrl().getValue() : null;
 
         try {
             // Validate against all-languages value set
--- a/fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
+++ b/fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
@@ -195,7 +195,7 @@ public class MemberOfFunction extends FHIRPathAbstractFunction {
      * Issues will be added to evaluation context if validation is not successful.
      * @param valueSet the value set
      * @param code the code to validate
-     * @param code the coding to validate
+     * @param coding the coding to validate
      * @param codeableConcept the codeable concept to validate
      * @param evaluationContext the evaluation context
      * @param elementNode the element node to validate
--- a/fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
+++ b/fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
@@ -33,6 +33,7 @@ import com.ibm.fhir.model.type.Quantity;
 import com.ibm.fhir.model.type.Uri;
 import com.ibm.fhir.model.type.code.IssueSeverity;
 import com.ibm.fhir.model.type.code.IssueType;
+import com.ibm.fhir.model.util.ValidationSupport;
 import com.ibm.fhir.path.FHIRPathElementNode;
 import com.ibm.fhir.path.FHIRPathNode;
 import com.ibm.fhir.path.FHIRPathTree;
@@ -48,6 +49,8 @@ import com.ibm.fhir.term.spi.ValidationOutcome;
  * is used to determine whether or not to add a warning to the evaluation context if the membership check fails.
  */
 public class MemberOfFunction extends FHIRPathAbstractFunction {
+    public static final String ALL_LANG_VALUE_SET_URL = "http://hl7.org/fhir/ValueSet/all-languages";
+
     @Override
     public String getName() {
         return "memberOf";
@@ -90,6 +93,8 @@ public class MemberOfFunction extends FHIRPathAbstractFunction {
         if (valueSet != null) {
             FHIRTermService service = FHIRTermService.getInstance();
             if (isExpanded(valueSet) || service.isExpandable(valueSet)) {
+
+                // Validate against expanded value set
                 if (element.is(Code.class)) {
                     Uri system = getSystem(evaluationContext.getTree(), elementNode);
                     Code code = element.as(Code.class);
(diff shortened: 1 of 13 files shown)


fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
```diff
@@ -119,16 +124,108 @@ public int getMaxArity() {
...
         return SINGLETON_TRUE;
     }
 
+    /**
+     * Determines if the value set is syntax-based.
+     * @param valueSet the value set
+     * @return true or false
+     */
+    private boolean isSyntaxBased(ValueSet valueSet) {
+        String valueSetUrl = valueSet.getUrl() != null ? valueSet.getUrl().getValue().split("\\|")[0] : null;
```
lmsurpre, line 153: what is the split for?

fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
```diff
@@ -119,16 +124,108 @@ public int getMaxArity() {
...
+        }
+        else if (element.is(CodeableConcept.class)) {
+            codeableConcept = element.as(CodeableConcept.class);
+        }
+        else if (element.is(Quantity.class)) {
+            Quantity quantity = element.as(Quantity.class);
+            coding = Coding.builder().system(quantity.getSystem()).code(quantity.getCode()).build();
+        }
+        else {
+            code = element.is(FHIR_STRING) ? Code.of(element.as(FHIR_STRING).getValue()) : Code.of(element.as(Uri.class).getValue());
```
lmsurpre, line 186: seems a bit odd not to handle the `element.is(Uri.class)` in the if/else block with the others, but I suppose its fine

fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
```diff
@@ -119,16 +124,108 @@ public int getMaxArity() {
...
+     * @param valueSet the value set
+     * @param code the code to validate
+     * @param code the coding to validate
+     * @param codeableConcept the codeable concept to validate
+     * @param evaluationContext the evaluation context
+     * @param elementNode the element node to validate
+     * @param strength the binding strength
+     * @return true if validation was successful, otherwise false
+     */
+    private boolean validateCodeAgainstSyntaxBasedValuedSet(ValueSet valueSet, Code code, Coding coding, CodeableConcept codeableConcept, EvaluationContext evaluationContext, FHIRPathElementNode elementNode, String strength) {
```
lmsurpre, line 205: maybe just combine it with the method above and take an Element?
then you just need one if/else block, right?  maybe a "horse apiece" though.

fhir-validation/src/test/resources/fhir/validation/test/package/.index.json
```diff
@@ -64,6 +64,60 @@
...
             "type": "FamilyMemberHistory"
         },
+        {
+            "filename": "test-language-primary-extension.json",
+            "resourceType": "StructureDefinition",
+            "id": "test-language-primary-extension",
+            "url": "http://ibm.com/fhir/StructureDefinition/test-language-primary-extension",
+            "version": "0.1.0",
+            "kind": "complex-type",
+            "type": "Extension"
```
lmsurpre, line 74: I didn't know this was supported in our registry; interesting

fhir-validation/src/test/java/com/ibm/fhir/validation/test/DeviceProfileTest.java
```diff
@@ -0,0 +1,235 @@
...
+import com.ibm.fhir.profile.ConstraintGenerator;
+import com.ibm.fhir.registry.FHIRRegistry;
+import com.ibm.fhir.validation.FHIRValidator;
+import com.ibm.fhir.validation.util.FHIRValidationUtil;
+
+public class DeviceProfileTest {
+    private static final String ENGLISH_US = "en-US";
+
+    @Test
+    public void testDeviceProfile() throws Exception {
```
lmsurpre, line 43: I wonder if one of these should include the word MaxValueSet or ConstraintGenerator to make it clear what the test is for

fhir-validation/src/test/java/com/ibm/fhir/validation/test/DeviceProfileTest.java
```diff
@@ -0,0 +1,235 @@
...
+        //-----[Device.safety]
+        // Choice: Yes; Optional: No; Repeatable: No
+        //----[test-language-primary-extension]
+        // Choice: Yes; Optional: No; Repeatable: Yes
+        //----[test-language-others-req-extension]
+        // Choice: Yes; Optional: Yes; Repeatable: No
+        //----[test-language-secondary-extension]
+        // Choice: Yes; Optional: Yes; Repeatable: Yes
+        //----[test-language-others-opt-extension]
+        //-----------------------------------------------
```
lmsurpre, line 67: this comment was very helpful for me, thanks

fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
tbieste: I kept the if/else same logic used in the apply(...) method for expanded value sets (around line 121) to be consistent.  I can change both places if there's a way that would be better.

fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
tbieste: Yeah, I went back and forth on that.  My goal was to minimize the number of lines of code to be added for any future syntax-based value sets (e.g. UCUM), so I thought this ended up looking cleaner.

fhir-validation/src/test/java/com/ibm/fhir/validation/test/DeviceProfileTest.java
tbieste: Good call.  I'll rename these to be more descriptive.

fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
```diff
@@ -119,16 +124,108 @@ public int getMaxArity() {
...
+        return validateCodeAgainstSyntaxBasedValuedSet(valueSet, code, coding, codeableConcept, evaluationContext, elementNode, strength);
+    }
+    
+    /**
+     * Validates the Code, Coding, CodableConcept against the syntax based value set.
+     * Only one of Code, Coding, or CodeableConept should be passed in.
+     * Issues will be added to evaluation context if validation is not successful.
+     * @param valueSet the value set
+     * @param code the code to validate
+     * @param code the coding to validate
```
michaelwschroeder, line 198: change param name to 'coding'

fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
tbieste: In case the profile specified a canonical with a version, I used the same split as done in CodeGenerator to just compare the URL part.

fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
tbieste: Fixed.

JohnTimm:
@tbieste Please download and install the code template and formatting rules for Eclipse here:
https://github.com/IBM/FHIR/wiki/Style-Guide

We are (mostly) consistent throughout our code base with:
```
if (condition) {
    doSomething();
} else {
    doSomethingElse();
}
```
and
```
try {
    doSomething();
} catch (Exception e) {
    // ...
}
```

tbieste:
@JohnTimm Thanks!  I imported the code template and formatting rules for Eclipse.

lmsurpre:
I was about to press merge on this one, but then I started thinking...

If a resource instance
1. has a Coding with a known "syntax-based" codeSystem, like all-langs or ucum; and
2. the code isn't valid per the syntax of those systems
then I'm thinking that should fail validation irregardless of the valueset binding or even the binding strength.

If a Coding claims to have system http://hl7.org/fhir/ValueSet/all-languages and isn't valid per that coding system, then this should be a validation error, full stop.  Why do we even need to involve the constraint generation?

lmsurpre:
Thought about it some more and realized that:
1. memberOf could get used outside of FHIRValidation and so we probably still want it to behave "properly" for these syntax-based valuesets; and
2. there's no need to do the CodeSystem validate-code if we've already validated its in a bound valueset (assuming the valueset was created properly)..so my comment above about taking a "shortcut" to avoid double-validation could go either way; and
3. we're not validating codes via CodeSystem validate-code at all just yet, and it seems beyond the scope of https://github.com/IBM/FHIR/issues/693

net: I'm still good with merging this and revisiting it if/when we validate CodeSystem codes

JohnTimm:
I agree with @lmsurpre that we need to visit the idea of validating arbitrary coded elements against a explicit code system (perhaps directly in the validator) but we really only need to do this for coded elements that are not bound to a checked value set and not already checked during object construction (if possible).

Issue #694, opened 2020-02-11
Validate UCUM unit strings


lmsurpre:
The valueset is defined at https://www.hl7.org/fhir/valueset-ucum-units.html

This has a required binding in these three resources:
```
Resource: ResearchElementDefinition.characteristic.unitOfMeasure (CodeableConcept / Required)
Resource: EffectEvidenceSynthesis.effectEstimate.unitOfMeasure (CodeableConcept / Required)
Resource: RiskEvidenceSynthesis.riskEstimate.unitOfMeasure (CodeableConcept / Required)
```

Per #601, we should validate that those values are valid when we validate those resource types.

Consider using https://github.com/FHIR/Ucum-java (or https://unitsofmeasurement.github.io)

lmsurpre:
note:  https://github.com/FHIR/Ucum-java  seems to be limited to just the common set of UCUM codes.
For a more complete implementation we could use "Indria" (JSR something), but its probably not needed.

lmsurpre:
note: we have some UCUM length unit support for our near search.  After we have improved UCUM support we should consider consolidating that code with whatever we introduce.

JohnTimm:
Using Indriya (the RI for JSR-385), basic format syntax validation can be achieved with something like the following:
```
SystemOfUnitsService service = ServiceProvider.current();
UnitFormat format = service.getFormatService().getUnitFormat();
Unit<?> unit = format.parse(unit)
```
`UnitFormat.parse` will throw a runtime exception (`MeasurementParseException`) if the syntax is not valid.

michaelwschroeder:
Validation of codes for the ucum-units value set is being done by the UCUM parser added under PR #1420. In addition to that value set, the http://hl7.org/fhir/ValueSet/all-distance-units and http://hl7.org/fhir/ValueSet/all-time-units value sets will also have to be validated syntactically. They are referenced as maxValueSets in the Age, Distance, and Duration types. That work will be done against this issue in a subsequent PR.

lmsurpre:
I submitted a RiskEvidenceSynthesis resource with a bogus ucum code, but the server said it was valid:
```{
  "resourceType": "RiskEvidenceSynthesis",
  "id": "example",
  "text": {
    "status": "generated",
    "div": "<div xmlns=\"http://www.w3.org/1999/xhtml\">[Put rendering here]</div>"
  },
  "status": "draft",
  "population": {
    "reference": "EvidenceVariable/example"
  },
  "outcome": {
[...]

PR #1420, opened 2020-08-18
issue #694 - Added ANTLR4 grammar for UCUM code syntax validation

tbieste:

This change adds ANTLR4 grammar and a utility that can be used for UCUM code syntax validation.


fhir-model/src/main/java/com/ibm/fhir/model/ucum/util/UCUMUtil.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    /**
+     * Determines if the string is a valid UCUM string.
+     * 
+     * @param ucumString
+     *            the string
+     * @return true or false
+     */
+    public static boolean isValidUcum(String ucumString) {
```
michaelwschroeder, line 44: Not familiar with antlr lexer/parser, so don't know if it's a big deal or not - is there not a way to avoid instantiating a new lexer/parser every time you want to parse a string?

tbieste: I followed the same flow as FHIRPathUtil.java, which instantiated a new lexer/parser each time, and that appears to be how it needs to work.

Issue #639, opened 2020-01-28
SqlIntegrityConstraintViolationException while using multiple data sources together


lmsurpre:
**Describe the bug**
While working on the db2 integration tests, I decided to used db2 for the `default` tenant, but keep `tenant1` using derby so that we don't need to deploy the schema 4 times.

Unfortunately, this caused an issue in our integration tests.  When the MultiDataStoreTest invokes a reequest with tenant=tenant1 and datastore=default, the config helper falls back to the default tenant config and so we end up using the db2 connection from default/default and that leads to:

```
[1/28/20 0:35:07:491 UTC] 0000268f FHIRRestServl I   Received request: tenantId:[tenant1] dsId:[default] user:[fhiruser] method:[PUT] uri:[https://localhost:9443/fhir-server/api/v4/Observation/9bab5ef9-888b-4a85-bcef-304a4d258c9d
[1/28/20 0:35:07:491 UTC] 0000268f FHIRProvider  > com.ibm.fhir.provider.FHIRProvider readFrom ENTRY
[1/28/20 0:35:07:492 UTC] 0000268f FHIRProvider  < com.ibm.fhir.provider.FHIRProvider readFrom RETURN
[1/28/20 0:35:07:492 UTC] 0000268f FHIRResource  > com.ibm.fhir.server.resources.FHIRResource update(String,String,Resource) ENTRY
[1/28/20 0:35:07:492 UTC] 0000268f FHIRResource  > com.ibm.fhir.server.resources.FHIRResource doPatchOrUpdate ENTRY
[1/28/20 0:35:07:492 UTC] 0000268f FHIRResource  1   Retrieved FHIRPersistenceHelper instance from servlet context: com.ibm.fhir.persistence.helper.FHIRPersistenceHelper@85377624
[1/28/20 0:35:07:492 UTC] 0000268f FHIRPersisten > com.ibm.fhir.persistence.helper.FHIRPersistenceHelper getFHIRPersistenceImplementation ENTRY
[1/28/20 0:35:07:492 UTC] 0000268f FHIRPersisten 1   Retrieved persistence factory property 'fhirServer/persistence/factoryClassname': com.ibm.fhir.persistence.jdbc.FHIRPersistenceJDBCFactory
[1/28/20 0:35:07:492 UTC] 0000268f FHIRPersisten 1   Using FHIR persistence factory class name: com.ibm.fhir.persistence.jdbc.FHIRPersistenceJDBCFactory
[...]

PR #1428, opened 2020-08-19
issue #639 datasources must come from tenant config, not fall back to…

punktilious:
… default

### DIFF
--- a/fhir-config/src/main/java/com/ibm/fhir/config/FHIRConfigHelper.java
+++ b/fhir-config/src/main/java/com/ibm/fhir/config/FHIRConfigHelper.java
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright IBM Corp. 2017,2019
+ * (C) Copyright IBM Corp. 2017,2020
  *
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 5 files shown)


fhir-config/src/main/java/com/ibm/fhir/config/FHIRConfigHelper.java
```diff
@@ -74,9 +74,16 @@ private static JsonValue getPropertyFromTenantOrDefault(String propertyName) {
         // let's try to find it in the default config.
         if (result == null && !FHIRConfiguration.DEFAULT_TENANT_ID.equals(tenantId)) {
             try {
-                pg = FHIRConfiguration.getInstance().loadConfiguration();
-                if (pg != null) {
-                    result = pg.getJsonValue(propertyName);
+                if (propertyName.startsWith(FHIRConfiguration.PROPERTY_DATASOURCES)) {
+                    // Issue #639. Prevent datasource lookups from falling back to 
+                    // the default datasource which breaks tenant isolation.
+                    result = null;
```
prb112, line 80: I don't think this is necessary.  line 75 checks... `result==null`
perhaps just make this an if without an else branch?

punktilious: Correct, this was just a style thing so that I could add that comment in a clear way rather than having to deal with negatives.

prb112: Good - it was intentional.  LGTM

Issue #1452, opened 2020-08-27
conformsTo generates validation errors even when used inside an "or"


lmsurpre:
**Describe the bug**
for one of the PlanNet extensions we generate the following constraint:
```
value.as(Reference).exists() and 
value.as(Reference).resolve().conformsTo('http://hl7.org/fhir/us/davinci-pdex-plan-net/StructureDefinition/plannet-PractitionerRole') or 
value.as(Reference).resolve().conformsTo('http://hl7.org/fhir/us/davinci-pdex-plan-net/StructureDefinition/plannet-OrganizationAffiliation') or 
value.as(Reference).resolve().conformsTo('http://hl7.org/fhir/us/davinci-pdex-plan-net/StructureDefinition/plannet-Location') or 
value.as(Reference).resolve().conformsTo('http://hl7.org/fhir/us/davinci-pdex-plan-net/StructureDefinition/plannet-Organization')
```

For a reference of type Organization, we'll execute the conformsTo statements in order and we'll hit this on the first 3:
```
generateIssue(evaluationContext, IssueSeverity.ERROR, IssueType.INVALID, "Conformance check failed: profile (or base definition) '" + url + "' is not applicable to type: " + ModelSupport.getTypeName(modelClass), node.path());
```

**To Reproduce**
Run ExamplesValidationTest from https://github.com/IBM/FHIR/pull/1441

**Expected behavior**
This should only generate an error if NONE of the or expressions are true.

**Additional context**
Similar, but different, from #1426

PR #1459, opened 2020-08-31
issue #1452 - split EvalContext issues and only add errorDetails if constraint failed

lmsurpre:
1. split EvaluationContext.issues into 2 lists:  supplementalWarnings and errorDetails
2. from FHIRValidator, only add errorDetails to the resulting issue list if the constraint actually fails

This addresses the issue because if the overall constraint is true, this will prevent us from adding the erroneous error details for the part of the "or" clause that failed.

### DIFF
--- a/fhir-path/src/main/java/com/ibm/fhir/path/function/ConformsToFunction.java
+++ b/fhir-path/src/main/java/com/ibm/fhir/path/function/ConformsToFunction.java
@@ -80,7 +80,7 @@ public class ConformsToFunction extends FHIRPathAbstractFunction {
             if (FHIRPathType.FHIR_UNKNOWN_RESOURCE_TYPE.equals(type)) {
                 if (!StructureDefinitionKind.RESOURCE.equals(structureDefinition.getKind())) {
                     // the profile (or base definition) is not applicable to type: UnknownResourceType
-                    generateIssue(evaluationContext, IssueSeverity.ERROR, IssueType.INVALID, "Conformance check failed: profile (or base definition) '" + url + "' is not applicable to type: UnknownResourceType", node.path());
+                    generateErrorDetail(evaluationContext, IssueType.INVALID, "Conformance check failed: profile (or base definition) '" + url + "' is not applicable to type: UnknownResourceType", node.path());
                     return SINGLETON_FALSE;
                 }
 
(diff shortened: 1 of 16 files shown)


fhir-path/src/main/java/com/ibm/fhir/path/function/ValidateCSFunction.java
```diff
@@ -95,17 +95,17 @@ private boolean validate(EvaluationContext evaluationContext, CodeSystem codeSys
                     return true;
                 }
             }
-            generateIssue(evaluationContext, IssueSeverity.ERROR, IssueType.CODE_INVALID, "CodeableConcept does not contain a coding element that matches the specified CodeSystem url and/or version", "%terminologies");
+            generateSupplementalWarning(evaluationContext, IssueSeverity.ERROR, IssueType.CODE_INVALID, "CodeableConcept does not contain a coding element that matches the specified CodeSystem url and/or version", "%terminologies");
```
michaelwschroeder, line 98: Should this be generateErrorDetail() since the issue it's replacing has a severity of ERROR (same comment on lines 104, 108)?

fhir-path/src/main/java/com/ibm/fhir/path/function/ConformsToFunction.java
```diff
@@ -116,7 +116,7 @@ public int getMaxArity() {
                     Collection<FHIRPathNode> result = evaluator.evaluate(evaluationContext, constraint.expression(), context);
                     if (evaluatesToBoolean(result) && isFalse(result)) {
                         // constraint validation failed
-                        generateIssue(evaluationContext, IssueSeverity.ERROR, IssueType.INVARIANT, constraint.id() + ": " + constraint.description(), node.path());
+                        generateSupplementalWarning(evaluationContext, IssueSeverity.ERROR, IssueType.INVARIANT, constraint.id() + ": " + constraint.description(), node.path());
```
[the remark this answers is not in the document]
lmsurpre: yes, it should.  updated.

fhir-path/src/main/java/com/ibm/fhir/path/function/ValidateCSFunction.java
lmsurpre: yes, it should. updated.

JohnTimm:
The more I think about this the more it seems like moving in this direction is making the semantics quite messy. I'm almost inclined to just to revert to the way it was before and change the severity from ERROR to WARNING or INFO in the places where this issue occurs.

lmsurpre:
closed in favor of #1465

PR #1465, opened 2020-09-01
issue #1452 - avoid adding context issues of severity error when constraint is true

lmsurpre:
An alternative to #1459. This PR does not attempt to model the difference between a issue detail which is propogated "up the stack" in the EvaludationContext and instead makes use of our assumption that "under no condition should we add a supplemental issue of type error when the overall constraint evaluates to true".


lmsurpre:
closed in favor of #1466

PR #1468, opened 2020-09-01
Enables the Unit Tests on Windows Build and Update windows.yml to catch early Windows Build Errors.

prb112:

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -25,7 +25,10 @@ jobs:
       run: mvn -B install --file fhir-examples/pom.xml --no-transfer-progress
     - name: Build parent without tests
       shell: powershell
-      run: mvn -B install --file fhir-parent/pom.xml --no-transfer-progress
+      run: |
+        mvn org.apache.maven.plugins:maven-dependency-plugin:3.1.2:go-offline -f fhir-parent -DexcludeReactor=true -Dmaven.wagon.http.retryHandler.count=3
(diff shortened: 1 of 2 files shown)


.github/workflows/windows.yml
```diff
@@ -25,7 +25,10 @@ jobs:
       run: mvn -B install --file fhir-examples/pom.xml --no-transfer-progress
     - name: Build parent without tests
       shell: powershell
-      run: mvn -B install --file fhir-parent/pom.xml -DskipTests --no-transfer-progress
+      run: |
+        mvn org.apache.maven.plugins:maven-dependency-plugin:3.1.2:go-offline -f fhir-parent -DexcludeReactor=true -Dmaven.wagon.http.retryHandler.count=3
+        mvn org.apache.maven.plugins:maven-dependency-plugin:3.1.2:resolve-plugins -f fhir-parent -DexcludeReactor=true -Dmaven.wagon.http.retryHandler.count=3
```
lmsurpre, line 30: can we add the `-B` and `--no-transfer-progress` to these to avoid some of the noise in the action's log?
```suggestion
        mvn -B org.apache.maven.plugins:maven-dependency-plugin:3.1.2:go-offline -f fhir-parent --no-transfer-progress -DexcludeReactor=true -Dmaven.wagon.http.retryHandler.count=3
        mvn -B org.apache.maven.plugins:maven-dependency-plugin:3.1.2:resolve-plugins -f fhir-parent --no-transfer-progress -DexcludeReactor=true -Dmaven.wagon.http.retryHandler.count=3
```

Issue #1216, opened 2020-06-10
FHIRValidator should check ValueSet membership for preferred/extensible bindings in the base specification


lmsurpre:
**Is your feature request related to a problem? Please describe.**
When you validate a resource instance against a given profile, one of the checks we perform is "valueset membership":
* for required bindings, if there is no corresponding coding we return an issue with severity error
* for preferred/extensible bindings, if there is no corresponding coding we return an issue with severity warning

However, if you validate a resource instance which has a preferred or extensible binding that is defined in the base spec, then there is no supplemental warning.

For example, validating a Patient resource that has a `maritalStatus` coding with a system of `http://terminology.hl7.org/CodeSystem/v2-0002`, there is no warning.

**Describe the solution you'd like**
Preferred and Extensible bindings from the base spec should behave similar to Preferred and Extensible bindings defined in profiles.  Construction of the objects should proceed, but FHIRValidation should return a supplemental warning when we validate them.

One option we could consider for this is to add an annotation with a generated constraint for each resource type that includes a code/Coding/CodeableConcept with preferred/extensible bindings.

Alternatively, we could try using the `@Binding` annotations that we already add at the field level from within the FHIRValidator.  For example, from Resource:
```
    @Binding(
        bindingName = "Language",
        strength = BindingStrength.ValueSet.PREFERRED,
        description = "A human language.",
        valueSet = "http://hl7.org/fhir/ValueSet/languages",
        maxValueSet = "http://hl7.org/fhir/ValueSet/all-languages"
    )
    protected final Code language;
```

**Describe alternatives you've considered**
I wonder if required bindings for Coding/CodeableConcept should be moved to FHIRValidator as well, or if we should continue to handle required bindings on object construction and preferred/extensible bindings on secondary validation...probably the latter since the required binding checks are already working.

**Additional context**
We got here by incrementally adding capability.  Initially, we only validated required codes (during object construction).  Then we added support for validating required bindings in profiles (during validation) and validating required Coding/CodeableConcept bindings in the base spec (during construction).  Finally, we enabled supplemental warnings for preferred/extensible bindings in profiles...but not for bindings from the base specification because we do those during object construction and we have no mechanism for providing supplemental warnings during object construction.

tbieste:
After looking at the code a bit, I think the approach of adding an annotation with a generated constraint for each resource type that includes a code/Coding/CodeableConcept with preferred/extensible bindings would be the way to go on this.

PR #1473, opened 2020-09-02
issue #1216 - Generate model constraints for value set bindings

tbieste:
This change adds generated constraints from the preferred/extensible bindings in the FHIR model, which enables validation warnings to be generated by the FHIR validator when those bindings are not followed, even when no profile is asserted.

### DIFF
--- a/fhir-tools/src/main/java/com/ibm/fhir/tools/CodeGenerator.java
+++ b/fhir-tools/src/main/java/com/ibm/fhir/tools/CodeGenerator.java
@@ -1523,6 +1523,7 @@ public class CodeGenerator {
                 valueMap.put("location", quote(location));
                 valueMap.put("description", quote(description));
                 valueMap.put("expression", quote(expression));
+                valueMap.put("generated", "true");
                 cb.annotation("Constraint", valueMap);
             }
         }
--- a/fhir-tools/src/main/java/com/ibm/fhir/tools/CodeGenerator.java
+++ b/fhir-tools/src/main/java/com/ibm/fhir/tools/CodeGenerator.java
@@ -915,7 +915,7 @@ public class CodeGenerator {
             }
 
             if (!nested) {
-                generateConstraintAnnotations(structureDefinition, cb);
+                generateConstraintAnnotations(structureDefinition, cb, className);
                 generateBindingAnnotation(structureDefinition, cb, className, structureDefinition.getJsonObject("snapshot").getJsonArray("element").getJsonObject(0));
                 cb.annotation("Generated", quote("com.ibm.fhir.tools.CodeGenerator"));
             }
@@ -1412,8 +1412,8 @@ public class CodeGenerator {
         }
         return sb.toString();
     }
-
-    private void generateConstraintAnnotations(JsonObject structureDefinition, CodeBuilder cb) {
+    
+    private void generateConstraintAnnotations(JsonObject structureDefinition, CodeBuilder cb, String className) {
         String name = structureDefinition.getString("name");
 
         List<JsonObject> allConstraints = new ArrayList<>();
@@ -1457,8 +1457,10 @@ public class CodeGenerator {
             }
         });
 
+        String lastId = null;
         for (JsonObject constraint : allConstraints) {
             String key = constraint.getString("key");
+            lastId = key;
             String path = pathMap.get(key);
             String severity = constraint.getString("severity");
             String human = constraint.getString("human");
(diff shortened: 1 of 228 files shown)


fhir-tools/src/main/java/com/ibm/fhir/tools/CodeGenerator.java
```diff
@@ -1475,6 +1477,154 @@ public int compare(JsonObject first, JsonObject second) {
...
+     * Generates constraint annotations from extensible and preferred bindings.
+     * 
+     * @param structureDefinition
+     *            the structure definition
+     * @param cb
+     *            the code builder
+     * @param className
+     *            the class name
+     * @param lastId
+     *            the ID of the last contraint
```
michaelwschroeder, line 1495: typo on 'contraint'

fhir-tools/src/main/java/com/ibm/fhir/tools/CodeGenerator.java
```diff
@@ -1475,6 +1477,154 @@ public int compare(JsonObject first, JsonObject second) {
...
+            String location = elementName.contains(".") ? elementName.replace(".div", ".`div`").replace("[x]", "") : "(base)";
+            String expressionElementName = "(base)".equals(location) ? elementName : "$this";
+
+            // Generate constraint for value set binding
+            String valueSet = binding.getString("valueSet", null);
+            if (valueSet != null) {
+                lastId = generateNextConstraintId(lastId, className);
+
+                String level = "Warning";
+                String description = (hasExtensibleBinding(elementDefinition) ? "SHALL, if possible, at least contain a code from value set "
```
michaelwschroeder, line 1515: wondering if you need the 'at least' part in the description - I think 'SHALL, if possible, contain a code from value set' and 'SHOULD contain a code from value set' works

tbieste:
Looks like there is a change to FHIR examples generator needed, so I will regenerate the FHIR examples as well.

fhir-tools/src/main/java/com/ibm/fhir/tools/CodeGenerator.java
```diff
@@ -1475,6 +1477,154 @@ public int compare(JsonObject first, JsonObject second) {
...
+                lastId = lastId.substring(0, lastId.length() - 1);
+            }
+            int nextKeyInt = Integer.parseInt(lastId) + 1;
+            nextSuffix = String.valueOf(nextKeyInt);
+        }
+        return camelCase(className) + "-" + nextSuffix;
+    }
+
+    /**
+     * Determines if any of the structure definition contains any elements with an extensible or preferred binding.
```
michaelwschroeder, line 1615: remove 'any of' before 'the structure definition'

fhir-tools/src/main/java/com/ibm/fhir/tools/CodeGenerator.java
tbieste: Fixed.

fhir-tools/src/main/java/com/ibm/fhir/tools/CodeGenerator.java
tbieste: Fixed.

fhir-tools/src/main/java/com/ibm/fhir/tools/CodeGenerator.java
tbieste: I can remove the words "at least".

JohnTimm:
@tbieste I think we should set generated = true on the constraints that have been added. We may also want to consider putting `generated` in the id of the constraint (i.e.`generated-library-1`). I want to make sure we distinguish what came from the spec out-of-the-box and what we are supplementing that with.

tbieste:
> @tbieste I think we should set `generated = true` on the constraints that have been added here. I want to make sure we distinguish what came from the spec out-of-the-box and what we are supplementing that with.
> 
> ```
> @Constraint(
>     id = "healthcareService-0",
>     level = "Warning",
>     location = "(base)",
>     description = "SHOULD contain a code from value set http://hl7.org/fhir/ValueSet/c80-practice-codes",
>     expression = "specialty.exists() implies (specialty.all(memberOf('http://hl7.org/fhir/ValueSet/c80-practice-codes', 'preferred')))",
>     generated = true
> )
> ```
Updated to add 'generated = true'.

JohnTimm (approved):
LGTM

lmsurpre (approved):
LGTM

Issue #1262, opened 2020-06-21
Support for wildcards in include and revinclude


lmsurpre:
**Is your feature request related to a problem? Please describe.**
From https://www.hl7.org/fhir/R4/search.html#revinclude:
> Both _include and _revinclude use the wild card "*" for the search parameter name, indicating that any search parameter of type=reference be included.

Currently, in strict mode, for a request like `{{url}}/Patient?_include=Patient:*` we return an error like:
```
{
    "resourceType": "OperationOutcome",
    "id": "ac-1e-42-f2-5efdab71-b044-46dc-b70c-f9b7b879e071",
    "issue": [
        {
            "severity": "fatal",
            "code": "invalid",
            "details": {
                "text": "Undefined Inclusion Parameter: Patient:*"
            },
            "expression": [
                "<empty>"
            ]
        }
    ]
}
```

**Describe the solution you'd like**
For both _include and _exclude search parameters, we should handle `*`. I think we could contain it to the fhir-search layer and handle it similar to a list of `_include`.
For example, for the Patient resource, that might be like:
`{{url}}/Patient?_include=Patient:general-practitioner&_include=Patient:link&_include=Patient:organization`.

However, we should ensure that we de-dup the resulting list in cases where the instance may have two different references to the same resource (hopefully rare).

**Describe alternatives you've considered**
Instead of implementing at the search processing layer (fhir-search) it may be possible to handle the `*` wildcard directly within the fhir-persistence implementation.  Reasons I prefer to keep it in fhir-search:
* JDBCQueryBuilder is already very complicated
* by implementing it in fhir-search we make it easier to implement alternative persistence layer implementations

**Additional context**
This search feature is especially important for the CARIN BlueButton implementation because there is discussion about whether it should be the *default* behavior used by clients.  See https://chat.fhir.org/#narrow/stream/204607-CARIN-Blue.20Button.20IG/topic/Call.20Recaps/near/201336959

prb112:
Merged, and back to @lmsurpre to close

lmsurpre:
I spot-checked this on our deployed server and it seems to be working well.  I like how the `self` URL that comes back is the full(expanded) query.

`Patient?_revinclude=ExplanationOfBenefit:*` became

```Patient?_count=10&_revinclude=ExplanationOfBenefit:payee:Patient&_revinclude=ExplanationOfBenefit:patient:Patient&_page=1```

`ExplanationOfBenefit?_include=ExplanationOfBenefit:*`
became
[...]

PR #1492, opened 2020-09-09
fix: double practitioner in SearchTest #1262

prb112:


michaelwschroeder (approved):
Thanks Paul!

Issue #1496, opened 2020-09-10
valueUrl is used incorrectly in the CapabilityStatement


prb112:
**Describe the bug**
valueUrl is used incorrectly in the CapabilityStatement
http://www.hl7.org/fhir/smart-app-launch/conformance/index.html#declaring-support-for-oauth2-endpoints

**To Reproduce**
1. Connect to the metadata endpoint
```
"extension": [
                    {
                        "extension": [
                            {
                                "url": "token",
                                "valueUrl": "https://cluster1-573846-250babbbe4c3000e15508cd07c1d282b-0000.us-east.containers.appdomain.cloud/auth/realms/master/protocol/openid-connect/token"
                            },
                            {
                                "url": "authorize",
                                "valueUrl": "https://cluster1-573846-250babbbe4c3000e15508cd07c1d282b-0000.us-east.containers.appdomain.cloud/auth/realms/master/protocol/openid-connect/auth"
                            },
                            {
                                "url": "register",
                                "valueUrl": "https://cluster1-573846-250babbbe4c3000e15508cd07c1d282b-0000.us-east.containers.appdomain.cloud/auth/realms/master/clients-registrations/openid-connect"
                            }
                        ],
                        "url": "http://fhir-registry.smarthealthit.org/StructureDefinition/oauth-uris"
                    }
```

**Expected behavior**
should be valueUri per spec http://www.hl7.org/fhir/smart-app-launch/conformance/index.html#declaring-support-for-oauth2-endpoints

**Additional context**
identified by Connectathon

PR #1497, opened 2020-09-10
valueUrl is used incorrectly in the CapabilityStatement #1496

prb112:


lmsurpre (approved):
LGTM

Issue #1500, opened 2020-09-10
Bundle processing validates resources more than once


punktilious:
**Describe the bug**
In FHIRRestHelper, bundle request entries are first validated in validateBundle. They are then validated a second time e.g. in doCreate. This second validation is redundant and wastes CPU.

**To Reproduce**
1. Enable tracing in server.xml e.g
```
<logging traceSpecification="*=info:com.ibm.fhir.*=finer" traceFormat="BASIC" maxFileSize="50" maxFiles="10"/>
```
2. Submit a bundle request
3. See trace logs and note that each resources is validated more than once.

**Expected behavior**
Each bundle entry only needs to be validated once.

**Additional context**
Modify the doCreate/doUpdate methods to support a flag which instructs the method to skipValidation. This can be set to true when called via the bundle processing path.

JohnTimm:
In the context of a bundle, we should perform validation but pass a flag to `doCreate` and `doUpdate` to indicate that validation has already been performed and can be skipped at the level.

punktilious:
As noted in the Additional context :-)

PR #1502, opened 2020-09-11
issue #1500 validate resources in bundles just once

punktilious:

### DIFF
--- a/fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
@@ -59,6 +59,7 @@ import com.ibm.fhir.server.listener.FHIRServletContextListener;
  */
 public class FHIRResource {
     private static final Logger log = java.util.logging.Logger.getLogger(FHIRResource.class.getName());
+    protected static final boolean NO_SKIP_VALIDATION = false; // do not skip validation
 
     public static final DateTimeFormatter HTTP_DATETIME_FORMATTER = new DateTimeFormatterBuilder()
             .appendPattern("EEE")
--- a/fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
@@ -59,7 +59,6 @@ import com.ibm.fhir.server.listener.FHIRServletContextListener;
  */
 public class FHIRResource {
     private static final Logger log = java.util.logging.Logger.getLogger(FHIRResource.class.getName());
-    protected static final boolean NO_SKIP_VALIDATION = false; // do not skip validation
 
     public static final DateTimeFormatter HTTP_DATETIME_FORMATTER = new DateTimeFormatterBuilder()
             .appendPattern("EEE")
--- a/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
@@ -124,7 +124,7 @@ public class FHIRRestHelper implements FHIRResourceHelpers {
 
     @Override
     public FHIRRestOperationResponse doCreate(String type, Resource resource, String ifNoneExist,
-            Map<String, String> requestProperties, boolean skipValidation) throws Exception {
+            Map<String, String> requestProperties, boolean doValidation) throws Exception {
         log.entering(this.getClass().getName(), "doCreate");
 
         FHIRRestOperationResponse ior = new FHIRRestOperationResponse();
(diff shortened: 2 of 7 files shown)


fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -59,6 +59,7 @@
  */
 public class FHIRResource {
     private static final Logger log = java.util.logging.Logger.getLogger(FHIRResource.class.getName());
+    protected static final boolean NO_SKIP_VALIDATION = false; // do not skip validation
```
prb112, line 62: small comment, no action necessary, just awareness. 
it comes off as a double negative and makes it slightly less readable

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -100,7 +100,9 @@
 
     private static final String EXTENSION_URL = "http://ibm.com/fhir/extension";
     private static final String LOCAL_REF_PREFIX = "urn:";
-
+    private static final boolean SKIP_VALIDATION = true; // constant for when we want to avoid repeating validation
+    private static final boolean NO_SKIP_VALIDATION = false; // do not skip validation
```
prb112, line 104: Since this is used in two spots, perhaps it should be reconciled into one constant

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -382,7 +385,7 @@ private FHIRRestOperationResponse doPatchOrUpdate(String type, String id, FHIRPa
             }
 
             // Validate the input and, if valid, start collecting supplemental warnings
-            List<Issue> warnings = new ArrayList<>(validateInput(newResource));
+            List<Issue> warnings = skipValidation ? new ArrayList<>() : new ArrayList<>(validateInput(newResource));
```
prb112, line 388: should it just be the validateInput(newResource)'s output not passing to a new arraylist?

operation/fhir-operation-document/src/main/java/com/ibm/fhir/operation/document/DocumentOperation.java
```diff
@@ -76,7 +76,7 @@ protected Parameters doInvoke(FHIROperationContext operationContext, Class<? ext
                     
                     if (persist) {
                         // FHIRResourceHelper resourceHelper = (FHIRResourceHelper) operationContext.getProperty(FHIROperationContext.PROPNAME_RESOURCE_HELPER;
```
prb112, line 78: should we drop this line?

prb112 (approved):
LGTM - small comments through out

JohnTimm:
What about creating two signatures in the interface with a default implementation that passes `true` for a `validateInput` parameter (or `false` for a skipValidation parameter)? My preference is for a parameter that reflects the default use case for calling `doCreate` and `doUpdate` such as `validateInput` or `performValidation` instead of the negative `skipValidation`. Having two signatures for `doCreate` and `doUpdate` will allow implementers of the `FHIROperation` interface to call `doCreate` and `doUpdate` as they do today and automatically get the desired behavior (to validate the input).

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
punktilious: I wanted to keep the semantics the same as the old code. I thought that warnings might be added to later on, hence the need to put it into its own list (the validateInput could be immutable). I think a better pattern would be to create the list, then pass it in for example: validateInput(warnings, newResource); The validateInput method could simply return the number of warnings it added.

punktilious:
> Minor comment: what about creating two signatures in the interface with a default implementation that passes `true` for a `validateInput` parameter (or `false` for a skipValidation parameter)? My preference is for a parameter that reflects the default use case for calling `doCreate` and `doUpdate` such as `validateInput` or `performValidation` instead of the negative `skipValidation`. Having two signatures for `doCreate` and `doUpdate` will allow implementers of the `FHIROperation` interface to call `doCreate` and `doUpdate` as they do today and automatically get the desired behavior (to validate the input).

Hmm, I toiled with this too...but my brain settled on skipValidation=true because you're telling it you want it to do something other than the norm (special case used for bundles). Using something like doValidation, it could suggest that it's an optional thing. At least, that's how I thought about it last night.

I'll hold off merging until I take a look at the interface/default stuff you mentioned. I've switched branches at the moment and am mid-change in fhir-bucket.

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
lmsurpre: FWIW I find `DO_VALIDATION = false` less mind-numbing than `NO_SKIP_VALIDATION = false`

punktilious: I've addressed this by taking John's suggestion of using a couple of default methods in the interface.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
punktilious: cleaned up

Issue #1494, opened 2020-09-10
Add Bundle.entry.search to Search Response


prb112:
**Is your feature request related to a problem? Please describe.**
Add Bundle.entry.search mode and score to Search Response
![image](https://user-images.githubusercontent.com/3016328/92753717-26b09600-f358-11ea-9f34-5ca7676c804a.png)

**Describe the solution you'd like**
1 - Downstream consumers need to know the state of the Search.

**Describe alternatives you've considered**
N/A

**Additional context**
Link to Spec - https://www.hl7.org/fhir/bundle.html and https://www.hl7.org/fhir/bundle-definitions.html

prb112:
PR 1503 identified an issue with Reflexsive references. 
The difficulty in including match versus outcome is hard to differentiate as we only return resources. 
To best differentiate, we may need to return back from the persistence layer the indications of INCLUDE or MATCH. 

Tests that must be done: 
1 - Search on Patient: _include=Observation:derived-from or _include=Patient:link on the Patient Resource
and show that the include is properly marks the resources included. 
2 - Test with only match of a single resource
3 - Test with match + include of different resources
```
I suspect a change has to be made at the persistence layer to do this efficiently.

**Comment by michaelwschroeder** [comment]
A potential solution is to replace the count query on include searches with a query that returns resource IDs. Those results could then be used to determine the set of MATCH resources vs. INCLUDE resources.

**Comment by d0roppe** [comment]
stretch goal for sprint 21

**Comment by michaelwschroeder** [comment]
I believe by adding a derived SORT_ORDER column to the search query to enable sorting of the MATCH resources before the INCLUDE resources will allow us, in combination with the `pageSize` and `totalCount` values in the searchContext, to determine which returned resources are MATCH vs. INCLUDE, regardless of resource type. For example, if we know all MATCH resources come first in the result set, then we can determine the number of MATCH resources based on `pageSize` and `totalCount`. If `totalCount` > `pageSize`, then the first `pageSize` resources in the result set are MATCH. If `pageSize` > `totalCount`, then the first `totalCount` resources in the result set are MATCH. Remaining resources in the result set are INCLUDE.

Here's an _include query:

```
SELECT
   RESOURCE_ID,
   LOGICAL_RESOURCE_ID,
   VERSION_ID,
   LAST_UPDATED,
   IS_DELETED,
   DATA,
   LOGICAL_ID 
FROM
   (
      SELECT
         RESOURCE_ID,
[...]

PR #1503, opened 2020-09-11
Add Bundle.entry.search to Search Response #1494

prb112:

### DIFF
--- a/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
@@ -45,12 +45,14 @@ import com.ibm.fhir.exception.FHIROperationException;
 import com.ibm.fhir.model.patch.FHIRPatch;
 import com.ibm.fhir.model.resource.Bundle;
 import com.ibm.fhir.model.resource.Bundle.Entry;
+import com.ibm.fhir.model.resource.Bundle.Entry.Search;
 import com.ibm.fhir.model.resource.OperationOutcome;
 import com.ibm.fhir.model.resource.OperationOutcome.Issue;
 import com.ibm.fhir.model.resource.Parameters;
 import com.ibm.fhir.model.resource.Resource;
 import com.ibm.fhir.model.type.Code;
 import com.ibm.fhir.model.type.CodeableConcept;
+import com.ibm.fhir.model.type.Decimal;
 import com.ibm.fhir.model.type.Extension;
 import com.ibm.fhir.model.type.UnsignedInt;
 import com.ibm.fhir.model.type.Uri;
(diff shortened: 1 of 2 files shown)


lmsurpre:
Some of the included formatting changes are good, but I'm not a fan of the ones where we take a multi-line fluent sequence and jam it into a single (very long) line.

JohnTimm:
> Some of the included formatting changes are good, but I'm not a fan of the ones where we take a multi-line fluent sequence and jam it into a single (very long) line.

I tend to agree with @lmsurpre on this one. Multiple lines was easier to read in this particular instance.

prb112:
Apologies, I'll fix that in the morning.  Definitely not the intent

prb112:
Fixed

lmsurpre (approved):
I'm not sold on the value of adding a score of "1.0" to every response entry, but the demarcation of MATCH vs OUTCOME vs INCLUDE is quite welcome.  LGTM

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -2208,9 +2212,33 @@ private Bundle createSearchBundle(List<Resource> resources, FHIRSearchContext se
...
                 throw new IllegalStateException("Returned resources must have an id.");
             }
+
+            // By default it's a MATCH, else if it's not a match it is of one of two types
+            // OUTCOME for operational outcomes embedded in the bundle
+            // INCLUDE for revinclude and include resources
+            SearchEntryMode mode = SearchEntryMode.MATCH;
+            String bundleEntryResourceType = resource.getClass().getSimpleName();
+            Search.Builder builder = Search.builder();
+            if (!resourceType.equals(bundleEntryResourceType)) {
```
michaelwschroeder, line 2222: I believe it's possible for an 'included' resource to be of the same type as the matched resource, for example _include=Observation:derived-from or _include=Observation:has-member could return other Observation resources. Think this would mark it as MATCH instead of INCLUDE.

prb112: unfortunately - we don't have enough details to differentiate.

prb112: We'd have to modify the underlying jdbc persistence layer, and I don't think that's reasonable.

prb112: Thoughts?

michaelwschroeder: I guess it comes down to whether an occasional inaccuracy in the typing is outweighed by the usefulness of having that typing. One other thought is that for include searches, maybe the count query be replaced with a query that actually returns the resource IDs, which you could then use to determine if the resource was a MATCH vs. INCLUDE. But I'm not familiar with the persistence layer yet, so don't know how big a deal that would be, and would probably be a slight performance hit as well.

prb112: Yeah - it'd be a pretty big change.  We could document this in the conformance document?

Issue #1506, opened 2020-09-14
fhir-path parser cannot parse datetime fractionalSeconds with more than 6 digits of precision


michaelwschroeder:
**Describe the bug**
When attempting to use a dateTime value with fractional seconds of more than 6 digits of precision in a FHIRPath expression, an exception is returned: `com.ibm.fhir.path.exception.FHIRPathException: java.time.format.DateTimeParseException: Text '2012-04-15T15:00:00.000000000Z' could not be parsed, unparsed text found at index 26`

**To Reproduce**
1. use a dateTime value with a fractional seconds precision of more than 6 digits in a FHIRPath expression

**Expected behavior**
In order to be consistent with parsing of dateTime and instant fields on object creation (see issue #1446), and date values in search queries (see issue #1505), up to 9 digits of precision should be allowed for fractional seconds in FHIRPath dateTime values,  and should be truncated to 6 digits of precision when evaluating the FHIRPath expression.

**Additional context**

lmsurpre:
Unit test coverage looks good.

PR #1513, opened 2020-09-18
Issue #1506 - enable parsing dateTime fractional seconds up to 9 digits

michaelwschroeder:

### DIFF
--- a/fhir-path/src/main/java/com/ibm/fhir/path/FHIRPathTimeValue.java
+++ b/fhir-path/src/main/java/com/ibm/fhir/path/FHIRPathTimeValue.java
@@ -13,10 +13,12 @@ import java.time.LocalTime;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeFormatterBuilder;
 import java.time.temporal.ChronoField;
+import java.time.temporal.ChronoUnit;
 import java.time.temporal.TemporalAmount;
 import java.util.Collection;
 import java.util.Objects;
 
+import com.ibm.fhir.model.util.ModelSupport;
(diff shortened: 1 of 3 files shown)


fhir-path/src/main/java/com/ibm/fhir/path/FHIRPathTimeValue.java
```diff
@@ -74,7 +76,7 @@ public LocalTime time() {
      *     a new FHIRPathTimeValue instance
      */
     public static FHIRPathTimeValue timeValue(String text) {
-        LocalTime time = LocalTime.parse(text, PARSER_FORMATTER);
+        LocalTime time = ModelSupport.truncateTime(LocalTime.parse(text, PARSER_FORMATTER), ChronoUnit.MICROS);
```
prb112, line 79: Is this automatically appending microseconds?

michaelwschroeder: Well, if a LocalTime instance is created without specifying any fractional seconds, the fields within that instance representing the fractional seconds will just be set to 0 (MILLI_OF_SECOND, MICRO_OF_SECOND, NANO_OF_SECOND), so in that sense, every TemporalAccessor always has a value for microseconds. The ModelSupport.truncateTime() method is calling the LocalTime.truncatedTo() method, which simply sets the fields within the LocalTime instance that are smaller than the specified unit to zero. So in this case, it will set the NANO_OF_SECOND field to 0, effectively truncating any precision greater than microseconds.

prb112 (approved):
LGTM

JohnTimm (approved):
LGTM

Issue #1480, opened 2020-09-05
Improve CapabilityStatement to support Security


prb112:
**Is your feature request related to a problem? Please describe.**
CapabilityStatement enables a Security feature which some clients are looking at to determine the authentication protocol and CORS support to use. Refer to https://www.hl7.org/fhir/CapabilityStatement.html

[Scenario 0: Demonstrate support authentication and authorization](https://confluence.hl7.org/pages/viewpage.action?pageId=91983711)
- Read Server's capability statement to decide if server requires authentication
- If server requires authentication
- Follow SMART App Launch IG to start client

**Describe the solution you'd like**
- Implements https://www.hl7.org/fhir/capabilitystatement-definitions.html#CapabilityStatement.rest.security
- Changes based on what is enabled on the server using fhir-config?

**Describe alternatives you've considered**
None

**Additional context**
Identified as part of a connectathon.

lmsurpre:
we already populate CapabilityStatement.rest.security with the OAuth endpoints that are configured in our fhir-server-config.  For example, here is a snippet from our CapabilityStatement for the upcoming connectathon:
```json
"rest": [
        {
            "mode": "server",
            "security": {
                "extension": [
                    {
                        "extension": [
                            {
                                "url": "token",
                                "valueUrl": "https://cluster1-573846-250babbbe4c3000e15508cd07c1d282b-0000.us-east.containers.appdomain.cloud/auth/realms/master/protocol/openid-connect/token"
                            },
                            {
                                "url": "authorize",
                                "valueUrl": "https://cluster1-573846-250babbbe4c3000e15508cd07c1d282b-0000.us-east.containers.appdomain.cloud/auth/realms/master/protocol/openid-connect/auth"
                            },
                            {
                                "url": "register",
                                "valueUrl": "https://cluster1-573846-250babbbe4c3000e15508cd07c1d282b-0000.us-east.containers.appdomain.cloud/auth/realms/master/clients-registrations/openid-connect"
                            }
                        ],
                        "url": "http://fhir-registry.smarthealthit.org/StructureDefinition/oauth-uris"
                    }
                ],
                "service": [
                    {
                        "coding": [
                            {
                                "system": "http://terminology.hl7.org/CodeSystem/restful-security-service",
                                "code": "SMART-on-FHIR"
                            }
                        ],
                        "text": "OAuth2 using SMART-on-FHIR profile (see http://docs.smarthealthit.org)"
                    }
                ]
            },
```

[...]

Issue #1020, opened 2020-05-04
No longer able to map unauthenticated users to the FHIRUsers security-role


albertwang-ibm:
We were using a configDropin to map special-subject EVERYONE to the FHIRUsers role (since authentication is being handled by a gateway).
Now, with the latest changes, this isn't working and it results in the health check failing with HTTP 401.

lmsurpre:
Reproduced locally just by adding a config override like this:
```
    <!-- webApp security is provided via IAM in IBM Cloud -->
    <webAppSecurity singleSignonEnabled="false" useAuthenticationDataForUnprotectedResource="false"/>
    <webApplication id="fhir-server-webapp">
        <application-bnd id="binding">
            <security-role id="users" name="FHIRUsers">
                <special-subject type="EVERYONE"/>
            </security-role>
        </application-bnd>
    </webApplication>
```
and hitting a protected endpoint.

But I was NOT able to reproduce locally for the `$healthcheck` endpoint, so not sure what is going on there.

lmsurpre:
I created a tiny reproducing case and openened it as an issue for OpenLiberty at https://github.com/OpenLiberty/open-liberty/issues/12050.
We might need to revert the change to the use of `@RolesAllowed` until that can get addressed (or else come up with an alternative for our cloud deploy).

lmsurpre:
Our workaround has been to add the Authorization header via our kubernetes ingress.  That said, I really thought the Liberty team would have this addressed by now; we might need to do something here...

prb112:
This feature is going to be in the next release.

lmsurpre:
I confirmed that this is working now by adding a configDropin with the following content to a local deploy;
```
<server>
    <webApplication id="fhir-server-webapp">
        <application-bnd id="bind">
            <security-role id="users" name="FHIRUsers">
                <special-subject type="EVERYONE"/>
            </security-role>
        </application-bnd>
    </webApplication>
</server>
```

with liberty 22.0.0.3 this is finally behaving as expected; I am able to invoke protected endpoints like `$healthcheck` without specifying a username/password:
```
$ curl -k -i 'https://localhost:9443/fhir-server/api/v4/$healthcheck'
HTTP/2 200 
date: Fri, 01 Apr 2022 19:05:37 GMT
content-length: 0
content-language: en-US
```

PR #1526, opened 2020-09-28
issue #1480 - introduce new fhir-server-config security group and use

lmsurpre:
1. combine our existing "oauth" and "authFilter" fhir-server-config property groups into a single new group called "security".

2. introduce new properties like:
    * fhirServer/security/basic (enabled by default)
    * fhirServer/security/cert (enabled by default)
    * fhirServer/security/smart (disabled by default, with customizable endpoints and scopes)
    * fhirServer/security/cors (default=true)

3. use these values to populate `CapabilityStatement.rest.security`
    * only serve up the Oauth extensions in the CapabilityStatement when fhirServer/security/service is set to `OAuth` or `SMART-on-FHIR`

4. only serve up the `.well-known/smart-capabilities` endpoint when fhirServer/security/service is set to `SMART-on-FHIR`

### DIFF
--- a/conformance/fhir-ig-davinci-pdex-plan-net/src/main/java/com/ibm/fhir/ig/davinci/pdex/plan/net/util/IndexGenerator.java
+++ b/conformance/fhir-ig-davinci-pdex-plan-net/src/main/java/com/ibm/fhir/ig/davinci/pdex/plan/net/util/IndexGenerator.java
@@ -23,7 +23,6 @@ import com.ibm.fhir.registry.util.Index.Entry;
 public class IndexGenerator {
     public static void main(String[] args) throws Exception {
         buildAndWriteIndex("hl7/fhir/us/davinci-pdex-plan-net");
-        buildAndWriteIndex("hl7/fhir/uv/vhdir");
     }
 
     private static void buildAndWriteIndex(String packagePath) throws IOException, FileNotFoundException {
--- a/fhir-server/liberty-config/config/default/fhir-server-config.json
+++ b/fhir-server/liberty-config/config/default/fhir-server-config.json
@@ -21,12 +21,15 @@
                     "authorizedClientCertIssuerOU": ""
                 }
             },
-            "smart": {
+            "oauth": {
                 "enabled": false,
                 "regUrl": "https://<host>:9443/oauth2/endpoint/oauth2-provider/registration",
                 "authUrl": "https://<host>:9443/oauth2/endpoint/oauth2-provider/authorize",
                 "tokenUrl": "https://<host>:9443/oauth2/endpoint/oauth2-provider/token",
-                "supportedScopes": ["openid", "profile", "fhirUser", "launch/patient", "patient/*.*", "offline_access"]
+                "smart": {
+                    "enabled": false,
+                    "scopes": ["openid", "profile", "fhirUser", "launch/patient", "patient/*.*", "offline_access"]
+                }
             }
         },
         "notifications": {
--- a/fhir-server/liberty-config/config/default/fhir-server-config.json
+++ b/fhir-server/liberty-config/config/default/fhir-server-config.json
@@ -8,15 +8,26 @@
             "conditionalDeleteMaxNumber": 10,
             "serverRegistryResourceProviderEnabled": true
         },
-        "authFilter": {
-            "enabled": false,
-            "authorizedClientCertClientCN": "",
-            "authorizedClientCertIssuerOU": ""
-        },
-        "oauth": {
-            "regUrl": "https://<host>:9443/oauth2/endpoint/oauth2-provider/registration",
-            "authUrl": "https://<host>:9443/oauth2/endpoint/oauth2-provider/authorize",
-            "tokenUrl": "https://<host>:9443/oauth2/endpoint/oauth2-provider/token"
+        "security": {
+            "cors": true,
+            "basic": {
+                "enabled": true
+            },
+            "certificates": {
+                "enabled": true,
+                "authFilter": {
+                    "enabled": false,
+                    "authorizedClientCertClientCN": "",
+                    "authorizedClientCertIssuerOU": ""
+                }
+            },
+            "smart": {
+                "enabled": false,
+                "regUrl": "https://<host>:9443/oauth2/endpoint/oauth2-provider/registration",
+                "authUrl": "https://<host>:9443/oauth2/endpoint/oauth2-provider/authorize",
+                "tokenUrl": "https://<host>:9443/oauth2/endpoint/oauth2-provider/token",
+                "supportedScopes": ["openid", "profile", "fhirUser", "launch/patient", "patient/*.*", "offline_access"]
+            }
         },
         "notifications": {
             "common": {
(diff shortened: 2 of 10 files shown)


conformance/fhir-ig-davinci-pdex-plan-net/src/main/java/com/ibm/fhir/ig/davinci/pdex/plan/net/util/IndexGenerator.java
```diff
@@ -23,7 +23,6 @@
 public class IndexGenerator {
     public static void main(String[] args) throws Exception {
         buildAndWriteIndex("hl7/fhir/us/davinci-pdex-plan-net");
-        buildAndWriteIndex("hl7/fhir/uv/vhdir");
```
lmsurpre, line 26: not related to the other changes, but thought I might as well include it anyway

JohnTimm (approved):
LGTM

conformance/fhir-ig-davinci-pdex-plan-net/src/main/java/com/ibm/fhir/ig/davinci/pdex/plan/net/util/IndexGenerator.java
prb112: sounds good

fhir-server/liberty-config/config/default/fhir-server-config.json
```diff
@@ -8,15 +8,29 @@
...
-            "enabled": false,
-            "authorizedClientCertClientCN": "",
-            "authorizedClientCertIssuerOU": ""
-        },
-        "oauth": {
-            "regUrl": "https://<host>:9443/oauth2/endpoint/oauth2-provider/registration",
-            "authUrl": "https://<host>:9443/oauth2/endpoint/oauth2-provider/authorize",
-            "tokenUrl": "https://<host>:9443/oauth2/endpoint/oauth2-provider/token"
+        "security": {
+            "cors": true,
```
prb112, line 12: If we require at least one auth mechanism to be selected, maybe adding a comment here to highlight it in the security block/json object

fhir-server/liberty-config/server.xml
```diff
@@ -99,6 +99,12 @@
     </dataSource>
 
     <webAppSecurity allowFailOverToBasicAuth="true" singleSignonEnabled="false"/>
+    
+    <cors domain="/"
+        allowedOrigins="*"
```
prb112, line 104: I don't think this is a best practice.  
We should probably limit to localhost or externalize the variable.

fhir-server/liberty-config/server.xml
```diff
@@ -99,6 +99,12 @@
...
+        allowedMethods="GET, DELETE, POST, PUT, HEAD"
+        allowedHeaders="*"
```
prb112, line 106: I'm not sure about the wildcard here, I think it's a-ok to have '*'

fhir-server/src/main/java/com/ibm/fhir/server/FHIRApplication.java
```diff
@@ -61,7 +63,9 @@ public FHIRApplication() {
                 classes.add(Search.class);
                 classes.add(Update.class);
                 classes.add(VRead.class);
-                classes.add(WellKnown.class);
+                if (FHIRConfigHelper.getBooleanProperty(FHIRConfiguration.PROPERTY_SECURITY_OAUTH_SMART_ENABLED, false)) {
+                    classes.add(WellKnown.class);
```
prb112, line 67: confirmation, if OAuth doesn't exist, we don't add and advertise?

fhir-server/src/main/java/com/ibm/fhir/server/resources/WellKnown.java
```diff
@@ -99,41 +107,36 @@ private JsonObject buildSmartConfig() throws Exception {
...
+        responseBuilder.add("capabilities", Json.createArrayBuilder() // required
+                .add("launch-standalone")
+                .add("client-public")
+                .add("client-confidential-symmetric")
+                .add("permission-offline")
+                // TODO: make this configurable
+                .add("context-standalone-patient")
+                .add("permission-patient")
+//                .add("context-standalone-encounter")
+//                .add("permission-user")
```
prb112, line 134: does this deserve a backlog issue?

prb112 (approved):
LGTM, left commentary, otherwise looks good. 
CORS filter allowedOrigins was my only concern

fhir-server/liberty-config/config/default/fhir-server-config.json
lmsurpre: Remember, this is the "advertised" security capabilities (shows up in our CapabilityStatement), which doesn't necessarily need to match the real security used from the liberty config (e.g. server.xml).
For example, even with issue #1020 unresolved, its possible to use basic auth on the server but to add that auth via some proxy so that it appears "open" to end users.
Thus, I think its allowed to advertise no security mechanism.

fhir-server/liberty-config/server.xml
lmsurpre: I've heard this before too, although I'm not familiar with the downsides of allowing all origins.
The spec specifically calls out `*` as the "easiest option", although it does stop short of saying that it is recommended / not recommended.
https://www.hl7.org/fhir/capabilitystatement-definitions.html#CapabilityStatement.rest.security.cors

fhir-server/liberty-config/server.xml
lmsurpre: Its allowed per the liberty docs.  Not sure if there are security ramifications or not, but I didn't see any reason to restrict which headers are allowed at this time.

fhir-server/src/main/java/com/ibm/fhir/server/FHIRApplication.java
lmsurpre: this properly lives UNDER oauth, so yes, if oauth field is missing, this smart field will be missing too (and thus false).

however, it *is* possible to have the fhir-server-config oauth field present and set to false, while the underlying smart_enabled field is set to true.
In this case, we will advertise the wellknown endpoint, but we won't advertise the oauth endpoints in the CapabilityStatement.  I considered changing this if to check that BOTH are enabled, but decided that this behavior would probably be acceptable (and maybe even useful in rare circumstances).

fhir-server/liberty-config/server.xml
prb112: interesting... https://advancedweb.hu/is-access-control-allow-origin-star-insecure/ it's not all that bad from this article

fhir-server/src/main/java/com/ibm/fhir/server/resources/WellKnown.java
lmsurpre: I could maybe just "fix" it by adding another property to the config.  let me just do that now.

prb112: no worries, just trying to take a critical eye to any change

lmsurpre: done

JohnTimm (approved):
LGTM

PR #1558, opened 2020-10-06
fix: stored procedures are not updatable due to a missing boolean = true

prb112:

### DIFF
--- a/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/BaseObject.java
+++ b/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/BaseObject.java
@@ -255,7 +255,7 @@ public abstract class BaseObject implements IDatabaseObject {
      */
     @Override
     public void applyVersion(IDatabaseAdapter target, IVersionHistoryService vhs) {
-        if (vhs.applies(getSchemaName(), getObjectType().name(), getObjectName(), version)) {
+        if (vhs.applies(getSchemaName(), getObjectType().name(), getObjectName(), version) || getObjectType().equals(DatabaseObjectType.PROCEDURE)) {
             logger.fine("Applying change [v" + version + "]: " + this.getTypeNameVersion());
 
             // Apply this change to the target database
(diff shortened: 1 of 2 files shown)


lmsurpre:
Seems fine to me.  A related question I have is whether it updates all FHIR_ADMIN schema stored procs (e.g. setTenant) or just the schema that is passed in.
The reason I ask is that our docs currently say "Update the stored procedures and functions for FHIRDATA (and not FHIR_ADMIN)" but I'm not sure that is right.
If it doesn't update FHIR_ADMIN stored procs, do we have a different means for updating that one?

prb112:
> Seems fine to me. A related question I have is whether it updates all FHIR_ADMIN schema stored procs (e.g. setTenant) or just the schema that is passed in.
> The reason I ask is that our docs currently say "Update the stored procedures and functions for FHIRDATA (and not FHIR_ADMIN)" but I'm not sure that is right.
> If it doesn't update FHIR_ADMIN stored procs, do we have a different means for updating that one?

so ... it turns out we never supported that
we just created the set_tenant and only updated with the --update-schema

punktilious:
This must be a regression because I used to update the stored procedures all the time when they were first being developed. In fact, the update option was added so that I could quickly update just the stored procs without having to process the rest of the model.

punktilious:
Also, I would've preferred to keep the various schemas in different PhysicalDataModel instances. I've considered a refactor of that stuff, but too much other stuff going on at the moment.

prb112:
> This must be a regression because I used to update the stored procedures all the time when they were first being developed. In fact, the update option was added so that I could quickly update just the stored procs without having to process the rest of the model.

Yes, it's a regression

lmsurpre (approved):
LGTM

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/model/BaseObject.java
```diff
@@ -255,7 +255,7 @@ public void applyTx(IDatabaseAdapter target, ITransactionProvider tp, IVersionHi
      */
     @Override
     public void applyVersion(IDatabaseAdapter target, IVersionHistoryService vhs) {
-        if (vhs.applies(getSchemaName(), getObjectType().name(), getObjectName(), version)) {
+        if (vhs.applies(getSchemaName(), getObjectType().name(), getObjectName(), version) || getObjectType().equals(DatabaseObjectType.PROCEDURE)) {
```
lmsurpre, line 258: wonder if we should add a comment to explaining the why.
basically we just want the procedures to get re-written every time (whether its a newer version or not)?

prb112: Right.  I can add a comment

Issue #1504, opened 2020-09-11
Evaluate local to external reference mapping in transaction Bundle processing


punktilious:
**Describe the bug**
For bundles with bundleType == TRANSACTION, the FHIR server remaps certain local references to server-defined references as defined at https://www.hl7.org/fhir/bundle.html#bundle-unique

Evaluate this processing for efficiency as well as ensuring the remapping is correct and complete regardless of bundle order.
- We suspect that we may not be capturing ALL of the references before we start storing any of the resources.
- Because ordering is not deterministic we think we'll need to do a full pass before any of the processing happens.

For example, collecting the `urn:uuid` values from the POSTed resources and replacing the references in the PUT resources with those (or similar).

Consider refactoring the core processing so that it can be unit-tested thoroughly without requiring the server to be running.

**To Reproduce**
Load large bundles with different resource relationships.

**Expected behavior**
Bundle processing should be efficient and correct.

Specifically, we need a unit test for confirming the current behavior is correct or for adding detail of the bug to this issue.

**Additional context**
N/A

michaelwschroeder:
Background on our current processing of local references in transaction bundles:
1. bundle entries can have a local reference specified in the Bundle.Entry.fullUrl field - the local reference must start with 'urn:'
2. for resources specified in POST and PUT requests, we look for reference fields containing references starting with 'urn:' and if they match a bundle entry local reference, we replace them with the actual resource reference of the resource in the bundle entry
3. we build a map of [local reference --> real resource reference] to enable the local reference resolution in the previous step. The map is built as follows:
    - after a POST request is processed, its [local ref --> real resource ref] mapping is added to the map (because it's only at that point that we know the logical ID the resource was created with)
    - after all POST resources have been processed, but before any PUT resources have been processed, all PUT [local ref --> real resource ref] mappings are added to the map (we can do this before processing because PUT requests are updates, so resources already have logical ID set)
4. processing order of requests is all POST requests first, then PUT requests

With current code, the following scenarios work:
1. any POST resource containing a local reference to a POST resource processed before it in the bundle will have the reference resolved properly because each POST resource has its mapping added to the map after being processed, meaning a mapping for a resource already processed will exist when a resource lower in the bundle is processed
2. any PUT resource containing a local reference to a POST resource will have the reference resolved properly because all POST resources will have their mappings added to the map before any PUT resources are processed
[...]

PR #1560, opened 2020-10-06
Issue #1504 - properly process local references in transaction bundles

michaelwschroeder:

### DIFF
--- a/docs/src/pages/guides/FHIRServerUsersGuide.md
+++ b/docs/src/pages/guides/FHIRServerUsersGuide.md
@@ -3,7 +3,7 @@ layout: post
 title:  IBM FHIR Server User's Guide
 description: IBM FHIR Server User's Guide
 Copyright: years 2017, 2020
-lastupdated: "2020-09-01"
+lastupdated: "2020-10-07"
 permalink: /FHIRServerUsersGuide/
 ---
 
@@ -906,7 +906,7 @@ Response resource:
 ## 4.8 Using local references within request bundles
 Inter-dependencies between resources are typically defined by one resource containing a field of type `Reference` which contains an _external reference_<sup id="a5">[5](#f5)</sup> to another resource. For example, an `Observation` resource could reference a `Patient` resource via the Observation's `subject` field. The value that is stored in the `Reference-type` field (for example, `subject` in the case of the `Observation` resource) could be an absolute URL, such as `https://fhirserver1:9443/fhir-server/api/v4/Patient/12345`, or a relative URL (for example, `Patient/12345`).
 
-In order to establish a reference to a resource, you must first know its resource identifier. However, if you are using a request bundle to create both the referenced resource (`Patient` in this example) and the resource which references it (`Observation`), then it is impossible to know the `Patient`resource identifier before the request bundle has been process (that is, before the new `Patient` resource is created).
+In order to establish a reference to a resource, you must first know its resource identifier. However, if you are using a request bundle to create both the referenced resource (`Patient` in this example) and the resource which references it (`Observation`), then it is impossible to know the `Patient`resource identifier before the request bundle has been processed (that is, before the new `Patient` resource is created).
 
 Thankfully, the HL7 FHIR specification defines a way to express a dependency between two resources within a request bundle by using a _local reference_<sup id="a6">[6](#f6)</sup>. In the following example, a request bundle contains a `POST` request to create a new `Patient` resource, along with a `POST` request to create a new `Observation` resource that references that `Patient`:
 
@@ -947,13 +947,18 @@ In order to reference a resource via a local reference, you must first define a 
 After you define a local identifier for the referenced resource, you can then define one or more references to that resource by using the local identifier instead of an external identifier. In the preceding example, you can see that the Observation's `subject.reference` field specifies the Patient's local identifier as specified in the `fullUrl` field of the Patient's request entry.
 
 ### 4.8.1 Processing rules
-There is one rule for the use of local references within a request bundle:  A local identifier must be defined via a request entry's `fullUrl` field before that local identifier can be used in a local reference.
+The following processing rules apply for the use of local references within a request bundle:
+1.  A local identifier must be defined via a request entry's `fullUrl` field in order for that local identifier to be used in a local reference.
+2.  Local references will only be recognized for local identifiers associated with requst entries with a request method of `POST` or `PUT`.
+3.  `POST` requests will be processed before `PUT` requests.
+4.  There is no order dependency within a request bundle for request entries defining local identifiers, and request entries which reference those local identifiers via local reference. The exception to this rule is for request entries which specify [conditional create](https://www.hl7.org/fhir/http.html#ccreate) or [conditional update](https://www.hl7.org/fhir/http.html#cond-update) requests.
+5.  If a request entry specifying a conditional create or update request defines a local identifier, that request entry must be processed before a request entry which references the local identifier in a local reference.
 
-In the example in [Section 4.8.0.1](#4801-example-1-observation-references-patient-via-local-reference), you can see that there are two POST requests and the `Patient` request entry appears in the bundle before the `Observation` request entry. This example satisfies the rule because the FHIR server will process the POST request entries in the order in which they appear within the request bundle.
+In the example in [Section 4.8.0.1](#4801-example-1-observation-references-patient-via-local-reference), you can see that there are two POST requests and the `Patient` request entry appears in the bundle before the `Observation` request entry. However, based on rule 4, it would still be a valid request bundle if the `Observation` request entry appeared before the `Patient` request entry, unless the `Patient` request entry specified a conditional create (rule 5).
 
-If, however, those entries were reversed, the FHIR server would return an error when processing the `Observation` request entry, because the `Patient` local identifier is not defined yet.
+If those entries were reversed, and the `Patient` request entry specified a conditional create, then the FHIR server would return an error when processing the `Observation` request entry, because the `Patient` local identifier would not be defined yet.
 
-The following example also satisfies the rule:
+The following examples also satisfy the local reference processing rules:
 
 #### 4.8.1.1 Example 2: Observation (PUT) appears before Patient (POST)
 ```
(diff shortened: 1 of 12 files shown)


fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -326,8 +324,8 @@ private FHIRRestOperationResponse doPatchOrUpdate(String type, String id, FHIRPa
                     ior.setPrevResource(null);
 
                     // if no id provided, then generate an id for the input resource
-                    if (newResource.getId() == null || newResource.getId() == null) {
-                        id = UUID.randomUUID().toString();
+                    if (newResource.getId() == null) {
```
prb112, line 327: strange logic there, generally when I see that I wonder why there were two, is newResource ever null? was the first check in case newResource was null?

fhir-persistence/src/main/java/com/ibm/fhir/persistence/FHIRPersistence.java
```diff
@@ -130,4 +132,13 @@
     default boolean isDeleteSupported() {
         return false;
     }
+
+    /**
+     * Generates a logical identity.
+     * 
+     * @return logical identity
+     */
+    default String getLogicalId() {
```
lmsurpre, line 141: I wonder if the method name should be `generateResourceId()` or some such to make it more clear that its no a logical id for the FHIRPersistence object but instead is used to generate resource ids.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -1512,432 +1526,477 @@ private Bundle processBundleEntries(Bundle requestBundle, Bundle responseBundle,
...
-            }
-
-            // Now visit each of the request entries using the list of indices obtained above.
-            // Use hashmap to store both the index and the according updated response bundle entry.
-            HashMap<Integer, Bundle.Entry> responseIndexAndEntries =
-                    new HashMap<Integer, Bundle.Entry>();
-            for (Integer entryIndex : entryIndices) {
+            // Visit each of the request entries, processing those with the specified request method.
+            // Use hashmap to store both the index and the accordingly updated response bundle entry.
+            HashMap<Integer, Bundle.Entry> responseIndexAndEntries = new HashMap<Integer, Bundle.Entry>();
```
prb112, line 1550:
```suggestion
            Map<Integer, Bundle.Entry> responseIndexAndEntries = new HashMap<Integer, Bundle.Entry>();
```
Shouldn't the interface be sufficient here?

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -1512,432 +1526,477 @@ private Bundle processBundleEntries(Bundle requestBundle, Bundle responseBundle,
...
+            for (int entryIndex=0; entryIndex<requestBundle.getEntry().size(); ++entryIndex) {
```
prb112, line 1551: try using `Entry<Integer,Bundle.Entry> entry : responseIndexAndEntries.entrySet() `

It'll save the subsequent update

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -1512,432 +1526,477 @@ private Bundle processBundleEntries(Bundle requestBundle, Bundle responseBundle,
...
+                    long initialTime = System.currentTimeMillis();
+                    
+                    try {
+                        FHIRUrlParser requestURL = new FHIRUrlParser(request.getUrl().getValue());
+
+                        if (log.isLoggable(Level.FINER)) {
+                            log.finer("Processing bundle request entry " + entryIndex + "; method="
+                                    + request.getMethod().getValue() + ", url="
+                                    + request.getUrl().getValue());
+                            log.finer("--> path: " + requestURL.getPath());
```
prb112, line 1569:
```suggestion
                            log.finer("--> path: '" + requestURL.getPath() + "'");
```

prb112: I suggest bracketing the log with a single quote as spaces are hard to see otherwise.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -1512,432 +1526,477 @@ private Bundle processBundleEntries(Bundle requestBundle, Bundle responseBundle,
...
+                            log.finer("--> query: " + requestURL.getQuery());
```
prb112, line 1570: same with this line --- 

I suggest bracketing the log with a single quote as spaces are hard to see otherwise.

fhir-server/src/test/java/com/ibm/fhir/server/test/FHIRRestHelperTest.java
```diff
@@ -0,0 +1,987 @@
...
+     * Test transaction bundle post with circular local reference dependency.
+     * Procedure has local reference to Encounter and Encounter has local reference to Procedure.
+     */
+    @Test
+    public void testTransactionBundlePostWithCircularDependency() throws Exception {
+        FHIRPersistence persistence = new MockPersistenceImpl();
+        FHIRRestHelper helper = new FHIRRestHelper(persistence);
+        
+        Encounter encounter = Encounter.builder().status(EncounterStatus.FINISHED).clazz(Coding.builder().code(Code.of("AMB")).build())
+                .reasonReference(Reference.builder().reference(com.ibm.fhir.model.type.String.of("urn:2")).build()).build();
```
lmsurpre, line 164: this style is totally fine, but often times we'll prefer a style with more whitespace like this:
```suggestion
        Encounter encounter = Encounter.builder()
                .status(EncounterStatus.FINISHED)
                .clazz(Coding.builder()
                    .code(Code.of("AMB"))
                    .build())
                .reasonReference(Reference.builder()
                    .reference(com.ibm.fhir.model.type.String.of("urn:2"))
                    .build())
                .build();
```

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -1512,432 +1526,477 @@ private Bundle processBundleEntries(Bundle requestBundle, Bundle responseBundle,
...
-                            Resource resource = requestEntry.getResource();
-                            if (resource == null) {
-                                String msg =
-                                        "BundleEntry.resource is required for bundled create requests.";
-                                throw buildRestException(msg, IssueType.NOT_FOUND);
-                            }
-
-                            // Convert any local references found within the resource to their
-                            // corresponding external reference.
+                        responseBuilder.status(string(Integer.toString(SC_NOT_FOUND)));
```
prb112, line 1608: since this if fixed, it may be great to make that string value a static final `string(Integer.toString(SC_NOT_FOUND))`

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -1512,432 +1526,477 @@ private Bundle processBundleEntries(Bundle requestBundle, Bundle responseBundle,
...
+                        responseIndexAndEntries.put(entryIndex, responseEntryBuilder.resource(FHIRUtil.buildOperationOutcome(e, false)).response(responseBuilder.build()).build());
+                        logBundleRequestCompletedMsg(requestDescription.toString(), initialTime, SC_NOT_FOUND);
+                    } catch (FHIRPersistenceResourceDeletedException e) {
+                        if (failFast) {
+                            String msg = "Error while processing request bundle.";
+                            throw new FHIRRestBundledRequestException(msg).withIssue(e.getIssues());
+                        }
+                        
+                        Bundle.Entry.Response.Builder responseBuilder = response.toBuilder();
+                        responseBuilder.status(string(Integer.toString(SC_GONE)));
```
prb112, line 1618: since this if fixed, it may be great to make that string value a static final string(Integer.toString(SC_GONE))

fhir-server/src/test/java/com/ibm/fhir/server/test/FHIRRestHelperTest.java
lmsurpre: additionally, we'll often statically import the `com.ibm.fhir.model.type.String.string` factory method so that we don't need to fully quality `com.ibm.fhir.model.type.String` for each usage.

lmsurpre: these style comments apply to many of the tests in this class...not planning to repeat it for each instance

fhir-server/src/test/java/com/ibm/fhir/server/test/FHIRRestHelperTest.java
```diff
@@ -0,0 +1,987 @@
+/*
+ * (C) Copyright IBM Corp. 2020
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.ibm.fhir.server.test;
```
prb112, line 7: It'd be great to have an integration test in fhir-server-test.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -2048,23 +2035,56 @@ public int compare(Integer indexA, Integer indexB) {
...
-        Bundle.Entry.Request request = entry.getRequest();
-        if (request != null) {
-            if (request.getUrl() != null && request.getUrl().getValue() != null) {
-                FHIRUrlParser requestURL = new FHIRUrlParser(request.getUrl().getValue());
-                path = requestURL.getPath();
+    private Map<String, String> buildLocalRefMap(Bundle requestBundle, Bundle responseBundle) throws Exception {
+        Map<String, String> localRefMap = new HashMap<>();
+
+        for (int entryIndex=0; entryIndex<requestBundle.getEntry().size(); ++entryIndex) {
+            Bundle.Entry requestEntry = requestBundle.getEntry().get(entryIndex);
```
prb112, line 2050: use an entryset

fhir-persistence/src/main/java/com/ibm/fhir/persistence/FHIRPersistence.java
```diff
@@ -130,4 +132,13 @@
...
+        return UUID.randomUUID().toString();
+    }
```
prb112, line 143: I think because this is in FHIRPersistence this is OK.... but it seems like something we shouldn't have as a default.

prb112: also the @override declares it as public.  best to mark this as Public in the interface (rather than the default package)

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -289,7 +289,7 @@ protected Action buildActionChain() {
             // system-generated value. For the update-or-create scenario, see update().
             // Default version is 1 for a brand new FHIR Resource.
             int newVersionNumber = 1;
-            logicalId = logicalIdentityProvider.createNewIdentityValue();
+            logicalId = getLogicalId();
```
prb112, line 292: What does this change get us?

docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -988,46 +993,87 @@ The following example also satisfies the rule:
...
+        "request" : {
+            "method" : "POST",
+            "url" : "Procedure"
+        }
+    } ]
+}
+```
 
-For example, in Example 1 from [Section 4.8.0.1](#4801-example-1-observation-references-patient-via-local-reference), the FHIR server detects the use of the local identifier in the `Patient` request entry (`urn:uuid:7113a0bb-d9e0-49df-9855-887409388c69`) and -- after creating the new `Patient` resource -- establishes a mapping between the local identifier and the resulting external reference associated with the new `Patient` (for example, `Patient/1cc5d299-d2be-4f93-8745-a121232ffe5b`).
+While processing a request bundle, but before processing individual request entries, the FHIR server will detect the use of a local identifier within any `POST` or `PUT` request entry's `fullUrl` field, and will establish a mapping between that local identifier and the corresponding external identifier that will result from performing the `POST` or `PUT` operation.
```
prb112, line 1035: use an active tense 
```suggestion
While processing a request bundle, but before processing individual request entries, the IBM FHIR Server detects the use of a local identifier within any `POST` or `PUT` request entry's `fullUrl` field, and establishes a mapping between that local identifier and the corresponding external identifier that resultsfrom performing the `POST` or `PUT` operation. 
```

fhir-server/src/test/java/com/ibm/fhir/server/test/FHIRRestHelperTest.java
```diff
@@ -0,0 +1,987 @@
...
+    @Test
+    public void testTransactionBundlePostWithDependency() throws Exception {
+        FHIRPersistence persistence = new MockPersistenceImpl();
+        FHIRRestHelper helper = new FHIRRestHelper(persistence);
+        
+        Patient patient = Patient.builder().id("1").build();
+        Bundle.Entry.Request bundleEntryRequest = Bundle.Entry.Request.builder().method(HTTPVerb.POST).url(Uri.of("Patient")).build();
+        Bundle.Entry bundleEntry = Bundle.Entry.builder().fullUrl(Uri.of("urn:1")).resource(patient).request(bundleEntryRequest).build();
+        
+        Procedure procedure = Procedure.builder().status(ProcedureStatus.COMPLETED).subject(Reference.builder().reference(com.ibm.fhir.model.type.String.of("urn:1")).build()).build();
```
lmsurpre, line 81: Building these test resources programmatically is totally fine, but an alternative you can use is to import a "minimal" example via code like this:
```
Procedure procedure = TestUtil.readExampleResource("json/ibm/minimal/Procedure-1.json")
procedure = procedure.toBuilder().
        subject(Reference.builder()
            .reference(com.ibm.fhir.model.type.String.of("urn:1"))
            .build())
        .build()
```
Its most useful if you have resource types with a lot of required fields.

lmsurpre: It might even be worth wrapping that pattern is a new helper method like `TestUtil.minimal(Procedure.class)` (which would return a new Builder for the passed resource type class, which has just the required fields populated).  Thoughts?

fhir-server/src/test/java/com/ibm/fhir/server/test/FHIRRestHelperTest.java
lmsurpre: Are the tests in com.ibm.fhir.server.test.Bundletest not sufficient?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
lmsurpre: getLogicalId was added to the FHIRPersistence interface, so that it can be called from outside the PL.  basically, he made that piece of it public.
so now the FHIRRestHelper can delegate to the FHIRPersistence to have it generate a resource id (separate from actually creating a resource)

prb112: a-ok

fhir-server/src/test/java/com/ibm/fhir/server/test/FHIRRestHelperTest.java
prb112: sounds fair Lee

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: newResource should not be null at this point - it's checked during bundle validation (or resource validation if request is not a bundle request)

fhir-persistence/src/main/java/com/ibm/fhir/persistence/FHIRPersistence.java
michaelwschroeder: I renamed to generateResourceId()

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: yes, fixed

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: In this case, I'm looping through the requestBundle, not the responseIndexAndEntries hashmap. I used this form of the for loop because I also need to get the corresponding entry in the responseBundle.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: Done.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: Done.

fhir-server/src/test/java/com/ibm/fhir/server/test/FHIRRestHelperTest.java
michaelwschroeder: I updated the testcases to use the style suggested, as well as importing `com.ibm.fhir.model.type.String.string`

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: Done.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: Done.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: Used this form of the for loop because I need to get both the requestBundle entry and the corresponding responseBundle entry.

fhir-persistence/src/main/java/com/ibm/fhir/persistence/FHIRPersistence.java
michaelwschroeder: I removed the `default` impl and updated the implementing classes to provide a generateResourceId() method. I also explicitly declared all methods in the interface `public`.

docs/src/pages/guides/FHIRServerUsersGuide.md
michaelwschroeder: Done - took your suggested changes.

fhir-server/src/test/java/com/ibm/fhir/server/test/FHIRRestHelperTest.java
michaelwschroeder: I did create a new TestUtil.getMinimalResource() method, per your suggestion. However, fhir-server does not currently pull in the fhir-model test-jar, and when I tried updating the pom to do so, I was getting some 'class not found' exceptions, so decided to just go with what I have. We can revisit if you think it's important.

lmsurpre: not important.  will be interested to see if we can make use of this new method over time...it seems generally useful to me; like a cleaner version of the pattern I described above

lmsurpre (approved):
LGTM

prb112 (approved):
LGTM

Issue #1273, opened 2020-06-23
Handle versioned references in chained and include/revinclude searches


lmsurpre:
**Is your feature request related to a problem? Please describe.**
https://www.hl7.org/fhir/search.html#versions says 
* If a resource has a reference that is versioned, and chaining is performed, the criteria should ideally be evaluated against the version referenced, but most systems will not be capable of this because search is only defined to function against the current version of a resource
* Where a search does not act on the referenced version, search results SHOULD contain a OperationOutcome with a warning that indicates the discrepancy
* If a resource has a reference that is versioned and _include is performed, the specified version SHOULD be provided.

**Describe the solution you'd like**
1. For chained search, ensure it works for searching the latest version of the versioned references (i.e. drop the version part) and add a warning to the OperationOutcome as suggested
2. For include, include the exact version that was references
3. For revinclude, do not include resources with versioned references to this instance unless the version is right

**Describe alternatives you've considered**
Alternatively, for number 1 we could change our schema to include a link to the version of the resource that the searchparameters were extracted from and stop deleting old search parameter values, but thats a large change with many implications.

lmsurpre:
Also relates to #1218 in that the reverse chaining should probably work similar to _revinclude... if the resource that references this one uses a versioned link then only follow that if the version is right.  That said, I think both of these are edge case...versioned references are not common.

kmbarton423:
stretch goal for sprint 21

michaelwschroeder:
Updated design: 
[issue1273design_v2.docx](https://github.com/IBM/FHIR/files/5874853/issue1273design_v2.docx)

d0roppe:
verified the issue based on the included design document.

PR #1571, opened 2020-10-08
Issue 1570 - Fix issues with Reference search and add comprehensive tests

prb112:


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -529,6 +527,7 @@ private SqlQueryData processReferenceParm(Class<?> resourceType, QueryParameter
             // Handle query parm representing this name/value pair construct:
             // <code>{name}:{Resource Type} = {resource-id}</code>
             if (queryParm.getModifier() != null && queryParm.getModifier().equals(Modifier.TYPE)) {
+                System.out.println(queryParm.getType());
```
lmsurpre, line 530: should it be a log statement?  or remove altogether?

prb112, line 530:
```suggestion
```

prb112: fixed  - it was meant  for debugging, and then I wanted to remove it.  Thanks for the eagle eyes

fhir-search/src/main/java/com/ibm/fhir/search/reference/value/UrlHandlerImpl.java
```diff
@@ -18,11 +18,11 @@
  * http://server/Patient/1
  * -> http://server/Patient/1
  * -> Patient/1
- * -> 1
+ * -> 1 (only when there is one target)
  */
 public class UrlHandlerImpl implements ParameterValueHandler {
 
-    private static final String REGEX = "/[A-z][a-z]{2,10}/[A-Za-z0-9\\-\\.]{1,64}$";
+    private static final String REGEX = "/([A-z][a-z]{2,36}/[A-Za-z0-9\\-\\.]{1,64})$";
```
lmsurpre, line 25: I assume the limit here is for performance reasons?
The longest resource type I found is 30 chars, but I still wonder if we should bump this to like 48 just to be catious.
Alternatively, we could check the actual list of resource names from the resourceTypes valueset and then use the one with max length, but thats probably overkill.

lmsurpre: we also have a related URL pattern in FHIRUtil.
```
sb.append(")\\/[A-Za-z0-9\\-\\.]{1,64}(\\/_history\\/[A-Za-z0-9\\-\\.]{1,64})?");
```

fhir-search/src/main/java/com/ibm/fhir/search/reference/value/TypeIdHandlerImpl.java
```diff
@@ -6,18 +6,26 @@
...
 
 /**
  * The supported cases are:
  * Patient/1
  * -> http://server/Patient/1
+ * -> 1 (only when there is one target)
  */
 public class TypeIdHandlerImpl implements ParameterValueHandler {
+
+    private static final String REGEX = "^([A-z][a-z]{2,36}/[A-Za-z0-9\\-\\.]{1,64})$";
```
lmsurpre, line 22: same comment about erring on the side of caution on referencetype length
```suggestion
    private static final String REGEX = "^([A-z][a-z]{2,64}/[A-Za-z0-9\\-\\.]{1,64})$";
```

lmsurpre:
Do the REGEX patterns in UrlHandlerImpl and TypeIdHandlerImpl fail when we get versioned references?
See https://www.hl7.org/fhir/search.html#versions for more info on these.
Not totally sure about our support for these to begin with, but probably want to at least consider them.

prb112:
> Do the REGEX patterns in UrlHandlerImpl and TypeIdHandlerImpl fail when we get versioned references?
> See https://www.hl7.org/fhir/search.html#versions for more info on these.
> Not totally sure about our support for these to begin with, but probably want to at least consider them. I know at the least I opened #1273 recently and pointed Robin at it for consideration as part of his updated db schema.

There is CanonicalUrlHandlerImpl that handles that case, it splits it off. 
TypeId does not yet support version and the regex would fail. 

I did apply your review suggestions.

lmsurpre:
> TypeId does not yet support version and the regex would fail.

Can we add this to the end of the REGEX for that one: 
`(\\/_history\\/[A-Za-z0-9\\-\\.]{1,64})?`

And then maybe a single test which exercises it?

prb112:
> > TypeId does not yet support version and the regex would fail.
> 
> Can we add this to the end of the REGEX (but before the `$`) for both that one and for UrlHandlerImpl:
> `(\\/_history\\/[A-Za-z0-9\\-\\.]{1,64})?`
> 
> And then maybe a single test which exercises it for each?

Sure - that's no problemo

prb112:
> > TypeId does not yet support version and the regex would fail.
> 
> Can we add this to the end of the REGEX (but before the `$`) for both that one and for UrlHandlerImpl:
> `(\\/_history\\/[A-Za-z0-9\\-\\.]{1,64})?`
> 
> And then maybe a single test which exercises it for each?

Updated

lmsurpre (approved):
LGTM

Issue #1508, opened 2020-09-14
Create performance measurements for CMS Payer Interop scenarios


punktilious:
**Is your feature request related to a problem? Please describe.**
Create access/search queries which can be executed at scale to drive read load into the FHIR server.

**Describe the solution you'd like**
Use random patient ids from the resource bundles loaded by fhir-bucket to drive CMS Payer Interop search/access scenarios.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

d0roppe:
Local Random Patient selection has now been implemented in JMeter test plan.

d0roppe:
Have added Random patient selector to the JMeter test form a CSV file I can download from the db console.  This seems to work well and quick for getting random patients as requested by this issue.

d0roppe:
JMeter test for this now in master branch.  should this task be closed?  @lmsurpre

PR #1573, opened 2020-10-09
issue #1508 - Initial delivery of fhir-meter project

d0roppe:
Detail: This is the initial delivery of the fhir-meter project
containing a readme and a JMeter test plan.  The test plan is for
searching a FHIR server with preloaded resources where you can get a
list of resource ID's into a CSV file before you run the test.  This is
meant to add load to the server through JMeter load testing.

Signed-off-by: Dag O. Roppe <roppe@us.ibm.com>

### DIFF
--- a/performance/fhir-meter/README.md
+++ b/performance/fhir-meter/README.md
@@ -1,0 +1,24 @@
+This is the README file for the the JMeter test plan searchTest.jmx.
+The searchTest.jmx test plan was written to just include several different searches over 19 different resources.  You can toggle any of the searches enabled or disabled to test and stress what you would like to do.  Look at the HTTP Request samplers in the Path: section to see what the search is doing.
+This test plan was created with JMeter version 5.3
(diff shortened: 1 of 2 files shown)


lmsurpre:
@prb112 do you think we need a copyright + license notice at the top of the jmx file?

prb112:
> @prb112 do you think we need a copyright + license notice at the top of the jmx file?

I think we should.  It should be best practice.

prb112:
@lmsurpre  @d0roppe 
As this is not packaged into a maven project, I don't think it falls under the `new-module`.  I want to double check the tag.  

Is this intended to be released to bintray or packaged for release?

lmsurpre:
> As this is not packaged into a maven project, I don't think it falls under the new-module. I want to double check the tag.
> Is this intended to be released to bintray or packaged for release?

good call.  i agree and removed the label.

lmsurpre (changes requested):
@d0roppe per the convo with paul, could you please add a copyright + license notice to the top of the jmx file?
probably via an xml comment like
```
<!-- 
(C) Copyright IBM Corp. 2020

SPDX-License-Identifier: Apache-2.0
-->
```

lmsurpre (approved):
I havn't run these myself, but I think it looks good to me.

PR #1582, opened 2020-10-14
Issue #1504 - add back in sort for PUT/DELETE requests

michaelwschroeder:

### DIFF
--- a/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
@@ -17,9 +17,13 @@ import java.net.URI;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeFormatterBuilder;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.UUID;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -1181,7 +1185,7 @@ public class FHIRRestHelper implements FHIRResourceHelpers {
 
             List<Bundle.Entry> responseList = new ArrayList<Bundle.Entry>();
             
-            List<String> localIdentifiers = new ArrayList<>();
+            Set<String> localIdentifiers = new HashSet<>();
 
             for (Bundle.Entry requestEntry : bundle.getEntry()) {
                 // Create a corresponding response entry and add it to the response bundle.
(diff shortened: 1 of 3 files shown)


fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -2023,6 +2044,98 @@ private Bundle reconstructResponseBundle(Bundle responseBundle,
...
+        }
+
+        @Override
+        public int compare(Integer indexA, Integer indexB) {
+            Bundle.Entry a = entries.get(indexA);
+            Bundle.Entry b = entries.get(indexB);
+            String pathA = getUrlPath(a);
+            String pathB = getUrlPath(b);
+
+            log.fine("Comparing request entry URL paths: " + pathA + ", " + pathB);
```
punktilious, line 2107: Need if isLoggable barrier here

punktilious (changes requested):

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
lmsurpre: like this:
```suggestion
            if (log.isLoggable(Level.FINE)) {
                log.fine("Comparing request entry URL paths: " + pathA + ", " + pathB);
            }
```

michaelwschroeder: Added check around all log.fine() statements that did not have check.

lmsurpre: FWIW, I don't think its always needed.  But it never hurts to have it, so I'm OK with adding it everywhere.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -1549,10 +1571,27 @@ private Bundle processEntriesForMethod(Bundle requestBundle, Bundle responseBund
...
+            // This list will contain the indices associated with only the entries for the specified http method.
+            List<Integer> entryIndices =
+                    getBundleRequestIndicesForMethod(requestBundle, responseBundle, httpMethod);
+            if (log.isLoggable(Level.FINER)) {
+                log.finer("Bundle request indices to be processed: " + entryIndices.toString());
+            }
+
+            // Next, for PUT and DELETE requests, we need to sort the indices by the request url path value.
+            if (httpMethod.equals(HTTPVerb.PUT) || httpMethod.equals(HTTPVerb.DELETE)) {
+                sortBundleRequestEntries(requestBundle, entryIndices);
```
lmsurpre, line 1584: I think @punktilious requested to add this one back in "to avoid deadlock issues", but I'm hoping someone can explain it to me because I'm not sure I get it.

lmsurpre: here is the problematic scenario:

within single transaction bundle:
```
bundle 1:
PUT A
PUT B
```

within a separate transaction bundle:
```
bundle 2:
PUT B
PUT A
```

the timing problem would be:
1. bundle 1 arrives first, it goes through entries and when it hits PUT A it locks A
2. bundle 2 arrives very soon afterward, and it starts processing the PUT B before bundle 1's PUT B (locking B)
3. deadlock

punktilious (approved):
LGTM

lmsurpre (approved):
LGTM

PR #1585, opened 2020-10-15
fix: add missing igs to the  maven profiles in the release process

prb112:


prb112:
Discussion with Lee, we're going to target 4.5.0 so we have an opportunity to work with the two new profiles.

prb112:
Follow on marked for `Add missing igs to the maven profiles in the release process #1606 `

lmsurpre (approved):
LGTM

Issue #1594, opened 2020-10-19
Posting bundles are not returning warnings in the OperationOutcome


mpayne2:
I have a bundle with 3 claim entries which each entry generates warnings if posted to FHIR individually, none of the warnings are returned in the OperationOutcome when I post them in the bundle.  This happens on both transaction and batch types.

The only time I get the warnings back for an entry is if the entry also had an error.
In my test, none of my claims have the narrative set which should generate a warning message.  When I set one of the claims to have an error, like setting the code but not the code system in claim.item.quantity, I get the error and warning for that entry.

I expected to get all the warnings in the OperationOutcome for all entries when I specify the Prefer header to return=OperationOutcome.

Here is my sample:
{
 "resourceType": "Bundle",
  "id": "bundle-transaction",
  "type": "batch",
  "entry": [ {
    "resource": {
      "resourceType": "Claim",
      "id": "0010",
      "meta": {
        "versionId": "1",
        "lastUpdated": "2020-03-24T19:42:52.014+00:00",
        "source": "#anRjJex8JdtnGhIV",
        "tag": [ {
          "system": "https://smarthealthit.org/tags",
          "code": "Covid19 synthetic population from Synthea"
        } ]
      },
      "status": "active",
      "type": {
        "coding": [ {
          "system": "http://terminology.hl7.org/CodeSystem/claim-type",
          "code": "institutional"
        } ]
      },
      "use": "claim",
      "patient": {
        "reference": "Patient/0001",
        "display": "Deon400 Lebsack687"
      },
      "billablePeriod": {
        "start": "2014-11-05T05:00:20-06:00",
        "end": "2014-11-05T05:15:20-06:00"
      },
      "created": "2014-11-05T05:15:20-06:00",
      "provider": {
        "reference": "Organization/7a208ea5-66f6-392a-8472-059d61de2410",
        "display": "MIDWEST ANESTHESIA PARTNERS LLC"
      },
      "priority": {
        "coding": [ {
          "system": "http://terminology.hl7.org/CodeSystem/processpriority",
          "code": "normal"
        } ]
      },
      "facility": {
        "reference": "Location/21bd6648-ed47-44ec-98a3-da7b6d47eab7",
        "display": "MIDWEST ANESTHESIA PARTNERS LLC"
      },
      "supportingInfo": [ {
        "sequence": 1,
        "category": {
          "coding": [ {
            "system": "http://terminology.hl7.org/CodeSystem/claiminformationcategory",
            "code": "info"
[...]

PR #1599, opened 2020-10-19
Issue #1594 - fix overwriting of validate warnings in bundle processing

michaelwschroeder:

### DIFF
--- a/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
@@ -106,7 +106,7 @@ public class FHIRRestHelper implements FHIRResourceHelpers {
     private static final com.ibm.fhir.model.type.String SC_GONE_STRING = string(Integer.toString(SC_GONE));
     private static final com.ibm.fhir.model.type.String SC_NOT_FOUND_STRING = string(Integer.toString(SC_NOT_FOUND));
     private static final com.ibm.fhir.model.type.String SC_OK_STRING = string(Integer.toString(SC_OK));
-    
+
     public static final DateTimeFormatter PARSER_FORMATTER = new DateTimeFormatterBuilder()
             .appendPattern("EEE")
             .optionalStart()
@@ -1190,7 +1190,7 @@ public class FHIRRestHelper implements FHIRResourceHelpers {
             List<OperationOutcome.Issue> issueList = new ArrayList<OperationOutcome.Issue>();
 
             List<Bundle.Entry> responseList = new ArrayList<Bundle.Entry>();
-            
+
             Set<String> localIdentifiers = new HashSet<>();
 
             for (Bundle.Entry requestEntry : bundle.getEntry()) {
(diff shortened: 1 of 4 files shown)


fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -2330,7 +2330,19 @@ private String getAbsoluteUri(String baseUri, String relativeUri) {
             if (HTTPReturnPreference.REPRESENTATION.equals(FHIRRequestContext.get().getReturnPreference())) {
                 bundleEntryBuilder.resource(resource);
             } else if (HTTPReturnPreference.OPERATION_OUTCOME.equals(FHIRRequestContext.get().getReturnPreference())) {
-                bundleEntryBuilder.resource(operationOutcome);
+                OperationOutcome responseEntryOutcome = (OperationOutcome)responseEntry.getResource();
+                if (responseEntryOutcome != null && !responseEntryOutcome.equals(FHIRUtil.ALL_OK)) {
```
prb112, line 2334: Since OperationOutcome overrides equals, I think this is OK.  we're not comparing two pointers necessarily. 

Confirming (I also double checked this)

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -2330,7 +2330,19 @@ private String getAbsoluteUri(String baseUri, String relativeUri) {
...
+                    if (operationOutcome.equals(FHIRUtil.ALL_OK)) {
+                        bundleEntryBuilder.resource(responseEntryOutcome);
```
prb112, line 2336: slight question here, should ALL_OK still be added along with the warning... Technically it's acceptable.

michaelwschroeder: the way it's currently working in doCreate and doPatchOrUpdate is that the OperationOutcome is generated by the FHIRUtil.buildOperationOutcome(issues) method, passing in any warnings found during the create or update. In that method, the ALL_OK is only added if there are no warnings. So that's the logic I used - will only return the ALL_OK if there are no warnings.

lmsurpre, line 2336: Does it matter whether the request was a `batch` vs a `transaction`?
I know for failed `transaction` requests we send back a single OperationOutcome (vs. sending one-per-resource in `batch`)...just want to make sure this logic works for both bundle types.

prb112: Works for me,

michaelwschroeder: Yes, this works for both types. This method will only get called after a successful call to the persistence layer, so the only change is that on a successful request, the OperationOutcome object may now contain warnings that it didn't contain before. In the case of a failed `transaction` request during validation, that code path is unchanged - it returns all validation warnings/errors for the failed bundle entry only. In the case of a failed `transaction` request during create or update, that code path is also unchanged from before. An exception is thrown and a bad request code is returned.

PR #1601, opened 2020-10-19
Use FHIRConfigHelper over fhirConfig to support per-tenant capabilities

lmsurpre:
`FHIRResource.fhirConfig` is always using the default tenant.
Instead, we should use FHIRConfigHelper for most properties so that we
get that nice tenant config -> default config fallback behavior.

This change focuses on the following properties:
* `fhirServer/core/capabilityStatementCacheTimeout`
* `fhirServer/security/*`

The former was marked as tenant-specific, but wasn't.  The latter was
correctly marked as `N|N` but there was no reason for this since they are just
about controlling the CapabilityStatement and smart-configuration info.

In addition to useful behavior, this change puts us in a better position
for testing various configs more easily.


prb112 (approved):
LGTM

Issue #1595, opened 2020-10-19
Error dropping FHIR schema for PostgreSQL


lmsurpre:
**Describe the bug**
fhir-persistence-schema drop-schema support isn't working for postgresql
related to #1167 

**To Reproduce**
1. create a postgresql schema using fhir-persistence-schema
2. run the tool's `--drop-schema-fhir` operation

Result:  it fails when it tries to drop the stored procedures (because those are only applicable to db2).

I think its just a matter of building the "PhysicalDataModel" more intelligently before we invoke the drop/delete.

**Expected behavior**
It should successfully drop the schema

**Additional context**
We also have the `--drop-schema` action which doesn't see to do anything for these schemas...needs to be revisited

lmsurpre:
We're past the reported errors, but this continues to fail for some postgres databases due to https://github.com/IBM/FHIR/issues/1631

PR #1607, opened 2020-10-20
DropSchema related changes for issue #1595

lmsurpre:
Previously, if `--drop-schema` was used without `--drop-schema-fhir`, `--drop-schema-oauth`, or `--drop-schema-jbatch` then it completed with success without actually doing anything.

I think what happened is that we were debating two different options for these args:
A. passing some kind of argument to `--drop-schema`, so it might be like `--drop-schema fhir` or `--drop-schema oauth jbatch`; or
B. using explicit flags for dropping the fhir, schema, or oauth tables, so usage would be like `--drop-schema-fhir --drop-schema-oauth`

But somehow we ended up with a combination of the two which I hope we can all agree is not very good.
This pull request modifies the current behavior and turns it into option B, removing support for the `--drop-schema` action.

Note: this PR does *not* actually address issue #1595 ...it only reworks the user experience of the cli so that others can avoid the error that I made initially.

### DIFF
--- a/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/postgresql/PostgreSqlTranslator.java
+++ b/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/postgresql/PostgreSqlTranslator.java
@@ -113,7 +113,7 @@ public class PostgreSqlTranslator implements IDatabaseTranslator {
 
     @Override
     public boolean isUndefinedName(SQLException x) {
-        return "42X05".equals(x.getSQLState());
+        return "42704".equals(x.getSQLState());
     }
 
     @Override
--- a/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/postgresql/PostgreSqlTranslator.java
+++ b/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/postgresql/PostgreSqlTranslator.java
@@ -113,7 +113,11 @@ public class PostgreSqlTranslator implements IDatabaseTranslator {
 
     @Override
     public boolean isUndefinedName(SQLException x) {
-        return "42704".equals(x.getSQLState());
+        String sqlState = x.getSQLState();
+        return "42704".equals(sqlState) ||
+               "42883".equals(sqlState) ||
+               "42P01".equals(sqlState) ||
+               "42P02".equals(sqlState);
     }
 
     @Override
--- a/fhir-persistence-schema/README.md
+++ b/fhir-persistence-schema/README.md
@@ -85,7 +85,7 @@ The following sections include common values for `OPTIONS`.
 ```
 --prop-file db2.properties
 --schema-name FHIRDATA
---drop-schema
+--drop-schema-fhir
 --drop-admin
 --confirm-drop
 ```
--- a/fhir-persistence-schema/README.md
+++ b/fhir-persistence-schema/README.md
@@ -30,7 +30,7 @@ su - db2inst1 -c "db2 \"connect to fhirdb\" && db2 \"grant connect on database T
 
 **Note 2:** When creating the database, `PAGESIZE` is important. So *do* use the statement below and not, e.g., the environment variable `DBNAME` of the [Db2 Docker container](https://hub.docker.com/r/ibmcom/db2) to generate the database; otherwise the step *Deploy new schema* below will fail with `SQLCODE=-286, SQLSTATE=42727`.  
 
-To create the PostgreSql database and database user, use the following commands:
+To create the PostgreSQL database and database user, use the following commands:
 
 ``` shell
 psql postgres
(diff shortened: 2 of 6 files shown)


fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/postgresql/PostgreSqlTranslator.java
```diff
@@ -113,7 +113,7 @@ public boolean isDuplicateSchema(SQLException x) {
 
     @Override
     public boolean isUndefinedName(SQLException x) {
-        return "42X05".equals(x.getSQLState());
```
prb112, line 116: Maybe this should be an OR with 

```
42883 | undefined_function
42P01 | undefined_table
42P02 | undefined_parameter
42704 | undefined_object
```

fhir-persistence-schema/README.md
```diff
@@ -85,7 +85,7 @@ The following sections include common values for `OPTIONS`.
 ```
 --prop-file db2.properties
 --schema-name FHIRDATA
---drop-schema
+--drop-schema-fhir
 --drop-admin
 --confirm-drop
 ```
```
prb112, line 91: maybe we should add all three in this example, or mention in a comment below this update that you can use --drop-schema-oauth and jbatch?

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
```diff
@@ -424,7 +423,7 @@ protected void dropSchema() {
                     JdbcTarget target = new JdbcTarget(c);
                     IDatabaseAdapter adapter = getDbAdapter(dbType, target);
 
-                    if (this.dropSchema) {
+                    if (dropFhirSchema || dropOauthSchema || dropJavaBatchSchema) {
```
prb112, line 426: is the adminDrop implicit here? what if we just call dropAdmin

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
```diff
@@ -1387,20 +1385,13 @@ protected void process() {
 
         if (addKeyForTenant != null) {
             addTenantKey();
-        } else if (this.dropSchema) {
+        } else if (this.dropAdmin || this.dropFhirSchema || this.dropJavaBatchSchema || this.dropOauthSchema) {
```
prb112, line 1388: See 426

prb112 (changes requested):
Left comments that I want to confirm before LGTM

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
lmsurpre: dropAdmin is covered separately in the if case just below this

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/postgresql/PostgreSqlTranslator.java
lmsurpre: empirically I found the `42704` is what we got back for undefined functions, tables, and indexes.  not sure how to reconcile that with the provided list, but its easy-enough for me to or them

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
punktilious: I think we should (later) revisit the whole option control logic in this Main...too much entropy.

fhir-persistence-schema/README.md
lmsurpre: sure, I can add the comment

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
prb112: +1 on Robin's comment.  it's definitely growing

punktilious (approved):
LGTM, subject to Paul's comments.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/postgresql/PostgreSqlTranslator.java
prb112: I wasn't sure if we should or not, it was more of a check-in, not necessarily that we had to or-then

prb112 (approved):

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/postgresql/PostgreSqlTranslator.java
lmsurpre: ok, i went ahead and added them

fhir-persistence-schema/README.md
lmsurpre: ended up adding two different subsections; one for dropping just the fhir schema and one for dropping them all

Issue #1351, opened 2020-07-22
Constrain Search based on SearchParameters and CapabilityStatements


prb112:
**Is your feature request related to a problem? Please describe.**
In the FHIR Search specification, there are complicating conditions for each [SearchParameter](https://www.hl7.org/fhir/r4/searchparameter.html) where multipleOr, multipleAnd, accepted comparators and chains. 

Also, one must consider profile specific constraints which limit the usage of particular codes and combinations of codes https://build.fhir.org/ig/HL7/carin-bb/search.html

**Describe the solution you'd like**

1. Update SearchUtil.java to enforce the following restrictions of a registered (or in extension-search-parameter.json) SearchParameter:

- multipleOr
- multipleAnd
- comparator
- modifier

2. Update SearchUtil.java to enforce the following restrictions of fhir-server-config.json:

- supported searches
- supported _includes
- supported _revincludes

**Describe alternatives you've considered**
Controlling these restrictions from fhir-server-config.json.  That can be done as follow-on work to this, in another GitHub issue, to amend the search restrictions enforced by this issue.

**Additional context**
None

kmbarton423:
Today had an Inferno search fail which was using "identifier" vs "id".  There is no enforcement of unique identifiers so in this situation, data for multiple patients was being retrieved by the search.  Search fails as unauthorized.  Not sure if there is anything programmatic that can be done here but this could be a very confusing.  Just something to consider when finalizing any interceptor behaviors.

prb112:
> Today had an Inferno search fail which was using "identifier" vs "id". There is no enforcement of unique identifiers so in this situation, data for multiple patients was being retrieved by the search. Search fails as unauthorized. Not sure if there is anything programmatic that can be done here but this could be a very confusing. Just something to consider when finalizing any interceptor behaviors.

HI Karen, I think that's a data quality issue.  Is there anything you need Troy or I to look at with regards to data retrieval and constraining the retrieval? Thanks, Paul

tbieste:
@prb112 
After reading through it again, trying to confirm my understanding if this is one requirement, or multiple things in one. If I'm off-base, I could use another call to talk through things about it.

> In the FHIR Search specification, there are complicating conditions for each SearchParameter where multipleOr, multipleAnd, accepted comparators and chains.
[...]

Issue #1272, opened 2020-06-23
Support conflict resolution for search parameter conflicts


lmsurpre:
**Is your feature request related to a problem? Please describe.**
For 4.3.0, I introduced a warning to surface search parameter conflicts (multiple parameters with the same code which are applicable to one or more of the same resource types).
For different definitions with the same code and the same expression, I log an info message.
For different definitions with the same code and a different expression, I log a warning message.

When I raised my concern about these conflicts on chat.fhir.org I was surprised by the response.
https://chat.fhir.org/#narrow/stream/179166-implementers/topic/Why.20do.20IGs.20redefine.20search.20parameters.20from.20the.20base.20spec.3F

Currently, there is no way for users to exclude the built-in search parameters other than to use a searchFilter in their fhir-server-config.  Unfortunately, because that filter config is done at the code level, it will exclude both the base spec SearchParameter AND the IG's SearchParameter.

We need to figure out how to let operators decide which search parameters we should use.

**Describe the solution you'd like**
I think we should update our searchFilter config to support filtering searchParameters by url 
instead of by code.

**Describe alternatives you've considered**
An alternative I considered is to move the base spec's SearchParameters out of the fhir-registry and move them to userlib and have the operator pick and chose which libraries to include.  That way, they could easily chose whether to include all base spec searchparameters or none of them; it would be up to the operator to package a set SearchParameters that don't conflict with one another.

tbieste:
Verified that testcases exist and run successfully, and successfully performed a search with a search parameter in the config file (with the new config file format), while a search with a search parameter not listed for the resource type failed (as expected).

PR #1608, opened 2020-10-20
issue #1272 - simplify searchParameter configuration

lmsurpre:
Initially we made each searchParameter its own object so that we could
add additional properties for issue #1351.
However, now we don't think that any additional fields are needed and so
we can simplify the config to be a simple map of SearchParameter  code
-> SearchParameter.url

This format will give us the flexibility of overriding a SearchParameter
code for a given resource type for a given deploy. However, currently
the code from the config is not used...it will only use the url and it
will auto-resolve any conflicting codes.


prb112 (approved):
LGTM, per our conversation this looks right

tbieste (approved):

tbieste:
>However, currently the code from the config is not used...it will only use the url and it
> will auto-resolve any conflicting codes.

I wonder if for clarity/consistency if the "code" was required to match the "code" of the SearchParameter pointed to by the URL. Perhaps we should add that in a follow-on PR.

lmsurpre:
> I wonder if for clarity/consistency if the "code" was required to match the "code" of the SearchParameter pointed to by the URL. Perhaps we should add that in a follow-on PR.

Good thought.  Eventually I don't want to require them to match because we want to enable overriding the one in the SearchParameter with the one in the config.  However, since thats a bigger lift, I think its worth seeing if there's an easy way to add the validation check you describe to this PR.  Looking...

PR #1610, opened 2020-10-20
issue #1272 - use new config approach while applying search param filter

lmsurpre:
also added security to the db2 and postgres sample configs

### DIFF
--- a/fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
+++ b/fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
@@ -198,7 +198,7 @@ public class SearchUtil {
         Map<String, ParametersMap> spBuiltin = ParametersUtil.getBuiltInSearchParametersMap();
 
         // Retrieve the current tenant's search parameter filtering rules.
-        Map<String, List<String>> filterRules = getFilterRules();
+        Map<String, Map<String, String>> filterRules = getFilterRules();
 
         // Retrieve the SPs associated with the specified resource type and filter per the filter rules.
         ParametersMap spMap = spBuiltin.get(resourceType);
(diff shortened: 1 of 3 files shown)


fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -270,19 +270,25 @@ public static void init() {
...
+                String code = sp.getCode().getValue();
+                String url = sp.getUrl().getValue();
 
-                    String url = sp.getUrl().getValue();
-                    if (includedSPs.contains(url)) {
+                if (includedSPs.containsKey(code)) {
+                    if (includedSPs.get(code).equals(url)) {
+                        results.add(sp);
+                    } else {
+                        log.info("Skipping search parameter with id='" + sp.getId() + "'. "
```
lmsurpre, line 284: had a hard time deciding whether this should be info-level or fine-level. opted for info-level since this behavior is new.
the issue with this is that we'll see these messages on each and every request that involves search parameters like this.

tbieste (approved):

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
prb112: Question: 
Executed on every search? if so, I'd say isLogLevel(fine)
If it's not, I'd say it's fine.

prb112 (approved):
LGTM

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
lmsurpre: I think it is executed for each search that targets a resource type with filtered parameters which match one of multiple possible target ids.  With our default config, we should never hit this.  The main motivator for keeping it info-level is that I think this will be the only log message you will get if you make an error while specifying the search parameter url in the config (so I don't want to hide that).
In the future, I think there is room for improving how we handle these search parameter filters.  For example, if we don't care about allowing dynamic updates to the config, we could check for all tenant configs during startup and validate the urls while we build a tenant-specific ParametersMap.

lmsurpre: Discussed with Paul and we think info level is ok for now given the above.

Issue #1612, opened 2020-10-22
Bulk Data $import and $export must inject the incomingURL to FHIRRequestContext


prb112:
Bulk Data $import and $export must inject the incomingURL to FHIRRequestContext using
    @Inject
    @BatchProperty(name = Constants.INCOMING_URL)
    String incomingUrl;

prb112:
Awaiting Review QA

PR #1613, opened 2020-10-22
Bulk Data $import and $export must inject the incomingURL to support Reference processing

prb112:
FHIRRequestContext #1612


lmsurpre (approved):
LGTM

Issue #1554, opened 2020-10-04
FHIR Server Startup - Bootstrapping Derby Error


ceekaay:
**Describe the bug**
I am using a Raspberry Pi 4 (4 GB) and followed the instructions in the blog entry (https://ibm.github.io/FHIR/blog/raspberrypi4howto). After a successful build and installing or using fhir-server-distribution.zip I get always the following error. I want to use the embedded derby db.

[console.log](https://github.com/IBM/FHIR/files/5324199/console.log)

[messages.log](https://github.com/IBM/FHIR/files/5324182/messages.log)

**To Reproduce**
1. Unzip fhir-server-distribution.zip in home directory
2. Run `./fhir-server-dist/install.sh  ~/fhir-server`
3. Edit the jvm.options file (-Xmx1024M and -Xms1024M)
4. Start server <WLP_HOME>/bin/server start fhir-server
5. Try to access https://<server.ip>:9443/fhir-server/api/v4/

**Expected behavior**
JSON Output with "ALL OK" message

**Additional context**
Raspberry Pi 4 (GB) - Raspberry Pi OS (32-bit) Lite (Debian Buster)
Tried Releases: 4.4.1 (own build) and 4.4.0 released version
Tried OpenJDK 8 and 11

prb112:
This error happens most frequently on memory constrained environments. 

```
[WARNING ] [ONCE] Not supported in Derby: ALTER TABLE APP.APPOINTMENT_STR_VALUES ALTER COLUMN ROW_ID SET CACHE 1000
[ERROR   ] DSRA0304E:  XAException occurred. XAException contents and details are: .
[ERROR   ] DSRA0302E:  XAException occurred.  Error code is: XA_RBROLLBACK (100).  Exception is: null
[ERROR   ] Encountered an exception while bootstrapping the FHIR database
java.sql.SQLNonTransientConnectionException: No current connection.
[ERROR   ] transaction rollback failed
No Global Transaction exists to commit.
[ERROR   ] Encountered an exception while initializing the servlet context.
javax.transaction.RollbackException
[ERROR   ] SRVE0283E: Exception caught while initializing context: java.lang.RuntimeException: Encountered an exception while initializing the servlet context.
	at com.ibm.fhir.server.listener.FHIRServletContextListener.contextInitialized(FHIRServletContextListener.java:199)
	at com.ibm.ws.webcontainer.webapp.WebApp.notifyServletContextCreated(WebApp.java:2433)
	at [internal classes]
[...]

PR #1642, opened 2020-10-28
#1554 - Set error response to not-found on missing code or system

csandersdev:
I updated the LookupOperation code so that a new type of FHIROperationException is thrown
with the code **not-found** when the code system lookup returns no result.
This will cause a 404 instead of a 400 error as detailed in the issue ticket and the referenced FHIR
document, but that is consistent with the behavior of the HAPI FHIR server, so it feels ok to me. 404
means NOT_FOUND vs. BAD_REQUEST which doesn't quite ring true.


prb112 (approved):
LGTM

prb112:
Thanks - the change looks good.  I'll refer to John and his expertise here.

JohnTimm (approved):
LGTM

lmsurpre:
@CoreySandersOH thanks for the pull request!  We like the change to set the error-code on the response body, but we discussed it as a team and we're thinking we'd prefer to keep the HTTP status code as a 400 (as opposed to changing it to a 404).
We didn't feel super strong about this, but I prefer the 400 because to me a POST to a 404 might imply that we got the endpoint wrong whereas in this case the endpoint is right and you just sent invalid parameters.  Combined with the fact we already return a 400, we think leaving that behavior more important than matching HAPI's (since spec doesn't seem to offer much advise on this case).
As you know, the default mapping for a not-found code is a 404, but we already have a pattern for overriding that default mapping.  Basically, we'd add an extension to the `not-found` code to indicate what it was that wasn't found.  Then, we'd look for that extension and choose a different code in the IssueTypeToHttpStatusMapper when we find it.
You can find an example of this in `IssueTypeToHttpStatusMapper.issueListToStatus(List<Issue> issues)` for the NOT_SUPPORTED code.

lmsurpre (commented):
see previous comment

csandersdev:
Just to be clear, the existing behavior without this pull request is not to send back 400, it is 500.  My goal was to fix that and make it clearer. The FHIR doc certainly recommends 400, but when I thought about what 404 would mean from an REST API standpoint, it is the same behavior if you attempted an action for a patient ID that is not in the system and that made sense to me. The fact that HAPI also sends back a 404 was enough to cement it in my mind. I can certainly go back and do the work to make it 400 if you guys want. I had already started down that path before I decided that 404 was just the right behavior and it wasn't worth the extra cycles to make it 400. Let me see what I can do.

csandersdev:
I was able to use the suggested extension strategy to change the response code, but the operation outcome renders that extension and is duplicate information in the response and in my opinion makes it less clear.

fhir-server-test/src/test/java/testng.xml
```diff
@@ -58,4 +58,10 @@
             <package name="com.ibm.fhir.server.test.performance" />
         </packages>
     </test>
-</suite>
\ No newline at end of file
+    <test name="TerminologyTests">
+        <packages>
+            <!-- All of these terminology tests are disabled by default, and enabled in test.properties -->
```
prb112, line 63: is this actually the case?  I don't think it can be disabled.

lmsurpre: good catch.  originally corey wrote it so that it can be enabled/disabled from test.properties.  I requested him to remove that since the terminology stuff gets baked into the server...there should be no need to disable them in this way
```suggestion
            <!-- All of these tests are automatically executed -->
```

lmsurpre (approved):
LGTM

lmsurpre:
I had to improve our windows workflow to get the maven output so that I could see that the build was crashing due to resource issues related to us accidentally committing the test.properties change for logging each request/response body.  That should be fixed now and this one is ready-to-rock.

csandersdev:
FYI, the final format of the response is ``HTTP 400`` with a body that looks like this...

```json
{
    "id": "ac-11-0-2-704ca5d6-fccc-46e6-bc3b-aea23d80904c",
    "issue": [
        {
            "severity": "error",
            "code": "not-found",
            "_code": {
                "extension": [
                    {
                        "url": "http://ibm.com/fhir/extension/not-found-detail",
                        "valueCode": "coding"
                    }
                ]
            },
            "details": {
                "text": "Code \"INVALID\" in System \"http://hl7.org/fhir/CodeSystem/example\" not found"
            }
        }
    ]
}
```

Issue #1218, opened 2020-06-10
reverse chaining (_has parameter)


prb112:
**Is your feature request related to a problem? Please describe.**
 reverse chaining (_has parameter)

**Describe the solution you'd like**
This feature was introduced to FHIR Searchin STU3. Its very useful for implementing queries like "find me all the patients that have a condition with a code of xyz".

https://www.hl7.org/fhir/search.html#has

the format is like: `Patient?_has:Condition:patient:code=xyz`

**Describe alternatives you've considered**
None, standards based

**Additional context**
Original Git Issue  `reverse chaining (_has parameter) #458`

lmsurpre:
I've deployed these changes and tested reverse chaining with the following queries like the following:
`get(base + '/Patient?gender=female' + '&' + '_has:Condition:patient:code=http://snomed.info/sct|44054006' + '&_count=1')`

PR #1666, opened 2020-11-03
Issue #1218 - enable reverse chain (_has) search

michaelwschroeder:

### DIFF
--- a/docs/src/pages/Conformance.md
+++ b/docs/src/pages/Conformance.md
@@ -113,8 +113,8 @@ In addition, the following search parameters are supported on all resources:
 
 These parameters can be used while searching any single resource type or while searching across resource types (whole system search).
 The `_type` parameter is special in that it is only applicable for whole system search. The `_has` parameter has two restrictions:
-It cannot be used in combination with the `_type` parameter, and the search parameter specified at the end of its chain cannot be
-a search result parameter.
+* `_has` cannot be used in combination with the `_type` parameter.
+* The search parameter specified at the end of its chain cannot be a search result parameter.
 
 The `_text`, `_content`, `_list`, `_query`, and `_filter` parameters are not supported at this time.
 
--- a/docs/src/pages/Conformance.md
+++ b/docs/src/pages/Conformance.md
@@ -2,7 +2,7 @@
 layout: post
 title:  Conformance
 description: Notes on the Conformance of the IBM FHIR Server
-date:   2020-10-05 01:00:00 -0400
+date:   2020-11-03 01:00:00 -0400
 permalink: /conformance/
 ---
 
@@ -109,11 +109,14 @@ In addition, the following search parameters are supported on all resources:
 * `_security`
 * `_source`
 * `_type`
+* `_has`
 
 These parameters can be used while searching any single resource type or while searching across resource types (whole system search).
-The `_type` parameter is special in that it is only applicable for whole system search.
+The `_type` parameter is special in that it is only applicable for whole system search. The `_has` parameter has two restrictions:
+It cannot be used in combination with the `_type` parameter, and the search parameter specified at the end of its chain cannot be
+a search result parameter.
 
-The `_text`, `_content`, `_list`, `_has`, `_query`, and `_filter` parameters are not supported at this time.
+The `_text`, `_content`, `_list`, `_query`, and `_filter` parameters are not supported at this time.
 
 Finally, the specification defines a set of "Search result parameters" for controlling the search behavior. The IBM FHIR Server supports the following:
 * `_sort`
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/JDBCConstants.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/JDBCConstants.java
@@ -36,8 +36,8 @@ public class JDBCConstants {
     public static final String DATE_END = "DATE_END";
     public static final String LATITUDE_VALUE = "LATITUDE_VALUE";
     public static final String LONGITUDE_VALUE = "LONGITUDE_VALUE";
-    public static final String RESOURCES = "_RESOURCES";
-    public static final String LOGICAL_RESOURCES = "_LOGICAL_RESOURCES";
+    public static final String _RESOURCES = "_RESOURCES";
+    public static final String _LOGICAL_RESOURCES = "_LOGICAL_RESOURCES";
     public static final String RESOURCE_ID = "RESOURCE_ID";
     public static final String LOGICAL_ID = "LOGICAL_ID";
     public static final String LOGICAL_RESOURCE_ID = "LOGICAL_RESOURCE_ID";
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/JDBCConstants.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/JDBCConstants.java
@@ -36,6 +36,14 @@ public class JDBCConstants {
     public static final String DATE_END = "DATE_END";
     public static final String LATITUDE_VALUE = "LATITUDE_VALUE";
     public static final String LONGITUDE_VALUE = "LONGITUDE_VALUE";
+    public static final String RESOURCES = "_RESOURCES";
+    public static final String LOGICAL_RESOURCES = "_LOGICAL_RESOURCES";
+    public static final String RESOURCE_ID = "RESOURCE_ID";
+    public static final String LOGICAL_ID = "LOGICAL_ID";
+    public static final String LOGICAL_RESOURCE_ID = "LOGICAL_RESOURCE_ID";
+    public static final String CURRENT_RESOURCE_ID = "CURRENT_RESOURCE_ID";
+    public static final String PARAMETER_NAME_ID = "PARAMETER_NAME_ID";
+    public static final String IS_DELETED_NO = "IS_DELETED = 'N'";
 
     // Generic SQL query string constants
     public static final String DOT = ".";
@@ -68,6 +76,8 @@ public class JDBCConstants {
     public static final char COMMA_CHAR = ',';
     public static final char QUOTE = '\'';
     public static final char PATH_CHAR = '/';
+    public static final String SELECT = " SELECT ";
+    public static final String AS = " AS ";
 
     // JDBC Operators
     public static final String EQ = " = ";
@@ -92,7 +102,7 @@ public class JDBCConstants {
     // MIN / MAX
     public static final String MAX = "MAX";
     public static final String MIN = "MIN";
-    
+
     // Db2 optimization hints
     public static final String SEARCH_REOPT = "search.reopt";
 
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
@@ -25,7 +25,6 @@ import static com.ibm.fhir.persistence.jdbc.JDBCConstants.JOIN;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.LEFT_PAREN;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.LIKE;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.LOGICAL_ID;
-import static com.ibm.fhir.persistence.jdbc.JDBCConstants.LOGICAL_RESOURCES;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.LOGICAL_RESOURCE_ID;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.MAX_NUM_OF_COMPOSITE_COMPONENTS;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.NE;
@@ -35,7 +34,6 @@ import static com.ibm.fhir.persistence.jdbc.JDBCConstants.OR;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.PARAMETER_NAME_ID;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.PARAMETER_TABLE_ALIAS;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.PERCENT_WILDCARD;
-import static com.ibm.fhir.persistence.jdbc.JDBCConstants.RESOURCES;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.RESOURCE_ID;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.RIGHT_PAREN;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.SELECT;
@@ -45,6 +43,8 @@ import static com.ibm.fhir.persistence.jdbc.JDBCConstants.STR_VALUE_LCASE;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.TOKEN_VALUE;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.UNDERSCORE_WILDCARD;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.WHERE;
+import static com.ibm.fhir.persistence.jdbc.JDBCConstants._LOGICAL_RESOURCES;
+import static com.ibm.fhir.persistence.jdbc.JDBCConstants._RESOURCES;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.modifierOperatorMap;
 
 import java.sql.Timestamp;
@@ -809,8 +809,8 @@ public class JDBCQueryBuilder extends AbstractQueryBuilder<SqlQueryData> {
 
         // Build this piece: FROM Device_RESOURCES CR1, Device_LOGICAL_RESOURCES CLR1, Device_TOKEN_VALUES_V CP1 WHERE
         whereClauseSegment.append(FROM)
-                .append(resourceTypeName).append(RESOURCES).append(SPACE).append(chainedResourceVar).append(COMMA)
-                .append(resourceTypeName).append(LOGICAL_RESOURCES).append(SPACE).append(chainedLogicalResourceVar);
+                .append(resourceTypeName).append(_RESOURCES).append(SPACE).append(chainedResourceVar).append(COMMA)
+                .append(resourceTypeName).append(_LOGICAL_RESOURCES).append(SPACE).append(chainedLogicalResourceVar);
 
         // If we're dealing with anything other than id, then proceed to add the parameters table.
         if (currentParm.getNextParameter() != null && !"_id".equals(currentParm.getNextParameter().getCode())) {
@@ -1478,7 +1478,7 @@ public class JDBCQueryBuilder extends AbstractQueryBuilder<SqlQueryData> {
             StringBuilder whereClauseSegment = new StringBuilder();
 
             if (parmIndex == 0) {
-                // BUild outer select:
+                // Build outer select:
                 // @formatter:off
                 //   SELECT CLR0.LOGICAL_ID
                 //     FROM <resource-type>_LOGICAL_RESOURCES AS CLR0
@@ -1487,8 +1487,8 @@ public class JDBCQueryBuilder extends AbstractQueryBuilder<SqlQueryData> {
                 //     WHERE
                 // @formatter:on
                 selectSegments.append(SELECT).append(prevChainedLogicalResourceVar).append(DOT).append(LOGICAL_ID)
-                                .append(FROM).append(resourceType.getSimpleName()).append(LOGICAL_RESOURCES).append(AS).append(prevChainedLogicalResourceVar)
-                                .append(JOIN).append(resourceType.getSimpleName()).append(RESOURCES).append(AS).append(prevChainedResourceVar)
+                                .append(FROM).append(resourceType.getSimpleName()).append(_LOGICAL_RESOURCES).append(AS).append(prevChainedLogicalResourceVar)
+                                .append(JOIN).append(resourceType.getSimpleName()).append(_RESOURCES).append(AS).append(prevChainedResourceVar)
                                 .append(ON).append(prevChainedResourceVar).append(DOT).append(RESOURCE_ID).append(EQ)
                                 .append(prevChainedLogicalResourceVar).append(DOT).append(CURRENT_RESOURCE_ID)
                                 .append(AND).append(prevChainedResourceVar).append(DOT ).append(IS_DELETED_NO)
@@ -1508,10 +1508,10 @@ public class JDBCQueryBuilder extends AbstractQueryBuilder<SqlQueryData> {
                     // @formatter:on
                     selectSegments.append(EXISTS).append(LEFT_PAREN).append("SELECT 1")
                                     .append(FROM).append(currentParm.getModifierResourceTypeName()).append("_TOKEN_VALUES_V").append(AS).append(chainedParmVar)
-                                    .append(JOIN).append(currentParm.getModifierResourceTypeName()).append(LOGICAL_RESOURCES).append(AS).append(chainedLogicalResourceVar)
+                                    .append(JOIN).append(currentParm.getModifierResourceTypeName()).append(_LOGICAL_RESOURCES).append(AS).append(chainedLogicalResourceVar)
                                     .append(ON).append(chainedLogicalResourceVar).append(DOT).append(LOGICAL_RESOURCE_ID).append(EQ)
                                     .append(chainedParmVar).append(DOT).append(LOGICAL_RESOURCE_ID)
-                                    .append(JOIN).append(currentParm.getModifierResourceTypeName()).append(RESOURCES).append(AS).append(chainedResourceVar)
+                                    .append(JOIN).append(currentParm.getModifierResourceTypeName()).append(_RESOURCES).append(AS).append(chainedResourceVar)
                                     .append(ON).append(chainedResourceVar).append(DOT).append(RESOURCE_ID).append(EQ)
                                     .append(chainedLogicalResourceVar).append(DOT).append(CURRENT_RESOURCE_ID)
                                     .append(AND).append(chainedResourceVar).append(DOT).append(IS_DELETED_NO);
@@ -1527,7 +1527,7 @@ public class JDBCQueryBuilder extends AbstractQueryBuilder<SqlQueryData> {
                         // @formatter:off
                         //   WHERE CPx.TOKEN_VALUE = CLR<x-1>.LOGICAL_ID
                         //     AND CPx.PARAMETER_NAME_ID = <parm-name-id>
-                        //     AND CPx.CDE_SYSTEM_ID = <code-system-id>
+                        //     AND CPx.CODE_SYSTEM_ID = <code-system-id>
                         //     AND
                         // @formatter:on
                         selectSegments.append(WHERE).append(chainedParmVar).append(DOT).append(TOKEN_VALUE).append(EQ)
@@ -1539,7 +1539,7 @@ public class JDBCQueryBuilder extends AbstractQueryBuilder<SqlQueryData> {
                         // @formatter:off
                         //   WHERE CPx.TOKEN_VALUE = CLR<x-1>.LOGICAL_ID
                         //     AND CPx.PARAMETER_NAME_ID = <parm-name-id>
-                        //     AND CPx.CDE_SYSTEM_ID = <code-system-id>
+                        //     AND CPx.CODE_SYSTEM_ID = <code-system-id>
                         // @formatter:on
                         whereClauseSegment.append(WHERE).append(chainedParmVar).append(DOT).append(TOKEN_VALUE).append(EQ)
                                             .append(prevChainedLogicalResourceVar).append(DOT).append(LOGICAL_ID).append(AND);
@@ -1602,7 +1602,7 @@ public class JDBCQueryBuilder extends AbstractQueryBuilder<SqlQueryData> {
                         //     ON CLRx.LOGICAL_RESOURCE_ID = CLR<x-1>.LOGICAL_RESOURCE_ID
                         //     AND
                         // @formatter:on
-                        whereClauseSegment.append(JOIN).append(previousParm.getModifierResourceTypeName()).append(LOGICAL_RESOURCES)
+                        whereClauseSegment.append(JOIN).append(previousParm.getModifierResourceTypeName()).append(_LOGICAL_RESOURCES)
                                             .append(AS).append(chainedLogicalResourceVar).append(ON)
                                             .append(chainedLogicalResourceVar).append(DOT).append(LOGICAL_RESOURCE_ID)
                                             .append(EQ).append(prevChainedLogicalResourceVar).append(DOT).append(LOGICAL_RESOURCE_ID).append(AND);
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
@@ -7,9 +7,11 @@
 package com.ibm.fhir.persistence.jdbc.util;
 
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.AND;
+import static com.ibm.fhir.persistence.jdbc.JDBCConstants.AS;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.BIND_VAR;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.CODE_SYSTEM_ID;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.COMMA;
+import static com.ibm.fhir.persistence.jdbc.JDBCConstants.CURRENT_RESOURCE_ID;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.DOT;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.EQ;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.ESCAPE_EXPR;
@@ -18,17 +20,26 @@ import static com.ibm.fhir.persistence.jdbc.JDBCConstants.ESCAPE_UNDERSCORE;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.EXISTS;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.FROM;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.IN;
+import static com.ibm.fhir.persistence.jdbc.JDBCConstants.IS_DELETED_NO;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.JOIN;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.LEFT_PAREN;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.LIKE;
+import static com.ibm.fhir.persistence.jdbc.JDBCConstants.LOGICAL_ID;
+import static com.ibm.fhir.persistence.jdbc.JDBCConstants.LOGICAL_RESOURCES;
+import static com.ibm.fhir.persistence.jdbc.JDBCConstants.LOGICAL_RESOURCE_ID;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.MAX_NUM_OF_COMPOSITE_COMPONENTS;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.NE;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.NOT;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.ON;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.OR;
+import static com.ibm.fhir.persistence.jdbc.JDBCConstants.PARAMETER_NAME_ID;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.PARAMETER_TABLE_ALIAS;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.PERCENT_WILDCARD;
+import static com.ibm.fhir.persistence.jdbc.JDBCConstants.RESOURCES;
+import static com.ibm.fhir.persistence.jdbc.JDBCConstants.RESOURCE_ID;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.RIGHT_PAREN;
+import static com.ibm.fhir.persistence.jdbc.JDBCConstants.SELECT;
+import static com.ibm.fhir.persistence.jdbc.JDBCConstants.SPACE;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.STR_VALUE;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.STR_VALUE_LCASE;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.TOKEN_VALUE;
@@ -114,6 +125,11 @@ public class JDBCQueryBuilder extends AbstractQueryBuilder<SqlQueryData> {
 
     // Hints to use for certain queries
     private final QueryHints queryHints;
+
+    // Table alias prefixes
+    private static final String CR = "CR";
+    private static final String CLR = "CLR";
+    private static final String CP = "CP";
 
     /**
      * Public constructor
@@ -629,9 +645,6 @@ public class JDBCQueryBuilder extends AbstractQueryBuilder<SqlQueryData> {
         final String METHODNAME = "processChainedReferenceParm";
         log.entering(CLASSNAME, METHODNAME, queryParm.toString());
 
-        final String CR = "CR";
-        final String CLR = "CLR";
-        final String CP = "CP";
         QueryParameter currentParm;
         int refParmIndex = 0;
         String chainedResourceVar = null;
@@ -745,7 +758,7 @@ public class JDBCQueryBuilder extends AbstractQueryBuilder<SqlQueryData> {
         List<Object> bindVariables = new ArrayList<>();
 
         whereClauseSegment
-            .append(chainedParmVar.replace("CP", "CLR")).append(DOT).append("LOGICAL_ID").append(" IN (");
+            .append(chainedParmVar.replace("CP", "CLR")).append(DOT).append(LOGICAL_ID).append(" IN (");
 
         List<QueryParameterValue> vals = currentParm.getValues();
         boolean add = false;
@@ -766,7 +779,7 @@ public class JDBCQueryBuilder extends AbstractQueryBuilder<SqlQueryData> {
     private void appendMidChainParm(StringBuilder whereClauseSegment, QueryParameter currentParm, String chainedParmVar)
             throws FHIRPersistenceDBConnectException, FHIRPersistenceDataAccessException, FHIRPersistenceException {
         Integer parameterNameId = identityCache.getParameterNameId(currentParm.getCode());
-        whereClauseSegment.append(chainedParmVar).append(DOT).append("PARAMETER_NAME_ID").append(EQ)
+        whereClauseSegment.append(chainedParmVar).append(DOT).append(PARAMETER_NAME_ID).append(EQ)
                 .append(parameterNameId);
 
         // TODO handle code system lookup failures here. Perhaps = NULL to make the query return no rows?
@@ -774,7 +787,7 @@ public class JDBCQueryBuilder extends AbstractQueryBuilder<SqlQueryData> {
         if (codeSystemName != null && !codeSystemName.equals("*")) {
             Integer codeSystemId = identityCache.getCodeSystemId(codeSystemName);
             if (codeSystemId != null) {
-                whereClauseSegment.append(AND).append(chainedParmVar + DOT).append(CODE_SYSTEM_ID).append(EQ)
+                whereClauseSegment.append(AND).append(chainedParmVar).append(DOT).append(CODE_SYSTEM_ID).append(EQ)
                         .append(codeSystemId);
             }
         }
@@ -784,24 +797,24 @@ public class JDBCQueryBuilder extends AbstractQueryBuilder<SqlQueryData> {
 
     private void appendInnerSelect(StringBuilder whereClauseSegment, QueryParameter currentParm,
             String resourceTypeName, String chainedResourceVar, String chainedLogicalResourceVar, String chainedParmVar) {
-        String chainedResourceTableAlias = chainedResourceVar + ".";
-        String chainedLogicalResourceTableAlias = chainedLogicalResourceVar + ".";
-        String chainedParmTableAlias = chainedParmVar + ".";
+        String chainedResourceTableAlias = chainedResourceVar + DOT;
+        String chainedLogicalResourceTableAlias = chainedLogicalResourceVar + DOT;
+        String chainedParmTableAlias = chainedParmVar + DOT;
 
         // Build this piece: SELECT 'resource-type-name' || '/' || CLRx.LOGICAL_ID
         // Note since #1366, we no longer need to prepend the resourceTypeName
-        whereClauseSegment.append("SELECT ").append(chainedLogicalResourceTableAlias).append("LOGICAL_ID");
+        whereClauseSegment.append(SELECT).append(chainedLogicalResourceTableAlias).append(LOGICAL_ID);
 
         QueryParameter nextParameter = currentParm.getNextParameter();
 
         // Build this piece: FROM Device_RESOURCES CR1, Device_LOGICAL_RESOURCES CLR1, Device_TOKEN_VALUES_V CP1 WHERE
         whereClauseSegment.append(FROM)
-                .append(resourceTypeName).append("_RESOURCES ").append(chainedResourceVar).append(", ")
-                .append(resourceTypeName).append("_LOGICAL_RESOURCES ").append(chainedLogicalResourceVar);
+                .append(resourceTypeName).append(RESOURCES).append(SPACE).append(chainedResourceVar).append(COMMA)
+                .append(resourceTypeName).append(LOGICAL_RESOURCES).append(SPACE).append(chainedLogicalResourceVar);
 
         // If we're dealing with anything other than id, then proceed to add the parameters table.
         if (currentParm.getNextParameter() != null && !"_id".equals(currentParm.getNextParameter().getCode())) {
-            whereClauseSegment.append(", ")
+            whereClauseSegment.append(COMMA)
                 .append(QuerySegmentAggregator.tableName(resourceTypeName, currentParm.getNextParameter()))
                 .append(chainedParmVar);
         }
@@ -823,19 +836,19 @@ public class JDBCQueryBuilder extends AbstractQueryBuilder<SqlQueryData> {
             }
         }
 
-        whereClauseSegment.append(" WHERE ");
+        whereClauseSegment.append(WHERE);
 
         // CR1.RESOURCE_ID = CLR1.CURRENT_RESOURCE_ID AND CR1.IS_DELETED <> 'Y' AND
         // CP1.LOGICAL_RESOURCE_ID = CLR1.LOGICAL_RESOURCE_ID AND
-        whereClauseSegment.append(chainedResourceTableAlias).append("RESOURCE_ID = ")
-                .append(chainedLogicalResourceTableAlias).append("CURRENT_RESOURCE_ID")
+        whereClauseSegment.append(chainedResourceTableAlias).append(RESOURCE_ID).append(EQ)
+                .append(chainedLogicalResourceTableAlias).append(CURRENT_RESOURCE_ID)
                 .append(AND)
                 .append(chainedResourceTableAlias)
-                .append("IS_DELETED").append(" <> 'Y'")
+                .append(IS_DELETED_NO)
                 .append(AND);
         if (currentParm.getNextParameter() != null && !"_id".equals(currentParm.getNextParameter().getCode())) {
-            whereClauseSegment.append(chainedParmTableAlias).append("LOGICAL_RESOURCE_ID = ")
-                .append(chainedResourceTableAlias).append("LOGICAL_RESOURCE_ID")
+            whereClauseSegment.append(chainedParmTableAlias).append(LOGICAL_RESOURCE_ID).append(EQ)
+                .append(chainedResourceTableAlias).append(LOGICAL_RESOURCE_ID)
                 .append(AND);
         }
     }
@@ -1293,7 +1306,7 @@ public class JDBCQueryBuilder extends AbstractQueryBuilder<SqlQueryData> {
         parameterNameId = identityCache.getParameterNameId(queryParmName);
 
         whereClauseSegment.append(LEFT_PAREN);
-        whereClauseSegment.append(parameterTableAlias + DOT).append("PARAMETER_NAME_ID=")
+        whereClauseSegment.append(parameterTableAlias + DOT).append(PARAMETER_NAME_ID).append(EQ)
                 .append(nullCheck(parameterNameId));
 
         log.exiting(CLASSNAME, METHODNAME);
@@ -1378,7 +1391,7 @@ public class JDBCQueryBuilder extends AbstractQueryBuilder<SqlQueryData> {
 
         whereClauseSegment.append("(SELECT 1 FROM " + valuesTable + WHERE);
         this.populateNameIdSubSegment(whereClauseSegment, queryParm.getCode(), valuesTable.toString());
-        whereClauseSegment.append(" AND " + valuesTable + ".LOGICAL_RESOURCE_ID = R.LOGICAL_RESOURCE_ID"); // correlate the [NOT] EXISTS subquery
+        whereClauseSegment.append(AND).append(valuesTable).append(".LOGICAL_RESOURCE_ID = R.LOGICAL_RESOURCE_ID"); // correlate the [NOT] EXISTS subquery
         whereClauseSegment.append(RIGHT_PAREN).append(RIGHT_PAREN);
 
         List<Object> bindVariables = new ArrayList<>();
@@ -1433,9 +1446,6 @@ public class JDBCQueryBuilder extends AbstractQueryBuilder<SqlQueryData> {
         final String METHODNAME = "processReverseChainedReferenceParm";
         log.entering(CLASSNAME, METHODNAME, queryParm.toString());
 
-        final String CR = "CR";
-        final String CLR = "CLR";
-        final String CP = "CP";
         String prevChainedResourceVar = null;
         String prevChainedLogicalResourceVar = null;
         String chainedResourceVar = null;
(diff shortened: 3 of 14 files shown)


docs/src/pages/Conformance.md
```diff
@@ -109,11 +109,14 @@ In addition, the following search parameters are supported on all resources:
 * `_security`
 * `_source`
 * `_type`
+* `_has`
 
 These parameters can be used while searching any single resource type or while searching across resource types (whole system search).
-The `_type` parameter is special in that it is only applicable for whole system search.
+The `_type` parameter is special in that it is only applicable for whole system search. The `_has` parameter has two restrictions:
+It cannot be used in combination with the `_type` parameter, and the search parameter specified at the end of its chain cannot be
+a search result parameter.
```
prb112, line 117:
```suggestion

* `_has` cannot be used in combination with the `_type` parameter.
* The search parameter specified at the end of its chain cannot be a search result parameter.
```

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
```diff
@@ -555,20 +555,35 @@ protected void buildWhereClause(StringBuilder whereClause, String overrideType)
...
+                            // Join on a select from logical resource table
+                            //   JOIN (
+                            //     SELECT CLR0.LOGICAL_ID FROM Observation_LOGICAL_RESOURCES AS CLR0
+                            //       ...
+                            //   ) AS param0 ON LR.LOGICAL_ID = param0.LOGICAL_ID
+                            whereClause.append(JOIN)
+                                        .append(LEFT_PAREN);
+                            whereClause.append(querySegment.getQueryString());
+                            whereClause.append(RIGHT_PAREN)
+                                        .append(" AS " + paramTableAlias)
```
prb112, line 569: Just as a form, since we are using a StringBuilder let's build the string with it. 

```suggestion
                                        .append(" AS ").append(paramTableAlias)
```

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
```diff
@@ -555,20 +555,35 @@ protected void buildWhereClause(StringBuilder whereClause, String overrideType)
...
+                                        .append(ON)
+                                        .append("LR.LOGICAL_ID = " + paramTableAlias + ".LOGICAL_ID");
```
prb112, line 571: Just as a form, since we are using a StringBuilder let's build the string with it.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchReverseChainTest.java
```diff
@@ -0,0 +1,1034 @@
...
+        // Get the location's logical id value.
+        locationId = getLocationLogicalId(response);
+
+        // Next, call the 'read' API to retrieve the new Location and verify it.
+        response   = target.path("Location/" + locationId).request(FHIRMediaType.APPLICATION_FHIR_JSON).get();
+        assertResponse(response, Response.Status.OK.getStatusCode());
+    }
+
+    @AfterClass
+    public void testDeleteResources() {
```
prb112, line 347:
```suggestion
    public void testDeleteResources() {
```

not a test per say?

fhir-search/src/main/java/com/ibm/fhir/search/uri/UriBuilder.java
```diff
@@ -279,4 +281,38 @@ private void appendChainedParm(QueryParameter param, StringBuilder returnString)
...
+                            .append(SearchConstants.COLON_DELIMITER)
+                            .append(revChainParam.getModifierResourceTypeName())
+                            .append(SearchConstants.COLON_DELIMITER)
+                            .append(revChainParam.getCode())
+                            .append(SearchConstants.COLON_DELIMITER);
+            } else if (revChainParam.isChained()) {
+                appendChainedParm(revChainParam, returnString);
+                returnString.append(SearchConstants.CHAINED_PARAMETER_CHARACTER);
+            } else {
+                appendNormalParameter(revChainParam, returnString);
```
prb112, line 313: If I follow... 

1st part
_has:Patient:patient:

It's always going to run the param.getChain (it'll have at least one entry)
Then loops... 

OK - this makes sense to me - I just needed to write this out, leaving here for others as they review.

fhir-search/src/test/java/com/ibm/fhir/search/test/RevChainParameterParseTest.java
```diff
@@ -0,0 +1,777 @@
...
+        queryParameters.put("_has:BadType:subject:code", Collections.singletonList("xxx"));
+        SearchUtil.parseQueryParameters(resourceType, queryParameters);
+    }
+
+    @Test(expectedExceptions = FHIRSearchException.class)
+    public void testReverseChainBadReferenceSearchParmException() throws Exception {
+        Map<String, List<String>> queryParameters = new HashMap<>();
+        Class<Patient> resourceType = Patient.class;
+
+        queryParameters.put("_has:Procedure:badSearchParm:code", Collections.singletonList("xxx"));
```
prb112, line 92: good test

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -756,20 +772,16 @@ public static FHIRSearchContext parseQueryParameters(Class<?> resourceType,
                           // Get the search parameter from our filtered set of applicable SPs for this resource type.
                           searchParameter = getSearchParameter(resType, parameterCode);
                           if (searchParameter == null) {
-                              String msg =
-                                      "Search parameter '" + parameterCode + "' for resource type '"
-                                              + resType + "' was not found.";
-                              throw SearchExceptionUtil.buildNewInvalidSearchException(msg);
+                              throw SearchExceptionUtil.buildNewInvalidSearchException(
+                                  String.format(SEARCH_PARAMETER_NOT_FOUND, parameterCode, resType));
```
prb112, line 776: slight comment

In Exception cases String.format, we as a team deem it to be OK, but it is actually slower than assembling a string.  (I mention this as John and I had a deep discussion about StringBuilding performance when I made a similar change in this class).

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -1589,6 +1605,168 @@ private static QueryParameter parseChainedParameter(Class<?> resourceType, Strin
...
 
+    /**
+     * Transforms the passed string representing reverse chain search criteria, into
+     * an actual chain of QueryParameter objects. This method consumes strings of this form:
+     * <pre>
+     *      +-------------------------------------------------------------------+
+     *      |                                                                   |
+     *      V                                                                   |
+     * >>---+--- "_has:{referenced-by-resource-type}:{reference-parameter}:" ---+--- "{search-parameter}" ---><
+     * </pre>
```
prb112, line 1616: mentioning this as it's the Eclipse Formatter control (turns on/off) 

```
@formatter:off 
@formatter:on
```

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -1370,6 +1382,10 @@ public static boolean isChainedParameter(String name) {
         return name.contains(SearchConstants.CHAINED_PARAMETER_CHARACTER);
     }
 
+    public static boolean isReverseChainedParameter(String name) {
```
prb112, line 1385: It's actually a code. 
```suggestion
    public static boolean isReverseChainedParameter(String code) {
```

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -1370,6 +1382,10 @@ public static boolean isChainedParameter(String name) {
...
+        return name.startsWith(SearchConstants.HAS);
```
prb112, line 1386:
```suggestion
        return code.startsWith(SearchConstants.HAS);
```

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -1589,6 +1605,168 @@ private static QueryParameter parseChainedParameter(Class<?> resourceType, Strin
...
+     *          The root of a parameter chain for the reverse chain criteria.
+     */
+    private static QueryParameter parseReverseChainedParameter(Class<?> resourceType, String reverseChainParameterString, String valuesString) throws Exception {
+
+        QueryParameter rootParameter = null;
+
+        try {
+            // Strip leading '_has:' and then split by ':_has:'
+            List<String> components = Arrays.asList(reverseChainParameterString
+                .replaceFirst(SearchConstants.HAS + SearchConstants.COLON_DELIMITER_STR, "")
```
prb112, line 1636: I think substring might be preferred here instead of replaceFirst. it'll always be a fixed length.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -1589,6 +1605,168 @@ private static QueryParameter parseChainedParameter(Class<?> resourceType, Strin
...
+                .split(SearchConstants.COLON_DELIMITER_STR + SearchConstants.HAS + SearchConstants.COLON_DELIMITER_STR));
```
prb112, line 1637: I'm almost positive in modern compilers this ends up getting inlined, however, it's worth making your own static final constant.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -1589,6 +1605,168 @@ private static QueryParameter parseChainedParameter(Class<?> resourceType, Strin
...
+                }
+
+                // Validate reference search parameter
+                String referenceSearchParameterName = subcomponents.get(1);
+                SearchParameter referenceSearchParameter = getSearchParameter(referencedByResourceType, referenceSearchParameterName);
+                if (referenceSearchParameter == null) {
+                    throw SearchExceptionUtil.buildNewInvalidSearchException(
+                        String.format(SEARCH_PARAMETER_NOT_FOUND, referenceSearchParameterName, referencedByResourceTypeName));
+                }
+                if (!Type.REFERENCE.equals(Type.fromValue(referenceSearchParameter.getType().getValue()))) {
```
prb112, line 1671: at this point in the code would referenceSearchParameter.getType ever be null? 
I don't think so at this point in the model this will be null as the `type | Σ | 1..1 | code`

This is good.  Leaving my note here.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -1589,6 +1605,168 @@ private static QueryParameter parseChainedParameter(Class<?> resourceType, Strin
...
+                    throw SearchExceptionUtil.buildNewInvalidSearchException(
+                        String.format(PARAMETER_TYPE_NOT_REFERENCE_FOR_REVERSE_CHAIN_SEARCH, referenceSearchParameterName));
+                }
+
+                // Validate resource type is one of the reference search parameter target resource types
+                if (!referenceSearchParameter.getTarget().contains(ResourceType.of(resourceType.getSimpleName()))) {
```
prb112, line 1677: This can only be an emptyList in the worst case.  I think this is good.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -1589,6 +1605,168 @@ private static QueryParameter parseChainedParameter(Class<?> resourceType, Strin
...
+                        String.format(TARGET_TYPE_OF_REFERENCE_PARAMETER_NOT_VALID_FOR_REVERSE_CHAIN_SEARCH,
+                            referenceSearchParameterName, resourceType.getSimpleName()));
+                }
+
+                // Create new QueryParameter
+                QueryParameter parameter = new QueryParameter(Type.REFERENCE, referenceSearchParameterName, Modifier.TYPE, referencedByResourceTypeName, false, true);
+                if (rootParameter == null) {
+                    rootParameter = parameter;
+                } else {
+                    if (rootParameter.getChain().isEmpty()) {
```
prb112, line 1688:
```suggestion
                } else if (rootParameter.getChain().isEmpty()) {
```

prb112: The Else if / Else seems like it could be reduced one level

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -1589,6 +1605,168 @@ private static QueryParameter parseChainedParameter(Class<?> resourceType, Strin
...
+
+                        if (modifierName != null) {
+                            if (ModelSupport.isResourceType(modifierName)) {
+                                modifier = Modifier.TYPE;
+                                modifierResourceTypeName = modifierName;
+                            } else {
+                                try {
+                                    modifier = Modifier.fromValue(modifierName);
+                                } catch (IllegalArgumentException e) {
+                                    String msg = "Undefined Modifier: " + modifierName;
```
prb112, line 1731: we should probably surround the modifierName with single quotes
We might want to URL encode it as well

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -1589,6 +1605,168 @@ private static QueryParameter parseChainedParameter(Class<?> resourceType, Strin
...
+                                    throw SearchExceptionUtil.buildNewInvalidSearchException(msg);
+                                }
+                            }
+                            if (!isAllowed(type, modifier)) {
+                                String msg = "Unsupported type/modifier combination: " + type.value() + "/" + modifier.value();
```
prb112, line 1736:
```suggestion
                                String msg = "Unsupported type/modifier combination: '" + type.value() + "/" + modifier.value() +"'";
```
Add single quotes

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -772,7 +774,7 @@ private void appendMidChainParm(StringBuilder whereClauseSegment, QueryParameter
         if (codeSystemName != null && !codeSystemName.equals("*")) {
             Integer codeSystemId = identityCache.getCodeSystemId(codeSystemName);
             if (codeSystemId != null) {
-                whereClauseSegment.append(AND).append(PARAMETER_TABLE_ALIAS + DOT).append(CODE_SYSTEM_ID).append(EQ)
+                whereClauseSegment.append(AND).append(chainedParmVar + DOT).append(CODE_SYSTEM_ID).append(EQ)
```
prb112, line 777:
```suggestion
                whereClauseSegment.append(AND).append(chainedParmVar).append(DOT).append(CODE_SYSTEM_ID).append(EQ)
```

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1384,4 +1386,242 @@ private SqlQueryData processMissingParm(Class<?> resourceType, QueryParameter qu
...
+     * @throws Exception
+     */
+    @Override
+    protected SqlQueryData processReverseChainedReferenceParm(Class<?> resourceType, QueryParameter queryParm) throws Exception {
+        final String METHODNAME = "processReverseChainedReferenceParm";
+        log.entering(CLASSNAME, METHODNAME, queryParm.toString());
+
+        final String CR = "CR";
+        final String CLR = "CLR";
+        final String CP = "CP";
```
prb112, line 1438: I'm almost positive these are repeated constants, maybe they should be top level constants for consistency?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1384,4 +1386,242 @@ private SqlQueryData processMissingParm(Class<?> resourceType, QueryParameter qu
...
+            chainedLogicalResourceVar = CLR + (parmIndex + 1);
+            chainedParmVar            = CP + (parmIndex + 1);
+            nextChainedResourceVar        = CR + (parmIndex + 2);
+            nextChainedLogicalResourceVar = CLR + (parmIndex + 2);
+            nextChainedParmVar        = CP + (parmIndex + 2);
+            StringBuilder whereClauseSegment = new StringBuilder();
+
+            if (parmIndex == 0) {
+                // BUild outer select: SELECT CLR0.LOGICAL_ID
+                //                       FROM <resource-type>_LOGICAL_RESOURCES AS CLR0
```
prb112, line 1472:
```suggestion

                // Builds outer select: 
                // @formatter:off
                // SELECT CLR0.LOGICAL_ID
                //                       FROM <resource-type>_LOGICAL_RESOURCES AS CLR0
```

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1384,4 +1386,242 @@ private SqlQueryData processMissingParm(Class<?> resourceType, QueryParameter qu
...
+                //                       JOIN <resource-type>_RESOURCES AS CR0
+                //                         ON CR0.RESOURCE_ID = CLR0.CURRENT_RESOURCE_ID AND CR0.IS_DELETED = 'N'
+                //                       WHERE
```
prb112, line 1475:
```suggestion
                //                       WHERE
                // @formatter:on
```

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1384,4 +1386,242 @@ private SqlQueryData processMissingParm(Class<?> resourceType, QueryParameter qu
...
+                selectSegments.append("SELECT " + prevChainedLogicalResourceVar + DOT + "LOGICAL_ID")
```
prb112, line 1476: let's break these into individual appends
Also SELECT and I think LogicalId are already in JDBCConstants

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1384,4 +1386,242 @@ private SqlQueryData processMissingParm(Class<?> resourceType, QueryParameter qu
...
+            }
+
+            if (parmIndex < lastParmIndex) {
+                if (currentParm.isReverseChained()) {
+                    // Build inner select joins: EXISTS (SELECT 1
+                    //                           FROM <modifierTypeResourceName>_TOKEN_VALUES_V AS CPx
+                    //                           JOIN <modifierTypeResourceName>_LOGICAL_RESOURCES AS CLRx
+                    //                             ON CLRx.LOGICAL_RESOURCE_ID = CPx.LOGICAL_RESOURCE_ID
+                    //                           JOIN <modifierTypeResourceName>_RESOURCES AS CRx
+                    //                             ON CRx.RESOURCE_ID = CLRx.CURRENT_RESOURCE_ID AND CRx.IS_DELETED = 'N'
```
prb112, line 1491:
```suggestion
                    // @formatter:off
                    //                           FROM <modifierTypeResourceName>_TOKEN_VALUES_V AS CPx
                    //                           JOIN <modifierTypeResourceName>_LOGICAL_RESOURCES AS CLRx
                    //                             ON CLRx.LOGICAL_RESOURCE_ID = CPx.LOGICAL_RESOURCE_ID
                    //                           JOIN <modifierTypeResourceName>_RESOURCES AS CRx
                    //                             ON CRx.RESOURCE_ID = CLRx.CURRENT_RESOURCE_ID AND CRx.IS_DELETED = 'N'
                    // @formatter:on
```

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1384,4 +1386,242 @@ private SqlQueryData processMissingParm(Class<?> resourceType, QueryParameter qu
...
+                    selectSegments.append("EXISTS " + LEFT_PAREN + "SELECT 1")
+                                    .append(FROM + currentParm.getModifierResourceTypeName() + "_TOKEN_VALUES_V AS " + chainedParmVar)
+                                    .append(JOIN + currentParm.getModifierResourceTypeName() + "_LOGICAL_RESOURCES AS " + chainedLogicalResourceVar)
+                                    .append(ON + chainedLogicalResourceVar + DOT + "LOGICAL_RESOURCE_ID" + EQ + chainedParmVar + DOT + "LOGICAL_RESOURCE_ID")
+                                    .append(JOIN + currentParm.getModifierResourceTypeName() + "_RESOURCES AS " + chainedResourceVar)
+                                    .append(ON + chainedResourceVar + DOT + "RESOURCE_ID" + EQ + chainedLogicalResourceVar + DOT + "CURRENT_RESOURCE_ID")
+                                    .append(AND + chainedResourceVar + DOT + "IS_DELETED" + EQ + "'N'");
```
prb112, line 1498: same comment about appends and the EXISTS I think is already in the JDBCConstants

same with RESOURCE_ID and IS_DELETED

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1384,4 +1386,242 @@ private SqlQueryData processMissingParm(Class<?> resourceType, QueryParameter qu
...
+                    if (parmIndex == 0) {
+                        referencedResourceType = resourceType.getSimpleName();
+                    } else {
+                        referencedResourceType = previousParm.getModifierResourceTypeName();
+                    }
+                    if (parmIndex < lastParmIndex - 1 && currentParm.getNextParameter().isReverseChained()) {
+                        // Build inner select where clause: WHERE CPx.TOKEN_VALUE = CLR<x-1>.LOGICAL_ID
+                        //                                    AND CPx.PARAMETER_NAME_ID = <parm-name-id>
+                        //                                    AND CPx.CDE_SYSTEM_ID = <code-system-id>
+                        //                                    AND
```
prb112, line 1510: Where you have the nice format, we can add the @formatter:on

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1384,4 +1386,242 @@ private SqlQueryData processMissingParm(Class<?> resourceType, QueryParameter qu
...
+                    // Build this piece: (SELECT 'resource-type-name' || '/' || CLR<x+1>.LOGICAL_ID ...
+                    selectSegments.append(LEFT_PAREN);
+                    appendInnerSelect(selectSegments, currentParm, currentParm.getModifierResourceTypeName(),
+                        nextChainedResourceVar, nextChainedLogicalResourceVar, nextChainedParmVar);
+                    whereClauseSegment.append(RIGHT_PAREN);
+
+                    if (!chainedParmProcessed) {
+                        chainedParmProcessed = true;
+
+                        // BUild ON clause for join: ) AS CPx ON CPx.LOGICAL_RESOURCE_ID = CLR<x-1>.LOGICAL_RESOURCE_ID
```
prb112, line 1546:
```suggestion
                        // Builds ON clause for join: ) AS CPx ON CPx.LOGICAL_RESOURCE_ID = CLR<x-1>.LOGICAL_RESOURCE_ID
```

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1384,4 +1386,242 @@ private SqlQueryData processMissingParm(Class<?> resourceType, QueryParameter qu
...
+                        //                    AND
+                        whereClauseSegment.append(JOIN + previousParm.getModifierResourceTypeName() + "_LOGICAL_RESOURCES")
+                                            .append(" AS " + chainedLogicalResourceVar + ON + chainedLogicalResourceVar + DOT + "LOGICAL_RESOURCE_ID");
+                    }
+                    whereClauseSegment.append(EQ + prevChainedLogicalResourceVar + DOT + "LOGICAL_RESOURCE_ID" + AND);
+                }
+
+                // Build the rest
+                SqlQueryData sqlQueryData;
+                if (!"_id".equals(currentParm.getCode())) {
```
prb112, line 1572: You could embed the !chainedParmProcessed block down here, and only test this branch at most once

Net, I'm suggesting a logic change where the test `if (!"_id".equals(currentParm.getCode())) { ` is not in some circumstances twice, and branching of !chainedParmProcessed is done here before the existing code in this branch

P.S. I think this works, please validate.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1384,4 +1386,242 @@ private SqlQueryData processMissingParm(Class<?> resourceType, QueryParameter qu
...
+        log.exiting(CLASSNAME, METHODNAME, queryData.getQueryString());
+        return queryData;
+    }
+
+    /**
+     * Populates the reference parameter name ID and code system ID sub-segment of the passed where clause segment.
+     *
+     * @param whereClauseSegment - the segment to which the sub-segment will be added
+     * @param queryParmName - the search parameter name
+     * @param resourceTypeName - the resource type of the reference being followed
```
prb112, line 1604: Not to change too much, but please note we use code not names. I think for consistency, we probably leave as name here.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1384,4 +1386,242 @@ private SqlQueryData processMissingParm(Class<?> resourceType, QueryParameter qu
...
+
+        Integer parameterNameId = identityCache.getParameterNameId(queryParmName);
+        Integer codeSystemId = getCodeSystemId(resourceTypeName);
+
+        // Build the segment:
+        // CPx.PARAMETER_NAME_ID = <parameter-name-id> AND CPx.CODE_SYSTEM_ID = <code-system_id>
+        whereClauseSegment.append(parameterTableAlias + DOT + "PARAMETER_NAME_ID=")
+                .append(nullCheck(parameterNameId))
+                .append(AND)
+                .append(parameterTableAlias + DOT + "CODE_SYSTEM_ID=")
```
prb112, line 1621: candidate for breaking into individual appends.

fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractReverseChainTest.java
```diff
@@ -0,0 +1,363 @@
...
+        savedDevice1 = persistence.create(getDefaultPersistenceContext(), savedDevice1).getResource();
+        savedDevice1 = savedDevice1.toBuilder().manufacturer(string("Updated Manufacturer")).build();
+        savedDevice1 = persistence.update(getDefaultPersistenceContext(), savedDevice1.getId(), savedDevice1).getResource();
+
+        // a Patient that will have no other resources referencing it
+        savedPatient4 = persistence.create(getDefaultPersistenceContext(), patient).getResource();
+    }
+
+    @AfterClass
+    public void deleteResources() throws Exception {
```
prb112, line 135: Random comment, you pick up the style of the code/section very quickly.  Thanks for matching the styles seemingly effortlessly.

fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractReverseChainTest.java
```diff
@@ -0,0 +1,363 @@
...
+//    @Test
+//    public void testReverseChainSingleResultWithCompositeParm() throws Exception {
+//        Map<String, List<String>> queryParms = new HashMap<String, List<String>>();
+//        queryParms.put("_has:Observation:encounter:code-value-string", Collections.singletonList("code$test"));
+//         List<Resource> resources = runQueryTest(Encounter.class, queryParms);
+//        assertNotNull(resources);
+//        assertEquals(1, resources.size());
+//        assertEquals("Encounter", resources.get(0).getClass().getSimpleName());
+//        assertEquals(savedEncounter1.getId(), resources.get(0).getId());
+//    }
```
prb112, line 354: Maybe comment why this commented out?

fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractReverseChainTest.java
```diff
@@ -0,0 +1,363 @@
...
+
+    private Reference reference(String reference) {
+        return Reference.builder().reference(string(reference)).build();
+    }
+
+    private HumanName humanName(String firstName, String lastName) {
+        return HumanName.builder().given(string(firstName)).family(string(lastName)).build();
+    }
```
prb112, line 362: Is it plausible / possible to have more than one _has on a reverse chain? 
Do we support it from the REST layer? 
What type of error do we throw?

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchReverseChainTest.java
```diff
@@ -0,0 +1,1034 @@
...
+                .meta(Meta.builder()
+                    .tag(Coding.builder()
+                        .code(Code.of(tag))
+                        .build())
+                    .build())
+                .generalPractitioner(Reference.builder().reference(of("PractitionerRole/" + tag)).build())
+                .managingOrganization(Reference.builder().reference(of("Organization/" + organization1Id)).build())
+                .birthDate(Date.of(now.toString().substring(0,10)))
+                .telecom(ContactPoint.builder().system(ContactPointSystem.PHONE).value(of("1" + tag)).build())
+               .build();
```
prb112, line 162: above is a good candidate for the formatter:off comment and end with formatter on

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchReverseChainTest.java
```diff
@@ -0,0 +1,1034 @@
...
+                .meta(Meta.builder()
+                    .tag(Coding.builder()
+                        .code(Code.of(tag))
+                        .build())
+                    .build())
+                .generalPractitioner(Reference.builder().reference(of("Practitioner/" + tag)).build())
+                .managingOrganization(Reference.builder().reference(of("Organization/" + organization2Id)).build())
+                .birthDate(Date.of(now.minus(1, ChronoUnit.DAYS).toString().substring(0,10)))
+                .telecom(ContactPoint.builder().system(ContactPointSystem.PHONE).value(of("2" + tag)).build())
+                .build();
```
prb112, line 196: same formatter comment

prb112: I'll leave you to decide to use it in the rest of the class (treat the formatter comments as FYI)

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchReverseChainTest.java
```diff
@@ -0,0 +1,1034 @@
...
+    public void testSearchReverseChainWithSearchParmUndefinedModifierError() {
+        WebTarget target = getWebTarget();
+        Response response =
+                target.path("Patient")
+                .queryParam("_has:Procedure:subject:code:badModifier", "test")
+                .request(FHIRMediaType.APPLICATION_FHIR_JSON)
+                .get();
+        assertResponse(response, Response.Status.BAD_REQUEST.getStatusCode());
+        assertExceptionOperationOutcome(response.readEntity(OperationOutcome.class),
+                "Undefined Modifier: badModifier");
```
prb112, line 519: Refer to corresponding comment in SearchUtil (this response should have 'badModifier' and possibly be URL Encoded.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchReverseChainTest.java
```diff
@@ -0,0 +1,1034 @@
...
+    public void testSearchReverseChainWithChainedSearchParmBadModifierError() {
+        WebTarget target = getWebTarget();
+        Response response =
+                target.path("Patient")
+                .queryParam("_has:Procedure:subject:encounter:contains.status", "test")
+                .request(FHIRMediaType.APPLICATION_FHIR_JSON)
+                .get();
+        assertResponse(response, Response.Status.BAD_REQUEST.getStatusCode());
+        assertExceptionOperationOutcome(response.readEntity(OperationOutcome.class),
+                "Modifier: 'CONTAINS' not allowed on chained parameter");
```
prb112, line 558: Why does this become uppercase, should we upper case it?

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchReverseChainTest.java
```diff
@@ -0,0 +1,1034 @@
...
+    public void testSearchReverseChainWithChainedSearchParmNotReferenceError() {
+        WebTarget target = getWebTarget();
+        Response response =
+                target.path("Patient")
+                .queryParam("_has:Procedure:subject:code.status", "test")
+                .request(FHIRMediaType.APPLICATION_FHIR_JSON)
+                .get();
+        assertResponse(response, Response.Status.BAD_REQUEST.getStatusCode());
+        assertExceptionOperationOutcome(response.readEntity(OperationOutcome.class),
+                "Type: 'TOKEN' not allowed on chained parameter");
```
prb112, line 571: Why does this become uppercase? I think it's from the fhir-search Type enum, and we should probably use the natural value of lower case token which is what's read fromt he Parameter

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchReverseChainTest.java
```diff
@@ -0,0 +1,1034 @@
...
+                target.path("Patient")
+                .queryParam("given", "1" + tag)
+                .queryParam("_has:Procedure:subject:status", ProcedureStatus.COMPLETED.getValue())
+                .request(FHIRMediaType.APPLICATION_FHIR_JSON)
+                .get();
+        assertResponse(response, Response.Status.OK.getStatusCode());
+        Bundle bundle = response.readEntity(Bundle.class);
+
+        assertNotNull(bundle);
+        assertTrue(bundle.getEntry().size() == 1);
```
prb112, line 790: I've seen this a couple of times. 
1 - assertEquals is probably preferred when doing the equals (that way you can see what the size value is
2 - If the delete fails for any reason, it leaves dirty data, so we typically do >= 1

If the delete fails, does it impact this test in any way?

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchReverseChainTest.java
```diff
@@ -0,0 +1,1034 @@
...
+        assertNotNull(bundle);
+        assertTrue(bundle.getEntry().size() == 2);
+        List<String> resourceIds = new ArrayList<>();
+        for (Bundle.Entry entry : bundle.getEntry()) {
+            resourceIds.add(entry.getResource().getId());
+        }
+        assertTrue(resourceIds.contains(patient1Id));
+        assertTrue(resourceIds.contains(procedure1Id));
+    }
+
```
prb112, line 1033: even if this fails, let's have a multiple _has test and capture the failure. 
Also probably should update conformance to state if multiple has is supported or not.

prb112 (commented):
Left some comments

docs/src/pages/Conformance.md
michaelwschroeder: done

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
michaelwschroeder: done

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
michaelwschroeder: done

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchReverseChainTest.java
michaelwschroeder: correct, just following pattern of one of the other tests to cleanup resources

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
michaelwschroeder: added annotations

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
michaelwschroeder: renamed to code

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
michaelwschroeder: done

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
michaelwschroeder: changed to use substring

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
michaelwschroeder: created constant

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
michaelwschroeder: done

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
michaelwschroeder: done

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
michaelwschroeder: done

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
michaelwschroeder: done

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
michaelwschroeder: moved out of this method and processChainedReferenceParm() and made them global

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
michaelwschroeder: added formatter annotations

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
michaelwschroeder: added formatter annotations

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
michaelwschroeder: broke into individual appends, also added several constants to JDBCConstants that were not there, and replaced string literals in this class with the new constants

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
michaelwschroeder: formatter annotations added

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
michaelwschroeder: done

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
michaelwschroeder: added formatter annotations

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
michaelwschroeder: done

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
michaelwschroeder: I think I did what you were suggesting here - please verify

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
michaelwschroeder: noted.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
michaelwschroeder: done

fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractReverseChainTest.java
michaelwschroeder: you're welcome!

fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractReverseChainTest.java
michaelwschroeder: added TODO indicating this should be uncommented when issue #1669 is resolved

fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractReverseChainTest.java
michaelwschroeder: yes, _has can be chained - the testcases starting with testMultiReverseChain... test this

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchReverseChainTest.java
michaelwschroeder: not clear on this - just add a comment before and after with the annotation?

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchReverseChainTest.java
michaelwschroeder: done

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchReverseChainTest.java
michaelwschroeder: changed the code in SearchUtil that constructs the exception to use Modifier.value() so we get the natural value

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchReverseChainTest.java
michaelwschroeder: changed the code in SearchUtil that constructs the exception to use Token.value() so we get the natural value

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchReverseChainTest.java
michaelwschroeder: Updated to use assertEquals. As far as the delete issue, I use a generated tag based on Instant.now() in all the resources and any fields that are searched contain some form of this value. So even if delete fails, these testcases should not find any dirty data.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchReverseChainTest.java
michaelwschroeder: Multiples are supported. The testcases starting with testSearchMultipleReverseChain... test this. It is stated in the FHIR spec that this is supported, but I can add a blurb to the conformance doc if you think I need to.

michaelwschroeder:
@prb112 thanks for the thorough review - appreciate it!

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
michaelwschroeder: fyi, changed back to using replaceFirst to handle case where the reverseChainParameterString might just be '_has'. If using substring, I'd have to first check the length of reverseChainParameterString before substringing past the expected initial '_has:' string.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/JDBCConstants.java
```diff
@@ -36,6 +36,14 @@
     public static final String DATE_END = "DATE_END";
     public static final String LATITUDE_VALUE = "LATITUDE_VALUE";
     public static final String LONGITUDE_VALUE = "LONGITUDE_VALUE";
+    public static final String RESOURCES = "_RESOURCES";
```
punktilious, line 39: I've been trying to keep the constant names identical to the string value for sanity. I don't really want to deviate from this. I think we should use a slightly different convention for prefix/postfix strings. Perhaps something as simple as _RESOURCES = "_RESOURCES"

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/JDBCConstants.java
```diff
@@ -36,6 +36,14 @@
...
+    public static final String LOGICAL_RESOURCES = "_LOGICAL_RESOURCES";
+    public static final String RESOURCE_ID = "RESOURCE_ID";
+    public static final String LOGICAL_ID = "LOGICAL_ID";
+    public static final String LOGICAL_RESOURCE_ID = "LOGICAL_RESOURCE_ID";
+    public static final String CURRENT_RESOURCE_ID = "CURRENT_RESOURCE_ID";
+    public static final String PARAMETER_NAME_ID = "PARAMETER_NAME_ID";
+    public static final String IS_DELETED_NO = "IS_DELETED = 'N'";
```
punktilious, line 46: Good idea.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1376,12 +1391,268 @@ private SqlQueryData processMissingParm(Class<?> resourceType, QueryParameter qu
...
+     * WHERE
+     *   EXISTS (
+     *     SELECT
+     *       1
+     *     FROM
+     *       Observation_TOKEN_VALUES_V AS CP1
+     *       JOIN Observation_LOGICAL_RESOURCES AS CLR1 ON CLR1.LOGICAL_RESOURCE_ID = CP1.LOGICAL_RESOURCE_ID
+     *       JOIN Observation_RESOURCES AS CR1 ON CR1.RESOURCE_ID = CLR1.CURRENT_RESOURCE_ID AND CR1.IS_DELETED = 'N'
+     *       JOIN Observation_TOKEN_VALUES_V AS CP2 ON CP2.LOGICAL_RESOURCE_ID = CLR1.LOGICAL_RESOURCE_ID
+     *       AND (
```
punktilious, line 1424: I don't think we should use parens unless required to control precedence.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1376,12 +1391,268 @@ private SqlQueryData processMissingParm(Class<?> resourceType, QueryParameter qu
...
+            chainedResourceVar        = CR + (parmIndex + 1);
+            chainedLogicalResourceVar = CLR + (parmIndex + 1);
+            chainedParmVar            = CP + (parmIndex + 1);
+            nextChainedResourceVar        = CR + (parmIndex + 2);
+            nextChainedLogicalResourceVar = CLR + (parmIndex + 2);
+            nextChainedParmVar        = CP + (parmIndex + 2);
+            StringBuilder whereClauseSegment = new StringBuilder();
+
+            if (parmIndex == 0) {
+                // BUild outer select:
```
punktilious, line 1481: Build

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1376,12 +1391,268 @@ private SqlQueryData processMissingParm(Class<?> resourceType, QueryParameter qu
...
+                        referencedResourceType = resourceType.getSimpleName();
+                    } else {
+                        referencedResourceType = previousParm.getModifierResourceTypeName();
+                    }
+                    if (parmIndex < lastParmIndex - 1 && currentParm.getNextParameter().isReverseChained()) {
+                        // Build inner select where clause:
+                        // @formatter:off
+                        //   WHERE CPx.TOKEN_VALUE = CLR<x-1>.LOGICAL_ID
+                        //     AND CPx.PARAMETER_NAME_ID = <parm-name-id>
+                        //     AND CPx.CDE_SYSTEM_ID = <code-system-id>
```
punktilious, line 1530: CODE_SYSTEM_ID

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1376,12 +1391,268 @@ private SqlQueryData processMissingParm(Class<?> resourceType, QueryParameter qu
...
+                        selectSegments.append(WHERE).append(chainedParmVar).append(DOT).append(TOKEN_VALUE).append(EQ)
+                                        .append(prevChainedLogicalResourceVar).append(DOT).append(LOGICAL_ID).append(AND);
+                        populateReferenceNameAndCodeSystemIdSubSegment(selectSegments, currentParm.getCode(), referencedResourceType, chainedParmVar);
+                        selectSegments.append(AND);
+                    } else {
+                        // Build final inner select where clause:
+                        // @formatter:off
+                        //   WHERE CPx.TOKEN_VALUE = CLR<x-1>.LOGICAL_ID
+                        //     AND CPx.PARAMETER_NAME_ID = <parm-name-id>
+                        //     AND CPx.CDE_SYSTEM_ID = <code-system-id>
```
punktilious, line 1542: CODE_SYSTEM_ID

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1376,12 +1391,268 @@ private SqlQueryData processMissingParm(Class<?> resourceType, QueryParameter qu
...
+            String resourceTypeName, String parameterTableAlias) throws FHIRPersistenceException {
+        final String METHODNAME = "populateReferenceNameAndCodeSystemIdSubSegment";
+        log.entering(CLASSNAME, METHODNAME, queryParmName);
+
+        Integer parameterNameId = identityCache.getParameterNameId(queryParmName);
+        Integer codeSystemId = getCodeSystemId(resourceTypeName);
+
+        // Build the segment:
+        // CPx.PARAMETER_NAME_ID = <parameter-name-id> AND CPx.CODE_SYSTEM_ID = <code-system_id>
+        whereClauseSegment.append(parameterTableAlias).append(DOT).append(PARAMETER_NAME_ID).append(EQ).append(nullCheck(parameterNameId))
```
punktilious, line 1652: nullCheck is redundant here. getParameterNameId will never return null.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
```diff
@@ -555,20 +556,41 @@ protected void buildWhereClause(StringBuilder whereClause, String overrideType)
...
-                        final String onFilter = querySegment.getQueryString().replaceAll(PARAMETER_TABLE_ALIAS + "\\.", paramTableAlias + ".");
-
-                        whereClause.append(JOIN);
-                        whereClause.append(tableName(overrideType, param));
-                        whereClause.append(" AS " + paramTableAlias);
-                        whereClause.append(ON);
-                        whereClause.append(onFilter);
-                        whereClause.append(" AND LR.LOGICAL_RESOURCE_ID = " + paramTableAlias + ".LOGICAL_RESOURCE_ID");
+                        if (param.isReverseChained()) {
+                            // Join on a select from logical resource table
```
punktilious, line 561: from Observation logical resources table. There's also a global (all resources) logical_resources table. Good to distinguish, because the join predicate would be slightly different

fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractReverseChainTest.java
punktilious: agreed

fhir-search/src/main/java/com/ibm/fhir/search/parameters/QueryParameter.java
```diff
@@ -44,6 +45,11 @@ public QueryParameter(Type type, String code, Modifier modifier, String modifier
         this.isInclusionCriteria = isInclusionCriteria;
     }
 
+    public QueryParameter(Type type, String code, Modifier modifier, String modifierResourceTypeName, boolean isInclusionCriteria, boolean isReverseChained) {
```
punktilious, line 48: I'll raise a backlog issue. QueryParameter could use some Javadoc love.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
punktilious: +1. And to add to this, although it's not the case here it's important exceptions are never used for flow control in the "happy path" case because they are expensive to build.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/JDBCConstants.java
michaelwschroeder: done - used your suggestion

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
michaelwschroeder: Agreed. Unfortunately in this case they're generated by the populateNameIdSubSegment() method, which gets called by many other methods - didn't want to take on making a change there and having to validate all of those existing cases.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
michaelwschroeder: done

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
michaelwschroeder: done

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
michaelwschroeder: done

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
michaelwschroeder: noted

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
michaelwschroeder: comment updated

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
prb112: looks great - thanks Mike

fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractReverseChainTest.java
prb112: Thanks Mike

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchReverseChainTest.java
prb112: That's right.  This was more of a heads up. 
It keeps it from getting reformatted accidently

prb112 (approved):
LGTM, I'll let Robin resolve the two remaining comments

lmsurpre (approved):
I didn't review the PR in great depth, but I did deploy it to one of our clusters and it appears to be functioning correctly.
One thing I noticed is that the queries aren't particularly speedy, but I'm guessing we should save that for the future because even some basic queries are slow for me right now (e.g. see https://github.com/IBM/FHIR/issues/1673).
I'm good with it whenever @punktilious is.

lmsurpre:
> One thing I noticed is that the queries aren't particularly speedy

To quantify this, on our connectathon system with 8112 Patients and 87364 Conditions, I'm seeing the following:

|search query|execution time|
|-------------|-------------|
| `/Condition?code=http://snomed.info/sct\|44054006&patient:Patient.gender=female&_count=1` | count=34ms, fetch=28ms |
| `/Patient?gender=female&_has:Condition:patient:code=http://snomed.info/sct\|44054006&_count=1` | count=13121ms, fetch=12791ms |

punktilious (approved):
LGTM

Issue #1536, opened 2020-09-30
reference fields in fhir-examples generated resources should include resource type


punktilious:
**Describe the bug**
Many reference fields in the fhir-examples generated resources use short random string values like `IRm0J8cmzs`. Such values do not conform to any reference format defined in the FHIR specification.

**To Reproduce**
Build the examples package and inspect the generated resources:

- mvn clean install -f fhir-examples-generator
- mvn com.ibm.fhir:fhir-examples-generator:generate-examples -f fhir-examples
- view fhir-examples/src/main/resources/json/ibm/complete-mock/Patient-1.json

**Expected behavior**
Reference field values should be a URL or of the form `<resourceType>/<logicalId>`

punktilious:
Implementation tip. In ModelSupport:

```
    /**
     * @param modelClass
     *            a model class which represents a FHIR resource or element
     * @param elementName
     *            the name of the reference element
     * @return a set of Strings which represent the the allowed target types for the reference
     */
    public static Set<String> getReferenceTargetTypes(Class<?> modelClass, String elementName) {
        ElementInfo elementInfo = getElementInfo(modelClass, elementName);
        if (elementInfo != null) {
            return elementInfo.getReferenceTypes();
        }
        return Collections.emptySet();
    }
```

dermatologist:
Would love to help out and give it a go!

prb112:
> Would love to help out and give it a go!

Fantastic.  We're happy to coach, and we are generally available vai Zulip on #ibm  https://chat.fhir.org/#narrow/stream/212434-ibm

dermatologist:
Thank you. I think to reproduce it should be in the file fhir-examples-generator folder. right?

* mvn clean install -f fhir-examples-generator
* mvn com.ibm.fhir:fhir-examples-generator:generate-examples -f **fhir-examples-generator**/pom.xml
* view fhir-examples/src/main/resources/json/ibm/complete-mock/Patient-1.json

lmsurpre:
you may need to build the whole project first since the fhir-examples-generator is currently configured to depend on the 4.4.1-SNAPSHOT version of our other modules.
* mvn clean install -f fhir-examples
* mvn clean install -f fhir-parent -DskipTests

prb112:
Unfortunately, no - the examples generator builds JSON files which are destined for `fhir-examples` not `fhir-examples-generator`

And yes, @lmsurpre is correct (dependencies must be built first)

[...]

PR #1667, opened 2020-11-04
Add reference target annotations, javadoc, and validation

lmsurpre:
Previously, we only had these for fields of type reference with a max
cardinality of 1. Now we have these for both repeating fields and for
choice types which include references as well.

I think there are many more cases we're missing wrt validation of choice
types and repeating fields of other types, but I was focused mainly on
ReferenceTargets for now (to address
https://github.com/IBM/FHIR/issues/1536 )


JohnTimm:
Can we move the instanceof check into the helper method instead of the constructor?

lmsurpre:
> Can we move the instanceof check into the helper method instead of the constructor?

done

fhir-model/src/main/java/com/ibm/fhir/model/util/ValidationSupport.java
```diff
@@ -285,6 +285,10 @@ public static void checkValue(String value, Pattern pattern) {
                 List<String> typeNameList = Arrays.stream(types).map(Class::getSimpleName).collect(Collectors.toList());
                 throw new IllegalStateException(String.format("Invalid type: %s for choice element: '%s' must be one of: %s", elementType.getSimpleName(), elementName, typeNameList.toString()));
             }
+            //do other validation here?
```
michaelwschroeder, line 288: I don't think you need to do this here since you're generating the ValidationSupport.checkReferenceType() call for reference-type choice elements now

michaelwschroeder (approved):

JohnTimm (approved):
LGTM

fhir-examples-generator/src/main/java/com/ibm/fhir/examples/DataCreatorBase.java
```diff
@@ -180,23 +186,52 @@ protected Object createArgument(Class<?> owningClass, Method builderMethod, Clas
...
+            if (Reference.class.equals(choiceType)){
+
+                // Handling references specially
+                Set<String> referenceTargetTypes = ModelSupport.getReferenceTargetTypes(owningClass, elementName);
+                if (!referenceTargetTypes.isEmpty()) {
+                    String[] targetTypes = new String[referenceTargetTypes.size()];
+                    referenceTargetTypes.toArray(targetTypes);
+
+                    // use the "choiceIndicator" to pick the reference type
+                    String targetType = targetTypes[(targetTypes.length - 1) % choiceIndicator];
```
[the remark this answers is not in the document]
lmsurpre: I'm thinking to take a rain-check on this one.  Lots going on and my bar is lower for this example-generator stuff than for core fhir server

fhir-model/src/main/java/com/ibm/fhir/model/util/ValidationSupport.java
lmsurpre: good call, meant to remove this.
I actually think there are other validation cases (for both choice types and list types) that we're still missing, but in this PR I was focused on just the referenceType-checking

fhir-model/src/main/java/com/ibm/fhir/model/util/ValidationSupport.java
```diff
@@ -285,6 +285,10 @@ public static void checkValue(String value, Pattern pattern) {
...
+//            if (element instanceof Reference) {
+//                checkReferenceType(element, elementName, ModelSupport.getReferenceTargetTypes(modelClass, elementName));
+//            }
```
lmsurpre, line 291:
```suggestion
```

fhir-model/src/main/java/com/ibm/fhir/model/util/ValidationSupport.java
lmsurpre: removed

PR #1668, opened 2020-11-04
Introduce Xhtml.from(String plainText) helper and add javadoc

lmsurpre:
This helper uses the OWASP Encoder (a small java dependency that we were
already using in the project, but not from `fhir-model`) to encode plain
text strings for use within HTML.
In addition to Xhtml javadoc, I also added javadoc for some of the code
subtype methods.

### DIFF
--- a/fhir-model/pom.xml
+++ b/fhir-model/pom.xml
@@ -25,6 +25,10 @@
             <artifactId>jcip-annotations</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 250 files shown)


JohnTimm:
Found a typo in the Javadoc `Inhereted` should be `Inherited`.

JohnTimm (approved):
LGTM

JohnTimm (approved):
LGTM

PR #1671, opened 2020-11-05
remove unused COS constants from main constants file

lmsurpre:
@punktilious not sure if you had a plan for these, but I had removed them locally just to keep my head straight.


punktilious (approved):
Yeah, this came over from bulk-load. Thanks for the cleanup. LGTM

PR #1675, opened 2020-11-06
fix: change to use a single RUN and change deprecated command in lieu of a label

prb112:

### DIFF
--- a/fhir-install/Dockerfile
+++ b/fhir-install/Dockerfile
@@ -23,7 +23,7 @@ RUN unzip -qq /tmp/fhir-server-distribution.zip -d /tmp && \
 FROM openliberty/open-liberty:20.0.0.9-full-java8-openj9-ubi
 COPY --chown=1001:0 --from=base /opt/ol/wlp /opt/ol/wlp
 
-MAINTAINER Lee Surprenant <lmsurpre@us.ibm.com>
+LABEL MAINTAINER="Lee Surprenant <lmsurpre@us.ibm.com>"
 
 ENV FHIR /opt/ol/wlp/usr/servers/fhir-server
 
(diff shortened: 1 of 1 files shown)


fhir-install/Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN unzip -qq /tmp/fhir-server-distribution.zip -d /tmp && \
 FROM openliberty/open-liberty:20.0.0.9-full-java8-openj9-ubi
 COPY --chown=1001:0 --from=base /opt/ol/wlp /opt/ol/wlp
 
-MAINTAINER Lee Surprenant <lmsurpre@us.ibm.com>
+LABEL MAINTAINER="Lee Surprenant <lmsurpre@us.ibm.com>"
```
lmsurpre, line 26: I'd be fine to totally drop this line...similar to how we decied now to include the `@author` tag on the source files

fhir-install/Dockerfile
```diff
@@ -36,15 +36,16 @@ ENV WLP_OUTPUT_DIR=/opt/ol/wlp/usr/servers
...
+# create the configDropins directories for the fhir-server
 USER root
 RUN ln -sfn ${FHIR} /output && \
     ln -sfn ${FHIR} /config && \
     rm -rf /opt/ol/wlp/output && \
-    rm -rf /opt/ol/wlp/usr/servers/defaultServer
+    rm -rf /opt/ol/wlp/usr/servers/defaultServer && \
+    mkdir -p /config/configDropins/defaults /config/configDropins/overrides && \
+    chown -R 1001:1001 /config/configDropins/ && \
+    yum clean all
```
lmsurpre, line 48: we don't do anything with yum in here at the moment...are you thinking to add a yum update too?

fhir-install/Dockerfile
```diff
@@ -36,15 +36,16 @@ ENV WLP_OUTPUT_DIR=/opt/ol/wlp/usr/servers
...
-# Replace links to defaultServer with links to fhir-server
+# Replace links to defaultServer with links to fhir-server and 
+# create the configDropins directories for the fhir-server
 USER root
 RUN ln -sfn ${FHIR} /output && \
     ln -sfn ${FHIR} /config && \
     rm -rf /opt/ol/wlp/output && \
-    rm -rf /opt/ol/wlp/usr/servers/defaultServer
+    rm -rf /opt/ol/wlp/usr/servers/defaultServer && \
+    mkdir -p /config/configDropins/defaults /config/configDropins/overrides && \
```
lmsurpre, line 46: I prefer this one to be down below because I prefer having this directory owned by the default user (and not root).  I don't think the extra layer is a big deal.

lmsurpre (approved):
LGTM

fhir-install/Dockerfile
prb112: OK - will remove

fhir-install/Dockerfile
prb112: Yes, I should have added it.

prb112: It's now added

Issue #1527, opened 2020-09-28
Private Memory on Db2 results in -930 errors at load


prb112:
**Describe the bug**
Private Memory on Db2 results in -930 errors at load

Further, the package cache and catalog cache runs out of space in a shared memory environment. 

**To Reproduce**
1. Go to create an instance of db2 in a shared environment such as IBM Cloud
2. Run a synthetic load with multiple bundles

```
[err] com.ibm.db2.jcc.am.SqlException: DB2 SQL Error: SQLCODE=-930, SQLSTATE=57011, SQLERRMC=null, DRIVER=4.26.14
[err] 	at com.ibm.db2.jcc.am.b7.a(b7.java:815)
[err] 	at com.ibm.db2.jcc.am.b7.a(b7.java:66)
[err] 	at com.ibm.db2.jcc.am.b7.a(b7.java:140)
[err] 	at com.ibm.db2.jcc.am.k4.b(k4.java:2471)
[err] 	at com.ibm.db2.jcc.am.k4.c(k4.java:2452)
[err] 	at com.ibm.db2.jcc.t4.ab.n(ab.java:914)
[err] 	at com.ibm.db2.jcc.t4.ab.f(ab.java:156)
[err] 	at com.ibm.db2.jcc.t4.p.e(p.java:81)
[err] 	at com.ibm.db2.jcc.t4.av.k(av.java:175)
[err] 	at com.ibm.db2.jcc.am.k4.ao(k4.java:2413)
[err] 	at com.ibm.db2.jcc.am.k5.a(k5.java:4686)
[err] 	at com.ibm.db2.jcc.am.CallableStatement.b(CallableStatement.java:134)
[err] 	at com.ibm.db2.jcc.am.CallableStatement.a(CallableStatement.java:103)
[err] 	at com.ibm.db2.jcc.am.CallableStatement.execute(CallableStatement.java:93)
[err] 	at com.ibm.ws.rsadapter.jdbc.WSJdbcPreparedStatement.execute(WSJdbcPreparedStatement.java:390)
[err] 	at com.ibm.fhir.persistence.jdbc.dao.impl.ResourceDAOImpl.insert(ResourceDAOImpl.java:528)
[err] 	at com.ibm.fhir.persistence.jdbc.impl.FHIRPersistenceJDBCImpl.create(FHIRPersistenceJDBCImpl.java:332)
[...]

PR #1677, opened 2020-11-07
Private Memory on Db2 results in -930 errors at load #1527

prb112:
This error is only on the Db2 and isolated to the Db2 stored procedure - add_any_resource.

Scenario
- Load a Bundle with multiple if not hundreds of resources in a simultaneous way within a single Connection
- Note, the Bundle can be set as Transaction or Batch.
- Db2 with 8GB of memory without INSTANCE_MEMORY=Automatic
- The simultaneous load with ~10 to 20 threads.
- The result is *SQL0930N There is not enough storage available to process the statement*.

I monitored the db2 memory under load using db2tmrk, db2pd and generated some tests which indicate the following is accurate (other memory expands to about 600GB). 

In the underlying Memory Allocation, the 2GB size cause a block of memory to be reserved, where it accumulates. So if a new large request comes up requiring substantial chunk of memory, this chunk can only be freshly acquired from so far unused range of addresses, or it may fail.  If the p_payload is lowered to 10M, then the code operates as expected and the same workload can function with 200 threads. Note, it's dependent on the number of CALL to add_any_resource.

This PR is opened to discuss in more detail on an approach.  
I'd like to lower the 2GB to 10M.  

Alternatives, 
1. If needed we can branch to a LARGE size Stored Procedure if the file is larger than 10M.
2. If needed we can raise to 100M, it'd give a decent overhead.

Reference: https://www.ibm.com/support/pages/db2-content-management-application-may-receive-sql0930n

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
@@ -57,6 +57,12 @@ import com.ibm.fhir.persistence.jdbc.util.SqlQueryData;
 public class ResourceDAOImpl extends FHIRDbDAOImpl implements ResourceDAO {
     private static final Logger log = Logger.getLogger(ResourceDAOImpl.class.getName());
     private static final String CLASSNAME = ResourceDAOImpl.class.getName();
+
+    // Per issue with private memory in db2, we have set this to 1M.
+    // Anything larger than 1M is then inserted into the db with an update.
+    private static final int SIZE_LIMIT = 1048576;
+    private static final String LARGE_BLOB = "UPDATE %s.%s_RESOURCES SET data = ? WHERE logical_resource_id = ?";
+
     public static final String DEFAULT_VALUE_REINDEX_TSTAMP = "1970-01-01 00:00:00";
 
 
@@ -535,7 +541,16 @@ public class ResourceDAOImpl extends FHIRDbDAOImpl implements ResourceDAO {
             stmt = connection.prepareCall(stmtString);
             stmt.setString(1, resource.getResourceType());
             stmt.setString(2, resource.getLogicalId());
-            stmt.setBytes(3, resource.getData());
+
+            boolean large = false;
+            if (SIZE_LIMIT < resource.getData().length) {
+                // Outside of the normal flow we have a BIG JSON or XML
+                stmt.setNull(3, java.sql.Types.NULL);
+                large = true;
+            } else {
+                // Normal Flow, we set the data
+                stmt.setBytes(3, resource.getData());
+            }
 
             lastUpdated = resource.getLastUpdated();
             stmt.setTimestamp(4, lastUpdated, UTC);
@@ -548,6 +563,19 @@ public class ResourceDAOImpl extends FHIRDbDAOImpl implements ResourceDAO {
             double dbCallDuration = (latestTime-dbCallStartTime)/1e6;
 
             resource.setId(stmt.getLong(7));
+
+            if (large) {
+                // Use the long id to update the record in the database.
+                String largeStmtString = String.format(LARGE_BLOB, getSchemaName(), resource.getResourceType());
+                PreparedStatement ps = connection.prepareStatement(largeStmtString);
+                ps.setLong(1, stmt.getLong(7));
+                long dbCallStartTime2 = System.nanoTime();
+                stmt.execute();
+                double dbCallDuration2 = (System.nanoTime() - dbCallStartTime2) / 1e6;
+                if (log.isLoggable(Level.FINE)) {
+                    log.fine("DB search by ids complete. SQL=[" + largeStmtString + "]  executionTime=" + dbCallDuration2 + "ms");
+                }
+            }
 
             // Parameter time
             // TODO FHIR_ADMIN schema name needs to come from the configuration/context
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
@@ -574,7 +574,7 @@ public class ResourceDAOImpl extends FHIRDbDAOImpl implements ResourceDAO {
                     stmt.execute();
                     double dbCallDuration2 = (System.nanoTime() - dbCallStartTime2) / 1e6;
                     if (log.isLoggable(Level.FINE)) {
-                        log.fine("DB search by ids complete. SQL=[" + largeStmtString + "]  executionTime=" + dbCallDuration2 + "ms");
+                        log.fine("DB update complete. SQL=[" + largeStmtString + "]  executionTime=" + dbCallDuration2 + "ms");
                     }
                 }
             }
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
@@ -567,13 +567,15 @@ public class ResourceDAOImpl extends FHIRDbDAOImpl implements ResourceDAO {
             if (large) {
                 // Use the long id to update the record in the database.
                 String largeStmtString = String.format(LARGE_BLOB, getSchemaName(), resource.getResourceType());
-                PreparedStatement ps = connection.prepareStatement(largeStmtString);
-                ps.setLong(1, stmt.getLong(7));
-                long dbCallStartTime2 = System.nanoTime();
-                stmt.execute();
-                double dbCallDuration2 = (System.nanoTime() - dbCallStartTime2) / 1e6;
-                if (log.isLoggable(Level.FINE)) {
-                    log.fine("DB search by ids complete. SQL=[" + largeStmtString + "]  executionTime=" + dbCallDuration2 + "ms");
+                try (PreparedStatement ps = connection.prepareStatement(largeStmtString);) {
+                    ps.setBytes(1, resource.getData());
+                    ps.setLong(2, stmt.getLong(7));
+                    long dbCallStartTime2 = System.nanoTime();
+                    stmt.execute();
+                    double dbCallDuration2 = (System.nanoTime() - dbCallStartTime2) / 1e6;
+                    if (log.isLoggable(Level.FINE)) {
+                        log.fine("DB search by ids complete. SQL=[" + largeStmtString + "]  executionTime=" + dbCallDuration2 + "ms");
+                    }
                 }
             }
 
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
@@ -56,58 +56,63 @@ import com.ibm.fhir.schema.control.FhirSchemaConstants;
  * and retrieving rows in the IBM FHIR Server resource tables.
  */
 public class ResourceDAOImpl extends FHIRDbDAOImpl implements ResourceDAO {
+
     private static final Logger log = Logger.getLogger(ResourceDAOImpl.class.getName());
     private static final String CLASSNAME = ResourceDAOImpl.class.getName();
 
     // Per issue with private memory in db2, we have set this to 1M.
     // Anything larger than 1M is then inserted into the db with an update.
-    private static final String LARGE_BLOB = "UPDATE %s_RESOURCES SET data = ? WHERE resource_id = ?";
+    private static final String LARGE_BLOB = "UPDATE %s_RESOURCES SET DATA = ? WHERE RESOURCE_ID = ?";
 
     public static final String DEFAULT_VALUE_REINDEX_TSTAMP = "1970-01-01 00:00:00";
 
-
     // Read the current version of the resource
     private static final String SQL_READ = "SELECT R.RESOURCE_ID, R.LOGICAL_RESOURCE_ID, R.VERSION_ID, R.LAST_UPDATED, R.IS_DELETED, R.DATA, LR.LOGICAL_ID " +
-                                            "FROM %s_RESOURCES R, %s_LOGICAL_RESOURCES LR WHERE " +
-                                           "LR.LOGICAL_ID = ? AND R.RESOURCE_ID = LR.CURRENT_RESOURCE_ID";
+            "FROM %s_RESOURCES R, %s_LOGICAL_RESOURCES LR WHERE " +
+            "LR.LOGICAL_ID = ? AND R.RESOURCE_ID = LR.CURRENT_RESOURCE_ID";
 
     // Read a specific version of the resource
-    private static final String SQL_VERSION_READ = "SELECT R.RESOURCE_ID, R.LOGICAL_RESOURCE_ID, R.VERSION_ID, R.LAST_UPDATED, R.IS_DELETED, R.DATA, LR.LOGICAL_ID " +
-                                                      "FROM %s_RESOURCES R, %s_LOGICAL_RESOURCES LR WHERE " +
-                                                      "LR.LOGICAL_ID = ? AND R.LOGICAL_RESOURCE_ID = LR.LOGICAL_RESOURCE_ID AND R.VERSION_ID = ?";
+    private static final String SQL_VERSION_READ =
+            "SELECT R.RESOURCE_ID, R.LOGICAL_RESOURCE_ID, R.VERSION_ID, R.LAST_UPDATED, R.IS_DELETED, R.DATA, LR.LOGICAL_ID " +
+                    "FROM %s_RESOURCES R, %s_LOGICAL_RESOURCES LR WHERE " +
+                    "LR.LOGICAL_ID = ? AND R.LOGICAL_RESOURCE_ID = LR.LOGICAL_RESOURCE_ID AND R.VERSION_ID = ?";
 
     // @formatter:off
     //                                                                                 0
-    //                                                                                 1 2 3 4 5 6 7
+    //                                                                                 1 2 3 4 5 6 7 8
     // @formatter:on
     // Don't forget that we must account for IN and OUT parameters.
-    private static final String SQL_INSERT_WITH_PARAMETERS = "CALL %s.add_any_resource(?,?,?,?,?,?,?)";
+    private static final String SQL_INSERT_WITH_PARAMETERS = "CALL %s.add_any_resource(?,?,?,?,?,?,?,?)";
 
     // Read version history of the resource identified by its logical-id
-    private static final String SQL_HISTORY = "SELECT R.RESOURCE_ID, R.LOGICAL_RESOURCE_ID, R.VERSION_ID, R.LAST_UPDATED, R.IS_DELETED, R.DATA, LR.LOGICAL_ID " +
-                                                 "FROM %s_RESOURCES R, %s_LOGICAL_RESOURCES LR WHERE " +
-                                                 "LR.LOGICAL_ID = ? AND R.LOGICAL_RESOURCE_ID = LR.LOGICAL_RESOURCE_ID " +
-                                              "ORDER BY R.VERSION_ID DESC ";
+    private static final String SQL_HISTORY =
+            "SELECT R.RESOURCE_ID, R.LOGICAL_RESOURCE_ID, R.VERSION_ID, R.LAST_UPDATED, R.IS_DELETED, R.DATA, LR.LOGICAL_ID " +
+                    "FROM %s_RESOURCES R, %s_LOGICAL_RESOURCES LR WHERE " +
+                    "LR.LOGICAL_ID = ? AND R.LOGICAL_RESOURCE_ID = LR.LOGICAL_RESOURCE_ID " +
+                    "ORDER BY R.VERSION_ID DESC ";
 
     // Count the number of versions we have for the resource identified by its logical-id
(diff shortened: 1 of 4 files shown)


prb112:
I'm hoping we can discuss this next week at some point.  No rush.

punktilious (approved):
LGTM

lmsurpre (changes requested):
This one-line change is actually a big deal because this effectively sets our max resource size on the server.  At the very least, we need to document this limit.  I'd also prefer a less drastic reduction (2GB to 10M is pretty huge).

prb112:
> This one-line change is actually a big deal because this effectively sets our max resource size on the server. At the very least, we need to document this limit. I'd also prefer a less drastic reduction (2GB to 10M is pretty huge).

I know it's a big change.  I'll leave this as WIP, I wanted to discuss with all of you. 
Yes - it's big, but it requires an architectural change otherwise.

punktilious:
Proposed solution: Limit the stored procedure to 1MB (to keep the memory requirements low). If the `data` payload is small (<1MB) then pass it to the procedure. If the payload is large, pass NULL to the procedure, and simply execute an UPDATE to the resources right after. This way, we keep the benefit of fewer round-trips to the database for the majority of cases, but aren't limiting our payload size due to the stored proc memory restrictions.

prb112:
> Proposed solution: Limit the stored procedure to 1MB (to keep the memory requirements low). If the `data` payload is small (<1MB) then pass it to the procedure. If the payload is large, pass NULL to the procedure, and simply execute an UPDATE to the resources right after. This way, we keep the benefit of fewer round-trips to the database for the majority of cases, but aren't limiting our payload size due to the stored proc memory restrictions.

updated per the proposed solution
I'll have to add a large JSON and IT

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
```diff
@@ -549,6 +564,19 @@ public Resource insert(Resource resource, List<ExtractedParameterValue> paramete
 
             resource.setId(stmt.getLong(7));
 
+            if (large) {
+                // Use the long id to update the record in the database.
+                String largeStmtString = String.format(LARGE_BLOB, getSchemaName(), resource.getResourceType());
+                PreparedStatement ps = connection.prepareStatement(largeStmtString);
+                ps.setLong(1, stmt.getLong(7));
```
lmsurpre, line 571: where does it actually set the bytes for the BLOB column?!

prb112: Thanks, silly me - I was doing this late at night

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
```diff
@@ -549,6 +564,21 @@ public Resource insert(Resource resource, List<ExtractedParameterValue> paramete
...
+                // Use the long id to update the record in the database.
+                String largeStmtString = String.format(LARGE_BLOB, getSchemaName(), resource.getResourceType());
+                try (PreparedStatement ps = connection.prepareStatement(largeStmtString);) {
+                    ps.setBytes(1, resource.getData());
+                    ps.setLong(2, stmt.getLong(7));
+                    long dbCallStartTime2 = System.nanoTime();
+                    stmt.execute();
+                    double dbCallDuration2 = (System.nanoTime() - dbCallStartTime2) / 1e6;
+                    if (log.isLoggable(Level.FINE)) {
+                        log.fine("DB search by ids complete. SQL=[" + largeStmtString + "]  executionTime=" + dbCallDuration2 + "ms");
```
prb112, line 577:
```suggestion
                        log.fine("DB update complete. SQL=[" + largeStmtString + "]  executionTime=" + dbCallDuration2 + "ms");
```

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
```diff
@@ -57,6 +57,12 @@
 public class ResourceDAOImpl extends FHIRDbDAOImpl implements ResourceDAO {
     private static final Logger log = Logger.getLogger(ResourceDAOImpl.class.getName());
     private static final String CLASSNAME = ResourceDAOImpl.class.getName();
+
+    // Per issue with private memory in db2, we have set this to 1M.
+    // Anything larger than 1M is then inserted into the db with an update.
+    private static final int SIZE_LIMIT = 1048576;
+    private static final String LARGE_BLOB = "UPDATE %s.%s_RESOURCES SET data = ? WHERE logical_resource_id = ?";
```
punktilious, line 64: That's going to overwrite the data for every version. You need to use `resource_id = ?` which is the PK for this table.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
```diff
@@ -57,6 +57,12 @@
 public class ResourceDAOImpl extends FHIRDbDAOImpl implements ResourceDAO {
     private static final Logger log = Logger.getLogger(ResourceDAOImpl.class.getName());
     private static final String CLASSNAME = ResourceDAOImpl.class.getName();
+
+    // Per issue with private memory in db2, we have set this to 1M.
+    // Anything larger than 1M is then inserted into the db with an update.
+    private static final int SIZE_LIMIT = 1048576;
```
punktilious, line 63: Move to FhirSchemaConstants (if fhir-persistence-schema is already a dependency).

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
lmsurpre, line 64: I think it should be:
```suggestion
    private static final String LARGE_BLOB = "UPDATE %s.%s_RESOURCES SET data = ? WHERE resource_id = ?";
```

Would like to see a test that creates a large resource, updates it, then reads both versions and confirms they are right.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
```diff
@@ -535,7 +541,16 @@ public Resource insert(Resource resource, List<ExtractedParameterValue> paramete
             stmt = connection.prepareCall(stmtString);
             stmt.setString(1, resource.getResourceType());
             stmt.setString(2, resource.getLogicalId());
-            stmt.setBytes(3, resource.getData());
+
+            boolean large = false;
+            if (SIZE_LIMIT < resource.getData().length) {
+                // Outside of the normal flow we have a BIG JSON or XML
+                stmt.setNull(3, java.sql.Types.NULL);
```
punktilious, line 548: You should specify the actual data-type here: java.sql.Types.BLOB.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
```diff
@@ -549,6 +564,21 @@ public Resource insert(Resource resource, List<ExtractedParameterValue> paramete
 
             resource.setId(stmt.getLong(7));
 
+            if (large) {
+                // Use the long id to update the record in the database.
+                String largeStmtString = String.format(LARGE_BLOB, getSchemaName(), resource.getResourceType());
```
punktilious, line 569: For table access, don't need to qualify with the schema name, although it doesn't hurt.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
```diff
@@ -549,6 +564,21 @@ public Resource insert(Resource resource, List<ExtractedParameterValue> paramete
...
+                try (PreparedStatement ps = connection.prepareStatement(largeStmtString);) {
```
punktilious, line 570: ; is redundant

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
```diff
@@ -549,6 +564,21 @@ public Resource insert(Resource resource, List<ExtractedParameterValue> paramete
...
+                    ps.setBytes(1, resource.getData());
+                    ps.setLong(2, stmt.getLong(7));
```
punktilious, line 572: I'd put stmt.getLong(7) into a local variable like dbResourceId just so we know what is being passed here.

fhir-persistence-schema/src/main/resources/db2/add_any_resource.sql
```diff
@@ -17,7 +17,7 @@
 -- ----------------------------------------------------------------------------
     ( IN p_resource_type                 VARCHAR( 36 OCTETS),
       IN p_logical_id                    VARCHAR(255 OCTETS), 
-      IN p_payload                          BLOB(2147483647),
+      IN p_payload                          BLOB(1M),
```
punktilious, line 20: That's definitely 1MiB not 1MB, right?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
prb112: Fixed

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
prb112: Thanks

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
prb112: Done

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
prb112: Done

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
prb112: Right - great point, fixed

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
prb112: Yeah - sometimes I get in the habit, as I often add second closeables, I've removed in this case

fhir-persistence-schema/src/main/resources/db2/add_any_resource.sql
prb112: Fixed to be explicit - I was under the impression it was 1MB (all capital)

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
```diff
@@ -547,7 +563,23 @@ public Resource insert(Resource resource, List<ExtractedParameterValue> paramete
             long latestTime = System.nanoTime();
             double dbCallDuration = (latestTime-dbCallStartTime)/1e6;
 
-            resource.setId(stmt.getLong(7));
+            long logicalResourceId = stmt.getLong(7);
```
lmsurpre, line 566: I know robin recommended to save this to a local variable, but I really don't like this variable name.  It makes it seem like it is the LOGICAL_RESOURCE_ID from this table but it is actually the RESOURCE_ID.

Aside:  I also don't love that we use the name RESOURCE_ID for this column...I'd have preferred something like `VERSIONED_RESOURCE_ID` or `RESOURCE_VERSION_ROW_ID` or similar, but we've had that since forever.
Since we probably don't want to touch the schema, maybe we can at least use that to name the variable?
```suggestion
            long versionedResourceRowId = stmt.getLong(7);
```

lmsurpre: Actually, scratch that, it *is* the logical resource id that we get back.  In this case the variable was named correctly but the update command won't.  We need to get the RESOURCE_ID (and not the LOGICAL_RESOURCE_ID) that was just written so that we can update just that row.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
```diff
@@ -178,7 +188,7 @@ protected ParameterTransactionDataImpl getTransactionData() {
 
     @Override
     public Resource read(String logicalId, String resourceType)
-            throws FHIRPersistenceDataAccessException, FHIRPersistenceDBConnectException {
```
lmsurpre, line 175: bad formatting change.  the double-indent signifies that its a continuation of the previous line so that it doesn't look like the first line of the method

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
```diff
@@ -200,7 +210,7 @@ public Resource read(String logicalId, String resourceType)
 
     @Override
     public Resource versionRead(String logicalId, String resourceType, int versionId)
-            throws FHIRPersistenceDataAccessException, FHIRPersistenceDBConnectException {
```
lmsurpre, line 197: bad formatting change.  the double-indent signifies that its a continuation of the previous line so that it doesn't look like the first line of the method

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
```diff
@@ -253,7 +265,7 @@ protected Resource createDTO(ResultSet resultSet) throws FHIRPersistenceDataAcce
 
     @Override
     public List<Resource> history(String resourceType, String logicalId, Timestamp fromDateTime, int offset, int maxResults)
-                                    throws FHIRPersistenceDataAccessException, FHIRPersistenceDBConnectException {
```
lmsurpre, line 250: bad formatting change.  should be a double-indent

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
```diff
@@ -281,13 +293,14 @@ protected Resource createDTO(ResultSet resultSet) throws FHIRPersistenceDataAcce
...
-            log.exiting(CLASSNAME, METHODNAME, Arrays.toString(new Object[] {resources}));
+            log.exiting(CLASSNAME, METHODNAME, Arrays.toString(new Object[] { resources }));
         }
         return resources;
     }
 
     @Override
-    public int historyCount(String resourceType, String logicalId, Timestamp fromDateTime) throws FHIRPersistenceDataAccessException, FHIRPersistenceDBConnectException {
+    public int historyCount(String resourceType, String logicalId, Timestamp fromDateTime)
+        throws FHIRPersistenceDataAccessException, FHIRPersistenceDBConnectException {
```
lmsurpre, line 303: double-indent

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
```diff
@@ -350,7 +363,7 @@ public void setPersistenceContext(FHIRPersistenceContext context) {
 
     @Override
     public Map<String, Integer> readAllResourceTypeNames()
-                                         throws FHIRPersistenceDBConnectException, FHIRPersistenceDataAccessException {
+        throws FHIRPersistenceDBConnectException, FHIRPersistenceDataAccessException {
```
lmsurpre, line 366: double-indent

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
```diff
@@ -720,7 +739,7 @@ public int searchCount(String sqlSelectCount) throws FHIRPersistenceDataAccessEx
 
     @Override
     public List<String> searchStringValues(SqlQueryData queryData)
-            throws FHIRPersistenceDataAccessException, FHIRPersistenceDBConnectException {
+        throws FHIRPersistenceDataAccessException, FHIRPersistenceDBConnectException {
```
lmsurpre, line 742: bad formatting change.  the double-indent signifies that its a continuation of the previous line so that it doesn't look like the first line of the method

fhir-persistence-schema/src/main/resources/db2/add_any_resource.sql
```diff
@@ -13,15 +13,17 @@
...
-      IN p_is_deleted                       CHAR(  1),
-      IN p_version                           INT,
-      OUT o_logical_resource_id            BIGINT
+    ( IN p_resource_type                VARCHAR( 36 OCTETS),
+      IN p_logical_id                   VARCHAR(255 OCTETS), 
+      IN p_payload                         BLOB(1048576),
+      IN p_last_updated               TIMESTAMP,
+      IN p_is_deleted                      CHAR(  1),
+      IN p_version                          INT,
+      OUT o_logical_resource_id          BIGINT,
```
lmsurpre, line 25: I'm still confused about how this can be a BIGINT.  I guess its *not* the same as the `Resource.id` field of the resource?

lmsurpre: ok, I looked this up.  its not the first time I've been tripped up on this:
```
X_LOGICAL_RESOURCES.LOGICAL_RESOURCE_ID = the unique key of the X_LOGICAL_RESOURCES table
X_LOGICAL_RESOURCES.LOGICAL_ID = the value of the Resource.id field for this resource
```

so o_logical_resource_id corresponds to `X_LOGICAL_RESOURCES.LOGICAL_RESOURCE_ID`

lmsurpre (approved):
left some formatting comments, but otherwise looks good

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
```diff
@@ -535,19 +558,46 @@ public Resource insert(Resource resource, List<ExtractedParameterValue> paramete
...
             stmt.registerOutParameter(7, Types.BIGINT);
+            stmt.registerOutParameter(8, Types.BIGINT);
 
             stmt.execute();
             long latestTime = System.nanoTime();
             double dbCallDuration = (latestTime-dbCallStartTime)/1e6;
 
             resource.setId(stmt.getLong(7));
+            long versionedResourceRowId = stmt.getLong(8);
+            if (large) {
```
punktilious, line 585: && resource.getData() != null

fhir-persistence-schema/src/main/resources/db2/add_any_resource.sql
```diff
@@ -198,4 +200,7 @@ BEGIN
   -- only the logical_resource_id is the target of any FK, so there's no need to return
```
punktilious, line 200: This comment no longer applied

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
prb112: it would have thrown a NPE prior to this.

prb112:
Updated per review

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
```diff
@@ -142,8 +142,7 @@
      * @param trxSyncRegistry
      */
     public ResourceDAOImpl(Connection c, String schemaName, FHIRDbFlavor flavor, TransactionSynchronizationRegistry trxSynchRegistry,
-            FHIRPersistenceJDBCCache cache, IResourceReferenceDAO rrd,
-            ParameterTransactionDataImpl ptdi) {
+        FHIRPersistenceJDBCCache cache, IResourceReferenceDAO rrd, ParameterTransactionDataImpl ptdi) {
```
lmsurpre, line 145: double-indent pls

punktilious (approved):

Issue #1673, opened 2020-11-06
Observation search performing slowly in PostgreSQL on IBM Cloud


lmsurpre:
I've loaded 1.5 million observations in an instance of the IBM Cloud Databases for Postgresql and given it 6 dedicated cores and 8 GB of RAM.

When I perform a query like `GET [base]/Observation?_summary=count` the first time, it takes about 30 seconds, then between 3 and 4 seconds after that.

However, when I remove the `_summary=count` and query just `GET [base]/Observation`, this jumps to 17-18 seconds (consistently).

I understand that the count might be expensive because we need to look at all the data, but I think https://github.com/IBM/FHIR/issues/1353 will make that better (so at least clients can avoid that cost for some queries if desired).

However, what concerns me is that fetching the first page of Observations (10 by default) is taking the other 12-13 seconds on top of that.  I would have expected that to be much faster.  Anything we can do?

P.S. I thought maybe with a more selective query it would be faster, but `Observation?code=http://loinc.org|85354-9` is taking about the same amount of time.

**To Reproduce**
1. Create a postgres database on IBM Cloud
2. Scale it to the levels mentioned above
3. Load 1.5 million observations (e.g. using fhir-bucket)
4. Perform search

**Expected behavior**
Non-count queries should return a page of results in 2 seconds or less.

**Additional context**
I'm not sure if its an issue with our IBM Cloud PostgreSQL in general, this specific instance, or if this is really just the expected performance.  I thought it was faster in 4.4.2 but I really don't know.

lmsurpre:
will likely benefit from #1385

lmsurpre:
the current query is something like this:
```
SELECT R.RESOURCE_ID,
	R.LOGICAL_RESOURCE_ID,
	R.VERSION_ID,
	R.LAST_UPDATED,
	R.IS_DELETED,
	R.DATA,
	LR.LOGICAL_ID
FROM OBSERVATION_RESOURCES R
JOIN
				(SELECT DISTINCT LR.LOGICAL_RESOURCE_ID,
						LR.LOGICAL_ID,
						LR.CURRENT_RESOURCE_ID
					FROM OBSERVATION_LOGICAL_RESOURCES LR
					JOIN OBSERVATION_RESOURCES R ON R.RESOURCE_ID = LR.CURRENT_RESOURCE_ID
[...]

PR #1678, opened 2020-11-09
issue #1673 - change default ordering for resource searches

lmsurpre:
Note: I'm opening this PR against Mike's branch for #1218 instead of master because I was already under that branch when I made the change.  If needed, I can cherry-pick it into a new branch off master, but my assumption is that we'll merge Mike's PR soon anyway and so it doesn't matter much.

Previously, we used a default sort of RESOURCE_ID ASC (the row identifier of the resource version being returned).
While investigating #1673, I noticed that removing this sort took this query from 10 seconds down to 3-4 seconds.

However, having *some* sort order is important, so that we get consistent page results.
Upon further inspection, I found that changing the default sort to LOGICAL_RESOURCE_ID (the row identifier of the logical resource row being returned)
yielded a similar performance boost and so this is the proposed change.

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/JDBCConstants.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/JDBCConstants.java
@@ -96,7 +96,7 @@ public class JDBCConstants {
     public static final String ASCENDING = "ASC";
     public static final String DESCENDING = "DESC";
 
-    public static final String DEFAULT_ORDERING = " ORDER BY RESOURCE_ID ASC ";
+    public static final String DEFAULT_ORDERING = " ORDER BY LOGICAL_RESOURCE_ID ASC ";
     public static final String DEFAULT_ORDERING_WITH_TABLE = " ORDER BY R.RESOURCE_ID ASC ";
 
     // MIN / MAX


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/JDBCConstants.java
```diff
@@ -96,7 +96,7 @@
     public static final String ASCENDING = "ASC";
     public static final String DESCENDING = "DESC";
 
-    public static final String DEFAULT_ORDERING = " ORDER BY RESOURCE_ID ASC ";
+    public static final String DEFAULT_ORDERING = " ORDER BY LOGICAL_RESOURCE_ID ASC ";
     public static final String DEFAULT_ORDERING_WITH_TABLE = " ORDER BY R.RESOURCE_ID ASC ";
```
prb112, line 100: just noticing this line, should it be updated?

prb112, line 100: just noticing this line, should it be updated?

prb112 (approved):
LGTM - with comment on

michaelwschroeder (approved):

michaelwschroeder:
LGTM

PR #1685, opened 2020-11-09
Expand visibility for ParametersMap

lmsurpre:
This allows it to be re-used in other places, which is especially useful
for the logic within ParametersMap.insert for detecting search code and
parameter definition url conflicts.


prb112 (approved):
LGTM

PR #1696, opened 2020-11-11
fix: labels are incorrect due to base image being from open-=libery and need to be overridden

prb112:


lmsurpre (approved):
LGTM

Issue #1695, opened 2020-11-11
Dockerfile license needs to be refreshed with IBM FHIR Server license


prb112:
**Describe the bug**
Dockerfile license needs to be refreshed with IBM FHIR license
Also make sure we only have one copy of license, and copy into the appropriate locations

*** FYI *** Requires some degree of automation.
Also link in the build-arg FHIR_VERSION

Also confirm that all licenses are properly referenced

**To Reproduce**
Dockerfile license needs to be refreshed with IBM FHIR license

**Expected behavior**
Needs to be the IBM FHIR Server license

**Additional context**
n/a

lmsurpre:
We should also investigate the proper way to list the license for dependencies, especially for the db2 jdbc driver that we package which is freely available from https://www.ibm.com/support/pages/db2-jdbc-driver-versions-and-downloads but not open-source.

PR #1697, opened 2020-11-11
fix: consistency with licenses #1695

prb112:


lmsurpre (approved):
eventually i'd like to see this come from https://github.com/IBM/FHIR/blob/master/LICENSE but LGTM

PR #1699, opened 2020-11-11
doc: escape the html block elements that should have been in code ele…

prb112:
…ment


punktilious (approved):
LGTM

Issue #1705, opened 2020-11-12
Search with reverse chaining performing slowly


d0roppe:
**Describe the bug**
The following search with _has takes over 2 minutes to run in the large postgres DB.
[base]/Patient?_has:Observation:patient:code-value-quantity=http://loinc.org|2339-0$68.6

**To Reproduce**

**Expected behavior**
The search should return faster than 2 minutes.   in most deploys it would timeout.

**Additional context**

lmsurpre:
I saw similar when we merged the initial PR for this feature:  https://github.com/IBM/FHIR/pull/1666#issuecomment-723742963
Will be great if we can speed it up.

d0roppe:
Running this search today with the latest code loaded on my local FHIR server linked to the large postgres DB.  this is now taking about 3.6 seconds to complete.

lmsurpre:
Closing based on Dag's comment.  Future enhancements can come as a result of #677 findings / fixes.

PR #1714, opened 2020-11-13
issue #1705 - simplified top-level join avoids bad postgresql plan for simple _has queries

lmsurpre:
The `R.LOGICAL_RESOURCE_ID = LR.LOGICAL_RESOURCE_ID` part of this join
is unnecessary because R.RESOURCE_ID is the unique identifier on the
Resources (R) table...so if LR.CURRENT_RESOURCE_ID = R.RESOURCE_ID then
this RESOURCES row will always have a LOGICAL_RESOURCE_ID which matches
the LOGICAL_RESOURCES row.

Making this change changed the processing results of the following query
from over 5 minutes down to under 1 second on my postgresql db with
~30,000 patients and ~300,000 conditions:
`GET [base]/Patient?_has:Condition:patient:code=http://snomed.info/sct|44054006`


michaelwschroeder (approved):
LGTM

punktilious (approved):
LGTM

Issue #1390, opened 2020-07-30
Refactoring of the fhirServer/bulkdata configuration


prb112:
**Is your feature request related to a problem? Please describe.**
Refactoring of the `fhirServer/bulkdata` Configuration

**Describe the solution you'd like**
- Refactor the bulkdata configuration for the following: 
1 - Multiple COS sources/destinations
2 - Remove unneeded parameters
3 - Remove internal code artifacts JobParameters and make it a bit more readable. 

**Describe alternatives you've considered**
None as of yet.

**Additional context**
None

lmsurpre:
There's a bit more I'd like to change in the bulkdata config as well.  For example, currently to use "HMAC" authentication with IBM Cloud Object Storage you have to set:
* iamAuthentication to false
* access token to the assigned access key
* service id to the assigned secret key

Instead, I think we should default to using HMAC authentication and introduce dedicated parameters that match the terms used by S3.

lmsurpre:
Testing all possible variations of all the config parameters is too difficult right now.  Instead, I will review the overall changes and help verify that the "multiple storageProvider" configs is working as expected.

lmsurpre:
Found and fixed a doc issue for specifying a custom outcome: #2085
Also opened https://github.com/IBM/FHIR/issues/2083 for some additional cleanup we could perform writing OperationOutcomes.

Overall, I think the reworked config changes are really good and I confirmed that the config options are covered in the User's Guide.

To me, just being able to set a per-tenant default for input, output, and outcomes is probably sufficient for the forseeable future…the ability to specify additional ones on a per-request basis (similar to our multi-datastore support) is just a nice bonus if folks end up needing it down the line.  https://github.com/IBM/FHIR/issues/2036 is the issue I opened for supporting the separate defaults via config.

PR #1735, opened 2020-11-19
Split COS endpointurl into internal and external

lmsurpre:
When running on VPC, the cos URL used by the application should be the
"direct" one.  However, we still need to use the "public" one for the
donwload urls in the final response.


prb112 (approved):
LGTM

lmsurpre:
Its not super clean because the externalUrl really doesn't need to be a job property.  We only need it at the very end while building the download urls.
However, this was just the fastest/easiest way I could think of to make it work for the DevDays event, and I think we have a separate task for cleaning up the config properties for bulkdata section anyway:
https://github.com/IBM/FHIR/issues/1390

I don't think I have time to clean it up now, so if needed we can just keep this one pending until I get back from vacation in December.

Issue #1629, opened 2020-10-27
No error message issued when passing in a incorrect resource on _revinclude.


d0roppe:
**Describe the bug**
When searching with the following get statement:
https://localhost:7000/fhir-server/api/v4/Patient?_id=<patient-id>&_revinclude=Provedance:*
there is no resource Provedance, but the fhir server did not issue an error.  it just dropped it and processed the search.
 
"url": "https://localhost:9443/fhir-server/api/v4/Patient?_count=10&_id=<patient-id>&_page=1"

**To Reproduce**

**Expected behavior**
With the fact that the resource type is an error it should give a error message vs dropping the _revinclude

**Additional context**

kmbarton423:
lenient vs strict mode consideration

michaelwschroeder:
If running in strict mode, an error will be returned if an invalid target resource type is specified in a `_revinclude` query parameter. If running in lenient mode, the error message will be logged and the invalid `_revinclude` query parameter will be ignored.

In addition, changes will be made to `_include`/`_revinclude` processing to handle errors more consistently when running in lenient mode. Currenty, if multiple `_include` or `_revinclude` query parameters are specified and an error is found in one of them, all will be ignored, with the exception of a bad search parameter being found. In that case, only the query parameter with the bad search parameter is ignored. With the new changes, when running in lenient mode, if an error is found processing a query parameter, only that query parameter will be ignored. Any valid `_include` or `_revinclude` query parameters will be included in the search query. Note that this applies only if running in lenient mode. If running in strict mode, any error found will result in a failed search request.

d0roppe:
Verified that in strict mode now this throws an error 400 and does not process the search. And in lenient mode the error part of the command is ignored and dropped, and the search is processed.

PR #1737, opened 2020-11-19
Issue #1629 - return error if bad resource type on revinclude

michaelwschroeder:

### DIFF
--- a/fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
+++ b/fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
@@ -700,14 +700,8 @@ public class SearchUtil {
                         if (ModelSupport.isResourceType(resType)) {
                             resourceTypes.add(resType);
                         } else {
-                            String msg = "_type search parameter has invalid resource type:" + resType;
-                            if (lenient) {
-                                // TODO add this to the list of supplemental warnings?
-                                log.log(Level.FINE, msg);
-                                continue;
-                            } else {
-                                throw SearchExceptionUtil.buildNewInvalidSearchException(msg);
-                            }
+                            manageException("_type search parameter has invalid resource type:" + resType, lenient);
+                            continue;
                         }
                     }
                 }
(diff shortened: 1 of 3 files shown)


michaelwschroeder:
In addition to now issuing error on bad target resource type, I've also made behavior consistent when running in lenient mode - see comment in the issue for details.

punktilious (approved):
LGTM

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -700,14 +700,8 @@ public static FHIRSearchContext parseQueryParameters(Class<?> resourceType,
                         if (ModelSupport.isResourceType(resType)) {
                             resourceTypes.add(resType);
                         } else {
-                            String msg = "_type search parameter has invalid resource type:" + resType;
-                            if (lenient) {
-                                // TODO add this to the list of supplemental warnings?
```
lmsurpre, line 705: I noticed we lost a couple TODOs in the scope of these changes.  We should probably talk about these and possibly open an enhancement request so that we don't lose this thought.
The idea is to include a OperationOutcome with an Issue of type warning with information about the ignored parameter.  However, I think that clients would only receive that OperationOutcome if either:
A. they specify a "return preference" of OperationOutcome (via the Prefer header); or
B. something else goes wrong causing us to return an OperationOutcome

For search, we would need to set the Bundle.entry.search.mode value of this OperationOutcome entry to `outcome`.

fhir-search/src/test/java/com/ibm/fhir/search/test/InclusionParameterParseTest.java
```diff
@@ -34,7 +34,7 @@
 /**
  * This TestNG test class contains methods that test the parsing of search result inclusion parameters (_include and
  * _revinclude) in the SearchUtil class.
- * 
+ *
  * @author markd
  * @author pbastide
```
lmsurpre, line 39: Once we went open source, we decided to remove @author tags from the code, but rather than all at once we just remove them as we find them.
```suggestion
```

lmsurpre (approved):
left a comment and a suggestion, but the changes look good

fhir-search/src/test/java/com/ibm/fhir/search/test/InclusionParameterParseTest.java
michaelwschroeder: done

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
michaelwschroeder: I think this makes sense. I assume it would apply to all instances where we're in lenient mode and we ignore a parameter. Are there other cases where we ignore parameters?

PR #1758, opened 2020-11-30
fix for nats error when greater than 1Mb per https://github.com/nats-…

prb112:
…io/nats-streaming-server/issues/745 and https://docs.nats.io/nats-server/configuration


lmsurpre (approved):
We're hoping that this will prevent the CI error which causes our `fhir-kafka-publisher` to serialize a large object that we use when testing, which polutes our CI output.  LGTM.

PR #1763, opened 2020-12-01
Update to PDEX Drug Formulary 1.0.1

lmsurpre:
This version isn't officially published yet, but the artifacts I pulled
from http://build.fhir.org/ig/HL7/davinci-pdex-formulary/ are expected
to be the ones published.

I updated the ResourceProcessor to fix the provided CapabilityStatement
and removed the other edits which are no longer needed.
I also removed the XML examples (of little value) and the IndexGenerator
which is no longer needed (now that the IG comes with a proper index).


prb112 (approved):
Makes sense to remove the XML samples.  looks good.

Issue #1724, opened 2020-11-18
Docker image of version 4.5.0 contains conflicting datasource definition files under `configDropins/overrides`


holger-stenzhorn:
**Describe the bug**
The Docker image of version 4.5.0 contains conflicting datasource definition files under `configDropins/overrides`. The `README.md` within that directory actually says...
> Just one of these datasource definition files should be copied into the target Liberty configDropins/overrides folder. If more than one of these is present at the same time it will break the Liberty configuration because the datasource ids and JNDI location are common among the files.

**To Reproduce**
n/a

**Expected behavior**
All datasource definition files now in `configDropins/overrides` should be located under `configDropins/disabled`.

**Additional context**
n/a

prb112:
Thank you for the bug, we will fix this in the docker image.

PR #1765, opened 2020-12-01
fix: automation update and modification for readme #1724

prb112:
- datasource.xml is created as the default which enables bootstrapping
- datasource-bootstrap.xml changed to reflect the referenced datasource
- updated README.md in the disabled folder to reflect the appropriate
behavior and expectation
- update the automation to reflect the dual usage.
	- copy-server-config.sh controls db2 is updated to use old method
	- postgres pre-integration-test.sh uses new method
	- pre-integration-test.sh uses new method (default derby)
- Update the User Guide to reflect the changes


build/persistence/postgres/pre-integration-test.sh
```diff
@@ -50,11 +50,12 @@ copy_server_config(){
...
-    echo "Replacing datasource content in server configDropins..."
-    rm -f $DIST/overrides/datasource-*.xml 2> /dev/null
+    # Note the overrides folder is specifically mounted to the docker image under configDropins/overrides
+    echo "Creating an overrides folder in $DIST"
     mkdir -p $DIST/overrides
-    cp -p ${WORKSPACE}/fhir-server/liberty-config/configDropins/overrides/datasource-bootstrap.xml $DIST/overrides
-    cp -p ${WORKSPACE}/fhir-server/liberty-config/configDropins/overrides/datasource-postgresql.xml $DIST/overrides
+    
+    # Copy and rename the postgres override
+    cp -p ${WORKSPACE}/fhir-server/liberty-config/configDropins/disabled/datasource-postgresql.xml $DIST/overrides/datasource.xml
```
prb112, line 58: I think this is what failed...

docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -450,7 +450,7 @@ The 'dataSourceJndiName' property can be removed as it is only used when 'enable
 
 When configured explicitly, the jndiName does not have to follow the standard naming convention, although this is not recommended.
 
-Continuing the above example, configure a drop-in server configuration file 'configDropins/overrides/datastore-default.xml' as follows:
+Continuing the above example, configure a drop-in server configuration file 'configDropins/overrides/datastore.xml' as follows:
```
lmsurpre, line 453: we're mixing datasource and datastore again...probably should match the others:
```suggestion
Continuing the above example, configure a drop-in server configuration file 'configDropins/overrides/datasource.xml' as follows:
```

build/persistence/postgres/pre-integration-test.sh
lmsurpre: I think we want the derby (bootstrap) config too because some of the server tests run against different tenants and, for reduced build times, we only provision a postgresql schema for the default tenant.

lmsurpre: never mind, I see the bootstrap datasources are included in the configDropins/defaults

docs/src/pages/guides/FHIRServerUsersGuide.md
lmsurpre: I included this change in my commit

build/persistence/postgres/pre-integration-test.sh
lmsurpre: I think I found the issue: a bad change to the datasource definitions.  I think the datasource properties need to be a peer of the jdbcDriver, not a child.  Also fixed indentations so we hopefully don't make that mistake again.

lmsurpre (approved):
LGTM

punktilious (approved):
LGTM

PR #1770, opened 2020-12-01
Corrected the command flag for tenant name reuse step

csandersdev:
Updated from --remove-tenant-meta to --delete-tenant-meta to reflect the actual flag in the CLI interface.


lmsurpre (approved):
GitHub Actions having some issues right now, but I confirmed the change is docs-only; should be safe to merge.

punktilious (approved):
LGTM. Thanks!

PR #1774, opened 2020-12-02
ci: update docker-push.sh to manage the paths to workspace correctly

prb112:

### DIFF
--- a/build/release/docker-push.sh
+++ b/build/release/docker-push.sh
@@ -11,7 +11,8 @@ docker login -u ${DOCKERHUB_USERNAME} -p ${DOCKERHUB_TOKEN}
 
 # Executes the build for the ibm-fhir-schematool
 # This build runs first so that the FAILURE comes before and stops the subsequent builds/releases.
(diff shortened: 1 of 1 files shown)


build/release/docker-push.sh
```diff
@@ -11,7 +11,8 @@ docker login -u ${DOCKERHUB_USERNAME} -p ${DOCKERHUB_TOKEN}
 
 # Executes the build for the ibm-fhir-schematool
 # This build runs first so that the FAILURE comes before and stops the subsequent builds/releases.
-${WORKSPACE}/fhir-install/src/main/docker/ibm-fhir-schematool/build.sh
+export WORKSPACE=$(pwd)
```
lmsurpre, line 14: is the export needed here if we're not using it.  wondering if we should export it in a parent shell script instead...

lmsurpre (approved):
LGTM

build/release/docker-push.sh
prb112: I opted to do it here, as I knew it worked relatively when I tested locally. 
Normally I have it at the workflow level, it's an option here.

PR #1779, opened 2020-12-02
fix: postgres/db2 refresh of the stored procedure

prb112:
- issue updating the stored procedure and function
- Verify the signature of the postgres procedure/function
```
SELECT md5(prosrc),proname FROM pg_proc WHERE proname =
'add_any_resource';
```

### DIFF
--- a/fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
+++ b/fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
@@ -374,6 +374,10 @@ public class Main {
         if (!MULTITENANT_FEATURE_ENABLED.contains(dbType) && newDb ) {
             populateResourceTypeAndParameterNameTableEntries(null);
         }
+
+        // Let's refresh the procedures and functions.
+        logger.info("Refreshing procedures and functions");
+        updateProcedures();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


michaelwschroeder (approved):

michaelwschroeder:
LGTM

punktilious (approved):
LGTM

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
```diff
@@ -374,6 +374,10 @@ protected void updateSchema() {
         if (!MULTITENANT_FEATURE_ENABLED.contains(dbType) && newDb ) {
             populateResourceTypeAndParameterNameTableEntries(null);
         }
+
+        // Let's refresh the procedures and functions.
+        logger.info("Refreshing procedures and functions");
+        updateProcedures();
```
lmsurpre, line 380: I was under the impression that the stored procs (or functions for postgres) should be part of the physical data model already and that they should get "refreshed" as part of the normal applyModel.
Obviously that wasn't working quite right, but now I'm not sure if it was because my understanding was wrong or if a premature commit (the change below) was the root cause.
I'm OK with adding this in to cover us (so for sure the issue will be addressed), but it would be good to know whether applyModel above is working properly or not wrt updating procedures.

lmsurpre (approved):
LGTM as a quick fix, but left a comment to hopefully better understand the behavior

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
prb112: yes, that's right they should however it doesn't exactly work on postgres, it does on db2.

PR #1782, opened 2020-12-03
fix: version history is case sensitive, and fails when customers chan…

prb112:
…ge the case sensisitivity of their schema names

### DIFF
--- a/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/GetLatestVersionDAO.java
+++ b/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/GetLatestVersionDAO.java
@@ -50,7 +50,7 @@ public class GetLatestVersionDAO implements IDatabaseSupplier<Map<String,Integer
         final String cols = DataDefinitionUtil.join(SchemaConstants.SCHEMA_NAME, SchemaConstants.OBJECT_TYPE, SchemaConstants.OBJECT_NAME);
         final String sql = "SELECT " + cols
                 + ", max(" + SchemaConstants.VERSION + ") FROM " + tbl
-                + " WHERE UCASE(" + SchemaConstants.SCHEMA_NAME + ") IN (UCASE(?), UCASE(?)) "
+                + " WHERE UPPER(" + SchemaConstants.SCHEMA_NAME + ") IN (UPPER(?), UPPER(?)) "
                 + " GROUP BY " + cols;
 
         // Fetch the current version history information for both the admin and specified data schemas
--- a/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/GetLatestVersionDAO.java
+++ b/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/GetLatestVersionDAO.java
@@ -50,7 +50,7 @@ public class GetLatestVersionDAO implements IDatabaseSupplier<Map<String,Integer
         final String cols = DataDefinitionUtil.join(SchemaConstants.SCHEMA_NAME, SchemaConstants.OBJECT_TYPE, SchemaConstants.OBJECT_NAME);
         final String sql = "SELECT " + cols
                 + ", max(" + SchemaConstants.VERSION + ") FROM " + tbl
-                + " WHERE " + SchemaConstants.SCHEMA_NAME + " IN (?, ?) "
+                + " WHERE UCASE(" + SchemaConstants.SCHEMA_NAME + ") IN (UCASE(?), UCASE(?)) "
                 + " GROUP BY " + cols;
 
         // Fetch the current version history information for both the admin and specified data schemas
@@ -59,7 +59,7 @@ public class GetLatestVersionDAO implements IDatabaseSupplier<Map<String,Integer
             ps.setString(2, schemaName);
             ResultSet rs = ps.executeQuery();
             while (rs.next()) {
-                String schema = rs.getString(1);
+                String schema = rs.getString(1).toUpperCase();
                 String type = rs.getString(2);
                 String name = rs.getString(3);
                 int version = rs.getInt(4);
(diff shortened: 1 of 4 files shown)


fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/VersionHistoryService.java
```diff
@@ -126,7 +126,7 @@ public void insertVersionHistoriesInTx(Collection<TypeNameVersion> versionHistor
      * @param version
      */
     public void insertVersionHistoryInTx(String objectSchema, String objectType, String objectName, int version) {
-        AddVersionDAO dao = new AddVersionDAO(adminSchemaName, objectSchema, objectType, objectName, version);
+        AddVersionDAO dao = new AddVersionDAO(adminSchemaName, objectSchema.toUpperCase(), objectType, objectName, version);
```
prb112, line 129: Going forward this would always ensure the schemaname is uppercase in the VHS table

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/VersionHistoryService.java
```diff
@@ -204,7 +204,7 @@ public void addVersion(String objectSchema, String objectType, String objectName
 
     @Override
     public boolean applies(String objectSchema, String objectType, String objectName, int changeVersion) {
-        String key = objectSchema + ":" + objectType + ":" + objectName;
+        String key = objectSchema.toUpperCase() + ":" + objectType + ":" + objectName;
```
prb112, line 207: Make sure we are forcing any incoming applies to be uppercase

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/GetLatestVersionDAO.java
```diff
@@ -59,7 +59,7 @@ public GetLatestVersionDAO(String adminSchemaName, String schemaName) {
             ps.setString(2, schemaName);
             ResultSet rs = ps.executeQuery();
             while (rs.next()) {
-                String schema = rs.getString(1);
+                String schema = rs.getString(1).toUpperCase();
```
prb112, line 62: I threw this in as uppercase for consistency

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/GetLatestVersionDAO.java
```diff
@@ -50,7 +50,7 @@ public GetLatestVersionDAO(String adminSchemaName, String schemaName) {
         final String cols = DataDefinitionUtil.join(SchemaConstants.SCHEMA_NAME, SchemaConstants.OBJECT_TYPE, SchemaConstants.OBJECT_NAME);
         final String sql = "SELECT " + cols
                 + ", max(" + SchemaConstants.VERSION + ") FROM " + tbl
-                + " WHERE " + SchemaConstants.SCHEMA_NAME + " IN (?, ?) "
+                + " WHERE UCASE(" + SchemaConstants.SCHEMA_NAME + ") IN (UCASE(?), UCASE(?)) "
```
prb112, line 53: I wrapped this in ucase for backwards compatiblity

prb112: swapped to  UPPER

punktilious: I don't think you should UCASE(?) the args like that...because it suggests you don't know what case the args are. We should be making sure the args are upper-case before that (because it's probably important for equals and HashMap lookups in other places). The first place I'd force case is when we parse the CLI argument in Main.

prb112: OK -  I sort of agree, However there is an inconsistency across versions I'd like too account for,  also schema names in postgres tend  to be lowercase versus uppercase  in db2.

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
```diff
@@ -146,11 +138,6 @@
     private boolean dropJavaBatchSchema = false;
     private boolean grantJavaBatchSchema = false;
 
-    // By default, the dryRun option is OFF, and FALSE
-    // When overridden, it simulates the actions.
-    @SuppressWarnings("unused")
-    private boolean dryRun = false;
-
```
prb112, line 153: I removed it as it was never used

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/GetLatestVersionDAO.java
lmsurpre: I was thinking normalizing casing here was critical for backwards compatibility.  If there is an existing version history service value thats been saved as lowercase, we need the new uppercase schema name to match it, right?

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/GetLatestVersionDAO.java
punktilious: Performance note: using WHERE UPPER(my_col) = ... can have performance implications because it often means that you can't use an index (on my_col). But in this case, it's unlikely to have a significant impact, because the row counts are relatively low.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/GetLatestVersionDAO.java
prb112: Correct, I do that when we process the result sets down further on in

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/GetLatestVersionDAO.java
prb112: good point.

punktilious (approved):
LGTM. I've made a mental note that we probably need some unit tests around the schema name handling and version history stuff.

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
lmsurpre: sounds good, though I'd still like to add it back in at some point!

lmsurpre (approved):
LGTM

Issue #1785, opened 2020-12-03
encode the error messages from FHIRProvider with Encode.forHtml


prb112:
encode the error messages from FHIRProvider with Encode.forHtml

PR #1786, opened 2020-12-03
fix: encode the error messages from FHIRProvider with Encode.forHtml

prb112:
- encode the error messages from FHIRProvider with Encode.forHtml #1785


JohnTimm:
Do we need to encode the path? The path is generated internally will never include external input.

JohnTimm:
Also, there may be other place we need to look out for where we mirror unrecognized element names, etc.

prb112:
> Do we need to encode the path? The path is generated internally and will never include external input.

I figure it didn't hurt to encode both.

prb112:
> Also, there may be other places we need to look out for where we mirror unrecognized element names, etc.

There might be other places, maybe we take a backlog item for that?

JohnTimm (approved):
LGTM

Issue #1790, opened 2020-12-03
FHIR_ADMIN.TENANTS should enforce uniqueness for tenant schema name


punktilious:
The schema name associated with a tenant is not tracked in the FHIR_ADMIN.TENANTS table. This can cause problems in Db2 multi-tenant scenarios if the wrong schema name is used in a schema update command, resulting in a new set of tables being created in a different schema for the same tenant.

By adding the schema_name to the FHIR_ADMIN.TENANTS table, the schema tool will know exactly which schema to use for a given tenant, without having to rely on a (possibly incorrect) CLI argument.

lmsurpre:
Relates to #1971

lmsurpre:
Should we allow a single tenant to have multiple schemas?
This would be needed to support multi-datastore tenants (single tenant with two different datastores in different schemas within the same db).

Issue #1789, opened 2020-12-03
Schema tool --refresh-tenants does not handle Db2 multi-tenant/multi-schema scenarios


punktilious:
With Db2, users must execute the following command after updating the schema for a release:

```
java -jar ./fhir-persistence-schema-<RELEASE-ID>-cli.jar --prop-file db2.properties --schema-name <SCHEMA-NAME> --refresh-tenants
```

This command processes each tenant from the FHIR_ADMIN.TENANTS table and adds any missing data partitions for tables which were added by the prior schema update step. This process is incorrect for databases with tenants spread over multiple schemas because the incorrect schema name is being used when attempting to add the new partitions.

The code is using the schema name given on the command-line instead of the schema name associated with the particular tenant.

The solution is to build the PhysicalDataModel individually for each tenant as it is processed, thus ensuring the correct schema name is used for each Table object.

Issue #1787, opened 2020-12-03
Incorrect classname used for logger in Db2ResourceReferenceDAO


punktilious:
The logger is initialized with the wrong classname:
```
public class Db2ResourceReferenceDAO extends ResourceReferenceDAO {
    private static final Logger logger = Logger.getLogger(PostgresResourceReferenceDAO.class.getName());
```

This may cause extreme panic and despair in developers when they see "postgres" log messages for a Db2 database interaction.

PR #1798, opened 2020-12-05
issue #1789 refresh-tenants does not handle Db2 multi-tenant/multi-schema scenarios

punktilious:
Also includes fixes for:

#1787 - incorrect classname in logger
#1790 - FHIR_ADMIN.TENANTS should enforce uniqueness for tenant schema name


prb112 (approved):
LGTM - makes sense

Issue #1803, opened 2020-12-07
IBM FHIR Server audit does not log bundles correctly.


prb112:
**Describe the bug**
IBM FHIR Server audit does not log bundles correctly.

**To Reproduce**
1. Setup FHIR Audit 
2. Post a Bundle with Audit Data
3. Never receive the resources

**Expected behavior**
Bundle should appear.

**Additional context**
n/a

PR #1804, opened 2020-12-07
IBM FHIR Server audit does not log bundles correctly. #1803

prb112:

### DIFF
--- a/fhir-audit/src/main/java/com/ibm/fhir/audit/logging/impl/WhcAuditCadfLogService.java
+++ b/fhir-audit/src/main/java/com/ibm/fhir/audit/logging/impl/WhcAuditCadfLogService.java
@@ -82,7 +82,7 @@ public class WhcAuditCadfLogService implements AuditLogService {
     private static CadfResource observerRsrc =
             new CadfResource.Builder("fhir-server", ResourceType.compute_node)
                             .geolocation(new CadfGeolocation.Builder(geoCity, geoState, geoCountry, null).build())
-                            .name("Fhir Audit")
+                            .name("IBM FHIR Server - Audit")
                             .host(System.getenv("HOSTNAME"))
                             .build();
 
@@ -174,7 +174,6 @@ public class WhcAuditCadfLogService implements AuditLogService {
         }
 
         logger.exiting(CLASSNAME, METHODNAME);
-
     }
 
     @Override
@@ -186,7 +185,7 @@ public class WhcAuditCadfLogService implements AuditLogService {
      * @param logEntry
      * @return
      * @throws IllegalStateException
-     * @throws IOException 
+     * @throws IOException
      */
     public static CadfEvent createCadfEvent(AuditLogEntry logEntry)
             throws IllegalStateException, IOException {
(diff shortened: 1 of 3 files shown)


fhir-audit/src/main/java/com/ibm/fhir/audit/logging/impl/WhcAuditCadfLogService.java
```diff
@@ -82,7 +82,7 @@
     private static CadfResource observerRsrc =
             new CadfResource.Builder("fhir-server", ResourceType.compute_node)
                             .geolocation(new CadfGeolocation.Builder(geoCity, geoState, geoCountry, null).build())
-                            .name("Fhir Audit")
+                            .name("IBM FHIR Server - Audit")
                             .host(System.getenv("HOSTNAME"))
```
punktilious, line 86: I know it's not part of the change, but is this the best way to get the name of the host? Do we ensure that HOSTNAME is part of the environment?  On Windows, COMPUTERNAME might be more appropriate. Let me know if you think we need a ticket for this.

fhir-audit/src/main/java/com/ibm/fhir/audit/logging/impl/WhcAuditCadfLogService.java
```diff
@@ -197,8 +196,10 @@ public static CadfEvent createCadfEvent(AuditLogEntry logEntry)
         Outcome cadfEventOutCome;
 
         // Cadf does't log config, so skip
-        if ((logEntry.getEventType() != AuditLogEventType.FHIR_CONFIGDATA.value()) && logEntry.getContext() != null
-                && logEntry.getContext().getAction() != null && logEntry.getContext().getApiParameters() != null) {
+        if ((logEntry.getEventType() != AuditLogEventType.FHIR_CONFIGDATA.value())
```
punktilious, line 199: String != String. Shouldn't this use equals(...)?

fhir-audit/src/main/java/com/ibm/fhir/audit/logging/impl/WhcAuditCadfLogService.java
prb112: Good point, I don't think it's super reliable. I think a ticket isn't bad, probably lower priority.

fhir-audit/src/main/java/com/ibm/fhir/audit/logging/impl/WhcAuditCadfLogService.java
prb112: Cleaned this up by addressing Lee's well-known request

fhir-server/src/main/java/com/ibm/fhir/server/util/RestAuditLogger.java
```diff
@@ -311,6 +313,21 @@ public static void logBundle(HttpServletRequest request, Bundle bundle, Date sta
                 .resourcesUpdated(updateCount).build());
         entry.setDescription("FHIR Bundle request");
 
+        // Previously we didn't set the Action which caused the logEntry to
+        // Skip over the actual logging, int his case, we're deciding by default
```
lmsurpre, line 317:
```suggestion
        // Skip over the actual logging, in this case, we're deciding by default
```

lmsurpre (approved):
LGTM

fhir-audit/src/main/java/com/ibm/fhir/audit/logging/impl/WhcAuditCadfLogService.java
prb112: Thanks

fhir-audit/src/main/java/com/ibm/fhir/audit/logging/impl/WhcAuditCadfLogService.java
lmsurpre: outdated now, but AuditLogEventType is an enum and enum==enum is just fine.

PR #1809, opened 2020-12-08
Prevent unauthorized access to Patient resource before it happens

lmsurpre:
This prevents us from leaking information about what patient resources
exist / don't exist on the system, which was deemed useful in case an
implementer uses sensitive data for the Resource.id of the Patient
resources (which still isn't advised).

### DIFF
--- a/fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
+++ b/fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
@@ -47,6 +47,19 @@ public class AuthzPolicyEnforcementPersistenceInterceptor implements FHIRPersist
     private static final String BEARER_TOKEN_PREFIX = "Bearer";
     private static final String PATIENT_REF_PREFIX = "Patient/";
 
+    private static final String REQUEST_NOT_PERMITTED = "Requested interaction is not permitted by any of the passed scopes.";
+
+    @Override
+    public void beforeRead(FHIRPersistenceEvent event) throws FHIRPersistenceInterceptorException {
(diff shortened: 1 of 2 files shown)


michaelwschroeder (approved):

michaelwschroeder:
LGTM

JohnTimm (approved):
LGTM

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
```diff
@@ -47,6 +47,19 @@
     private static final String BEARER_TOKEN_PREFIX = "Bearer";
     private static final String PATIENT_REF_PREFIX = "Patient/";
 
+    private static final String REQUEST_NOT_PERMITTED = "Requested interaction is not permitted by any of the passed scopes.";
+
+    @Override
+    public void beforeRead(FHIRPersistenceEvent event) throws FHIRPersistenceInterceptorException {
```
lmsurpre, line 53: expand to beforeVread and beforeHistory as well

lmsurpre: done

d0roppe (commented):
LGTM from what I understand.  Message is consistent?

lmsurpre:
> Message is consistent?

Since we're not divulging any information that the user hasn't directly passed us, I opted for a more verbose message.  Example:
> Interaction with 'Patient/123abc' is not permitted under patient context 'xyz'.

This contrasts with the more generic message we return in the other cases where the resource(s) involved in the interaction belong to some compartment that we (and potentially the end user) doesn't know until it is executed.

PR #1820, opened 2020-12-11
fix: it tests were not running for drug formulary and c4bb

prb112:


lmsurpre (approved):
LGTM

lmsurpre:
the changes look good, but what do we need to do to enable them to run in the CI?  i didn't see anything here that would change that...

prb112:
> ~the changes look good, but what do we need to do to enable them to run in the CI? i didn't see anything here that would change that...~
> 
> UPDATE: now I actually looked and I see the inclusion of the tests is dependent on their presence on the server (as determined by `ProfilesTestBase.checkProfileExistsOnServer()`)...clever

That's right.

Issue #1739, opened 2020-11-20
Update checking of _include and _revinclude parameter values against tenant configuration to account for specification of target resource type


michaelwschroeder:
**Is your feature request related to a problem? Please describe.**
Validation of _include and _revinclude parameters will fail if the parameter value contains a target resource type but the configuration value in the `fhirServer/resources/<resourceType>/searchIncludes` or `fhirServer/resources/<resourceType>/searchRevIncludes` list does not contain a target resource type.

**Describe the solution you'd like**
If the configuration value specifies a target resource type, then the parameter value should match exactly.
If the configuration value does not specify a target resource type, then the parameter value should only need to match on its first two components (regardless of target resource type specified)

**Describe alternatives you've considered**
None.

**Acceptance Criteria**
At least one acceptance criteria is included.

1.  
  GIVEN a `fhirServer/resources/<resourceType>/searchIncludes` or `fhirServer/resources/<resourceType>/searchRevIncludes` list item contains a target resource type
    WHEN an `_include` or `_revinclude` search parameter value is validated that matches the first 2 components of the config value, but does not contain a target resource type
    THEN an error should be issued

2.  
  GIVEN a `fhirServer/resources/<resourceType>/searchIncludes` or `fhirServer/resources/<resourceType>/searchRevIncludes` list item does not contain a target resource type
    WHEN an `_include` or `_revinclude` search parameter value is validated that matches the first 2 components of the config value, but also contains a target resource type
    THEN the parameter should validate successfully

**Additional context**
Example 1: this should generate an error - config value contains target resource type but parameter value does not:
`fhirServer/resources/<resourceType>/searchIncludes` config value: `Patient:general-practitioner:Practitioner`
`_include` search parameter value: `Patient:general-practitioner`

Example 2: this should not generate an error - config value does not contain target resource type but parameter value does:
`fhirServer/resources/<resourceType>/searchIncludes` config value: `Patient:general-practitioner`
`_include` search parameter value: `Patient:general-practitioner:Practitioner`

lmsurpre:
stretch goal for sprint 22

michaelwschroeder:
Fix validated in local environment running 4.5.3 level code.

PR #1836, opened 2020-12-17
issue #1739 - Update _include and _revinclude parameter checking

tbieste:
The validity checking of the _include and _revinclude parameters has been updated so if the configuration value does not specify a target resource type, then the parameter value only needs to match on its first two components (regardless of target resource type specified).


lmsurpre (approved):
LGTM

PR #1837, opened 2020-12-18
HREX - Update .index.json

prb112:
Warning about invalid entry when discovering HREX profile. 
An example snuck into the .index.json. 

Fixes this warning on startup
fhir-server_1  | [WARNING ] Skipping index entry 2 due to one or more missing required fields, beginning with: url


lmsurpre (approved):
LGTM

Issue #1839, opened 2020-12-23
Use search parameter filtering for search parameter disambiguation


lmsurpre:
**Is your feature request related to a problem? Please describe.**
In https://github.com/IBM/FHIR/issues/1272 we introduced url-based search parameter configuration in `fhir-server-config.json`.

However, when we build the built-in ParametersMap object, this is done outside of the context of any tenant and so we cannot really consult the tenant-specific search filter.

Further, the current ParametersMap only supports having a single search parameter per code (per resourceType).
There is some logic to warn about cases where there are multiple search parameters with the same code but differing expressions, but I thought that having multiple search parameters with the same code and equivalent expressions would be fine.

However, for our 4.5.3 release, we added some logic to `SearchUtil.filterSearchParameters` where, if the search parameter chosen during the initialization of the ParametersMap has a URL that differs from the one used in the tenant configuration, then the system skips over this parameter altogether.

**Describe the solution you'd like**
1. Update the ParametersMap to support having multiple search parameters with the same code
2. Emit warnings whenever we have conflicting search parameters and ask operators to disambiguate via `fhir-server-config.json`

For number 2, we used to do just this.  However, since we had no mechanism for disambiguating, the logs were very frustrating because we were packaging multiple IGs with conflicting parameters.  Now that we don't prepackage any IGs, I think it is reasonable to expect operators to do the disambiguation in the config when they package IGs that conflict with base spec search params.

**Describe alternatives you've considered**
https://github.com/IBM/FHIR/issues/1625 outlines an alternative idea I had which was to build a per-tenant ParametersMap object.  If we did that, then we probably wouldn't need to have ParametersMap support having multiple search parameters with the same code.  However, that change seems slightly larger.

**Additional context**
The workaround for this issue is to avoid packaging conflicting search parameters.

d0roppe:
Started the server and noted these messages in the log file:
[1/25/21, 17:56:11:995 UTC] 00000035 com.ibm.fhir.search.util.SearchUtil                          W Skipping search parameter with id='location-address'. Found multiple search parameters for code 'address' on resource type 'Location'; use search parameter filtering to disambiguate.
[1/25/21, 17:56:11:995 UTC] 00000035 com.ibm.fhir.search.util.SearchUtil                          W Skipping search parameter with id='location-address-state'. Found multiple search parameters for code 'address-state' on resource type 'Location'; use search parameter filtering to disambiguate.

Added the following to the fhir-server-config.json
        "resources": {
[...]

Issue #1743, opened 2020-11-23
SearchUtil#getFilteredBuiltinSearchParameters uses List#contains


punktilious:
**Describe the bug**
List#contains performs a linear search, which could have performs implications if the list is large and called repeatedly.

**To Reproduce**
See code

**Expected behavior**
A performant implementation.

**Additional context**
Review usage to see if a HashSet would be more appropriate in this context.

lmsurpre:
See https://github.com/IBM/FHIR/issues/1625 for some related thoughts on improving performance of getting the filtered set of search parameters.

punktilious:
LGTM. Verified through inspection.

Issue #1596, opened 2020-10-19
Deprecate `extension-search-parameters.json` in favor of the `fhir-registry`


prb112:
**Is your feature request related to a problem? Please describe.**
Deprecate `extension-search-parameters.json` in favor of the `fhir-registry`

**Describe the solution you'd like**
Currently, customers use `extension-search-parameters` to define search-parameters outside the scope of SearchParameter loaded in Implementation Guides and the Specification.

Note, there are significant updates to tests needed.

**Describe alternatives you've considered**
Keep the existing framework.

**Additional context**
n/a

prb112:
Ran into an issue related to this... https://github.com/IBM/FHIR/issues/2495
Search.init is called before the Registry, however, the registry relies on Search.init to populate the codes for Search, so that the SPs can be processed.  To use this it'd require some rethinking on the initialization and caches that are used.

lmsurpre:
All search parameters (including tenant-specific ones) are now loaded from the registry.
I did *not* formally deprecate the extension-search-parameters.json mechanism...instead I introduced a new RegistryResourceProvider that can read those files and bridge their contents into the registry.

It is possible to use a different method to load tenant-specific search parameters (e.g. using the ServerRegistryResourceProvider), but this was easy enough to continue supporting for now so that we don't needlessly break anyone that may be using this pattern.

prb112:
It looks great.  Good job.

PR #1840, opened 2020-12-23
issues #1839 and #1743 - support search parameter disambiguation

lmsurpre:
1. Update ParametersMap to support storing multiple search parameters
with the same code
2. Address minor perf concern from #1743 by collecting to a map instead of a list
3. Update SearchUtil.getSearchParameter to lookup the search parameter
by URI from the config if possible (instead of applying the filter to
the full set of built-in parameters).
4. Update the docs to reflect that search parameter filtering now
applies to tenant-specific search parameters as well. This should help
us move toward https://github.com/IBM/FHIR/issues/1596

Also fixed a bad trace message and did some minor formatting / javadoc.


prb112 (approved):
LGTM - makes sense to me.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -125,28 +125,17 @@
...
-     * SearchParameter name (e.g. "_lastUpdated").
-     * When getSearchParameter(resourceType, name) is called, we'll need to first
-     * search in the current tenant's map,
-     * then if not found, look in the "built-in" tenant's map. Also, when
-     * getSearchParameters(resourceType) is called,
-     * we'll need to return a List that contains SearchParameters from the current
-     * tenant's map (if present) plus those
+     * This is our in-memory cache of SearchParameter objects. The cache is organized at the top level by tenant-id,
+     * with the built-in (FHIR spec-defined) SearchParameters stored under the "built-in" pseudo-tenant-id.
+     * SearchParameters contained in the default tenant's extension-search-parameters.xml file are stored under the
```
michaelwschroeder, line 130: want to use extension-search-parameters.json instead of extension-search-parameters.xml ?

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -125,28 +125,17 @@
...
+     * "default" tenant-id, and other tenants' SearchParameters (defined in their tenant-specific
+     * extension-search-parameters.xml files) will be stored under their respective tenant-ids as well. The objects
```
michaelwschroeder, line 132: same as above

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -301,23 +282,27 @@ public static void init() {
                 if (includedSPs.containsKey(code)) {
                     String configuredUrl = includedSPs.get(code);
                     if (configuredUrl != null && configuredUrl.equals(url)) {
-                        results.add(sp);
-                    } else {
+                        results.put(code, sp);
+                    } else if (log.isLoggable(Level.FINE)) {
                         log.info("Skipping search parameter with id='" + sp.getId() + "'. "
```
michaelwschroeder, line 287: Do you want to use log.fine here, since that's what you're checking against in the line above?

michaelwschroeder (approved):

michaelwschroeder:
LGTM - left a couple minor comments

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
lmsurpre:
```suggestion
     * SearchParameters contained in the default tenant's extension-search-parameters.json file are stored under the
```

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
lmsurpre:
```suggestion
     * extension-search-parameters.json files) will be stored under their respective tenant-ids as well. The objects
```

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
lmsurpre:
```suggestion
                        log.fine("Skipping search parameter with id='" + sp.getId() + "'. "
```

tbieste (approved):

punktilious:
Are there any unit tests which should be updated or added?

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersMap.java
```diff
@@ -97,12 +71,12 @@ public void insert(String code, SearchParameter parameter) {
...
      */
     public void insertAll(ParametersMap map) {
-        for (Entry<String, SearchParameter> entry : map.codeMap.entrySet()) {
-            insert(entry.getKey(), entry.getValue());
+        for (SearchParameter sp : map.urlMap.values()) {
+            insert(sp.getCode().getValue(), sp);
         }
     }
 
-    public SearchParameter lookupByCode(String searchParameterCode) {
```
punktilious, line 105: Minor. This Set is mutable. Could make this immutable on return. That may have performance implications given how much this stuff gets used. Perhaps use a custom class which is a HashSet and implements an interface with only the getters you need.

lmsurpre: Good call.  I think wrapping the set via Collections.immutableSet should be fine.

lmsurpre:
@punktilious just pushed an update per your comments.  will plan to merge it once the tests pass.

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersMap.java
lmsurpre: done

PR #1842, opened 2021-01-04
fix for hardcoded year from SearchLastUpdatedIdTest

lmsurpre:
more specifically, I get the current year (in UTC), and use that.  in places where we already had a list of years, i just prepend the current year to the existing list.


prb112 (approved):
LGTM

Issue #1602, opened 2020-10-20
Add support for deploying via Helm chart


chgl:
**Is your feature request related to a problem? Please describe.**

The server is currently not available as a Helm chart, making installation on Kubernetes less convenient.

**Describe the solution you'd like**

Provide a Helm chart the make it easier to install the application in Kubernetes.

**Describe alternatives you've considered**

**Additional context**

I could definitely work on this starting in 1-2 weeks. I think it should be fairly straight-forward, the biggest challenge will probably be setting up the database/applying the schema using the persistence cli tools.

lmsurpre:
We definitely have a lot of this working on our internal deploys...I can see about opening some of that up and pushing it to github.  I also recently read the following post:  https://helm.sh/blog/helm-turns-five/ and so I think the timing is good.
* Helm stable and incubator chart repositories will now be hosted by GitHub
* There is now an official Helm Chart Releaser GitHub Action:  https://github.com/marketplace/actions/helm-chart-releaser

chgl:
> We're also using helm for a few of our internal deploys...I can see about opening some of that up and pushing it to github.

That would be great, thank you!

+1 for helm-chart-releaser & https://github.com/norwoodj/helm-docs

Let me know if I can help setting it up or give feedback at some point.

luisgarcc:
For what is worth we have a bare bones FHIR chart, with limited configurability to suit some demos we have been doing.

It's available in the [alvearie health-patterns](https://github.com/Alvearie/health-patterns/tree/main/clinical-ingestion/helm-charts/fhir) repository in case someone might want to use it in the mean time, while the official chart from the FHIR team is published.

lmsurpre:
Extracting a standalone helm chart from the one we use within our internal deploy pipeline proved a bit more difficult than I was expecting.  I created a draft pull request #1602 with a "clean base" but left quite a few TODOs in the description.  Will try getting back to this one when I have a chance.

lmsurpre:
The initial version of this chart is now committed to https://github.com/Alvearie/alvearie-helm/tree/main/ibm-fhir-server.
Additional enhancements will be handled via issues on that other repo.

PR #1848, opened 2021-01-06
issue #1602 - initial commit of helm chart

lmsurpre:
Introduces a helm chart for the IBM FHIR Server under `/charts`.

Also introduces a github action for publishing the updated chart on our Releases tab
> This uses @helm/chart-releaser-action to turn your GitHub project into a self-hosted Helm chart repo. It does this by checking each chart in your project, and whenever there's a new chart version, creates a corresponding GitHub release named for the chart version, adds Helm chart artifacts to the release, and creates or updates an index.yaml file with metadata about those releases, which is then hosted on GitHub Pages

The chart uses helm hooks to execute the `ibmcom/ibm-fhir-schematool` image to create and update a database schema prior to deploying the updated application.

TODOs
1. Make it work with Db2


charts/README.md
```diff
@@ -0,0 +1,43 @@
...
+To install the IBM FHIR Server via helm, you must first have a database.
+
+To install a PostgreSQL database to the same cluster using helm, you can run the following command:
+```
+$ helm install my-release bitnami/postgresql
+```
+
+For example if you target the default namespace and name the release "postgres",
+then that would produce something like this:
+```sh
```
prb112, line 14: not clear. where. this ends...

charts/fhir-server/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+description: Helm chart for the IBM FHIR Server
+name: ibm-fhir-server
+version: 0.0.1
+appVersion: 4.8.3
```
prb112, line 5: can this be externalized? part of a strategy.matrix?

charts/fhir-server/templates/config-internal.yaml
```diff
@@ -0,0 +1,209 @@
...
+data:
+  datasources.xml: >
+    {{- if eq .Values.db.type "db2" }}
+    TODO
+    {{- else }}
+    <server>
+      <!-- ============================================================== -->
+      <!-- TENANT: default; DSID: default; TYPE: read-write               -->
+      <!-- ============================================================== -->
+      <dataSource id="fhirDefaultDefault" jndiName="jdbc/fhir_default_default" type="javax.sql.XADataSource" statementCacheSize="200" syncQueryTimeoutWithTransactionTimeout="true" validationTimeout="30s">
```
prb112, line 14: I know. we. added `isolationLevel="TRANSACTION_READ_COMMITTED"` for. db2, do we need this for postgres?

prb112 (commented):
Looks good, makes sense. in another. repo

charts/README.md
lmsurpre: it ends on line 43, but I probably need to add some more stuff below there to describe how to actually use the chart.  For now I just wanted to document a way of getting a postgresql in your cluster since I wanted to figure that out anyway

charts/fhir-server/templates/config-internal.yaml
lmsurpre: I'm pretty sure that Robin said its not needed for postgresql...something about the default being different for db2 and derby.

lmsurpre:
After discussing this with Paul and others, we've decided that it makes more sense to keep this helm chart out of this repo for now (so that it can be released on a separate cadence without polluting our Releases tab).  Thus we have created a new repository under the Alvearie project and I've moved the content of this pull request to https://github.com/Alvearie/alvearie-helm/pull/1

Issue #1849, opened 2021-01-06
Update AuthZ interceptor to validate search requests and to convert non-compartment searches to compartment searches


michaelwschroeder:
**Is your feature request related to a problem? Please describe.**
Currently, the AuthZ interceptor does not fully validate search requests, leading to the possibility of inferring data about patient that caller does not have access to.

For example, a search request with specific parameters designed to look for "Observations for patient 1234 with a code of abc", then looking for whether the search result is empty vs receiving a 403.

**Describe the solution you'd like**
The AuthZ interceptor's `beforeSearch()` method should do the following validation/processing of search requests:

If compartment search:
1. if not a Patient compartment search: fail with 403 error
2. else if Patient compartment search, but compartment ID does not match patient ID from token: fail with 403 error
3. else if Patient compartment search, but resource type being searched is not member of Patient compartment: fail with 400 error
4. else valid Patient compartment search: do nothing, allow search to continue

If not compartment search:
1. if resource type being searched is Patient: fail with 403 error
2. else if resource type being searched is not member of Patient compartment: do nothing, allow search to continue
3. else if resource type being searched is a member of Patient compartment: convert the search to a Patient compartment search of the caller's current patient context by adding the necessary inclusionCriteria search parameters to the searchContext

**Describe alternatives you've considered**
1. Turn 410 response code into a 404 response code
2. For search, return 200 with an empty bundle instead of 403
3. Disable Patient search altogether
4. ?

**Acceptance Criteria**
At least one acceptance criteria is included.

1.  
    GIVEN a compartment search request that is not a Patient compartment search
    WHEN the search request is executed
    THEN return a 403 FORBIDDEN http response
2.  
    GIVEN a compartment search request that is a Patient compartment search AND compartment ID does not match patient ID that caller has access to
    WHEN the search request is executed
    THEN return a 403 FORBIDDEN http response
3.  
    GIVEN a compartment search request that is a Patient compartment search AND compartment ID matches patient ID the caller has access to, but resource type being searched is not a member of the Patient compartment
    WHEN the search request is executed
    THEN return a 400 BAD REQUEST http response
4.  
    GIVEN a compartment search request that is a Patient compartment search AND compartment ID matches patient ID that caller has access to AND resource type being searched is a member of the Patient compartment
    WHEN the search request is executed
    THEN return a 200 http response with search results
5.  
    GIVEN a non-compartment search request that is a Patient search
    WHEN the search request is executed
[...]

PR #1850, opened 2021-01-06
Issue #1849 - AuthZ interceptor validate/convert search requests

michaelwschroeder:

### DIFF
--- a/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright IBM Corp. 2016, 2020
+ * (C) Copyright IBM Corp. 2016, 2021
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -223,7 +223,7 @@ public class FHIRRestHelper implements FHIRResourceHelpers {
 
             // First, invoke the 'beforeCreate' interceptor methods.
             FHIRPersistenceEvent event =
-                    new FHIRPersistenceEvent(resource, buildPersistenceEventProperties(type, null, null, requestProperties));
+                    new FHIRPersistenceEvent(resource, buildPersistenceEventProperties(type, null, null, null));
             getInterceptorMgr().fireBeforeCreateEvent(event);
 
             FHIRPersistenceContext persistenceContext =
@@ -413,7 +413,7 @@ public class FHIRRestHelper implements FHIRResourceHelpers {
 
             // First, create the persistence event.
             FHIRPersistenceEvent event = new FHIRPersistenceEvent(newResource,
-                    buildPersistenceEventProperties(type, newResource.getId(), null, requestProperties));
+                    buildPersistenceEventProperties(type, newResource.getId(), null, null));
 
             // Next, set the "previous resource" in the persistence event.
             event.setPrevFhirResource(ior.getPrevResource());
@@ -603,7 +603,7 @@ public class FHIRRestHelper implements FHIRResourceHelpers {
                     resourceToDelete = entry.getResource();
                     // First, invoke the 'beforeDelete' interceptor methods.
                     FHIRPersistenceEvent event =
-                            new FHIRPersistenceEvent(null, buildPersistenceEventProperties(type, id, null, requestProperties));
+                            new FHIRPersistenceEvent(null, buildPersistenceEventProperties(type, id, null, null));
                     event.setFhirResource(resourceToDelete);
                     getInterceptorMgr().fireBeforeDeleteEvent(event);
 
@@ -711,7 +711,7 @@ public class FHIRRestHelper implements FHIRResourceHelpers {
 
             // First, invoke the 'beforeRead' interceptor methods.
             FHIRPersistenceEvent event =
-                    new FHIRPersistenceEvent(contextResource, buildPersistenceEventProperties(type, id, null, requestProperties));
+                    new FHIRPersistenceEvent(contextResource, buildPersistenceEventProperties(type, id, null, searchContext));
             getInterceptorMgr().fireBeforeReadEvent(event);
 
             FHIRPersistenceContext persistenceContext =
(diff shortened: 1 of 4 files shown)


fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
```diff
@@ -51,20 +54,79 @@
 
     @Override
     public void beforeRead(FHIRPersistenceEvent event) throws FHIRPersistenceInterceptorException {
-        enforceDirectPatientAcess(event);
+        enforceDirectPatientAccess(event);
```
lmsurpre, line 57: good typo fix

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
```diff
@@ -51,20 +54,79 @@
...
 
     @Override
     public void beforeHistory(FHIRPersistenceEvent event) throws FHIRPersistenceInterceptorException {
-        enforceDirectPatientAcess(event);
+        enforceDirectPatientAccess(event);
     }
 
-    private void enforceDirectPatientAcess(FHIRPersistenceEvent event) throws FHIRPersistenceInterceptorException {
+    @Override
+    public void beforeSearch(FHIRPersistenceEvent event) throws FHIRPersistenceInterceptorException {
```
lmsurpre, line 71: thinking we should document what this does somewhere, but not sure where.  either here or at the class level i think.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -2635,10 +2635,19 @@ private String getRequestBaseUri(String type) throws Exception {
     /**
      * Builds a collection of properties that will be passed to the persistence interceptors.
      *
+     * @param type
+     *            the resource type
```
lmsurpre, line 2639: super minor, but at some point we should agree on consistent formatting for our javadoc...we're pretty much all over with this at the moment

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -2635,10 +2635,19 @@ private String getRequestBaseUri(String type) throws Exception {
...
+     *            the resource logical ID
+     * @param version
+     *            the resource version
+     * @param searchContext
+     *            the request search context
+     * @return a map of persistence event properties
      * @throws FHIRPersistenceException
      */
     private Map<String, Object> buildPersistenceEventProperties(String type, String id,
-            String version, Map<String, String> requestProperties) throws FHIRPersistenceException {
```
lmsurpre, line 2641: at first i was worried about this, but i see that it was never used within the method and should be safe to remove.

the requestProperties argument to our various FHIRRestHelper methods has some interesting history.  Back in the old days we didn't have this anywhere.
then, for a specific solution, they wanted a way to pass different HTTP header values to different entries within the *same* batch/transaction bundle.
we were sort of forced to accommodate that and we did so by:
1. introducing this requestProperties parameter all over
2. defining an extension that clients could set at the batch/transaction bundle entry level for setting/overriding specific "header" values
3. adding logic to set the requestProperties from these extensions with this to something non-null while we process the batch/transaction

when we went to open source, I think we lost parts 2 and 3 of this since it was deemed specific to that one client.  requestProperties lingers in our FHIRRestHelper methods, but i'm not sure whether it ever gets set at all at this point.  I'll open an issue to look into that...we'll probably need to decide whether to:
A. bring back that bundle-entry HTTP headers feature; or
B. remove requestProperties from all over

None of that needs to happen now, just thought I'd share the history :-)

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
```diff
@@ -51,20 +54,79 @@
...
+                try {
+                    if (CompartmentUtil.getCompartmentResourceTypes("Patient").contains(event.getFhirResourceType())) {
+                        // NOTE: We currently do not support OR'd compartment searches, nor do we currently expect more than one patient ID to be
+                        // specified in the authorization token (see getPatientIdFromToken()). We will use the first ID specified for the compartment search.
+                        FHIRSearchContext compartmentSearchContext = SearchUtil.parseQueryParameters("Patient", patientIdFromToken.get(0),
+                            ModelSupport.getResourceType(event.getFhirResourceType()), Collections.emptyMap(), searchContext.isLenient());
+                        searchContext.getSearchParameters().addAll(compartmentSearchContext.getSearchParameters());
+                    }
+                } catch (Exception e) {
+                    log.log(Level.WARNING, "Unexpected exception while converting search to Patient compartment search", e);
```
lmsurpre, line 123: I'm thinking we should wrap and rethrow in this case.  If not, then we would continue to execute the search and could potentially leak whether the resources exist or not.  Granted, we should never get here anyway, so its not too critical.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
lmsurpre: opened https://github.com/IBM/FHIR/issues/1852 for it

lmsurpre (approved):
Left some comments, but overall LGTM

JohnTimm (approved):
LGTM pending resolution of Lee's comments

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
michaelwschroeder: Documented at method level.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: Agreed. In this case I just went with the format being used on the other methods in the class.

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
michaelwschroeder: Done - wrap and rethrow.

PR #1860, opened 2021-01-08
fix: target and grant schemas #1858

### LINKED ISSUE #1858: fhir-persistence-schema doesn't behave as expected.
**Describe the bug**
1. ibm-fhir-schematool doesn't specify the target schema to be granted permissions for oauth/jbatch.
2. ibm-fhir-schematool doesn't specify the target schema to be dropped for oauth/jbatch or fhirdata.
3. fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java doesn't allow schema names to be set.

**To Reproduce**

1. Create a postgres db
```
docker run -itd \
    --rm --name postgresfhir \
    --shm-size=256m \
    --sysctl net.core.somaxconn=256 \
    --sysctl net.ipv4.tcp_syncookies=0 \
    --hostname db \
    -e POSTGRES_USER=postgres \
    -e POSTGRES_PASSWORD=change-password \
    -e PGDATA=/var/lib/postgresql/data/pgdata \
    -e POSTGRES_DB=fhirdb \
    -p 5432:5432 \
    --privileged \
    -v "$PWD/my-postgres.conf":/etc/postgresql/postgresql.conf \
    postgres:12.5 postgres -c 'config_file=/etc/postgresql/postgresql.conf'
```

Be sure max_locks_per_transaction = 10000 is set.

2. Run the following
- `docker run --rm  --env ENV_TOOL_INPUT=$(cat persistence-onboard-example.json | base64) ibm-fhir-schematool:latest`
- `docker run --rm  --env ENV_TOOL_INPUT=$(cat persistence-offboard-example.json | base64) ibm-fhir-schematool:latest`

**Expected behavior**
The grant should work properly and the drop-schema should be dropped.

**Additional context**
n/a

prb112:
closes #1858

### DIFF
--- a/fhir-persistence-schema/README.md
+++ b/fhir-persistence-schema/README.md
@@ -275,6 +275,9 @@ For PostgreSQL:
 --db-type postgresql
 ```
 
+Alternatively, you can drop specific schemas with `--drop-schema-batch schema-name-to-drop` and
+`--drop-schema-oauth schema-name-to-drop`
+
 ## Alternative: Setting up a shared Db2 with separate schemas for each tenant
 
 For those using multiple schemas for each customer, for instance, customer 2 needs to be separately configured with the database and schema. 
--- a/fhir-install/src/main/docker/ibm-fhir-schematool/run.sh
+++ b/fhir-install/src/main/docker/ibm-fhir-schematool/run.sh
@@ -338,22 +338,29 @@ function grant_to_dbuser {
     then
         error_warn "Target User is not set and we are skipping the grant phase, it is recommended to run"
     else
+        # Pick off the schemas
+        SCHEMA_OAUTH=$(get_property schema.name.oauth .persistence[0].schema.oauth)
+        if [ -z "${SCHEMA_OAUTH}" ] || [ "null" = "${SCHEMA_OAUTH}" ]
+        then
+            SCHEMA_OAUTH="FHIR_OAUTH"
+        fi
+        SCHEMA_BATCH=$(get_property schema.name.batch .persistence[0].schema.batch)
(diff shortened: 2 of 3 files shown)


fhir-persistence-schema/README.md
```diff
@@ -275,6 +275,9 @@ For PostgreSQL:
 --db-type postgresql
 ```
 
+Alternatively, you can drop specific schemas with `--drop-schema-batch schema-name-to-drop` and
+`--drop-schema-oauth schema-name-to-drop`
```
lmsurpre, line 279: It seems a little odd to me because we don't allow the user to specify the schema name during the create.

prb112:
```
            case "--create-schema-oauth":
                this.createOauthSchema = true;
                if (nextIdx < args.length && !args[nextIdx].startsWith("--")) {
                    schema.setOauthSchemaName(args[nextIdx]);
                    i++;
                }
                break;

prb112: if the next value is not a name, but a value it puts it as the oauth schema name, same for fhir, and batch

lmsurpre: interesting.  i've always just created the schemas with `--create-schemas` which takes the fhir schema name from `--schema-name` property and uses the hardcoded values for the other two.  i do vaguely remember adding the alternative now and listing them under https://github.com/IBM/FHIR/tree/master/fhir-persistence-schema#alternative-setting-up-a-shared-db2-with-separate-schemas-for-each-tenant

lmsurpre (approved):
LGTM

Issue #1615, opened 2020-10-22
Enforce FHIR server interaction limitations in fhir-server REST layer


lmsurpre:
In https://github.com/IBM/FHIR/issues/1449 we introduced a new configuration property group called `resources`.
In this property group, server administrators can now configure which interactions and search parameters should show up for which resource types in the CapabilityStatement.
However, currently, this only drives the CapabilityStatement and not the actual behavior of the server.
It should be possible to enforce that configuration in the fhir-server REST layer, probably by isolating this logic in the FHIRRestHelper class so that it applies to batch/transaction entries as well.

This task is very similar to #1351 except that it should apply to the resource type interactions instead of the searches that are allowed.

There is also a related enhancement request #819 for supporting a subset of resources in the fhir-persistence-schema tool, so that we don't need to create and manage all the tables for resource types that don't need to be supported.

lmsurpre:
stretch goal for sprint 22

lmsurpre:
I verified this one by deploying a server with a list of resources with open=false and executing requests like:
* POST `[base]/Patient` (where only read actions are supported for the Patient resource)
* GET `[base]/Claim` (where Claim is not a supported resource type)
* GET `[base]/Patient/[id]/Claim` (where Claim is not a supported resource type)

In each case, the server behaved as expected.

PR #1862, opened 2021-01-11
Issue #1615 - Enforce configured interactions in REST layer

michaelwschroeder:

### DIFF
--- a/docs/src/pages/Conformance.md
+++ b/docs/src/pages/Conformance.md
@@ -21,53 +21,7 @@ The HL7 FHIR specification is more than just a data format. It defines an [HTTP 
 
 The IBM FHIR Server implements a linear versioning scheme for resources and fully implements the `vread` and `history` interactions, as well as version-aware updates.
 
-### Configuring allowable API interactions
-By default, the IBM FHIR Server allows all supported API interactions (`create`, `read`, `vread`, `history`, `search`, `update`, `patch`, `delete`). However, it is possible to configure which of these interactions are allowed on a per resource basis through a set of interaction rules specified via the `fhirServer/resources/<resourceType>/interactions` property in `fhir-server-config.json`. The following snippet shows the general form for specifying interaction rules:
-
-```
-"resources": {
-    "open": true,
-    "Condition": {
-        "interactions": ["create", "read", "vread", "history", "search", "update", "patch", "delete"]
-    },
-    "Observation": {
-        "interactions": ["create", "read", "vread", "history", "delete"]
-    },
-    "Patient": {
-        "interactions": ["read", "vread", "history", "search"]
-    },
-    "Procedure": {
-        "interactions": ["create", "read", "vread", "history", "delete"]
-    }
-}
-```
-
-The `fhirServer/resources/<resourceType>/interactions` property is a JSON array of strings that represent the RESTful interactions allowed for the given resource type. If an interaction is not in the list of strings specified for a resource type, that interaction will not be allowed for that resource type. In the example above, the following interactions are allowed for the `Observation` resource type: `create`, `read`, `vread`, `history`, `delete`. This means a user will not be able to search for `Observation` resources because the `search` interaction is not specified in the list of allowed interactions.
-
-Omitting this property is equivalent to allowing all interactions for a given resource type. An empty array, `[]`, can be used to indicate that no interactions are allowed. Additionally, to define the set of interactions allowed for resource types which are not specifically configured in the `fhirServer/resources` property group, the base type of `Resource` may be specified:
-
-```
-"resources": {
-    "open": true,
-    "Condition": {
-        "interactions": ["create", "read", "vread", "history", "search", "update", "delete"]
-    },
-    "Observation": {
-        "interactions": ["create", "read", "vread", "history", "delete"]
-    },
-    "Patient": {
-        "interactions": ["read", "vread", "history", "search"]
-    },
-    "Procedure": {
-        "interactions": ["create", "read", "vread", "history", "delete"]
-    },
-    "Resource": {
-        "interactions": ["create", "read", "vread", "history", "search", "update", "patch", "delete"]
-    }
-}
-```
-
-In the example above, for any resource type which is not specifically configured, such as `Encounter`, all of the interactions listed for the `Resource` resource type will be allowed.
+By default, the IBM FHIR Server allows all supported API interactions (`create`, `read`, `vread`, `history`, `search`, `update`, `patch`, `delete`). However, it is possible to configure which of these interactions are allowed on a per resource basis through a set of interaction rules. See the [user guide](https://ibm.github.io/FHIR/guides/FHIRServerUsersGuide#412-fhir-rest-api) for details.
 
 ### Extended operations
 The HL7 FHIR specification also defines a mechanism for extending the base API with [extended operations](https://www.hl7.org/fhir/R4/operations.html).
--- a/docs/src/pages/Conformance.md
+++ b/docs/src/pages/Conformance.md
@@ -2,7 +2,7 @@
 layout: post
 title:  Conformance
 description: Notes on the Conformance of the IBM FHIR Server
-date:   2020-11-03 01:00:00 -0400
+date:   2021-01-11 01:00:00 -0400
 permalink: /conformance/
 ---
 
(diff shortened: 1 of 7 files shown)


docs/src/pages/Conformance.md
```diff
@@ -21,6 +21,54 @@ The HL7 FHIR specification is more than just a data format. It defines an [HTTP
 
 The IBM FHIR Server implements a linear versioning scheme for resources and fully implements the `vread` and `history` interactions, as well as version-aware updates.
 
+### Configuring allowable API interactions
+By default, the IBM FHIR Server allows all supported API interactions (`create`, `read`, `vread`, `history`, `search`, `update`, `patch`, `delete`). However, it is possible to configure which of these interactions are allowed on a per resource basis through a set of interaction rules specified via the `fhirServer/resources/<resourceType>/interactions` property in `fhir-server-config.json`. The following snippet shows the general form for specifying interaction rules:
```
lmsurpre, line 25: I very much like having this section here, but in my opinion I think we should:
1. keep it high-level and more client-focused here
2. move the "how to configure it in fhir-server-config.json" stuff into the user's guide

lmsurpre: Maybe this section just becomes a paragraph that says its configurable and then links to the user's guide for the detail?

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -668,24 +677,45 @@ public Resource doRead(String type, String id, boolean throwExcOnNull, boolean i
...
+     * @param includeDeleted
+     *            if true, return resource even if deleted
      * @param requestProperties
      *            additional request properties which supplement the HTTP headers associated with this request
+     * @param contextResource
+     *            a FHIR resource associated with this request
+     * @param queryParameters
+     *            for supporting _summary for resource read
+     * @param doValidation
+     *            if true, validate the interaction; if false, assume the interaction does not need to be validated
```
lmsurpre, line 705: to me, it is a bit ambiguous as to whether this means it controls whether the resource will be validated or not (in addition to whether the interaction is allowed or not).

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -2822,4 +2879,48 @@ public int doReindex(FHIROperationContext operationContext, OperationOutcome.Bui
...
+                log.fine("Allowed interactions: " + interactions);
+            }
+        } catch (Exception e) {
+            throw buildRestException("Error retrieving interactions configuration.", IssueType.UNKNOWN, IssueSeverity.ERROR);
+        }
+
+        // Perform validation of specified interaction against specified resourceType
+        if (interactions != null && !interactions.contains(interaction)) {
+            throw buildRestException("The requested interaction of type '" + interaction + "' is not allowed for resource type '" + resourceType + "'",
+                IssueType.BUSINESS_RULE, IssueSeverity.ERROR);
```
lmsurpre, line 2922: Do you know what kind of error code this will result in?

lmsurpre: Looks like it will be an HTTP 400.
I wondering if that is suitable for all cases, or if we should attempt to make it a 404 if `fhirServer/resources/open` is set to false and the resource type is not listed.

docs/src/pages/Conformance.md
michaelwschroeder: That makes sense. I was waffling between putting it in the conformance guide and the user's guide. Wasn't quite sure where it would go in the user's guide. Do you think a new sub-section in Chap.4? Maybe titled "FHIR HTTP API" or "FHIR REST API"?

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: Will change the parm name to checkInteractionAllowed.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: Wondering if 404 might be a little confusing, although they'd have the OperationOutcome indicating the error. Would 403 FORBIDDEN make sense, or is that more reserved for authorization type issues?

docs/src/pages/Conformance.md
michaelwschroeder: Discussed with Lee - will add a new section in Chap. 4.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: Discussed with Lee - will use IssueType.NOT_FOUND (which maps to 404) if `fhirServer/resources/open` is set to false and resource type not listed,

docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -2042,6 +2043,55 @@ select * from cos://us-south/fhir-audit-dev0 stored as json where EVENTTIME BETW
...
+    },
+    "Resource": {
+        "interactions": ["create", "read", "vread", "history", "search", "update", "patch", "delete"]
+    }
+}
+```
+
+In the example above, for any resource type which is not specifically configured, such as `Encounter`, or for any resource type which is configured but does not specify the `fhirServer/resources/<resourceType>/interactions` property, such as `Procedure`, all of the interactions listed for the `Resource` resource type will be allowed. One final consideration when configuring interactions is the `fhirServer/resources/open` property. If this property is specified and its value is set to `false`, then no interactions will be allowed for resource types which are not specifically listed in the `fhirServer/resources` property group. 
+
+In addition to interaction configuration, the `fhirServer/resources` property group also provides the ability to configure search parameter filtering and profile validation. See [Search configuration](https://ibm.github.io/FHIR/guides/FHIRSearchConfiguration#12-filtering) and [Resource validation](#44-resource-validation) respectively for details.
```
lmsurpre, line 2093: 👍

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -2903,13 +2905,23 @@ private void validateInteraction(String interaction, String resourceType) throws
...
+                    // Get the 'Resource' interaction property
+                    List<String> defaultInteractions = FHIRConfigHelper.getStringListProperty(defaultInteractionsConfigPath.toString());
+                    if (defaultInteractions != null) {
+                        interactions = defaultInteractions;
+                    }
                 }
             }
 
-            if (log.isLoggable(Level.FINE)) {
+           if (log.isLoggable(Level.FINE)) {
```
lmsurpre, line 2924: inadvertent whitespace deletion.  fix if you want

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
lmsurpre: The updated logic looks OK to me

docs/src/pages/Conformance.md
lmsurpre: Updated doc looks good to me

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
lmsurpre: param name is changed

lmsurpre (approved):
LGTM

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -137,6 +138,9 @@ public FHIRRestOperationResponse doCreate(String type, Resource resource, String
             Map<String, String> requestProperties, boolean doValidation) throws Exception {
         log.entering(this.getClass().getName(), "doCreate");
 
+        // Validate that interaction is allowed for given resource type
+        validateInteraction("create", type);
```
punktilious, line 142: "create" should be a constant or preferably an enum.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -2822,4 +2880,62 @@ public int doReindex(FHIROperationContext operationContext, OperationOutcome.Bui
...
+
+            if (log.isLoggable(Level.FINE)) {
+                log.fine("Allowed interactions: " + interactions);
+            }
+        } catch (Exception e) {
+            throw buildRestException("Error retrieving interactions configuration.", IssueType.UNKNOWN, IssueSeverity.ERROR);
+        }
+
+        // Perform validation of specified interaction against specified resourceType
+        if (interactions != null && !interactions.contains(interaction)) {
```
punktilious, line 2932: Linear search through a list, which logically is a Set. Although the list is very small, so perhaps a linear search won't be that much more expensive. But we also end up doing all this property work on every request. I think at some point we need to take a look at our property model to see if there's a way to streamline this stuff. But it probably starts with taking a look at our CPU usage overall.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: fixed

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: added enum for interaction types

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: In this case, since the FHIRConfigHelper returns lists, we'd have to convert to sets, which probably isn't worth the effort given the very small set size. Agreed on the property model. There are now several places we're reading just the 'resources' section of the config (here, search parm filtering, profile validation stuff). Might make sense to read once for a request and cache somewhere.

punktilious: Perhaps a new issue for the backlog?

punktilious (approved):
LGTM. Just one general thing - do we have any tests to examine behavior when the fhir-server-config is wrong. For example, if I wanted to limit the interactions for Patient but I mistyped it as Patlent...would this be caught/flagged?

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: I've opened a new issue to investigate options: https://github.com/IBM/FHIR/issues/1865

michaelwschroeder:
I do not know of any code that validates the contents of the config file. I believe in general we look for specific sections or properties by name, so if something invalid is specified, it's just ignored. For your example, we would look for interactions specified for the Patient resource and not find the Patient section, thus by default would allow all interaction types. The invalid Patlent section would not generate any type of error in and of itself.

PR #1873, opened 2021-01-19
Issue #1494 - add Bundle.entry.search to search results

michaelwschroeder:

### DIFF
--- a/docs/src/pages/Conformance.md
+++ b/docs/src/pages/Conformance.md
@@ -130,7 +130,7 @@ Finally, the specification defines a set of "Search result parameters" for contr
 
 The `_count` parameter can be used to request up to 1000 resources matching the search criteria. An attempt to exceed this `_count` limit will not be honored and returned resources will be capped at 1000. Any associated `_include` or `_revinclude` resources are not considered in the `_count` limit. 
 
-The `_include` and `_revinclude` parameters can be used to return resources related to the primary search results, in order to reduce the overall network delay of repeated retrievals of related resources. The number of `_include` or `_revinclude` resources returned for a single page of primary search results will be limited to 1000. If the number of resources to be returned exceeds 1000, the search will fail.  
+The `_include` and `_revinclude` parameters can be used to return resources related to the primary search results, in order to reduce the overall network delay of repeated retrievals of related resources. The number of `_include` or `_revinclude` resources returned for a single page of primary search results will be limited to 1000. If the number of included resources to be returned exceeds 1000, the search will fail. For example, if the primary search result is one resource and the number of included resources is 1000, the search will succeed. However, if the primary search result is one resource and the number of included resources is 1001, the search will fail.
 
 The `:iterate` modifier is not supported for the `_include` parameter (or any other).
 
--- a/docs/src/pages/Conformance.md
+++ b/docs/src/pages/Conformance.md
@@ -2,7 +2,7 @@
 layout: post
 title:  Conformance
 description: Notes on the Conformance of the IBM FHIR Server
-date:   2021-01-11 01:00:00 -0400
+date:   2021-01-19 01:00:00 -0400
 permalink: /conformance/
 ---
 
@@ -128,7 +128,9 @@ Finally, the specification defines a set of "Search result parameters" for contr
 * `_summary`
 * `_elements`
 
-The `_count` parameter can be used to request up to 1000 records matching the search criteria.  An attempt to exceed this `_count` limit will not be honored and returned records will be capped at 1000.  Any associated `_include` records are not considered in the `_count` limit. 
+The `_count` parameter can be used to request up to 1000 resources matching the search criteria. An attempt to exceed this `_count` limit will not be honored and returned resources will be capped at 1000. Any associated `_include` or `_revinclude` resources are not considered in the `_count` limit. 
+
+The `_include` and `_revinclude` parameters can be used to return resources related to the primary search results, in order to reduce the overall network delay of repeated retrievals of related resources. The number of `_include` or `_revinclude` resources returned for a single page of primary search results will be limited to 1000. If the number of resources to be returned exceeds 1000, the search will fail.  
 
 The `:iterate` modifier is not supported for the `_include` parameter (or any other).
 
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/InclusionQuerySegmentAggregator.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/InclusionQuerySegmentAggregator.java
@@ -9,9 +9,12 @@ package com.ibm.fhir.persistence.jdbc.util;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.AND;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.COMBINED_RESULTS;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.COMMA;
+import static com.ibm.fhir.persistence.jdbc.JDBCConstants.FETCH_FIRST;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.LEFT_PAREN;
+import static com.ibm.fhir.persistence.jdbc.JDBCConstants.LIMIT;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.QUOTE;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.RIGHT_PAREN;
+import static com.ibm.fhir.persistence.jdbc.JDBCConstants.ROWS_ONLY;
 
 import java.util.ArrayList;
 import java.util.HashSet;
@@ -437,9 +440,9 @@ public class InclusionQuerySegmentAggregator extends QuerySegmentAggregator {
         }
 
         if (this.parameterDao.isDb2Database()) {
-            queryString.append(" LIMIT ").append(limit);
+            queryString.append(LIMIT).append(limit);
         } else {
-            queryString.append(" FETCH FIRST ").append(limit).append(" ROWS ONLY");
+            queryString.append(FETCH_FIRST).append(limit).append(ROWS_ONLY);
         }
     }
 }
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/InclusionQuerySegmentAggregator.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/InclusionQuerySegmentAggregator.java
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright IBM Corp. 2018, 2020
+ * (C) Copyright IBM Corp. 2018, 2021
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -14,7 +14,9 @@ import static com.ibm.fhir.persistence.jdbc.JDBCConstants.QUOTE;
 import static com.ibm.fhir.persistence.jdbc.JDBCConstants.RIGHT_PAREN;
 
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.logging.Logger;
 
 import com.ibm.fhir.persistence.exception.FHIRPersistenceException;
@@ -22,6 +24,7 @@ import com.ibm.fhir.persistence.jdbc.connection.QueryHints;
 import com.ibm.fhir.persistence.jdbc.dao.api.JDBCIdentityCache;
 import com.ibm.fhir.persistence.jdbc.dao.api.ParameterDAO;
 import com.ibm.fhir.persistence.jdbc.dao.api.ResourceDAO;
+import com.ibm.fhir.search.SearchConstants;
 import com.ibm.fhir.search.parameters.InclusionParameter;
 
 /**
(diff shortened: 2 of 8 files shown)


fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -2444,13 +2447,32 @@ private Bundle createSearchBundle(List<Resource> resources, FHIRSearchContext se
...
+            Bundle.Entry entry = Bundle.Entry.builder()
+                    .fullUrl(Uri.of(getRequestBaseUri(type) + "/" + resource.getClass().getSimpleName() + "/" + resource.getId()))
+                    .resource(resource)
+                    .search(Search.builder()
+                        .mode(matchResourceCount-- > 0 ? SearchEntryMode.MATCH : SearchEntryMode.INCLUDE)
+                        .score(Decimal.of("1"))
+                        .build())
+                    .build();
 
             bundleBuider.entry(entry);
```
lmsurpre, line 2477: not your typo, but if you happen to make another round of edits, please fix this variable name bundleBuider -> bundleBuilder

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchRevIncludeTest.java
```diff
@@ -0,0 +1,774 @@
...
+        WebTarget target = getWebTarget();
+        Response response =
+                target.path("Patient")
+                .queryParam("_revinclude", "Procedure:patient")
+                .queryParam("_sort", "active")
+                .request(FHIRMediaType.APPLICATION_FHIR_JSON)
+                .get();
+        assertResponse(response, Response.Status.BAD_REQUEST.getStatusCode());
+        assertExceptionOperationOutcome(response.readEntity(OperationOutcome.class),
+                "_sort search result parameter not supported with _include or _revinclude.");
```
lmsurpre, line 477: in my opinion, this is an artificial limitation.  that said, its not really related to this PR and so I might open a separate issue on it (if anything)

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -2444,13 +2447,32 @@ private Bundle createSearchBundle(List<Resource> resources, FHIRSearchContext se
...
+        // Calculate how many resources are 'match' mode
+        int pageSize = searchContext.getPageSize();
+        int offset = (searchContext.getPageNumber() - 1) * pageSize;
+        int matchResourceCount = pageSize;
+        if (totalCount.getValue() < offset + pageSize) {
+            matchResourceCount = totalCount.getValue() - offset;
+        }
+
+        // Check if too many included resources
+        if (resources.size() > matchResourceCount + SearchConstants.MAX_PAGE_SIZE) {
```
lmsurpre, line 2459: the resources list can't ever include any OperationOutcome right?  i assume that gets added in to the response bundle at some later point...

lmsurpre (approved):
LGTM

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: Correct. At this point, the resources list will only contain resources returned from the query. For issue #1273 we might add OperationOutcome resources, but that could be done after this check.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -2444,13 +2447,32 @@ private Bundle createSearchBundle(List<Resource> resources, FHIRSearchContext se
...
+            String msg = "Number of returned 'include' resources exceeds allowable limit of " + SearchConstants.MAX_PAGE_SIZE;
```
prb112, line 2460: seems like this could actually be a final and possibly even final static at the class level.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -2444,13 +2447,32 @@ private Bundle createSearchBundle(List<Resource> resources, FHIRSearchContext se
...
                 throw new IllegalStateException("Returned resources must have an id.");
             }
-            Bundle.Entry entry = Bundle.Entry.builder().fullUrl(Uri.of(getRequestBaseUri(type) + "/"
-                    + resource.getClass().getSimpleName() + "/"
-                    + resource.getId())).resource(resource).build();
+            Bundle.Entry entry = Bundle.Entry.builder()
+                    .fullUrl(Uri.of(getRequestBaseUri(type) + "/" + resource.getClass().getSimpleName() + "/" + resource.getId()))
+                    .resource(resource)
+                    .search(Search.builder()
+                        .mode(matchResourceCount-- > 0 ? SearchEntryMode.MATCH : SearchEntryMode.INCLUDE)
```
prb112, line 2472: why are we decrementing `--` matchResourceCount (is this an off by one problem?)  If that's the case, shouldn't it be prefixed `--`?

prb112: also since matchResourceCount is calculated one time, why not create a local variable searchEntryMode instead of looping the calculation?

docs/src/pages/Conformance.md
```diff
@@ -128,7 +128,9 @@ Finally, the specification defines a set of "Search result parameters" for contr
 * `_summary`
 * `_elements`
 
-The `_count` parameter can be used to request up to 1000 records matching the search criteria.  An attempt to exceed this `_count` limit will not be honored and returned records will be capped at 1000.  Any associated `_include` records are not considered in the `_count` limit. 
+The `_count` parameter can be used to request up to 1000 resources matching the search criteria. An attempt to exceed this `_count` limit will not be honored and returned resources will be capped at 1000. Any associated `_include` or `_revinclude` resources are not considered in the `_count` limit. 
+
+The `_include` and `_revinclude` parameters can be used to return resources related to the primary search results, in order to reduce the overall network delay of repeated retrievals of related resources. The number of `_include` or `_revinclude` resources returned for a single page of primary search results will be limited to 1000. If the number of resources to be returned exceeds 1000, the search will fail.
```
prb112, line 133: Maybe make a clarifying statement... e.g. if there are 1 resource and 1000 resources included, the search succeeds, if 1 resource and 1001 resources are included the search fails? It may make the above more concrete with an example.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/InclusionQuerySegmentAggregator.java
```diff
@@ -412,4 +423,23 @@ private void addBindVariables(List<Object> bindVariables) {
...
+     * @param queryString A query string buffer.
+     */
+    private void addLimitClause(StringBuilder queryString) {
+        int limit = Integer.MAX_VALUE;
+        if (this.pageSize <= Integer.MAX_VALUE - (SearchConstants.MAX_PAGE_SIZE + 1)) {
+            limit = this.pageSize + SearchConstants.MAX_PAGE_SIZE + 1;
+        }
+
+        if (this.parameterDao.isDb2Database()) {
+            queryString.append(" LIMIT ").append(limit);
```
prb112, line 440: do we have a Constant for this ? I thought there was one? maybe we add one for 440,442's references

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchIncludeTest.java
```diff
@@ -0,0 +1,585 @@
...
+        // Next, call the 'read' API to retrieve the new procedure and verify it.
+        response = target.path("Procedure/" + procedure3Id).request(FHIRMediaType.APPLICATION_FHIR_JSON).get();
+        assertResponse(response, Response.Status.OK.getStatusCode());
+    }
+
+    @AfterClass
+    public void testDeleteResources() {
+        WebTarget target = getWebTarget();
+        if (patient1Id != null) {
+            Response response   = target.path("Patient/" + patient1Id).request(FHIRMediaType.APPLICATION_FHIR_JSON).delete();
```
prb112, line 271: if this fails, the following entries are never executed, perhaps split them out?

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchIncludeTest.java
```diff
@@ -0,0 +1,585 @@
...
+                .queryParam("_tag", tag)
+                .queryParam("gender", "female")
+                .queryParam("_include", "Patient:*")
+                .request(FHIRMediaType.APPLICATION_FHIR_JSON)
+                .get();
+        assertResponse(response, Response.Status.OK.getStatusCode());
+        Bundle bundle = response.readEntity(Bundle.class);
+
+        assertNotNull(bundle);
+        assertEquals(4, bundle.getEntry().size());
```
prb112, line 462: would this fail after @AfterClass fails to execute with the removal of prior test failures?

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchIncludeTest.java
```diff
@@ -0,0 +1,585 @@
...
+                .queryParam("_tag", tag)
+                .queryParam("gender", "female")
+                .queryParam("_include", "Patient:link")
+                .request(FHIRMediaType.APPLICATION_FHIR_JSON)
+                .get();
+        assertResponse(response, Response.Status.OK.getStatusCode());
+        Bundle bundle = response.readEntity(Bundle.class);
+
+        assertNotNull(bundle);
+        assertEquals(2, bundle.getEntry().size());
```
prb112, line 414: This is pretty locked in to a specific result.  If there is bad data, is this test resilient?

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchIncludeTest.java
prb112: I won't comment on the rest of them, just food for thought

prb112 (commented):
LGTM, however want to confirm a couple points where I left comments. specifically on the `--` usage with matchResourceCount.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: Regardless of how we do it, we have to do a compare of the iteration we're on inside the loop to 'matchResourceCount' in order to set the mode properly. I'm simply decrementing 'matchResourceCount' each time through the loop, and when we get to 0, we know we have to switch the mode from 'match' to 'include'. I don't think there's a more efficient way to do it, but if you think the code is more readable using a local variable for the mode I can do that.

prb112: Got it.  Do you mind adding a comment on the logic to the code? I see this as a hidden bit of logic, and it'd be good to surface it in a comment.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchIncludeTest.java
michaelwschroeder: The searches all get scoped to the set of resources created for a single test run by using a tag based on an Instant.now() value, so prior test artifacts should not be an issue, even if not cleaned up properly.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchIncludeTest.java
michaelwschroeder: See above comment. I believe the tests are resilient based on scoping of the searches to only those resources created in the current test run.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: Fixed.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: Moved to final static at class level.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: Comment added.

docs/src/pages/Conformance.md
michaelwschroeder: Example added.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/InclusionQuerySegmentAggregator.java
michaelwschroeder: Constants added for LIMIT, OFFSET, FETCH FIRST, FETCH NEXT, ROWS, ROWS ONLY

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchIncludeTest.java
michaelwschroeder: Done.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchIncludeTest.java
prb112: excellent!

prb112 (approved):
LGTM - thanks Mike

Issue #1874, opened 2021-01-20
Document validation of resource types in fhir-server-config.json with whole-system search


michaelwschroeder:
**Is your feature request related to a problem? Please describe.**
We need to document what must be specified in fhir-server-config.json in order for whole-system search to be allowed.

**Describe the solution you'd like**
Documentation should include an explanation of what is checked and example configurations.

**Describe alternatives you've considered**

**Acceptance Criteria**
At least one acceptance criteria is included.

1.  
  GIVEN [a precondition]
    AND [another precondition] WHEN [test step]
    AND [test step]
    THEN [verification step]
    AND [verification step]

**Additional context**

lmsurpre:
doc-only change; docs were reviewed

PR #1876, opened 2021-01-21
Issue #1874 - document resource type validation

michaelwschroeder:


docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -2088,7 +2088,26 @@ Omitting the `fhirServer/resources/<resourceType>/interactions` property is equi
...
+    },
+    "Patient": {
+        "interactions": ["read", "vread", "history", "search"]
+    }
+}
+```
+
+In this case, since the `fhirServer/resources/open` property is set to `false`, only the resource types listed (`Condition`, `Observation`, `Patient`) are allowed to be interacted with via the FHIR REST API. For example, a `create` request of a `Procedure` resource will fail since that resource type is not specified. Whole-system search is a special case of this resource type validation, since no resource type is specified on a whole-system search request. In this case, validation will be done against the `Resource` resource type. In the above configuration example, a whole-system search request such as `GET [base]?_lastUpdated=gt2020-01-01` will fail because the `Resource` resource type is not specified. If the configuration were to have the `fhirServer/resources/open` property set to `true`, or if the `Resource` resource type were specified in the `fhirServer/resources` property group, then the whole-system search request would be allowed, assuming the `search` interaction was valid for the `Resource` resource type.
```
lmsurpre, line 2110: I recommend putting the whole-system search description in a separate paragraph so it stands out a little better:
```suggestion
In this case, since the `fhirServer/resources/open` property is set to `false`, only the resource types listed (`Condition`, `Observation`, `Patient`) are allowed to be interacted with via the FHIR REST API. For example, a `create` request of a `Procedure` resource will fail since that resource type is not specified.

Whole-system search is a special case of this resource type validation, since no resource type is specified on a whole-system search request. In this case, validation will be done against the `Resource` resource type. In the above configuration example, a whole-system search request such as `GET [base]?_lastUpdated=gt2020-01-01` will fail because the `Resource` resource type is not specified. If the configuration were to have the `fhirServer/resources/open` property set to `true`, or if the `Resource` resource type were specified in the `fhirServer/resources` property group, then the whole-system search request would be allowed, assuming the `search` interaction was valid for the `Resource` resource type.
```

lmsurpre (approved):
LGTM

Issue #1179, opened 2020-06-01
Remove CXF from the server webapp


lmsurpre:
Liberty already provides JAX-RS 2.1 support (using CXF) and so we really shouldn't need to package our own version.

The one thing preventing it is our use of CXF-specific classes from FHIRUrlParser in fhir-provider:
```java
import org.apache.cxf.jaxrs.impl.MetadataMap;
import org.apache.cxf.jaxrs.utils.JAXRSUtils;
```

One option is to flip it to provided, but I'm not sure about that since liberty's use of CXF is more of an internal detail.

It only seems to be used from a couple places, so hopefully it wouldn't be too difficult to stop using these CXF-specific APIs:
![image.png](https://images.zenhubusercontent.com/5d6ff0e89320850001a28a98/018cae0e-4884-4372-b0e7-624c62056089)

Note: we DO want to continue packaging a working JAX-RS client implementation with our fhir-client, so I think

prb112:
Team discussion: There may be performance implications for it. 
If we chose not to do this, let's upgrade CXF.

prb112:
I poked at this one a bit today. 
We could mark them provided, and not bundle and it works. 

However, there are a couple more locations that reference CXF 

```
./fhir-provider/src/main/java/com/ibm/fhir/provider/util/FHIRUrlParser.java:11:import org.apache.cxf.jaxrs.impl.MetadataMap;
./fhir-provider/src/main/java/com/ibm/fhir/provider/util/FHIRUrlParser.java:12:import org.apache.cxf.jaxrs.utils.JAXRSUtils;
./fhir-client/src/main/java/com/ibm/fhir/client/impl/FHIRClientImpl.java:37:import org.apache.cxf.feature.LoggingFeature;
./fhir-client/src/main/java/com/ibm/fhir/client/impl/FHIRClientImpl.java:782:            // true: If need, tell Apache CXF to use the Async HTTP conduit for PATCH operation as the
./fhir-client/src/main/java/com/ibm/fhir/client/FHIRClient.java:100:     * Indicates whether or not to enable to CXF Logging feature which will log all request and response messages
```

And pom.xml references

```
./fhir-server-test/pom.xml:90:            <groupId>org.apache.cxf</groupId>
./fhir-server-test/pom.xml:91:            <artifactId>cxf-rt-frontend-jaxrs</artifactId>
./fhir-provider/pom.xml:41:            <groupId>org.apache.cxf</groupId>
[...]

PR #1884, opened 2021-01-25
upgrade liberty, postgresql, db2, and cxf

lmsurpre:
Please hold off on merging this until we decide whether to cut a 4.5.5

### DIFF
--- a/fhir-install/Dockerfile
+++ b/fhir-install/Dockerfile
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # ----------------------------------------------------------------------------
 
-FROM openliberty/open-liberty:20.0.0.9-full-java8-openj9-ubi as base
+FROM openliberty/open-liberty:20.0.0.12-full-java8-openj9-ubi as base
 
 ENV LICENSE accept
 
@@ -20,9 +20,9 @@ RUN unzip -qq /tmp/fhir-server-distribution.zip -d /tmp && \
 
 # ----------------------------------------------------------------------------
 
-FROM openliberty/open-liberty:20.0.0.9-full-java8-openj9-ubi
+FROM openliberty/open-liberty:20.0.0.12-full-java8-openj9-ubi
 
-ARG FHIR_VERSION=4.5.1
+ARG FHIR_VERSION=4.6.0
 
 # The following labels are required: 
 LABEL name='IBM FHIR Server'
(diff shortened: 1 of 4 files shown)


fhir-install/Dockerfile
```diff
@@ -20,9 +20,9 @@ RUN unzip -qq /tmp/fhir-server-distribution.zip -d /tmp && \
 
 # ----------------------------------------------------------------------------
 
-FROM openliberty/open-liberty:20.0.0.9-full-java8-openj9-ubi
+FROM openliberty/open-liberty:20.0.0.12-full-java8-openj9-ubi
 
-ARG FHIR_VERSION=4.5.1
+ARG FHIR_VERSION=4.6.0
```
lmsurpre, line 25: Our build scripts must overwrite this value because obviously 4.5.1 wasn't right, but I thought it would still be good to try keeping this default value up-to-date

prb112: yes they do

fhir-parent/pom.xml
```diff
@@ -10,11 +10,11 @@
...
     <properties>
-        <liberty.version>20.0.0.9</liberty.version>
+        <liberty.version>20.0.0.12</liberty.version>
         <derby.version>10.14.2.0</derby.version>
-        <db2.version>11.5.0.0</db2.version>
-        <postgresql.version>42.2.12</postgresql.version>
-        <cxf.version>3.3.6</cxf.version>
+        <db2.version>11.5.5.0</db2.version>
+        <postgresql.version>42.2.18</postgresql.version>
+        <cxf.version>3.3.9</cxf.version>
```
prb112, line 17: Should we go up to 3.4.2?

lmsurpre: We probably could if we want to.  I stayed on the more catious side since it looked like most of the important fixes were backported to 3.3.x anyway.  I still want to see if we can remove the embedded CXF version altogether since we should be able to get everything we need from Liberty.

prb112: maybe we should have an issue to track that work? and target 4.6.0?

lmsurpre: https://github.com/IBM/FHIR/issues/1179 ;-)

prb112 (approved):
LGTM

ldettman (commented):
Looks good.

PR #1887, opened 2021-01-26
Make the 'compartment param not found' warning conditional

lmsurpre:
If `useStoredCompartmentParam` is set to true, then we shouldn't need the compartment inclusion parameters to be externally visible.


punktilious (approved):
LGTM

Issue #1661, opened 2020-11-02
Search with no error or warning, just ignoring what I passed in on the search.


d0roppe:
**Describe the bug**
I sent a search like below to the FHIR server:  and it treated it as a patient read, and ignored the search parameters.  with no warning or error. 
fhir server should issue some error if you send information that is not understood. instead of just dropping it. 

**To Reproduce**
1. Start the IBM FHIR Server
2. Execute the request `[base]/Patient/[id]/?_lastUpdated=ge2020-11-02`

**Expected behavior**
If there is more data in an API call that is not understood. fhir should issue an error. 

**Additional context**
n/a

prb112:
I don't see this as a bug, rather a feature.  I suggest an enhancement request, not a bug.

lmsurpre:
I also think its more of a feature request than a bug.   The request is that we support "whole-system" search scoped to a particular compartment.  The desire is that:
* `[base]/Patient/[id]/ResourceType?param=value` searches for ResourceTypes within the patient compartment
* `[base]/Patient/[id]/?_lastUpdated=ge2020-11-02` search ALL resource types within the patient compartment

Note: overlaps with #264

lmsurpre:
The idea on this one is just to emit some kind of message to indicate that we don't support searching across multiple resource types within a compartment ( #1658)

tbieste:
Since there are no characters (except query parameters) after the final '/', the JAX-RS processing is sending this down the "read" code path, instead of the "search" code path, so the FHIR Server never even sees the final '/'.

One option is to add @Path("{compartment}/{compartmentId}/") to Search.java, so the "search" code path is invoked and will see that the resource type was not specified, and send the following error:
```
{
    "resourceType": "OperationOutcome",
    "id": "c0-a8-1-b-2a83384e-b3d8-446b-a557-d02fe5f409f3",
    "issue": [
        {
            "severity": "fatal",
            "code": "not-supported",
            "_code": {
                "extension": [
                    {
                        "url": "http://ibm.com/fhir/extension/not-supported-detail",
                        "valueCode": "resource"
                    }
                ]
            },
            "details": {
                "text": "'' is not a valid resource type."
            }
        }
    ]
}
```

tbieste:
Updated the read operation to handle the _elements parameters (see https://www.hl7.org/fhir/http.html#read), and to return an error if any non-general search parameters are used -- instead of just ignoring them (or only returning an error if that are not a valid search parameter for searching that resource type).

[...]

PR #1896, opened 2021-02-01
Issue #1661 - Read returns error when invalid search parameters used

tbieste:
Updated read to support the _elements search parameter, and return an error when invalid search parameters are used.

### DIFF
--- a/fhir-server-test/src/test/java/com/ibm/fhir/server/test/ReadTest.java
+++ b/fhir-server-test/src/test/java/com/ibm/fhir/server/test/ReadTest.java
@@ -6,7 +6,17 @@
 
 package com.ibm.fhir.server.test;
 
+import static com.ibm.fhir.model.type.String.string;
+import static com.ibm.fhir.model.type.Uri.uri;
+import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertNull;
+import static org.testng.Assert.assertTrue;
+
+import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 
 import javax.ws.rs.client.Entity;
 import javax.ws.rs.client.WebTarget;
@@ -18,17 +28,25 @@ import com.ibm.fhir.core.FHIRMediaType;
 import com.ibm.fhir.model.resource.OperationOutcome;
 import com.ibm.fhir.model.resource.Patient;
 import com.ibm.fhir.model.test.TestUtil;
+import com.ibm.fhir.model.type.Code;
+import com.ibm.fhir.model.type.Coding;
 import com.ibm.fhir.model.type.code.AdministrativeGender;
+import com.ibm.fhir.model.util.FHIRUtil;
 
 public class ReadTest extends FHIRServerTestBase {
     private String patientId;
+    private Coding subsettedTag =
+            Coding.builder().system(uri("http://terminology.hl7.org/CodeSystem/v3-ObservationValue"))
+            .code(Code.of("SUBSETTED"))
+            .display(string("subsetted"))
+            .build();
 
     @Test(groups = { "server-read" })
     public void testCreatePatient() throws Exception {
         WebTarget target = getWebTarget();
 
         // Build a new Patient and then call the 'create' API.
-        Patient patient = TestUtil.readLocalResource("Patient_JohnDoe.json");
+        Patient patient = TestUtil.readLocalResource("patient-example-a.json");
 
         patient = patient.toBuilder().gender(AdministrativeGender.MALE).build();
         Entity<Patient> entity =
@@ -51,7 +69,7 @@ public class ReadTest extends FHIRServerTestBase {
     }
 
     @Test(groups = { "server-read" }, dependsOnMethods = {"testCreatePatient" })
-    public void testReadPatient_elements() {
+    public void testReadPatient_elements() throws Exception {
         WebTarget target = getWebTarget();
         Response response =
                 target.path("Patient/" + patientId)
@@ -60,10 +78,11 @@ public class ReadTest extends FHIRServerTestBase {
         assertResponse(response, Response.Status.OK.getStatusCode());
         Patient patient = response.readEntity(Patient.class);
         assertNotNull(patient);
+        checkSubsetted(patient, Collections.singletonList("getActive"));
     }
 
     @Test(groups = { "server-read" }, dependsOnMethods = {"testCreatePatient" })
-    public void testReadPatient_summary() {
+    public void testReadPatient_summary() throws Exception {
         WebTarget target = getWebTarget();
         Response response =
                 target.path("Patient/" + patientId)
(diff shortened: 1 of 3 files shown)


fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -740,6 +741,10 @@ private Resource doRead(String type, String id, boolean throwExcOnNull, boolean
 
             FHIRSearchContext searchContext = null;
             if (queryParameters != null) {
+                if (!queryParameters.keySet().stream().allMatch(k -> FHIRConstants.GENERAL_PARAMETER_NAMES.contains(k))) {
+                    throw SearchExceptionUtil.buildNewInvalidSearchException(
```
michaelwschroeder, line 745: Wondering if we should take into account the 'lenient' vs. 'strict' setting, and only log the issue if in lenient mode.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/ReadTest.java
```diff
@@ -0,0 +1,89 @@
...
+        response = target.path("Patient/"
+                + patientId).request(FHIRMediaType.APPLICATION_FHIR_JSON)
+                .get();
+        assertResponse(response, Response.Status.OK.getStatusCode());
+        Patient responsePatient = response.readEntity(Patient.class);
+        TestUtil.assertResourceEquals(patient, responsePatient);
+    }
+
+    @Test(groups = { "server-read" }, dependsOnMethods = {"testCreatePatient" })
+    public void testReadPatient_elements() {
```
michaelwschroeder, line 54: for this test and the testReadPatient_summary test, should probably validate fields within the returned resource to verify _elements (or _summary) processing was completed successfully.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/ReadTest.java
```diff
@@ -0,0 +1,89 @@
...
+        WebTarget target = getWebTarget();
+        Response response =
+                target.path("Patient/" + patientId)
+                .queryParam("_lastUpdated", "ge2020-11-02")
+                .request(FHIRMediaType.APPLICATION_FHIR_JSON).get();
+        assertResponse(response, Response.Status.BAD_REQUEST.getStatusCode());
+        assertExceptionOperationOutcome(response.readEntity(OperationOutcome.class),
+            "Read only supports general search parameters.");
+    }
+
```
michaelwschroeder, line 88: maybe add a test where you specify both _elements and _summary and verify _elements gets processed.

tbieste: Added test.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/ReadTest.java
tbieste: Added validating of returned fields.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
tbieste: Updated to only log the issue if in lenient mode.

michaelwschroeder (approved):

michaelwschroeder:
LGTM

fhir-server-test/src/test/java/com/ibm/fhir/server/test/ReadTest.java
```diff
@@ -0,0 +1,155 @@
...
+    @Test(groups = { "server-read" }, dependsOnMethods = {"testCreatePatient" })
+    public void testReadPatient_nonGeneralSearchParameter() {
+        WebTarget target = getWebTarget();
+        Response response =
+                target.path("Patient/" + patientId)
+                .queryParam("_lastUpdated", "ge2020-11-02")
+                .request(FHIRMediaType.APPLICATION_FHIR_JSON).get();
+        assertResponse(response, Response.Status.BAD_REQUEST.getStatusCode());
+        assertExceptionOperationOutcome(response.readEntity(OperationOutcome.class),
+            "Read only supports general search parameters.");
```
lmsurpre, line 121: I wonder if we should just list them here since there's only 2.  Is "general search parameters" something defined in the spec?

tbieste: There's a section in the spec called General parameters, https://www.hl7.org/fhir/http.html#parameters, for ones that apply to any interaction. (_format, _pretty, _summary, _elements)  I could just mention _summary and _elements.  Just wasn't sure what would be the least confusing way to phrase it.

lmsurpre: ah, i do think we support `_format` and `_pretty` for all requests (including reads).  if so, i think your language is good.

lmsurpre: sorry, thought about this some more and I still think we should either:
A. list the supported parameters; or
B. tell the user which parameter they used which is invalid

tbieste: Updated to indicate the invalid parameter(s).

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -754,12 +754,13 @@ private Resource doRead(String type, String id, boolean throwExcOnNull, boolean
                 boolean lenient = HTTPHandlingPreference.LENIENT.equals(requestContext.getHandlingPreference());
 
                 // Determine if any non-general search parameters are specified
-                if (!queryParameters.keySet().stream().allMatch(k -> FHIRConstants.GENERAL_PARAMETER_NAMES.contains(k))) {
-                    FHIRSearchException se = SearchExceptionUtil.buildNewInvalidSearchException("Read only supports general search parameters.");
+                List<String> nonGeneralParams = queryParameters.keySet().stream().filter(k -> !FHIRConstants.GENERAL_PARAMETER_NAMES.contains(k)).collect(Collectors.toList());
+                for (String nonGeneralParam : nonGeneralParams) {
```
michaelwschroeder, line 758: Might be nice to show all the failing parms in the exception msg, since the way it's written you'll only get the first one before you throw the exception. Rather than a for loop, maybe a check if list is not empty, then update the exception msg to be "Search parameters " + nonGeneralParams + " not supported by read". It'll format the parm strings inside brackets. Then for the log msg, can use your prev msg since the exception will contain the failing parms.

tbieste: Yeah, I went this route since this is how all the current invalid search parameter checking works.  Which is, as soon as an invalid search parameter (or value) is found, it throws an exception for the single parameter and parsing ends.  I can have just this error work differently, but thought consistency would be better.

michaelwschroeder: I'm fine with leaving it the way you have it.

tbieste:
Comments addressed. Ready to merge.

lmsurpre (approved):
LGTM

PR #1901, opened 2021-02-02
Convert ImportOperation Parameter Processing to use FHIR Path #951

### LINKED ISSUE #951: BulkData $import API improvement - Convert Parameter Processing to use FHIR Path
**BulkData $import API improvement - Convert Parameter Processing to use FHIR Path**

The parameters rely on a non-Normative pattern for transmitting $import operation. 
The intent is to add dynamicity, and convert Parameter Processing to use FHIR Path in AbstractOperation

Reference #671

prb112:
Closes #951 

### DIFF
--- a/operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataImportUtil.java
+++ b/operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataImportUtil.java
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright IBM Corp. 2019, 2020
+ * (C) Copyright IBM Corp. 2019, 2021
  *
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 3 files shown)


operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataImportUtil.java
```diff
@@ -31,66 +31,79 @@
...
+     *
+     * @return
+     * @throws FHIROperationException
+     */
+    public String retrieveInputFormat() throws FHIROperationException {
+        try {
+            Collection<FHIRPathNode> result = evaluator.evaluate(evaluationContext, "parameter.where(name = 'inputFormat').value");
+            Iterator<FHIRPathNode> iter = result.iterator();
+            while (iter.hasNext()) {
+                FHIRPathElementNode node = (FHIRPathElementNode) iter.next();
```
lmsurpre, line 58: really no different, but since we have it:
```suggestion
                FHIRPathElementNode node = iter.next().as(FHIRPathElementNode.class);
```

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataImportUtil.java
```diff
@@ -31,66 +31,79 @@
...
+                String val = node.asElementNode().element().as(com.ibm.fhir.model.type.String.class).getValue();
+                if (BulkDataConstants.INPUT_FORMATS.contains(val)) {
+                    return val;
                 }
             }
+        } catch (ClassCastException ce) {
+            throw buildExceptionWithIssue("$import invalid parameter name in 'inputFormat'", ce, IssueType.INVALID);
+        } catch (FHIRPathException e) {
+            throw buildExceptionWithIssue("$import invalid parameters value type in 'inputFormat'", e, IssueType.INVALID);
```
lmsurpre, line 67: the error messages are a little confusing to me; would something like this be better?
```suggestion
        } catch (ClassCastException ce) {
            throw buildExceptionWithIssue("$import invalid parameter value in 'inputFormat'", ce, IssueType.INVALID);
        } catch (FHIRPathException e) {
            throw buildExceptionWithIssue("Unexpected error while processing $import parameter 'inputFormat'", e, IssueType.INVALID);
```

lmsurpre: because I think we'd typically hit the first one when the inputFormat parameter value isn't a FHIR string and I wouldn't expect to ever hit the second one.  Or is there some case that throws FHIRPathException that I'm missing?

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataImportUtil.java
```diff
@@ -31,66 +31,79 @@
...
-                }
+    /**
+     * processes the retrieveInputSource from the parameters object and evaluationContext.
+     *
+     * @implNote If there are multiple entries, the processing only takes the first entry.
+     *
+     * @return
+     * @throws FHIROperationException
+     */
+    public String retrieveInputSource() throws FHIROperationException {
```
lmsurpre, line 81: looks pretty similar to the previous method...wonder if it could be a single helper

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataImportUtil.java
```diff
@@ -31,66 +31,79 @@
...
+            Collection<FHIRPathNode> result = evaluator.evaluate(evaluationContext, "parameter.where(name = 'inputSource').value");
+            Iterator<FHIRPathNode> iter = result.iterator();
+            while (iter.hasNext()) {
+                FHIRPathElementNode node = (FHIRPathElementNode) iter.next();
+                return node.asElementNode().element().as(com.ibm.fhir.model.type.Uri.class).getValue();
             }
+        } catch (ClassCastException ce) {
+            throw buildExceptionWithIssue("$import invalid parameter name in 'inputSource'", ce, IssueType.INVALID);
+        } catch (FHIRPathException e) {
+            throw buildExceptionWithIssue("$import found invalid parameter type while processing inputSource'", e, IssueType.INVALID);
```
lmsurpre, line 92:
```suggestion
        } catch (ClassCastException | FHIRPathException e) {
            throw buildExceptionWithIssue("invalid $import parameter value in 'inputSource'", e, IssueType.INVALID);
```

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataImportUtil.java
```diff
@@ -244,11 +254,23 @@ public static StorageDetail retrieveStorageDetails(Parameters parameters) throws
...
+            throws FHIROperationException {
+        OperationOutcome.Issue ooi = FHIRUtil.buildOperationOutcomeIssue(msg, issueType);
+        return new FHIROperationException(msg).withIssue(ooi);
+    }
+
+    public FHIROperationException buildExceptionWithIssue(String msg, Throwable cause, IssueType issueType)
+            throws FHIROperationException {
+        OperationOutcome.Issue ooi = FHIRUtil.buildOperationOutcomeIssue(msg, issueType);
+        return new FHIROperationException(msg, cause).withIssue(ooi);
+    }
```
lmsurpre, line 275: should these be private?

also, I wonder if we have these in enough places to promote them to FHIRUtil...but I'd save that for some other time

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataImportUtil.java
lmsurpre: probably not worth it if we only have 2

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataImportUtil.java
prb112: That is true.

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataImportUtil.java
prb112: Yeah - I left them separate, but very similar as they are different types.

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataImportUtil.java
prb112: I just moved these existing ones around, and yes, they probably should be in a util, but... I think in this case it'd go to FHIROperationUtil since it throws FHIROperationException or we could generalize it.

prb112: I can do that in this one, or leave as-is

lmsurpre: i see it now.  used to be at the top of the class (lines 38 and 44 respectively).  lets call this one good.

lmsurpre (approved):
LGTM

prb112:
merged as I'll incorporate testing it with my other open issues

PR #1902, opened 2021-02-02
Control the writing of operational outcomes with https $import

prb112:

### DIFF
--- a/docs/src/pages/guides/FHIRServerUsersGuide.md
+++ b/docs/src/pages/guides/FHIRServerUsersGuide.md
@@ -3,7 +3,7 @@ layout: post
 title:  IBM FHIR Server User's Guide
 description: IBM FHIR Server User's Guide
 Copyright: years 2017, 2021
-lastupdated: "2021-01-21"
+lastupdated: "2021-02-22"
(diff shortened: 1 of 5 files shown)


lmsurpre:
The changes look good to me, but it makes me wonder if we should at least provide a high-level summary of the resources which failed to import...any thoughts/ideas on that?

prb112:
> The changes look good to me, but it makes me wonder if we should at least provide a high-level summary of the resources which failed to import...any thoughts/ideas on that?

Good point... I think we could do that. It shouldn't be so much of a problem

prb112:
```

**Comment by prb112** [comment]
```
{
    "transactionTime": "2021-02-05T14:00:57.717-05:00",
    "request": "$import",
    "requiresAccessToken": false,
    "output": [
        {
            "type": "OperationOutcome",
            "url": "test-import.ndjson_oo_success.ndjson",
            "count": 4
        }
    ],
    "error": [
        {
            "type": "OperationOutcome",
            "url": "test-import.ndjson_oo_errors.ndjson",
            "count": 0
        }
    ]
* Connection #0 to host localhost left intact
}*
```

This is what is returned.

Issue #1557, opened 2020-10-05
A search with _include returns the same included resource multiple times if it matches more than once.


d0roppe:
**Describe the bug**
In you search a resource and use the _include and the include matching resource is identical for more than one matching.  it is included more than once.  We should try to only return distinct include matching resources to reduce the data returned. 
**To Reproduce**

**Expected behavior**
a included resource should be unique for the search result and not returned multiple times. 
**Additional context**
Example would be searching the Coverage resource, and using _include=Coverage:*  the same patient could be the policyHolder and subscriber, and that patient resource should be returned only once.

prb112:
Team discussion: 
- Core Problem: Duplicate on the Single Page
- Design work and team agreement must be reached before proceeding with the broader change of duplicates being returned cross pages.

d0roppe:
stretch goal for sprint 21

d0roppe:
verified that _include does only return 1 copy of the included resource, version specific. in case there is another version of that resource matching the include, the other version is also returned, so still unique.  Closing issue.

PR #1910, opened 2021-02-04
Issue #1557 - remove dups, ignore _elements for include resources

michaelwschroeder:

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
@@ -28,6 +28,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -681,7 +682,43 @@ public class FHIRPersistenceJDBCImpl implements FHIRPersistence, SchemaNameSuppl
                         }
                     } else {
                         unsortedResultsList = resourceDao.search(query);
-                        resources = this.convertResourceDTOList(unsortedResultsList, resourceType, elements);
+                        List<com.ibm.fhir.persistence.jdbc.dto.Resource> matchResultList = unsortedResultsList;
+                        List<com.ibm.fhir.persistence.jdbc.dto.Resource> includeResultList = new ArrayList<>();
+
+                        // Check if _include or _revinclude search. If so, remove duplicates from 'include' resources
+                        // (duplicates of both 'match' and 'include' resources) and make sure _elements processing
+                        // is not done for 'include' resources.
+                        if (searchContext.hasIncludeParameters() || searchContext.hasRevIncludeParameters()) {
+                            // Calculate 'match' results count
+                            int pageSize = searchContext.getPageSize();
+                            int offset = (searchContext.getPageNumber() - 1) * pageSize;
+                            int matchResultCount = pageSize;
+                            if (searchResultCount < offset + pageSize) {
+                                matchResultCount = searchResultCount - offset;
+                            }
+
+                            // Split results list into 'match' list and 'include' list
+                            if (unsortedResultsList.size() > matchResultCount) {
+                                matchResultList = unsortedResultsList.subList(0, matchResultCount);
+
+                                // Remove duplicates from 'include' list
+                                Set<Long> resultIds = new HashSet<>();
+                                for (com.ibm.fhir.persistence.jdbc.dto.Resource resource : matchResultList) {
+                                    resultIds.add(resource.getId());
+                                }
+                                for (int i=matchResultCount; i<unsortedResultsList.size(); ++i) {
+                                    com.ibm.fhir.persistence.jdbc.dto.Resource resource = unsortedResultsList.get(i);
+                                    if (!resultIds.contains(resource.getId())) {
+                                        resultIds.add(resource.getId());
+                                        includeResultList.add(resource);
+                                    }
+                                }
+                            }
+                        }
+
+                        // Convert resources
+                        resources = this.convertResourceDTOList(matchResultList, resourceType, elements);
+                        resources.addAll(this.convertResourceDTOList(includeResultList, resourceType, null));
                     }
                 }
             }
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
@@ -606,8 +606,6 @@ public class FHIRPersistenceJDBCImpl implements FHIRPersistence, SchemaNameSuppl
         MultiResourceResult.Builder<Resource> resultBuilder = new MultiResourceResult.Builder<>();
         FHIRSearchContext searchContext = context.getSearchContext();
         JDBCQueryBuilder queryBuilder;
-        List<Long> sortedIdList;
-        List<com.ibm.fhir.persistence.jdbc.dto.Resource> unsortedResultsList;
         int searchResultCount = 0;
         SqlQueryData countQuery;
         SqlQueryData query;
@@ -677,12 +675,11 @@ public class FHIRPersistenceJDBCImpl implements FHIRPersistence, SchemaNameSuppl
                         if (resourceType.equals(Resource.class)) {
                            resources = this.convertResourceDTOList(resourceDao.search(query), resourceType, elements);
                         } else {
-                            sortedIdList = resourceDao.searchForIds(query);
-                            resources = this.buildSortedFhirResources(resourceDao, context, resourceType, sortedIdList, elements);
+                            resources = this.buildSortedFhirResources(resourceDao, context, resourceType, resourceDao.searchForIds(query), elements);
                         }
                     } else {
-                        unsortedResultsList = resourceDao.search(query);
-                        List<com.ibm.fhir.persistence.jdbc.dto.Resource> matchResultList = unsortedResultsList;
+                        List<com.ibm.fhir.persistence.jdbc.dto.Resource> resultsList = resourceDao.search(query);
+                        List<com.ibm.fhir.persistence.jdbc.dto.Resource> matchResultList = resultsList;
                         List<com.ibm.fhir.persistence.jdbc.dto.Resource> includeResultList = new ArrayList<>();
 
                         // Check if _include or _revinclude search. If so, remove duplicates from 'include' resources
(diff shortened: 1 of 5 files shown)


michaelwschroeder:
This PR is for both issue 1557 (A search with _include returns the same included resource multiple times if it matches more than once) and for issue 1516 (_elements option should not be applied to the _include option on a search)

fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractIncludeRevincludeTest.java
```diff
@@ -7,6 +7,7 @@
 package com.ibm.fhir.persistence.test.common;
 
 import static com.ibm.fhir.model.type.String.string;
+import static org.junit.Assert.assertNull;
```
prb112, line 10: let's use the testng assert 

`import static org.testng.AssertJUnit.assertNull;`

fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractIncludeRevincludeTest.java
```diff
@@ -385,6 +475,44 @@ else if (resource instanceof Patient) {
...
+        for (Resource resource : resources) {
+            if (resource instanceof Observation) {
+                assertEquals(savedObservation6.getId(), resource.getId());
+                // check elements are not filtered out
+                assertEquals("Device/" + savedDevice1.getId() + "/_history/1", resource.as(Observation.class).getSubject().getReference().getValue());
+                assertEquals("Device/" + savedDevice1.getId() + "/_history/3", resource.as(Observation.class).getDevice().getReference().getValue());
+                assertEquals("Device/" + savedDevice1.getId() + "/_history/2", resource.as(Observation.class).getFocus().get(0).getReference().getValue());
+                assertEquals("Specimen/1", resource.as(Observation.class).getSpecimen().getReference().getValue());
+            }
+            else if (resource instanceof Device) {
```
prb112, line 503: note the pattern of continued lines, not line breaks 

} else if ... 

we can get that next time we're in the code.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchIncludeTest.java
```diff
@@ -7,6 +7,7 @@
 package com.ibm.fhir.server.test;
 
 import static com.ibm.fhir.model.type.String.of;
+import static org.junit.Assert.assertNull;
```
prb112, line 10: convert to testng (per the prior example in the review)

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchRevIncludeTest.java
```diff
@@ -7,6 +7,7 @@
 package com.ibm.fhir.server.test;
 
 import static com.ibm.fhir.model.type.String.of;
+import static org.junit.Assert.assertNull;
```
prb112, line 10: per the prior coment ocnvert to testng's version of AssertJunit

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -681,7 +682,43 @@ private ParameterDAO makeParameterDAO(Connection connection) throws FHIRPersiste
...
+                                // Remove duplicates from 'include' list
+                                Set<Long> resultIds = new HashSet<>();
+                                for (com.ibm.fhir.persistence.jdbc.dto.Resource resource : matchResultList) {
+                                    resultIds.add(resource.getId());
+                                }
+                                for (int i=matchResultCount; i<unsortedResultsList.size(); ++i) {
+                                    com.ibm.fhir.persistence.jdbc.dto.Resource resource = unsortedResultsList.get(i);
+                                    if (!resultIds.contains(resource.getId())) {
+                                        resultIds.add(resource.getId());
+                                        includeResultList.add(resource);
```
prb112, line 713: I've stared at this logic for a few moments. 

Could you use a LinkedHashMap (preserve order) and use a computeIfNotPresent? 

@punktilious thoughts on optimizations?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -681,7 +682,43 @@ private ParameterDAO makeParameterDAO(Connection connection) throws FHIRPersiste
...
+                            int pageSize = searchContext.getPageSize();
+                            int offset = (searchContext.getPageNumber() - 1) * pageSize;
+                            int matchResultCount = pageSize;
+                            if (searchResultCount < offset + pageSize) {
+                                matchResultCount = searchResultCount - offset;
+                            }
+
+                            // Split results list into 'match' list and 'include' list
+                            if (unsortedResultsList.size() > matchResultCount) {
+                                matchResultList = unsortedResultsList.subList(0, matchResultCount);
```
lmsurpre, line 702: I'm confused by the variable name `unsortedResultsList`.  If its unsorted, then how do we know the `match` resources come first?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -681,7 +682,43 @@ private ParameterDAO makeParameterDAO(Connection connection) throws FHIRPersiste
...
+
+                                // Remove duplicates from 'include' list
+                                Set<Long> resultIds = new HashSet<>();
+                                for (com.ibm.fhir.persistence.jdbc.dto.Resource resource : matchResultList) {
+                                    resultIds.add(resource.getId());
+                                }
+                                for (int i=matchResultCount; i<unsortedResultsList.size(); ++i) {
+                                    com.ibm.fhir.persistence.jdbc.dto.Resource resource = unsortedResultsList.get(i);
+                                    if (!resultIds.contains(resource.getId())) {
```
lmsurpre, line 711: Question:  if a resource is both a `match` *and* and `include`, should we return it just once (as a match), or twice (once as a match and once as an include)?
This code seems to support the former:  a given resource will only be in the results once and if its both a match and an include then it will only be labeled as a match.

fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractIncludeRevincludeTest.java
michaelwschroeder: Done.

fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractIncludeRevincludeTest.java
michaelwschroeder: Done.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchIncludeTest.java
michaelwschroeder: Done.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchRevIncludeTest.java
michaelwschroeder: Done.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
michaelwschroeder: This is a holdover from the existing code. It's unsorted in that the _sort parameter was not specified for this search. However, the match results are ensured to be sorted before the include results based on the SQL sort I implemented for issue #1494.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
michaelwschroeder: Yes, I went by the tenet that a resource should only appear once in a search result bundle. Can certainly go the other way if that's the consensus.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
michaelwschroeder: not quite sure how I would take advantage of that here...I have to build the includeResultList to pass to the convertResourceDTOList() on line 721

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -681,7 +682,43 @@ private ParameterDAO makeParameterDAO(Connection connection) throws FHIRPersiste
...
                     } else {
                         unsortedResultsList = resourceDao.search(query);
-                        resources = this.convertResourceDTOList(unsortedResultsList, resourceType, elements);
+                        List<com.ibm.fhir.persistence.jdbc.dto.Resource> matchResultList = unsortedResultsList;
+                        List<com.ibm.fhir.persistence.jdbc.dto.Resource> includeResultList = new ArrayList<>();
+
+                        // Check if _include or _revinclude search. If so, remove duplicates from 'include' resources
+                        // (duplicates of both 'match' and 'include' resources) and make sure _elements processing
+                        // is not done for 'include' resources.
+                        if (searchContext.hasIncludeParameters() || searchContext.hasRevIncludeParameters()) {
```
tbieste, line 691: Unless it's mentioned elsewhere (or well-known), I didn't initially know that all the matches come first in the returned search results, and are then followed by the includes.  Maybe a quick comment mentioning that, which explains why the code assumes that order.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
tbieste: I cannot find guidance in the spec of which option is preferred.  Give that, I think the approach that a resource should only appear once in a search result bundle is a good one.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
michaelwschroeder: I renamed this variable to avoid confusion, and added a comment in the section of code where I'm splitting the results to clarify that the 'match' results have been sorted ahead of the 'include' results.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
michaelwschroeder: Done.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
lmsurpre: I'm good with it too, just thought it was an important decision worth pondering (and maybe including somewhere either in the design doc or in our conformance.md doc)

lmsurpre (approved):
LGTM

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
michaelwschroeder: I documented in the conformance doc.

tbieste (approved):
LGTM

Issue #1206, opened 2020-06-08
FHIRNotificationEvent does not support multi-tenancy


rgupta3349:
**Is your feature request related to a problem? Please describe.**
FHIRNotificationEvent does not support multi-tenancy. 
Fhir Server can published Create/Update event as FHIRNotificationEvent but tenantid is not part of FHIRNotificationEvent object. For multi tenancy it would be really helpful to know which event is intended for which tenant.

**Describe the solution you'd like**
FHIRNotificationEvent object to include tenantid.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

rgupta3349:
1214 quote: In its current state, it should only be used for single-tenant, single-server deployments and, at the very least, the docs should make this extremely clear.

Is there any plan in your road map for Fhir notification to support multi-tenancy?

prb112:
This issue is not in our current sprint plan.

prb112:
**Team Discussion:** 
Only scope this to the tenantId update in the FHIRNotificationEvent. 
If there needs to be follow on work with specific operator/user use cases, these are to be addressed in separate issues.

PR #1919, opened 2021-02-08
FHIRNotificationEvent does not support multi-tenancy #1206

prb112:
Closing #1206 

### DIFF
--- a/fhir-notification/src/main/java/com/ibm/fhir/notification/FHIRNotificationService.java
+++ b/fhir-notification/src/main/java/com/ibm/fhir/notification/FHIRNotificationService.java
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright IBM Corp. 2016, 2020
+ * (C) Copyright IBM Corp. 2016, 2021
  *
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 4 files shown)


fhir-notification/src/main/java/com/ibm/fhir/notification/FHIRNotificationService.java
```diff
@@ -207,6 +208,12 @@ private FHIRNotificationEvent buildNotificationEvent(String operation, FHIRPersi
             event.setResourceId(resource.getId());
             event.setResource(resource);
 
+            // Adds the tenant id and datastore id
+            String tenantId = FHIRRequestContext.get().getTenantId();
+            String dsId = FHIRRequestContext.get().getDataStoreId();
+            event.setDatasourceId(dsId);
+            event.setTenantId(tenantId);
```
lmsurpre, line 215: I wonder if we should avoid setting these when the tenant and datastore are default/default, which would be typical for a single-tenant deployment.

prb112: I think we should include them as a matter of practice.  It'll allow someone to recreate a request and retrieve the full data and use it purposefully (and not by accident)

lmsurpre (approved):
LGTM

Issue #1844, opened 2021-01-05
Do not log full stack trace for a simple 404 not found


punktilious:
**Describe the bug**
When requesting a resource which doesn't exist, the FHIR server responds with a 404 not found, but it also logs an exception with a full stack trace. This is not necessary, because the code is behaving as expected.

**Expected behavior**
Return the 404 which is already logged. Do not log the exception/stack trace.

prb112:
I cannot replicate this issue.

1 - Try Getting a Random Resource

```
curl --location --request GET 'https://localhost:9443/fhir-server/api/v4/Patient/AAAAAA' --header 'Authorization: Basic ZmhpcnVzZXI6Y2hhbmdlLXBhc3N3b3Jk' --data-raw '' -k
{"resourceType":"OperationOutcome","id":"c0-a8-56-16-e25f7d41-09a6-4bf1-9dd4-fa35eb2991ce","issue":[{"severity":"fatal","code":"not-found","details":{"text":"Resource 'Patient/AAAAAA' not found."},"expression":["<empty>"]}]}
```

If OK, we close. 

Else can provide more details on the request?

prb112:
Oh - it only shows up in the trace.log

```
[2/8/21, 11:39:59:549 EST] 00000078 FHIRResource  I   Resource 'Patient/AAAAAA' not found.
                                 com.ibm.fhir.persistence.exception.FHIRPersistenceResourceNotFoundException: Resource 'Patient/AAAAAA' not found.  [probeId=c0-a8-56-16-0eef7803-bbfe-40ed-86c2-a1d1bdc6dc11]
	at com.ibm.fhir.server.util.FHIRRestHelper.doRead(FHIRRestHelper.java:766)
	at com.ibm.fhir.server.util.FHIRRestHelper.doRead(FHIRRestHelper.java:699)
	at com.ibm.fhir.server.resources.Read.read(Read.java:76)
	at com.ibm.fhir.server.resources.Read$Proxy$_$$_WeldClientProxy.read(Unknown Source)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #1920, opened 2021-02-08
Do not log full stack trace for a simple 404 not found #1844

prb112:

### DIFF
--- a/fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
@@ -226,7 +226,7 @@ public class FHIRResource {
                 log.log(Level.WARNING, msg, e);
             }
 
-            List<Bundle.Entry> toAdd = new ArrayList<Bundle.Entry>();
+            List<Bundle.Entry> toAdd = new ArrayList<>();
             // replace bundle entries that have an empty response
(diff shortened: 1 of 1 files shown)


fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -280,7 +280,7 @@ protected Response exceptionResponse(Exception e, Status status) {
     }
 
     protected Response exceptionResponse(OperationOutcome oo, Status status) {
-        if (log.isLoggable(Level.FINE)) {
+        if (log.isLoggable(Level.FINE) && !Status.NOT_FOUND.equals(status)) {
```
lmsurpre, line 283: I don't understand how preventing the FINE-level log here prevents the stack trace described at https://github.com/IBM/FHIR/issues/1844#issuecomment-775277681

prb112: That's why I assigned Robin - he didn't include any curl to replicate

I tried with the described resource that doesn't exist and it does not show up except in a trace.log (which is expectd\ed)

lmsurpre:
to me, logging a stack trace for a 404 isn't an issue if its at the FINE level (not INFO).  we reopened the issue for further clarification

Issue #1907, opened 2021-02-04
link urls are wrong when originalRequestUriHeader contains unencoded pipe


lmsurpre:
**Describe the bug**
We are fronting the FHIR server with a reverse proxy and setting an originalRequestUriHeader to the frontend URI that was requested to the gateway.

On some requests, the search bundle response links come back with links that match our front-end url as expected.  For example:
`https://frontend-host/Observation?_count=10&code=57698-3&_page=10`

But for other requests, its coming back with the backend server url which is not what we want:
`https://backend-host/fhir-server/api/v4/Observation?_count=10&code=http://loinc.org|57698-3&_page=10`

**To Reproduce**
1. Configure IBM FHIR Server with an originalRequestUriHeaderName
2. Make a request to the server and set this header to a value with an unescaped `|`
3. Note that the URL that comes back has the backend-url and not the desired front-end URL

**Expected behavior**
All generated links should have the frontend-url in them, not the backend-url

**Additional context**
Specifically, its failing on `URI requestUri = new URI(requestUriString);` in UriBuilder.toSearchSelfUri.
If the originalRequestUriHeader value was properly encoded, it would work as expected.  However, we should be more robust in our handling of cases where it is not.  Furthermore, we don't even use the query portion of the originalRequestUriHeader (just the hostname and path), and so we should be able to avoid most encoding issues just by stripping off the query portion.

d0roppe:
I tested a few scenarios where this X-FHIR-FORWARDED-URI header produces really odd behavior
[1907.xlsx](https://github.com/IBM/FHIR/files/5984306/1907.xlsx)
here is a file with the 4 odd behaviors all of these seem to be wrong.

lmsurpre:
I confirm that the problem in the description of this issue still exists.  I did address it in one place (as tested in our updated UriBuilderTest), but I missed the fact that we also try to parse the incoming header value within FHIRRestServletFilter.

As for the strange behavior on the fullUrl values of a search response bundle, I would consider that a separate (but related) issue.
The originalRequestUri header value must be set appropriate for the incoming request, otherwise you're going to get strange results like this.
However, we have pretty limited documentation around what an "appropriate" header value would be, and absolutely no documentation wrt what the behavior will be (nor what it should be) for inappropriate header values.

Here is the extent of the current documentation:
> This optional config parameter is provided for cases where the server is deployed behind a reverse proxy that overwrites the host and/or path portions of the original request.

[...]

PR #1949, opened 2021-02-16
issues #1907 and #1964 - strip the query from originalRequestUri before parsing and fix baseUrl for whole-system queries

lmsurpre:
1. Previously I updated our UriBuilder for this case (and introduced tests
to ensure it is properly functioning). However, I missed the fact that
FHIRRestServletFilter parses the originalRequestUri header value into a
URI and, when this fails, it simply reverts to the actual URL of the
request. Therefor, when the request contained a character like `|`, this
header value was never being used in downstream processing like
UriBuilder.buildSelfUri.toSearchSelfUri().

2. Set originalRequestUriHeaderName to "X-FHIR-FORWARDED-URL" in our
default fhir-server-config (so that this can get tested by our default
CI without flags)
3. Added OriginalRequestRewriteServerTest to ensure this feature is
working as expected
4. Expanded SearchAllTest to include a check for the fullUrl entries in the
response bundle
5. Fix FHIRRestHelper.getRequestBaseUri(type) in two cases:
  * for certain cases we were setting the baseUri to requestUri (losing our strip of the query string)
  * strip the `/_search` from the end of the whole-system search URL

6. Also performed some minor cleanup.

### DIFF
--- a/fhir-server-test/src/test/java/com/ibm/fhir/server/test/OriginalRequestRewriteServerTest.java
+++ b/fhir-server-test/src/test/java/com/ibm/fhir/server/test/OriginalRequestRewriteServerTest.java
@@ -1,0 +1,100 @@
+/*
+ * (C) Copyright IBM Corp. 2021
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.ibm.fhir.server.test;
+
+import static com.ibm.fhir.model.type.Code.code;
+import static com.ibm.fhir.model.type.Uri.uri;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertTrue;
+
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.client.WebTarget;
+import javax.ws.rs.core.Response;
+
+import org.testng.annotations.Test;
+
+import com.ibm.fhir.core.FHIRMediaType;
+import com.ibm.fhir.model.resource.Bundle;
+import com.ibm.fhir.model.resource.Patient;
+import com.ibm.fhir.model.test.TestUtil;
+import com.ibm.fhir.model.type.Coding;
+import com.ibm.fhir.model.type.Meta;
+
+/**
+ * Basic sniff test of the FHIR Server.
+ */
+public class OriginalRequestRewriteServerTest extends FHIRServerTestBase {
+    private static final String TAG = "OriginalRequestRewriteServerTest";
+    private static final String ORIGINAL_URI = "https://frontend-url/base";
+
+    /**
+     * Create a Patient and ensure the Location header reflects the OriginalRequestUri
+     */
+    @Test
+    public void testCreatePatient_LocationRewrite() throws Exception {
+        WebTarget target = getWebTarget();
+
+        // Build a new Patient and then call the 'create' API.
+        Patient patient = TestUtil.readLocalResource("Patient_JohnDoe.json");
+        Meta.Builder meta = patient.getMeta() == null ? Meta.builder() : patient.getMeta().toBuilder();
+        meta.tag(Coding.builder().system(uri("http://ibm.com/fhir/tag")).code(code(TAG)).build());
+        patient = patient.toBuilder()
+                .meta(meta.build())
+                .build();
+        Entity<Patient> entity = Entity.entity(patient, FHIRMediaType.APPLICATION_FHIR_JSON);
+        Response response = target.path("Patient").request()
+                .header("X-FHIR-Forwarded-URL", ORIGINAL_URI + "/Patient")
+                .post(entity, Response.class);
+        assertResponse(response, Response.Status.CREATED.getStatusCode());
+        assertTrue(response.getLocation().toString().startsWith(ORIGINAL_URI),
+                response.getLocation().toString() + " should begin with " + ORIGINAL_URI);
+    }
+
+    /**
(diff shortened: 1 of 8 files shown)


michaelwschroeder (approved):
LGTM

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
```diff
@@ -316,41 +316,51 @@ private String getRequestURL(HttpServletRequest request) {
     /**
      * Get the original request URL from either the HttpServletRequest or a configured Header (in case of re-writing proxies).
      *
+     * If configured as a header, the header value will only be used for the protocol, server name, port number, and server path;
```
prb112, line 319: Copyright year

prb112 (approved):
LGTM - one file needs a copyright year update

fhir-server/liberty-config/config/default/fhir-server-config.json
```diff
@@ -4,6 +4,7 @@
         "core": {
             "tenantIdHeaderName": "X-FHIR-TENANT-ID",
             "datastoreIdHeaderName": "X-FHIR-DSID",
+            "originalRequestUriHeaderName": "X-FHIR-FORWARDED-URL",
```
prb112, line 7: one additional thought here... we prefix with X-FHIR, it's generally discouraged. I do like the consistency with the other headers, and wanted to just leave this comment for posterity/ future thought.

https://tools.ietf.org/html/rfc6648
https://stackoverflow.com/a/3561399/1873438

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -2841,14 +2841,16 @@ private String getRequestBaseUri(String type) throws Exception {
...
+        }
+
+        // Strip any path segments for whole-system interactions (in case of whole-system search, "Resource" is passed as the type)
+        if (type == null || type.isEmpty() || "Resource".equals(type)) {
+            if (baseUri.endsWith("/_search")) {
                 baseUri = baseUri.substring(0, baseUri.length() - "/_search".length());
+            } else if (baseUri.endsWith("/_history")) {
+                baseUri = baseUri.substring(0, baseUri.length() - "/_history".length());
             } else if (baseUri.contains("/$")) {
                 baseUri = baseUri.substring(0, baseUri.lastIndexOf("/$"));
```
prb112, line 2855: Thiss one gets called when processing a bundle.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/OriginalRequestRewriteServerTest.java
```diff
@@ -0,0 +1,100 @@
+/*
+ * (C) Copyright IBM Corp. 2017, 2020
+ *
```
prb112, line 3: 2021

prb112 (approved):
LGTM (minus CI failures)

prb112:
```
Error:  Failures: 
Error:    OriginalRequestRewriteServerTest.testCreatePatient_LocationRewrite:53 https://localhost:9443/fhir-server/api/v4/Patient/177b25f6c63-fdb38467-b6c6-4950-851a-098907ae3d8d/_history/1 should begin with https://frontend-url/base expected [true] but found [false]
```

fhir-server/liberty-config/config/default/fhir-server-config.json
lmsurpre: Interesting.  I tend to agree with the first commenter on that answer:
> Just as there are many kids that will never end up as professional athletes, many custom headers will never end up as standards. I'm inclined to keep the "X-" on those.

Especially given that we already have the two others with the "X-FHIR" prefix.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
lmsurpre: Not sure if you were suggesting that something should change or not; I mostly just repositioned this and made it run when the type is "Resource".  Without this change, I found that the fullUrl in our search response bundles contained a trailing `_search` (when the search was invoked via `POST [base]/_search`).

fhir-server-test/src/test/java/com/ibm/fhir/server/test/OriginalRequestRewriteServerTest.java
lmsurpre: fixed

fhir-server/liberty-config/config/default/fhir-server-config.json
prb112: I like the consistency, but wanted to bring it up.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
prb112: Got it.

michaelwschroeder (approved):
LGTM

d0roppe:
LGTM - when passing in the full URI

d0roppe (approved):
LGTM - when passing in the full URI

Issue #1553, opened 2020-10-02
Relative References with the Same Logical Identifier should throw an error


prb112:
**Describe the bug**
Relative References with the Same Logical Identifier should throw an error on Search

For searches on subject=1 that use a SearchParameter like the following: 

```
    {
      "fullUrl": "http://hl7.org/fhir/SearchParameter/Account-subject",
      "resource": {
        "resourceType": "SearchParameter",
        "id": "Account-subject",
          "description": "The entity that caused the expenses",
        "code": "subject",
        "base": [
          "Account"
        ],
        "type": "reference",
        "expression": "Account.subject",
        "target": [
          "Practitioner",
          "Organization",
          "Device",
          "Patient",
          "HealthcareService",
          "PractitionerRole",
          "Location"
        ]
      }
    }
```
*abbreviated* to highlight the problem

The target is extracted and matrixed with the input identifier to generate a logical reference, such as: 
- `Practitioner/1`
- `Organization/1`
- et cetra

In the specific case where there are TWO relative references with the same Logical Identifier should throw an error, as stated in https://www.hl7.org/fhir/search.html#reference

```
Some references may point to more than one type of resource; e.g. subject: Reference(Patient|Group|Device|..). In these cases, multiple resources may have the same logical identifier. Servers SHOULD reject a search where the logical id refers to more than one matching resource across different types.
```

**To Reproduce**
1. Create a Practitioner Resource with ID 1
2. Create a Organization Resource with ID 1
3. Search with `subject=1`

**Expected behavior**
Throw an error in the case of the reference ID being duplicate.

**Additional context**
n/a

michaelwschroeder:
After discussion with the team, the following was agreed upon with respect to references that only contain a logical ID:

1. A logical ID-only reference is not valid to be stored in a resource’s reference field, but since we currently allow it, we need to make it a configuration option as to whether or not an error is generated at create/update time in order not to break existing resources.
2. If configured to accept logical ID-only, we should index as an absolute reference at create/update time.
3. Logical ID-only references should be treated as absolute references, but to resources that don't exist on the server.
4. Logical ID-only values are valid as reference search parameter values, and should match on any resources of valid target type with that logical ID. If resources with multiple target types are returned, we will fail the search (per this issue).
5. Logical ID-only references used in chain or include or compartment type searches should not return any matches (based on #3 above).

Given this set of rules, the search scenarios identified for this issue will be handled as follows:

1. Standard search where reference parameter value is a logical ID
[...]

PR #1970, opened 2021-02-18
Issue #1553 - check multiple resource types on logical ID ref search

michaelwschroeder:

### DIFF
--- a/fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractCompartmentTest.java
+++ b/fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractCompartmentTest.java
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright IBM Corp. 2016,2019
+ * (C) Copyright IBM Corp. 2016,2021
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -39,54 +39,68 @@ public abstract class AbstractCompartmentTest extends AbstractPersistenceTest {
     Practitioner savedPractitioner;
     RelatedPerson savedRelatedPerson;
     Observation savedObservation;
-    
+    Observation savedObservation2;
+
     /**
      * Builds and saves an Observation with the following references:
-     * 
+     *
      * Observation.subject = Patient
      * Observation.device = Device
      * Observation.encounter = Encounter
      * Observation.performer[0] = Patient
      * Observation.performer[1] = Practitioner
      * Observation.performer[2] = RelatedPerson
+     *
+     * Builds and saves an Observation with the following logical ID-only reference:
+     *
+     * Observation.subject = Patient
      */
     @BeforeClass
     public void createResources() throws Exception {
         Observation.Builder observationBuilder = ((Observation) TestUtil.readExampleResource("json/ibm/minimal/Observation-1.json")).toBuilder();
-        
+        Observation.Builder observation2Builder = ((Observation) TestUtil.readExampleResource("json/ibm/minimal/Observation-1.json")).toBuilder();
+
         Patient patient = TestUtil.readExampleResource("json/ibm/minimal/Patient-1.json");
         savedPatient = persistence.create(getDefaultPersistenceContext(), patient).getResource();
         observationBuilder.subject(buildReference(savedPatient));
         observationBuilder.performer(buildReference(savedPatient));
-        
+        observation2Builder.subject(Reference.builder().reference(string(savedPatient.getId())).build());
+
         Device device = TestUtil.readExampleResource("json/ibm/minimal/Device-1.json");
         savedDevice = persistence.create(getDefaultPersistenceContext(), device).getResource();
         observationBuilder.device(buildReference(savedDevice));
-        
+
         Encounter encounter = TestUtil.readExampleResource("json/ibm/minimal/Encounter-1.json");
         savedEncounter = persistence.create(getDefaultPersistenceContext(), encounter).getResource();
         observationBuilder.encounter(buildReference(savedEncounter));
-        
+
         Practitioner practitioner = TestUtil.readExampleResource("json/ibm/minimal/Practitioner-1.json");
         savedPractitioner = persistence.create(getDefaultPersistenceContext(), practitioner).getResource();
         observationBuilder.performer(buildReference(savedPractitioner));
-        
+
         RelatedPerson relatedPerson = TestUtil.readExampleResource("json/ibm/minimal/RelatedPerson-1.json");
         savedRelatedPerson = persistence.create(getDefaultPersistenceContext(), relatedPerson).getResource();
         observationBuilder.performer(buildReference(savedRelatedPerson));
-        
+
         savedObservation = persistence.create(getDefaultPersistenceContext(), observationBuilder.build()).getResource();
         assertNotNull(savedObservation);
         assertNotNull(savedObservation.getId());
         assertNotNull(savedObservation.getMeta());
         assertNotNull(savedObservation.getMeta().getVersionId().getValue());
         assertEquals("1", savedObservation.getMeta().getVersionId().getValue());
+
+        savedObservation2 = persistence.create(getDefaultPersistenceContext(), observation2Builder.build()).getResource();
+        assertNotNull(savedObservation2);
+        assertNotNull(savedObservation2.getId());
+        assertNotNull(savedObservation2.getMeta());
+        assertNotNull(savedObservation2.getMeta().getVersionId().getValue());
+        assertEquals("1", savedObservation2.getMeta().getVersionId().getValue());
     }
 
     private Reference buildReference(Resource resource) {
         assertNotNull(resource);
         assertNotNull(resource.getId());
-        
+
         String resourceTypeName = FHIRUtil.getResourceTypeName(resource);
         return Reference.builder()
                         .reference(string(resourceTypeName + "/" + resource.getId()))
@@ -95,38 +109,45 @@ public abstract class AbstractCompartmentTest extends AbstractPersistenceTest {
 
     @Test
     public void testPatientCompartment() throws Exception {
-        List<Resource> results = runQueryTest("Patient", savedPatient.getId(), 
+        List<Resource> results = runQueryTest("Patient", savedPatient.getId(),
                                     Observation.class, "_id", savedObservation.getId());
         // This currently returns 2 due to https://github.com/IBM/FHIR/issues/265
 //        assertEquals(1, results.size());
         assertTrue(results.size() > 0);
     }
-    
+
     @Test
     public void testDeviceCompartment() throws Exception {
-        List<Resource> results = runQueryTest("Device", savedDevice.getId(), 
+        List<Resource> results = runQueryTest("Device", savedDevice.getId(),
                                     Observation.class, "_id", savedObservation.getId());
         assertEquals(1, results.size());
     }
-    
+
     @Test
     public void testEncounterCompartment() throws Exception {
-        List<Resource> results = runQueryTest("Encounter", savedEncounter.getId(), 
+        List<Resource> results = runQueryTest("Encounter", savedEncounter.getId(),
                                     Observation.class, "_id", savedObservation.getId());
         assertEquals(1, results.size());
     }
-    
+
     @Test
     public void testPractitionerCompartment() throws Exception {
-        List<Resource> results = runQueryTest("Practitioner", savedPractitioner.getId(), 
+        List<Resource> results = runQueryTest("Practitioner", savedPractitioner.getId(),
                                     Observation.class, "_id", savedObservation.getId());
         assertEquals(1, results.size());
     }
-    
+
     @Test
     public void testRelatedPersonCompartment() throws Exception {
-        List<Resource> results = runQueryTest("RelatedPerson", savedRelatedPerson.getId(), 
+        List<Resource> results = runQueryTest("RelatedPerson", savedRelatedPerson.getId(),
                                     Observation.class, "_id", savedObservation.getId());
         assertEquals(1, results.size());
+    }
+
+    @Test
+    public void testPatientCompartmentViaLogicalId() throws Exception {
+        List<Resource> results = runQueryTest("Patient", savedPatient.getId(),
+                                    Observation.class, "_id", savedObservation2.getId());
+        assertEquals(0, results.size());
     }
 }
--- a/fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractCompartmentTest.java
+++ b/fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractCompartmentTest.java
@@ -7,7 +7,6 @@
 package com.ibm.fhir.persistence.test.common;
 
 import static com.ibm.fhir.model.type.String.string;
-import static org.testng.Assert.assertTrue;
 import static org.testng.AssertJUnit.assertEquals;
 import static org.testng.AssertJUnit.assertNotNull;
 
@@ -111,9 +110,7 @@ public abstract class AbstractCompartmentTest extends AbstractPersistenceTest {
     public void testPatientCompartment() throws Exception {
         List<Resource> results = runQueryTest("Patient", savedPatient.getId(),
                                     Observation.class, "_id", savedObservation.getId());
-        // This currently returns 2 due to https://github.com/IBM/FHIR/issues/265
-//        assertEquals(1, results.size());
-        assertTrue(results.size() > 0);
+        assertEquals(1, results.size());
     }
 
     @Test
--- a/fhir-search/src/main/java/com/ibm/fhir/search/util/ReferenceUtil.java
+++ b/fhir-search/src/main/java/com/ibm/fhir/search/util/ReferenceUtil.java
@@ -141,7 +141,6 @@ public class ReferenceUtil {
                         version = Integer.parseInt(tokens[3]);
                     }
                 }
-                // TODO: should the reference type be used if not null and we only have a logical ID?
             } else if (value != null && value.startsWith(HTTP) || value.startsWith(HTTPS) || value.startsWith(URN)) {
                 // - Absolute reference. We only know the type if it is given by the type field
                 referenceType = ReferenceType.LITERAL_ABSOLUTE;
@@ -161,7 +160,6 @@ public class ReferenceUtil {
                         version = Integer.parseInt(tokens[3]);
                     }
                 }
-                // TODO: should the reference type be used if not null and we only have a logical ID?
             }
         }
 
--- a/fhir-search/src/main/java/com/ibm/fhir/search/util/ReferenceUtil.java
+++ b/fhir-search/src/main/java/com/ibm/fhir/search/util/ReferenceUtil.java
@@ -62,56 +62,7 @@ public class ReferenceUtil {
         Integer version = null;
 
         if (ref.getReference() != null && ref.getReference().getValue() != null) {
-            // LITERAL REFERENCE
-            // * an absolute URL
-            // * relative URL, which is relative to the Service Base URL, or, if processing a resource
-            //   from a bundle, which is relative to the base URL implied by the Bundle.entry.fullUrl
-            // Note that fragment (internal) references are not relevant here, because bundle
-            // processing will already have resolved them, replacing them with the relative values
-            value = ref.getReference().getValue();
-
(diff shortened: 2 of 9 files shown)


michaelwschroeder:
Please take a look at issue #1553 for the new scope of this work after discussions about how we should handle logical ID-only references as references in resources and as search parameter values.

fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractCompartmentTest.java
```diff
@@ -95,38 +109,45 @@ private Reference buildReference(Resource resource) {
 
     @Test
     public void testPatientCompartment() throws Exception {
-        List<Resource> results = runQueryTest("Patient", savedPatient.getId(), 
+        List<Resource> results = runQueryTest("Patient", savedPatient.getId(),
                                     Observation.class, "_id", savedObservation.getId());
         // This currently returns 2 due to https://github.com/IBM/FHIR/issues/265
 //        assertEquals(1, results.size());
```
lmsurpre, line 115: not really related to you changes, but I see 265 is closed now.  maybe remove the comment and uncomment the assert?
```suggestion
        assertEquals(1, results.size());
```

fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractCompartmentTest.java
```diff
@@ -95,38 +109,45 @@ private Reference buildReference(Resource resource) {
...
         assertTrue(results.size() > 0);
```
lmsurpre, line 116: not really related to you changes, but I see 265 is closed now. maybe remove the comment and uncomment the assert?
```suggestion
        assertEquals(1, results.size());
```

fhir-search/src/main/java/com/ibm/fhir/search/util/ReferenceUtil.java
```diff
@@ -62,18 +62,68 @@ public static ReferenceValue createReferenceValueFrom(Reference ref, String base
...
+     * <li>LITERAL_ABSOLUTE: the full URI of the reference</li>
+     * <li>INVALID: null</li>
+     * </ol>
+     *
+     * @param refValue a reference value
+     * @param refType a reference resource type (used for LITERAL_ABSOLUTE only)
+     * @param baseUrl the base URL used to determine whether to convert absolute references to relative references
+     * @return a structured representation of the reference value that varies by its inferred reference type
+     */
+    public static ReferenceValue createReferenceValueFrom(String refValue, String refType, String baseUrl) {
```
lmsurpre, line 108: it not super-clear to me when you would call this variant vs the other createReferenceValueFrom from reading the javadoc

fhir-search/src/main/java/com/ibm/fhir/search/util/ReferenceUtil.java
```diff
@@ -91,13 +141,13 @@ public static ReferenceValue createReferenceValueFrom(Reference ref, String base
                         version = Integer.parseInt(tokens[3]);
                     }
                 }
+                // TODO: should the reference type be used if not null and we only have a logical ID?
```
lmsurpre, line 144: would be nice to resolve these TODOs if we can.  i'm not sure what the answer is but I think we have a better shot of getting it right now than in the future (since its all in our heads)

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -2591,51 +2602,115 @@ private Bundle createSearchBundle(List<Resource> resources, FHIRSearchContext se
...
+     *
+     * @param resourceType
+     *            The search resource type.
+     * @param queryParameters
+     *            The list of reference query parameters that only specified a logical ID.
+     * @param resources
+     *            The list of resources to check for multiple resource types.
+     * @throws Exception
+     */
+    private void checkForMatchLogicalIdDuplicates(String resourceType, List<String> queryParameters, List<Resource> resources) throws Exception {
```
lmsurpre, line 2669: ugh, I really wish this were simpler; I'm just not sure how to do it without just saying "screw it" :-/

lmsurpre:
Does this pull request implement this part of the decision for #1553?
```A logical ID-only reference is not valid to be stored in a resource’s reference field, but since we currently allow it, we need to make it a configuration option as to whether or not an error is generated at create/update time in order not to break existing resources.```
I didn't see any config parameter, but maybe I just missed it?

**Comment by michaelwschroeder** [comment]
Please take a look at issue #1553 for the new scope of this work after discussions about how we should handle logical ID-only references as references in resources and as search parameter values.

> 
> 
> Does this pull request implement this part of the decision for #1553?
> 
> > A logical ID-only reference is not valid to be stored in a resource’s reference field, but since we currently allow it, we need to make it a configuration option as to whether or not an error is generated at create/update time in order not to break existing resources.
> 
> I didn't see any new config parameter, but maybe I just missed it?

No, this issue does not address the config parameter. Issue #1961 has been opened to address it.

File: fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java (hunk 5 above)
Comment (michaelwschroeder) [reply fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java:2669]: Yeah, I just don't think there's an easy way to do it. Good thing is that it's only going to get executed for searches in which there's a logical ID-only reference value specified.

File: fhir-search/src/main/java/com/ibm/fhir/search/util/ReferenceUtil.java (hunk 3 above)
Comment (michaelwschroeder) [reply fhir-search/src/main/java/com/ibm/fhir/search/util/ReferenceUtil.java:108]: Updated the javadoc.

File: fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractCompartmentTest.java (hunk 2 above)
Comment (michaelwschroeder) [reply fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractCompartmentTest.java]: Done.

File: fhir-search/src/main/java/com/ibm/fhir/search/util/ReferenceUtil.java (hunk 4 above)
Comment (michaelwschroeder) [reply fhir-search/src/main/java/com/ibm/fhir/search/util/ReferenceUtil.java:144]: I think it would make sense to use the type in the case where we only have a logical ID, but looking at the spec at  https://www.hl7.org/fhir/references.html#type, I see this line:

  `When the type is provided directly, it SHALL agree with the type determined by resolving the resource.`

Given that we don't attempt to resolve the reference at ingestion time, maybe this isn't a concern.

Comment (lmsurpre) [reply fhir-search/src/main/java/com/ibm/fhir/search/util/ReferenceUtil.java:144]: I think we should handle it similar to how we handle absolute references.  Judging from the new lines just below, I guess we do use it.

But I remember @punktilious sharing his concern about potentially using these for compartment and/or chaining/includes. Hmmm...

How do we ensure that we don't attempt those for absolute references?  Maybe we should add a new value to the ReferenceType enum for this class of reference.  I *wouldn't* call it "Logical Id-only" but instead something like "non-FHIR relative" because the value wouldn't necessarilly need to match the resource's id...its basically just any relative reference that doesn't match the pattern for FHIR relative references.

File: fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractCompartmentTest.java [hunk 6]
```diff
@@ -39,54 +38,68 @@
...
         Observation.Builder observationBuilder = ((Observation) TestUtil.readExampleResource("json/ibm/minimal/Observation-1.json")).toBuilder();
-        
+        Observation.Builder observation2Builder = ((Observation) TestUtil.readExampleResource("json/ibm/minimal/Observation-1.json")).toBuilder();
+
         Patient patient = TestUtil.readExampleResource("json/ibm/minimal/Patient-1.json");
         savedPatient = persistence.create(getDefaultPersistenceContext(), patient).getResource();
         observationBuilder.subject(buildReference(savedPatient));
         observationBuilder.performer(buildReference(savedPatient));
-        
+        observation2Builder.subject(Reference.builder().reference(string(savedPatient.getId())).build());
```
prb112, line 66: Does the prior line buildReference accomplish the same as the code on 66? might be best to reuse if it does the same thing.

prb112: we seem to bounce between static import and this method, maybe something to reconcile next time we're in the code. 

Treat this as a thought for the future.

fhir-search/src/main/java/com/ibm/fhir/search/util/ReferenceUtil.java
```diff
@@ -62,18 +62,68 @@ public static ReferenceValue createReferenceValueFrom(Reference ref, String base
         Integer version = null;
 
         if (ref.getReference() != null && ref.getReference().getValue() != null) {
+            return createReferenceValueFrom(ref.getReference().getValue(), ref.getType() != null ? ref.getType().getValue() : null, baseUrl);
+        } else if (ref.getIdentifier() != null && ref.getIdentifier().getValue() != null) {
+            // LOGICAL REFERENCE
+            value = ref.getIdentifier().getValue().getValue();
```
prb112, line 68: In very very rare circumstances the .getValue can be null. 
Should there be a check on 66 to see it's null 

Example of null. 

```
"identifier": [
             ...
            "_value": {
                "extension": [
                    {
                        "url": "http://hl7.org/fhir/StructureDefinition/data-absent-reason",
                        "valueCode": "unknown"
                    }
                ]
            }...
        }
    ]
```

You may want to double check me on the extension part.

fhir-search/src/main/java/com/ibm/fhir/search/util/ReferenceUtil.java
```diff
@@ -62,18 +62,68 @@ public static ReferenceValue createReferenceValueFrom(Reference ref, String base
...
+            referenceType = ReferenceType.LOGICAL;
+            if (ref.getType() != null) {
+                // We need
+                targetResourceType = ref.getType().getValue();
```
prb112, line 72: What happens when the type is null? similar to prior comment?

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -1591,39 +1594,41 @@ private static QueryParameter parseChainedParameter(HashSet<String> resourceType
                 }
 
                 HashSet<String> modifierResourceTypeNameForResourceTypes = new HashSet<>();
```
prb112, line 1596: Any particular reason for HashSet versus Set declaration?

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -1707,8 +1727,12 @@ private static QueryParameter parseChainedParameter(Class<?> resourceType, Strin
                 }
 
                 searchParameter = getSearchParameter(resourceType, parameterName);
-                type = Type.fromValue(searchParameter.getType().getValue());
+                if (searchParameter == null) {
+                    throw SearchExceptionUtil.buildNewInvalidSearchException(
+                        String.format(SEARCH_PARAMETER_NOT_FOUND, parameterName, resourceType.getSimpleName()));
+                }
```
prb112, line 1733: maybe a decent fit for a small helper method for the 3 places its called?

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -1755,6 +1783,19 @@ private static QueryParameter parseChainedParameter(Class<?> resourceType, Strin
...
+            if (checkForLogicalId) {
+                // For last search parameter, if type REFERENCE and not scoped to single target resource type, check if
+                // value is logical ID only.
+                for (QueryParameterValue value : valueList) {
+                    ReferenceValue refVal = ReferenceUtil.createReferenceValueFrom(value.getValueString(), null, ReferenceUtil.getBaseUrl(null));
+                    if (refVal.getType() == ReferenceType.LITERAL_RELATIVE && refVal.getTargetResourceType() == null) {
+                        throw SearchExceptionUtil.buildNewInvalidSearchException(
+                            String.format(LOGICAL_ID_VALUE_NOT_ALLOWED_FOR_REFERENCE_SEARCH, rootParameter.getChain().getLast().getCode(), value.getValueString()));
+                    }
+                }
```
prb112, line 1796: You may want to refactor this into a new method. It's very similar to the prior method.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -1900,6 +1941,16 @@ private static QueryParameter parseReverseChainedParameter(Class<?> resourceType
...
                             modifier, modifierResourceTypeName, valuesString);
 
+                        if (Type.REFERENCE == type && searchParameter.getTarget().size() > 1 && modifierResourceTypeName == null) {
+                            for (QueryParameterValue value : valueList) {
+                                ReferenceValue refVal = ReferenceUtil.createReferenceValueFrom(value.getValueString(), null, ReferenceUtil.getBaseUrl(null));
+                                if (refVal.getType() == ReferenceType.LITERAL_RELATIVE && refVal.getTargetResourceType() == null) {
+                                    throw SearchExceptionUtil.buildNewInvalidSearchException(
+                                        String.format(LOGICAL_ID_VALUE_NOT_ALLOWED_FOR_REFERENCE_SEARCH, parameterName, value.getValueString()));
+                                }
+                            }
```
prb112, line 1951: Similar to 1790-1796

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchReferenceTest.java
```diff
@@ -33,18 +34,29 @@
...
-    public String createPatientWithReference(String reference) throws Exception {
+    public String createPatientWithReference(String field, String reference) throws Exception {
         WebTarget target = getWebTarget();
 
         // Build a new Patient and then call the 'create' API.
         Patient patient = TestUtil.readLocalResource("Patient_JohnDoe.json");
 
-        patient =
-                patient.toBuilder().gender(AdministrativeGender.MALE).managingOrganization(Reference.builder().display(com.ibm.fhir.model.type.String.of("Test Organization")).reference(com.ibm.fhir.model.type.String.of(reference)).build()).build();
+        if (field.equals("organization")) {
```
prb112, line 51: a cleaner way to write this is 

"organization".equals(field)

Refer to https://stackoverflow.com/a/10911432/1873438

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchReferenceTest.java
```diff
@@ -69,18 +81,25 @@ public String createPatientWithReference(String reference) throws Exception {
     public void testCreatePatient() throws Exception {
         // references without an explicit type will be interpreted as literals (e.g. externals) not local references
         // we will not infer their type, which means that searches specifying a type will not match them by design
-        patientResourceIdWithReferenceId = createPatientWithReference("3001");
+        patientResourceIdWithReferenceId = createPatientWithReference("organization", "3001");
```
prb112, line 84: For consistency, we may want to mark this in the proper case for a FHIR Resource 'Organization'

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
prb112: It feels like a lot of processing over resources multiple times. 

Could we add a flag to turn this feature off or on?  specifically on line 2670, 2616
it feels like something we may want to have as lenient or strict.

fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractCompartmentTest.java
michaelwschroeder: The buildReference() method is not the same - it builds a valid relative reference containing both resource type and logical ID. On line 66, I want just the logical ID.

fhir-search/src/main/java/com/ibm/fhir/search/util/ReferenceUtil.java
michaelwschroeder: In the case of an identifier, it is not required that the type is specified. The identifier is valid on it's own.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchReferenceTest.java
michaelwschroeder: This parameter actually refers to a search parameter code, so I think case is appropriate.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: I'll make a couple points and then we can decide how to proceed:
1. I've combined checkForMatchLogicalIdDuplicates() and checkForVersionedReference() into one method: performSearchReferenceChecks(), so now we'll only loop through 'match' resources, and only a maximum of one time
2. We'll only do this processing on a chain search or on a search of a reference field where the search parameter value is a logical ID-only, the former being uncommon and the latter being rare.

Given the above, I think a flag would probably be overkill, particularly given the fact that we're doing this processing to be compliant with the FHIR spec.

fhir-search/src/main/java/com/ibm/fhir/search/util/ReferenceUtil.java
michaelwschroeder: added the null check

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
michaelwschroeder: changed to Set

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
michaelwschroeder: created helper method

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
michaelwschroeder: Refactored into new method.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
michaelwschroeder: Refactored into new method.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchReferenceTest.java
michaelwschroeder: Done.

fhir-search/src/main/java/com/ibm/fhir/search/util/ReferenceUtil.java
michaelwschroeder: The existing code (prior to this change) does use the type for absolute references. I see no issue in treating logical ID-only references the same way. Basically it would then make the following two References equivalent:
```
 "subject": {
   "reference" : "Patient/123"
 }
```

```
 "subject": {
   "reference" : "123",
   "type" : "Patient"
 }
```

A logical ID-only reference without a type value would continue to be treated as it is today. In that case there's no concern with compartment/chain/include type searches, because the SQL generated for those searches always looks for a code_system_id of a specific resource type, so the logical ID-only reference would never match.

lmsurpre: yeah, my concern was that I don't know that we want a "logical id-only reference + a type" to work the same as an actual valid relative reference.

prb112 (approved):

fhir-search/src/main/java/com/ibm/fhir/search/util/ReferenceUtil.java
lmsurpre: team decision:  we don't want to handle these as the same;  in future our default will be to reject the "reference": "123" anyway.

michaelwschroeder: TODOs removed.

Issue #1971, opened 2021-02-18
schema name case normalization is not working as expected


lmsurpre:
**Describe the bug**
In the fix for https://github.com/IBM/FHIR/issues/1783 we started normalizing schema names to upper case.

Unfortunately, for users that have lowercase characters in their schema name, we are not properly updating the FHIR_ADMIN.VERSION_HISTORY table.

Specifically, instead of updating the existing objects in the table, we seem to be creating a new entry in the table (this one with an upper-case schema name).

**To Reproduce**
1. Create and deploy our schema using a version of fhir-persistence-schema prior to https://github.com/IBM/FHIR/pull/1782 (which I believe was released in 4.5.1) and pass a lower-case schema name (e.g. "fhirdata")
2. Run the `--update-schema` action using the latest fhir-persistence-schema project

`select * from fhir_admin.version_history` and note that *_Resource_Token_Refs tables now have two entries:
* one version for schemaname "fhirdata"
* version 8 for schemaname "FHIRDATA"

Also note that if the `--update-schema` is executed a second time, it now fails with errors like:
```
WARNING ls.pool.PoolConnectionProvider Rolling back transaction
2021-02-18 15:37:13.495 wh-cmsiop-cthon-fhir-update-schema-5rhqr fhir-schematool SEVERE 00000017  SEVERE .fhir.task.core.impl.TaskGroup GROUP:FHIRDATA.COVERAGEELIGIBILITYRESPONSE_RESOURCE_TABLE_GROUP:0
2021-02-18 15:37:13.495 wh-cmsiop-cthon-fhir-update-schema-5rhqr fhir-schematool ERROR 
org.postgresql.util.PSQLException: ERROR: index "idx_coverageeligibilityresponse_resource_token_refs_tvlr" does not exist
com.ibm.fhir.database.utils.api.UndefinedNameException: org.postgresql.util.PSQLException: ERROR: index "idx_coverageeligibilityresponse_resource_token_refs_tvlr" does not exist
	at com.ibm.fhir.database.utils.postgres.PostgresTranslator.translate(PostgresTranslator.java:100)
	at com.ibm.fhir.database.utils.common.DropIndex.run(DropIndex.java:46)
	at com.ibm.fhir.database.utils.common.CommonDatabaseAdapter.runStatement(CommonDatabaseAdapter.java:459)
	at com.ibm.fhir.database.utils.postgres.PostgresAdapter.runStatement(PostgresAdapter.java:266)
[...]

PR #1972, opened 2021-02-18
issue #1971 - handle duplicate entries in version_history with grace

lmsurpre:
This should workaround the root cause of issue #1971 and prevent
`--update-schema` from trying to apply updates that have already been
applied.

The root cause will need to be addressed in a separate PR.

### DIFF
--- a/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/GetLatestVersionDAO.java
+++ b/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/GetLatestVersionDAO.java
@@ -70,7 +70,7 @@ public class GetLatestVersionDAO implements IDatabaseSupplier<Map<String,Integer
                     if (currentValue != null) {
                         // version can't be null due to NOT NULL db constraint
                         newValue = Integer.max(currentValue, version);
-                        logger.warning("Version history entry " + schemaTypeName + " exists with multiple values [" + currentValue
+                        logger.fine("Version history entry " + schemaTypeName + " exists with multiple values [" + currentValue
                                 + ", " + version + "]; using " + newValue + ". Check schema name casing.");
                     }
                     return newValue;
--- a/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/GetLatestVersionDAO.java
+++ b/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/GetLatestVersionDAO.java
@@ -71,7 +71,7 @@ public class GetLatestVersionDAO implements IDatabaseSupplier<Map<String,Integer
                         // version can't be null due to NOT NULL db constraint
                         newValue = Integer.max(currentValue, version);
                         logger.warning("Version history entry " + schemaTypeName + " exists with multiple values [" + currentValue
-                                + ", " + version + "]; using " + newValue);
+                                + ", " + version + "]; using " + newValue + ". Check schema name casing.");
                     }
                     return newValue;
                 });
--- a/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/GetLatestVersionDAO.java
+++ b/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/GetLatestVersionDAO.java
@@ -65,7 +65,16 @@ public class GetLatestVersionDAO implements IDatabaseSupplier<Map<String,Integer
                 int version = rs.getInt(4);
 
                 String schemaTypeName = schema + ":" + type + ":" + name;
-                result.put(schemaTypeName, version);
+                result.compute(schemaTypeName, (key, currentValue) -> {
+                    Integer newValue = version;
+                    if (currentValue != null) {
+                        // version can't be null due to NOT NULL db constraint
+                        newValue = Integer.max(currentValue, version);
+                        logger.warning("Version history entry " + schemaTypeName + " exists with multiple values [" + currentValue
+                                + ", " + version + "]; using " + newValue);
+                    }
+                    return newValue;
+                });
             }
         }
         catch (SQLException x) {


fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/GetLatestVersionDAO.java
```diff
@@ -65,7 +65,16 @@ public GetLatestVersionDAO(String adminSchemaName, String schemaName) {
                 int version = rs.getInt(4);
 
                 String schemaTypeName = schema + ":" + type + ":" + name;
-                result.put(schemaTypeName, version);
+                result.compute(schemaTypeName, (key, currentValue) -> {
```
prb112, line 68: basically you are reconciling conflicts in the Version_HISTORY? 

Wouldn't it be better to reconcile in the sql? What if this table is migrated and the restored insert order is different?

prb112 (approved):
LGTM (discussed on team call, Lee to add comment)

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/GetLatestVersionDAO.java
lmsurpre: what its doing:  if we already have an entry for this `shema:type:name` in our map (`currentValue != null`), then that means that we have two entries that only vary by case (which is the root of the issue).

so, in that case, we log a warning and then take the higher of the two versions.

for example, consider there are two entries in the table:
|schema|type|name|version|
|--|--|--|--|
my_schema | TABLE | MY_TABLE | 6
MY_SCHEMA | TABLE | MY_TABLE | 8

because we UCASE them above, then both become `MY_SCHEMA:TABLE:MY_TABLE`.
previously we were always overwriting the map entry which is non-deterministic.  when it came back 6, we were trying to reapply the 6->8 migration which failed.
with this change, we will always use the higher value, which will prevent us from trying the same migration over and over, even when the table contains 2 entries for a single schema name (due to the casing issue described in #1971)

punktilious: This should be done in the DAO query. The query should select upper(schema_name). This also needs to be added to the group by: 
```        final String cols = DataDefinitionUtil.join("UPPER(" + SchemaConstants.SCHEMA_NAME + ")", SchemaConstants.OBJECT_TYPE, SchemaConstants.OBJECT_NAME);
```

punktilious: I would also UPPER(object_name) too, just to be on the safe-side, because: 
```
                String name = rs.getString(3).toUpperCase();
```

punktilious: Although having the warning message is useful, which you wouldn't get by changing the query. So I'd just suggest mentioning something about "case" in the warning message, to remind us what's going on the next time this crops up.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/GetLatestVersionDAO.java
```diff
@@ -65,7 +65,16 @@ public GetLatestVersionDAO(String adminSchemaName, String schemaName) {
...
+                    Integer newValue = version;
+                    if (currentValue != null) {
+                        // version can't be null due to NOT NULL db constraint
+                        newValue = Integer.max(currentValue, version);
+                        logger.warning("Version history entry " + schemaTypeName + " exists with multiple values [" + currentValue
+                                + ", " + version + "]; using " + newValue);
```
lmsurpre, line 74:
```suggestion
                                + ", " + version + "]; using " + newValue + ". Check schema name casing.");
```

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/GetLatestVersionDAO.java
lmsurpre: OK, I added "Check schema name casing." to the end of the warning message.  If that sounds good, please approve the PR and we can close out this issue.

prb112: > This should be done in the DAO query. The query should select upper(schema_name). This also needs to be added to the group by:
> 
> ```
>  final String cols = DataDefinitionUtil.join("UPPER(" + SchemaConstants.SCHEMA_NAME + ")", SchemaConstants.OBJECT_TYPE, SchemaConstants.OBJECT_NAME);
> ```

@punktilious I thought we did this and decided it wasn't the right approach.  I had used UCASE in the query maybe a few months back.  I even added the toUpperCase.  

I do think we might be rehashing.

lmsurpre: > I do think we might be rehashing.
Partially my fault; just trying to be clear on the desired behavior.

The advantage to doing it in the java code here is we can give a nice warning message.  However, what isn't clear to me is what the expected user action should be when they see this message.

If we want them to just ignore it, then I’m thinking we should either 
A. just fix the query / revert my proposed change; or
B. change the log message to info level (or lower)

If we want them to fix their version history table up, then I should probably update the warning message to indicate that (or point to somewhere that we document that this is safe for them to do).  The message would become something like
> To avoid this warning, please normalize the schema and object names to UPPER case in FHIR_ADMIN.VERSION_HISTORY.

Which one do we want?

lmsurpre: team decision:  we don't want to ask the user to clean up the VERSION_HISTORY table at this time, and so I will reduce the log level of this message of FINE

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/version/GetLatestVersionDAO.java
```diff
@@ -65,7 +65,16 @@ public GetLatestVersionDAO(String adminSchemaName, String schemaName) {
                 int version = rs.getInt(4);
 
                 String schemaTypeName = schema + ":" + type + ":" + name;
-                result.put(schemaTypeName, version);
+                result.compute(schemaTypeName, (key, currentValue) -> {
+                    Integer newValue = version;
+                    if (currentValue != null) {
+                        // version can't be null due to NOT NULL db constraint
+                        newValue = Integer.max(currentValue, version);
+                        logger.warning("Version history entry " + schemaTypeName + " exists with multiple values [" + currentValue
```
lmsurpre, line 73:
```suggestion
                        logger.fine("Version history entry " + schemaTypeName + " exists with multiple values [" + currentValue
```

Issue #1897, opened 2021-02-02
Move db bootstrapping out of server startup


lmsurpre:
**Is your feature request related to a problem? Please describe.**
We've had a few issues where users report that the default docker container is failing to start due to a timeout that occurs during bootstrapping of the derby databases. See https://github.com/IBM/FHIR/issues/1554 and https://github.com/IBM/FHIR/issues/1888

The problem is that the bootstrapping code inherits the global transaction timeout (120s by default) that we use for serving user requests.  At the very least, the bootstrapping should have a different transaction timeout.

We've made it easy(ish) to get past the issue by setting the default transaction timeout via an env var (#1666).
However, the only reason we have the bootstrapping code to begin with is to make it dead simple to get our server up and running, so this kind of defeats the purpose.

**Describe the solution you'd like**
1. Database bootstrapping should be moved to a bash script
2. Package this bootstrapping script in our zip installer (and thus our docker image as well)
3. Write a custom entrypoint for the docker image to make it simple to:
  * control whether or not bootstrapping happens
  * specify which dbs are bootstrapped, as per #1408

Initially we can just focus on derby, but I think it would make sense to extend that behavior to Db2 and PostgreSQL as well.

**Describe alternatives you've considered**
Either:
A. never perform bootstrapping during image startup; or
B. keep bootstrapping scoped to derby and force the use of the schema-tool cli and/or image for the 'real' dbs.

**Acceptance Criteria**
At least one acceptance criteria is included.

1.  
  GIVEN a global transaction timeout of 60s
    AND derby bootstrapping enabled
    WHEN a user runs `docker run -p9443:9443 ibmcom/ibm-fhir-server`
    THEN the server starts successfully
    AND has a functional embedded derby database

To confirm it:
1. build the project:  `mvn clean install -f fhir-parent -DskipTests`
2. build the docker image `docker build fhir-install -t ibmcom/ibm-fhir-server`
3. run the docker image, ask it to use the bootstrap DB, and set a lower transaction timeout

`docker run -e BOOTSTRAP_DB=true -e FHIR_TRANSACTION_MANAGER_TIMEOUT=60s ibmcom/ibm-fhir-server`

**Additional context**

tbieste:
[...]

Issue #1408, opened 2020-08-08
Bootstrapping of Derby database should be driven by configuration


punktilious:
**Is your feature request related to a problem? Please describe.**
The server bootstraps 4 Derby database schemas on initial startup:

            bootstrapFhirDb(utx, "default", "default", ds);
            bootstrapFhirDb(utx, "tenant1", "profile", ds);
            bootstrapFhirDb(utx, "tenant1", "reference", ds);
            bootstrapFhirDb(utx, "tenant1", "study1", ds);

This aligns with the initial configuration, and is intended to make the FHIR server easy to get up an running.

**Describe the solution you'd like**
Drive the bootstrapping from the fhir-server-config configuration. The bootstrap can then be skipped where a Derby database isn't required (e.g. in production instances).

**Describe alternatives you've considered**
Disable bootstrapping entirely and require database schemas to be provisioned before the server starts.

**Additional context**
N/A

lmsurpre:
I would vote to remove all bootsrapping code from the server and move it into a different shell script.  Then, for our docker image, we could run that to bootstrap derby (conditionally, based on some argument/config)

lmsurpre:
Currently I'm thinking to keep this simple.  We would *only* support bootstrapping of the default tenant's default datasource.
For our FHIR CI's multi-tenant tests, we'd need to do something special if we intend to continue supporting the 3 datastores for tenant1.

tbieste:
Wanted to get thoughts on the ENVVARs to use for the new script to be called by the fhir-server Dockerfile.

|Ibm-fhir-schematool |property	Ibm-fhir-schematool Purpose|	ENVVAR in ibm-fhir-server script|
|---------|-----------------|-----|
| |	|	DB_BOOTSTRAP: Whether to bootstrap DB. [true/false] |
|tool.behavior|	Switches from the Onboarding BEHAVIOR to Offboarding BEHAVIOR flow: [onboard/offboard/debug/custom].|	N/A|
|db.type|	The database type - [postgresql/db2]|	DB_TYPE: The DB type. [db2/derby/postgresql] |
|db.host|	The database server hostname|	DB_HOST: The DB host.|
|db.port|	The database server port|	DB_PORT: The DB port.|
|db.database|	The name of the database|	DB_DATABASE: The DB name.|
|user|	A username with connect and admin permissions on the target database|	DB_ADMIN_USER: The DB user.|
[...]

Issue #819, opened 2020-03-19
Feature request: option to specify list of include / exclude resources


smhgit:
It will be very helpful to have the ability to define which resources to include / exclude. Schema will be generated only for the included resources.

Hopefully this feature will help to reduce the database resources (storage / performance etc.)

lmsurpre:
The persistence tool already supports it.  What we need is a way to drive that from the cli, good documentation for that, and testing.

lmsurpre:
Note:  this could inhibit our ability to handle the `:contained` search parameter modifier where an unsupported resource type is contained within a supported one.

punktilious:
Also worth looking at catalog tuning (for Db2 and PostgreSQL) as the schema is pretty big, and this could cause cache pressure if there are multiple instances of our schema inside one database.

lmsurpre:
As part of this one we should also revisit JDBCIdentityCache.getResourceTypeNames and its callers.  Currently I think it returns Resource and DomainResource but it probably shouldn't?

lmsurpre:
> The persistence tool already supports it. What we need is a way to drive that from the cli, good documentation for that, and testing.

I added that as an experimental/undocumented feature some time ago (e.g. see https://github.com/IBM/FHIR/blob/main/demo/docker-compose.yml#L61 for usage), but unfortunately its still frustratingly easy to shoot yourself in the foot with this feature.

For example, an `_include` or `_revinclude` search will still try looking in all of the potential resource tables that could be referenced.  So a search like `GET [base]/Obseration?_include=Observation:subject` can fail with an SQL error like 
```
[INFO] Table/View 'PROCEDURE_LOGICAL_RESOURCES' does not exist.
```
We probably need to make our server aware of which resource tables are in the db to make this foolproof.  Til then, I think we'll just keep it experimental/undocumented.

PR #1986, opened 2021-02-24
Issue #1897 - Add bootstrap script as entrypoint

tbieste:
1. Issue #1897 - Add bootstrap script as entrypoint
2. Issue #1408 - move advanced bootstrapping into CI
3. Issue #819 - add undocumented 'resourceTypes' property to fhir-persistence-schema


fhir-install/Dockerfile
```diff
@@ -32,6 +39,7 @@ LABEL summary="Image for IBM FHIR Server with OpenJ9 and UBI 8"
 LABEL description="The IBM FHIR Server is a modular Java implementation of version 4 of the HL7 FHIR specification with a focus on performance and configurability."
 
 COPY --chown=1001:0 --from=base /opt/ol/wlp /opt/ol/wlp
```
tbieste, line 41: Copy /opt

fhir-install/src/main/docker/ibm-fhir-server/bootstrap.sh
```diff
@@ -0,0 +1,65 @@
...
+
+##############################################################################
+# Description:
+# This script is the entrypoint used by the ibm-fhir-server docker image, and 
+# optionally bootstraps a derby database prior to running the IBM FHIR Server.  
+
+set -e -o pipefail
+
+##############################################################################
+# The global variables used are:
```
tbieste, line 17: List the current directory.

fhir-install/Dockerfile
tbieste: Updated.

fhir-install/src/main/docker/ibm-fhir-server/bootstrap.sh
tbieste: Updated.

lmsurpre (approved):
LGTM

build/notifications/kafka/docker-compose.yml
```diff
@@ -63,6 +63,22 @@ services:
...
+          --db-type derby --prop db.database=/opt/ol/wlp/usr/servers/fhir-server/derby/reference --prop db.create=Y
+          --prop resourceTypes=Patient,Group,Practitioner,PractitionerRole,Device,Organization,Location,Medication,Observation,MedicationAdministration,StructureDefinition,ElementDefinition,CodeSystem,ValueSet,Resource
+          --update-schema &&
+        java -jar /opt/ibm-fhir-server/tools/fhir-persistence-schema-*-cli.jar
+          --db-type derby --prop db.database=/opt/ol/wlp/usr/servers/fhir-server/derby/study1 --prop db.create=Y
+          --prop resourceTypes=Patient,Group,Practitioner,PractitionerRole,Device,Organization,Location,Encounter,AllergyIntolerance,Observation,Condition,CarePlan,Provenance,Medication,MedicationAdministration,StructureDefinition,ElementDefinition,CodeSystem,ValueSet,Resource
+          --update-schema &&
+        /opt/ol/wlp/bin/server run fhir-server "
     healthcheck:
       start_period: 60s
```
prb112, line 83: maybe we should standardize to 120s (we have a variety of 60 and 120s)

docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -1491,7 +1388,6 @@ The global configuration contains non-tenant specific configuration parameters (
             },
             "jdbc":{
                 "__comment":"Configuration properties for the JDBC persistence implementation",
-                "bootstrapDb":false
```
prb112, line 1494: Does this make sense to have this section. now? or maybe update the comment to indicate it is a placeholder?

fhir-install/src/main/assembly/distribution.xml
```diff
@@ -106,5 +106,14 @@
...
         </dependencySet>
+        <dependencySet>
+            <useProjectArtifact>false</useProjectArtifact>
+            <outputDirectory>tools</outputDirectory>
+            <directoryMode>0755</directoryMode>
+            <fileMode>0755</fileMode>
+            <includes>
+                <include>${project.groupId}:fhir-persistence-schema:jar:cli</include>
+            </includes>
+        </dependencySet>
```
prb112, line 117: May want to simplify and add the other clis mentioned here, and we. can. have one deliverable to upload? 
https://github.com/IBM/FHIR/wiki/Release-Guide#release

fhir-install/src/main/docker/ibm-fhir-server/bootstrap.sh
```diff
@@ -0,0 +1,70 @@
...
+
+# _call_derby_db - local function to call derby database
+# ARGUMENTS:
+#   String of additional parameters
+function _call_derby_db {
+    /opt/java/openjdk/bin/java -jar ${FHIR_PERSISTENCE_SCHEMA_CLI_LOCATION}/fhir-persistence-schema-*-cli.jar \
+        --prop "db.create=Y" \
+        --prop "db.database=derby/fhirDB" \
+        --db-type derby \
+        ${1} 2>&1
```
prb112, line 46: Why would we take `${1}` what are the. additional parameters we would want to take?

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
```diff
@@ -200,6 +200,11 @@ protected Connection createConnection() {
      * can perform the DDL deployment in parallel
      */
     protected void configureConnectionPool() {
+        if ("derby".equals(dbType) && maxConnectionPoolSize > 1) {
```
prb112, line 203: Maybe we should use the DbType.DERBY.equals to ensure consistency

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
```diff
@@ -215,7 +220,16 @@ protected void configureConnectionPool() {
...
     protected void buildCommonModel(PhysicalDataModel pdm, boolean fhirSchema, boolean oauthSchema, boolean javaBatchSchema) {
         if (fhirSchema) {
-            FhirSchemaGenerator gen = new FhirSchemaGenerator(schema.getAdminSchemaName(), schema.getSchemaName(), isMultitenant());
+            String resourceTypesString = properties.getProperty("resourceTypes");
+
+            FhirSchemaGenerator gen;
+            if (resourceTypesString == null || resourceTypesString.isEmpty()) {
+                gen = new FhirSchemaGenerator(schema.getAdminSchemaName(), schema.getSchemaName(), isMultitenant());
+            } else {
+                Set<String> resourceTypes = new HashSet<>(Arrays.asList(resourceTypesString.split(",")));
```
prb112, line 229: You could use .stream().filter(x -> ModelSupport.isResourceType(x)).collect(Collection.toList())

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
```diff
@@ -1490,6 +1502,14 @@ protected void process() {
             checkCompatibility();
         }
 
+        if (translator.isDerby() && !"APP".equals(schema.getSchemaName())) {
+            if (!schema.isDefaultSchemaName()) {
```
prb112, line 1506: 1505 the last conditional ... is that the same as the 1506?

fhir-server-webapp/pom.xml
```diff
@@ -115,6 +115,22 @@
...
+            <exclusions>
+                <!-- exclude artifacts packaged in shared library dir -->
+                <exclusion>
+                    <groupId>com.ibm.db2</groupId>
+                    <artifactId>jcc</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.apache.derby</groupId>
+                    <artifactId>derby</artifactId>
+                </exclusion>
```
prb112, line 131: Maybe these should be marked provided at the top of the dependency list? 
That way they don't sneak in.

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
lmsurpre: we want a set, not a list.  we could use streams and Collection.toSet() but I think its equivalent.  any reason you think the stream version is better or just a style preference?

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
lmsurpre: no.  1505 checks whether schema.getSchemaName = APP.
if so, then we don't need to do anything.
if not, then we want to set it to APP (and log a warning iff the value was explicitly passed)

build/notifications/kafka/docker-compose.yml
lmsurpre: we only have 120s in one spot and there is good reason for it...its because that particular one does extra bootstrapping during container start

docs/src/pages/guides/FHIRServerUsersGuide.md
lmsurpre: I don't think it hurts to have it as a placeholder; and I'd prefer not to touch the user's guide any more in this change set because I have extensive changes in https://github.com/IBM/FHIR/pull/1989

fhir-install/src/main/assembly/distribution.xml
lmsurpre: i think this is the only one that we need at runtime; thus to me it makes sense to keep the others outside of the distribution (and users can continue to get them independent of the distribution zip)

fhir-install/src/main/docker/ibm-fhir-server/bootstrap.sh
lmsurpre: initially it was used so we could call `create-schemas` and then `update-schema`, but then I had troy remove the `create-schemas` step since it wasn't needed for embedded derby
personally, I think its fine to keep the helper this way in case we want to adapt the script to bootstrap other dbs in the future

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
lmsurpre: good call; in fact for enums we can just use `==`
```suggestion
        if (dbType == DbType.DERBY && maxConnectionPoolSize > 1) {
```

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
lmsurpre: oh i see you are proposing to validate the input here as well, but I made the decision to pass whatever the user passes to us on and I do the resourceType validation right in FhirSchemaGenerator.
specifically, I decided to generate tables for whatever they pass us (but log a warning if its not a known resource type).
if we want to reverse that decision, then we should at least log a warning if we're gonna ignore the input.
the option i considered is to just throw if they pass an invalid resource type name.

fhir-server-webapp/pom.xml
lmsurpre: will address this as part of https://github.com/IBM/FHIR/pull/1989

Issue #1982, opened 2021-02-23
Enforce StorageType source and destination for BulkData using FHIR_BULKDATA_ALLOWED_TYPES


prb112:
**Is your feature request related to a problem? Please describe.**
Enforce StorageType source and destination for BulkData using FHIR_BULKDATA_ALLOWED_TYPES

**Describe the solution you'd like**
Enforce StorageType source and destination for BulkData using FHIR_BULKDATA_ALLOWED_TYPES

**Describe alternatives you've considered**
Allow all

**Acceptance Criteria**
At least one acceptance criteria is included.

1.  Export
  GIVEN Env FHIR_BULKDATA_ALLOWED_TYPES=ibm-cos,aws-s3
    AND bulkdata is configured WHEN an export with file is executed
    THEN the export fails.

1.  Import
  GIVEN Env FHIR_BULKDATA_ALLOWED_TYPES=ibm-cos,aws-s3
    AND bulkdata is configured WHEN an import with file is executed
    THEN the import fails.

**Additional context**
n/a

kmbarton423:
Set environment variable FHIR_BULKDATA_ALLOWED_TYPES=ibm-cos,aws-s3
Configured FHIR server bulkdata:storageProviders:default:type=file
Attempted both $import and $export.
Received the following in response:

{
    "resourceType": "OperationOutcome",
    "id": "ac-11-0-2-6b5b11da-3c01-487f-b0c9-8753360c2e48",
    "issue": [
        {
            "severity": "fatal",
            "code": "not-supported",
            "details": {
                "text": "storageType is disallowed"
            },
            "expression": [
                "<empty>"
            ]
        }
    ]
}

Issue #1978, opened 2021-02-22
Ambiguous dependencies when starting IBM FHIR Server 4.6.0


tbieste:
**Describe the bug**
Exceptions about ambiguous dependencies when starting IBM FHIR Server (4.6.0-snapshot).

**To Reproduce**

1. Start Liberty Server (w/ fhir-bulkimportexport-webapp, fhir-openapi, fhir-server-webapp)

2. 14 exceptions similar to the following are shown in the Liberty Console:
```
[ERROR   ] CWWKZ0002E: An exception occurred while starting the application fhir-bulkimportexport-webapp. The exception message was: com.ibm.ws.container.service.state.StateChangeException: org.jboss.weld.exceptions.DeploymentException: Exception List with 14 exceptions:
Exception 0 :
org.jboss.weld.exceptions.DeploymentException: WELD-001409: Ambiguous dependencies for type String with qualifiers @Default
  at injection point [BackedAnnotatedField] @Inject @BatchProperty com.ibm.fhir.jbatch.bulkdata.export.system.ChunkReader.incomingUrl
  at com.ibm.fhir.jbatch.bulkdata.export.system.ChunkReader.incomingUrl(ChunkReader.java:0)
  Possible dependencies: 
  - Producer Method [String] with qualifiers [@Any @Default] declared as [[UnbackedAnnotatedMethod] @Produces @Dependent @ManualTrace @Claim public com.ibm.ws.security.mp.jwt.cdi.ClaimProducer.getString(InjectionPoint)],
  - Producer Method [String] with qualifiers [@Any @Default] declared as [[UnbackedAnnotatedMethod] @Produces @BatchProperty @Dependent public com.ibm.ws.jbatch.cdi.BatchProducerBean.produceProperty(InjectionPoint)]

	at org.jboss.weld.bootstrap.Validator.validateInjectionPointForDeploymentProblems(Validator.java:386)
	at org.jboss.weld.bootstrap.Validator.validateInjectionPoint(Validator.java:290)
	at org.jboss.weld.bootstrap.Validator.validateGeneralBean(Validator.java:143)
	at org.jboss.weld.bootstrap.Validator.validateRIBean(Validator.java:164)
	at org.jboss.weld.bootstrap.Validator.validateBean(Validator.java:526)
	at org.jboss.weld.bootstrap.ConcurrentValidator$1.doWork(ConcurrentValidator.java:64)
[...]

Issue #1363, opened 2020-07-27
Job still in STARTED status after pod crash


lmsurpre:
**Describe the bug**
We deploy IBM FHIR Server on Kubernetes.  I had a pod that was executing a batch job go down and I expected the liberty batch framework to mark the job as paused/stopped/failed, but instead it just stays in STARTED mode and we have no way to resume it from the external API.

**To Reproduce**

Basically you just need to make it crash mid-job.  Here's one way:
1. Deploy server to kubertnetes (or similar platform) that will enforce disk quota:
* with a disk size limit (e.g. 1GB)
* with FINER logging
2. Invoke a big import

The pod should get evicted.

Now check the status of the job that was running and it will be in a state of STARTED

**Expected behavior**
We need to recognize this case and restart/resume the job from a new worker node.

**Additional context**
For https://github.com/IBM/FHIR/issues/767 we decided to resume the job when the status endpoint is invoked, but that logic only runs when the job is paused.  Otherwise, we have no idea if its being already worked or not.

Paul mentioned to me that there is a way to resume a job by issues a PUT to the underlying jobexecution via the liberty batch api.  However, currently our code doesn't do this and we'd have no clue that we should since the job status is never paused or stopped.

PR #1999, opened 2021-02-27
Improve ImportExport with JavaBatch

prb112:
- Change Runtime Check for ANTLR Version to 4.9.1
- Change confusing $import error message to reflect actual error
- Check the HttpStatusCode for $import/$export job creation for 201,
send useful message back
- Normalize the directory structure for the jbatch webapp
- Associate the jbatch webapp with the operation to normalize the code
- Move the IndexGenerator the the src/test/java to the tools package
- Remove prior FHIRBasicAuthenticator
- Tests Clients must use custom vm arguments to use the TLS key
    -Djavax.net.ssl.keyStore=/wlp/usr/servers/fhir-server/resources/security/fhirTrustStore.p12
    -Djavax.net.ssl.trustStore=/wlp/usr/servers/fhir-server/resources/security/fhirTrustStore.p12
    -Djavax.net.ssl.keyStoreType=p12
    -Djavax.net.ssl.trustStoreType=p12
    -Djavax.net.ssl.trustStorePassword=change-password
    -Djavax.net.ssl.keyStorePassword=change-password
- Remove the implied dependency on Apache CXF client in the bulkdata
client in test
- Remove DummpyImportExportImpl as it is unused
- Refactor the BulkDataClient to use Apache Http Client
- Created an Adapter to manage changes in the Configuration
- Added an Enabled Configuration Setting that is per tenant
- Added logging for the BulkData Client
- Refactor the batch-jobs definitions to include a consistent and
minimally sensitive Configuration.
- Normalize the OperationFields into a registry
- Update JobParameters to reflect the OperationFields
- Refactor Duplicate StorageType classes
- Move the tools to src/test/java
- Mark fhir-server provided in the webapp for bulkdata
- Bulk Data Audit Logging #1332
	- Add BulkAuditLogger and fhir-audit dependency
	- Add to Import
- NullPointer in ImportJobListener.afterJob #1361
	- Refactore the AfterJob ExitStatus and Reporter
	- Replicate and harden the partition handling in the
ImportPartitionMapper which feeds the transientUserData object
	- Occasionally we don't set the importResourceType on the
PartitionPlan.  I've added code to make sure this is set.
- Enforce StorageType source and destination for BulkData using
FHIR_BULKDATA_ALLOWED_TYPES #1982
	- Introduce configuration from System.env FHIR_BULKDATA_ALLOWED_TYPES
	- Add Operation specific checks for enabled/verify allowed source types
- Added Preflight Configuration Check for S3, Https, File
- Ambiguous dependencies when starting IBM FHIR Server 4.6.0 #1978
	- Handles the JobContext, StepContext to BulkDataContext using a
BatchContextAdapter
	- InjectionPoints are now specific versus generic String injection
points which cause the conflict.
	- Must inject the PartionPlan properties
- Remove the default Bucketname
- Refactoring of the fhirServer/bulkdata configuration #1390
	- Introduce Legacy Configuration
	- Introduce V2 Configuration
- Refactor Unique Calls to S3 Client into a Single Control Wrapper
- Exclude database-utils derby/db2 from the webapp.
- Job still in STARTED status after pod crash #1363
	- Add web.xml to block uncovered HTTP/S methods
	- Add ServletContextListener to Log Startup and hook into the local
BatchRuntime
- Added Logging Shunts where a try-catch are surrounding and rethrowing
the implemented JavaBatch Methods so the logging can push a message to
the main System.Out warning.
	- $import is now covered
	- Introduced StepChunkListener to unify the logging for Export/Import
- Increase logging in bulkdataclient
- Adding to TrustStore for S3
- Remove hardcoded password in the HttpWrapper
- Support bulk import/export from the local filesystem #1055
	- FileWrapper and a new file configuration enable writing to a local
file.
	- FilePreflight checks to see if the configuration supports write or
read
	- Also Add export to local file support to bulkdata export batchjob
#1211


operation/fhir-operation-bulkdata/pom.xml
```diff
@@ -48,5 +48,13 @@
             <artifactId>jsonassert</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
lmsurpre, line 51: not really related to this PR, but can we flip the `jakarta.ws.rs-api` above to provided?

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/audit/BulkAuditLogger.java
```diff
@@ -0,0 +1,312 @@
...
+     * @param eventType - the type of event
+     * @param resource
+     * - The Resource object.
+     * @param startTime
+     * - The start time of the request execution.
+     * @param endTime
+     * - The end time of the request execution.
+     * @param responseStatus
+     * - The response status.
+     * @param location
```
tbieste, line 93: Applies to several methods in this file:  Location and users params do not have hyphen in comments like other params.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/export/group/resource/GroupHandler.java
```diff
@@ -0,0 +1,150 @@
...
+import com.ibm.fhir.persistence.FHIRPersistence;
+import com.ibm.fhir.persistence.context.FHIRPersistenceContext;
+import com.ibm.fhir.persistence.context.FHIRPersistenceContextFactory;
+import com.ibm.fhir.persistence.helper.FHIRTransactionHelper;
+import com.ibm.fhir.search.SearchConstants;
+import com.ibm.fhir.search.context.FHIRSearchContext;
+import com.ibm.fhir.search.util.SearchUtil;
+
+/**
+ *
```
tbieste, line 30: Please add class description.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/export/patient/resource/PatientResourceHandler.java
```diff
@@ -0,0 +1,213 @@
...
+import com.ibm.fhir.persistence.FHIRPersistence;
+import com.ibm.fhir.persistence.context.FHIRPersistenceContext;
+import com.ibm.fhir.persistence.context.FHIRPersistenceContextFactory;
+import com.ibm.fhir.search.SearchConstants;
+import com.ibm.fhir.search.compartment.CompartmentUtil;
+import com.ibm.fhir.search.context.FHIRSearchContext;
+import com.ibm.fhir.search.util.SearchUtil;
+
+/**
+ *
```
tbieste, line 37: Please add class description.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/export/writer/SparkParquetWriter.java
```diff
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
```
tbieste, line 5: Update copyright.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/control/JobControllerBean.java
```diff
@@ -0,0 +1,34 @@
...
+
+import javax.annotation.PostConstruct;
+import javax.annotation.security.RunAs;
+import javax.batch.operations.JobOperator;
+import javax.batch.runtime.BatchRuntime;
+import javax.ejb.Startup;
+import javax.inject.Inject;
+
+/**
+ *
```
tbieste: Please add class description.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/data/CheckPointUserData.java
```diff
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
```
tbieste, line 5: Update copyright.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/data/TransientUserData.java
```diff
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
```
tbieste, line 4: Update copyright.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/audit/BulkAuditLogger.java
prb112: Fixed, coming in PR

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/export/writer/SparkParquetWriter.java
prb112: in next commit

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/export/group/resource/GroupHandler.java
prb112: next commit

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/export/patient/resource/PatientResourceHandler.java
prb112: next commit

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/control/JobControllerBean.java
prb112: I've since removed this class

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/system/ChunkWriter.java
```diff
@@ -0,0 +1,110 @@
+/*
+ * (C) Copyright IBM Corp. 2019, 2020
+ *
```
tbieste, line 3: Update copyright.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/system/ExportJobListener.java
```diff
@@ -0,0 +1,151 @@
+/*
+ * (C) Copyright IBM Corp. 2020
```
tbieste, line 2: Update copyright.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/data/TransientUserData.java
prb112: next commit

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/data/CheckPointUserData.java
prb112: next commit

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/system/ExportJobListener.java
prb112: Fixed in next commit

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/system/SystemExportPartitionMapper.java
```diff
@@ -4,22 +4,26 @@
  * SPDX-License-Identifier: Apache-2.0
```
tbieste, line 4: Update copyright.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/load/data/ImportCheckPointData.java
```diff
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
```
tbieste, line 4: Update copyright.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/load/data/ImportTransientUserData.java
```diff
@@ -4,16 +4,19 @@
  * SPDX-License-Identifier: Apache-2.0
```
tbieste, line 4: Update copyright.

fhir-bulkdata-webapp/src/test/java/com/ibm/fhir/bulkcommon/BulkDataUtilsTest.java
```diff
@@ -14,15 +14,15 @@
```
tbieste, line 14: Update copyright.

fhir-bulkdata-webapp/src/test/java/com/ibm/fhir/bulkcommon/SparkParquetWriterTest.java
```diff
@@ -16,7 +16,7 @@
 import org.apache.commons.io.FileUtils;
 import org.testng.annotations.Test;
 
-import com.ibm.fhir.jbatch.bulkdata.export.common.SparkParquetWriter;
+import com.ibm.fhir.bulkdata.export.writer.SparkParquetWriter;
```
tbieste, line 19: Update copyright.

fhir-bulkdata-webapp/src/test/java/com/ibm/fhir/jbatch/bulkdata/tools/Main.java
```diff
@@ -0,0 +1,339 @@
+/*
+ * (C) Copyright IBM Corp. 2020
```
tbieste, line 2: Is this file needed, since it is commented out?  Otherwise, update copyright.

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/GroupExportOperation.java
```diff
@@ -18,7 +18,7 @@
  * <code>$export</code>
```
tbieste, line 18: Update copyright.

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/PatientExportOperation.java
```diff
@@ -18,7 +18,7 @@
  * <code>$export</code>
```
tbieste, line 18: Update copyright.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/system/SystemExportPartitionMapper.java
prb112: Fixed in next commit

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/load/data/ImportCheckPointData.java
prb112: Fixed in next commit

fhir-bulkdata-webapp/src/test/java/com/ibm/fhir/bulkcommon/BulkDataUtilsTest.java
prb112: Fixed in next commit

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/load/data/ImportTransientUserData.java
prb112: Fixed in next commit

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/system/ChunkWriter.java
prb112: Fixed in next commit

fhir-bulkdata-webapp/src/test/java/com/ibm/fhir/jbatch/bulkdata/tools/Main.java
prb112: I'm going to remove it.  It was a prior test file.

fhir-bulkdata-webapp/src/test/java/com/ibm/fhir/bulkcommon/SparkParquetWriterTest.java
prb112: Fixed

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/GroupExportOperation.java
prb112: fixed

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/PatientExportOperation.java
prb112: fixed

operation/fhir-operation-bulkdata/src/test/java/com/ibm/fhir/operation/bulkdata/model/BulkExportJobRestTest.java
```diff
@@ -304,18 +304,11 @@ public void testBulkExportJobInstanceResponseParserAndGeneratorRoundTripEmpty()
     public void testBulkExportJobInstanceRequestWithSensitive() throws Exception {
         String jsonString =
```
tbieste, line 305: Update copyright.

operation/fhir-operation-bulkdata/src/test/java/com/ibm/fhir/operation/bulkdata/tool/helpers/dynamic/AgeRangeBloodPressureGroup.java
```diff
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package com.ibm.fhir.operation.bulkdata.helpers;
+package com.ibm.fhir.operation.bulkdata.tool.helpers.dynamic;
```
tbieste, line 7: Update copyright.

prb112: I'll do all at the same time.  Thanks

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/common/BulkDataUtils.java
```diff
@@ -204,7 +153,7 @@ private static int getFhirResourceFromBufferReader(BufferedReader resReader, int
                 try {
                     fhirResources.add(FHIRParser.parser(Format.JSON).parse(new StringReader(resLine)));
                     exported++;
-                    if (exported == Constants.IMPORT_NUMOFFHIRRESOURCES_PERREAD) {
+                    if (exported == adapter.getImportNumberOfFhirResourcesPerRead(null)) {
```
tbieste, line 156: Maybe change to ">=" just in case somewhere/somehow getImportNumberOfFhirResourcesPerRead
 returned a different value and we loop forever.

tbieste (commented):
Left minor comments primarily around copyright year updates, etc.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/common/BulkDataUtils.java
prb112: it is a constant underneath.  However, it could change based on the interface.  Many thanks for the catch.

prb112: Fixed

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/audit/BulkAuditLogger.java
```diff
@@ -0,0 +1,311 @@
...
+     * @param responseStatus
+     * The response status.
+     * @param location
+     * the destination or source for the export or import
+     * @param users
+     * the principals that initiated the request
+     * @return AuditLogEntry an audit log entry.
+     */
+    private AuditLogEntry createAuditLogEntry(AuditLogEventType eventType, Resource resource, Date startTime, Date endTime, Response.Status responseStatus,
+        String location, String users) {
```
lmsurpre, line 99: Applies to a number of long method declarations:  please use double-indent to offset from the first expression of the method

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/audit/BulkAuditLogger.java
```diff
@@ -0,0 +1,311 @@
...
+     *            The response status.
+     * @param location
+     *            the destination or source for the export or import
+     * @param users
+     *            the principals that initiated the request
+     * @throws Exception
+     */
+    public void logUpdateOnImport(Resource oldResource, Resource updatedResource, Date startTime, Date endTime, Response.Status responseStatus, String location,
+        String users)
+        throws Exception {
```
lmsurpre, line 230: Applies to a number of long method declarations:  please use double-indent to offset from the first expression of the method

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/common/BulkDataUtils.java
```diff
@@ -59,30 +45,33 @@
...
-import com.ibm.fhir.jbatch.bulkdata.load.ImportTransientUserData;
 import com.ibm.fhir.model.format.Format;
 import com.ibm.fhir.model.parser.FHIRParser;
 import com.ibm.fhir.model.parser.exception.FHIRParserException;
 import com.ibm.fhir.model.resource.OperationOutcome;
 import com.ibm.fhir.model.resource.Resource;
 import com.ibm.fhir.model.util.FHIRUtil;
 import com.ibm.fhir.model.util.ModelSupport;
+import com.ibm.fhir.operation.bulkdata.config.ConfigurationAdapter;
+import com.ibm.fhir.operation.bulkdata.config.ConfigurationFactory;
```
lmsurpre, line 58: I've always wanted some common spot for the stuff thats shared between fhir-operation-bulkdata and fhir-bulkdata-webapp; maybe another day :-)

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/checkpoint/CheckPointAlgorithm.java
```diff
@@ -0,0 +1,73 @@
...
+import javax.inject.Inject;
+
+import com.ibm.fhir.bulkdata.jbatch.export.data.TransientUserData;
+import com.ibm.fhir.operation.bulkdata.config.ConfigurationFactory;
+
+/**
+ * BulkData Export Custom CheckpointAlgorithm which considers COS size requirements while checkpointing.
+ */
+@Dependent
+public class CheckPointAlgorithm implements CheckpointAlgorithm {
```
lmsurpre, line 23: having the instance class name vary from the interface name only by casing is a little odd.  is there any distinguishing feature from other checkpoint algorithm which we can use in the name?  maybe `BulkExportCheckpointAlgorithm`?

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/data/CheckPointUserData.java
```diff
@@ -1,18 +1,17 @@
...
 import java.util.List;
 
 import com.ibm.cloud.objectstorage.services.s3.model.PartETag;
 
 /**
- * Bulk export Chunk implementation - job check point data.
- *
+ * Bulk Export Job Checkpoint data.
  */
 public class CheckPointUserData implements java.io.Serializable {
```
lmsurpre, line 16: unrelated to your change, but wondering if we should follow the framework's precedence here and use `CheckpointUserData` rather than `CheckPointUserData`

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/fast/NopWriter.java
```diff
@@ -27,7 +27,7 @@
     @Inject
     JobContext jobContext;
 
-    /**
+    /*
```
lmsurpre, line 30: why not javadoc?

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/FileProvider.java
```diff
@@ -0,0 +1,153 @@
...
+        this.fhirResourceType = fhirResourceType;
+
+        String fileName = cosBucketPathPrefix + "_" + fhirResourceType + "_" + chunkData.getUploadCount() + ".ndjson";
+        String base = configuration.getBaseFileLocation(source);
+
+        String fn = base + "/" + fileName;
+        Path p1 = Paths.get(fn);
+        try {
+            // This is a trap. Be sure to mark CREATE and APPEND.
+            // We do create empty files.
```
lmsurpre, line 117: under what conditions do we create empty files?

fhir-path/src/main/java/com/ibm/fhir/path/FHIRPathLexer.java
```diff
@@ -1,13 +1,17 @@
...
+import org.antlr.v4.runtime.Vocabulary;
+import org.antlr.v4.runtime.VocabularyImpl;
+import org.antlr.v4.runtime.atn.ATN;
+import org.antlr.v4.runtime.atn.ATNDeserializer;
+import org.antlr.v4.runtime.atn.LexerATNSimulator;
+import org.antlr.v4.runtime.atn.PredictionContextCache;
 import org.antlr.v4.runtime.dfa.DFA;
-import org.antlr.v4.runtime.misc.*;
+
+//Manually changed to 4.9.1
```
lmsurpre, line 14: seems a bit odd to have these fhirpath antlr changes mixed in with the other ones.  I was thinking John already addressed this, but I guess that was just for the model?

fhir-server-test/src/test/resources/test.properties
```diff
@@ -39,11 +39,12 @@ fhirclient.http.return.pref = minimal
 test.client.oauth.enabled = false
 test.websocket.enabled = false
 test.kafka.enabled = false
-test.bulkdata.import.enabled = false
-test.bulkdata.export.enabled = false
-test.bulkdata.useminio = false
-test.bulkdata.minio.username = minio
-test.bulkdata.minio.password = change-password
```
lmsurpre, line 46: do we still exercise minio during the CI, or is it only the file-based storage provider that gets tested now?

lmsurpre:
question:  what is the change to `fhir-server/liberty-config/resources/security/fhirTrustStore.p12` ?

fhir-server/liberty-config/server.xml
```diff
@@ -14,7 +14,7 @@
         <!-- mpJwt-1.1 isn't used by default,
              but we include it here to avoid NoClassDefFound in our classes that *can* use it -->
         <feature>mpJwt-1.1</feature>
-    </featureManager>
+	</featureManager>
```
lmsurpre, line 17: looks like a tab snuck in.  I'm assuming its accidental:
```suggestion
    </featureManager>
```

operation/fhir-operation-bulkdata/src/docs/Configuration.md
```diff
@@ -0,0 +1,79 @@
+The following is a brief template of the V2 Configuration.
```
lmsurpre, line 1: I think its the first time I've seen us put info under `src/docs`.  Do we want to adopt this pattern going forward?
If not, maybe just stuff it into the fhir-operation-bulkdata README?

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/BulkDataResourceProvider.java
```diff
@@ -1,16 +0,0 @@
...
- * (C) Copyright IBM Corp. 2020
- *
- * SPDX-License-Identifier: Apache-2.0
- */
-
-package com.ibm.fhir.operation.bulkdata;
-
-import com.ibm.fhir.registry.util.PackageRegistryResourceProvider;
-
-public class BulkDataResourceProvider extends PackageRegistryResourceProvider {
```
lmsurpre, line 11: I believe this registry resource provider is what we were using to advertise our support of the bulkdata spec in our CapabilityStatement.  Are you still advertising that capability, but in a new way?
See https://github.com/IBM/FHIR/issues/1315 for more info.

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/client/BulkDataClient.java
```diff
@@ -652,53 +771,81 @@ public String submitImport(String inputFormat, String inputSource, List<Input> i
...
+            jobPost.releaseConnection();
+            jobResponse.close();
+        }
+        cli.close();
+
+        return baseUri + "/$bulkdata-status?job=" + transformer.endcodeJobId(jobId);
     }
 
-    /**
+    /*
```
lmsurpre, line 831: why not javadoc?

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/BulkDataResourceProvider.java
lmsurpre: ah, now I see it was just moved to `ResourceProvider` under subpackage `provider`.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/data/CheckPointUserData.java
prb112: I have no preferences - which one ? you pick ;-)

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/checkpoint/CheckPointAlgorithm.java
prb112: yes - totally agree.  I'll make that change

fhir-server-test/src/test/resources/test.properties
prb112: It's  only the file-base provider. i can  enable minio too  - I'd just have to include a different set of tests. 
We could do that with a separate issue.  Let me  know and I can create it.

fhir-path/src/main/java/com/ibm/fhir/path/FHIRPathLexer.java
prb112: oh - good find.   yeah - I  did it prior to John, it turns out there really aren't any changes with 4.9.1

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/fast/NopWriter.java
prb112: it's a private method

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/FileProvider.java
prb112: 0 resourced files. 
E.g. export all resources and we have zero for parameters. 

I decided too put it in writeResources.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/fast/NopWriter.java
prb112: the next  method was private

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/FileProvider.java
prb112: I just changed this slightly so we don't create empty files.

prb112:
> question: what is the change to `fhir-server/liberty-config/resources/security/fhirTrustStore.p12` ?

I added the key's signer to the truststore. (it was an annoying thing with httpclient) 
I think I also added the ibm  cloud db ca

operation/fhir-operation-bulkdata/src/docs/Configuration.md
prb112: moved/changed

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/client/BulkDataClient.java
prb112: it really isn't intended  too be publicaly readable.  It's also a private method.  thus the `/*`

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/common/BulkDataUtils.java
prb112: This BulkDataUtils is really focused on S3 operations. 
I pushed the helper methods in the corresponding provider classes (StorageProvider)

operation/fhir-operation-bulkdata/src/test/sh/export-group-1.sh
```diff
@@ -0,0 +1,30 @@
...
+# < HTTP/2 202
+# < content-location: https://localhost:9443/fhir-server/api/v4/$bulkdata-status?job=eikWD%2BJszJJ_DkN6HWMLYQ
+
+# 3 - Set the Job Id and execute the request
+# Repeat until 200(OK)
+curl --location --request GET 'https://localhost:9443/fhir-server/api/v4/$bulkdata-status?job=O5X6d9LOYDZV35OjHkt5wA' \
+    --header 'Content-Type: application/fhir+json' -k \
+    -u "fhiruser:${PASS}" -v --header "X-FHIR-TENANT-ID: ${TENANT_ID}"
+
+# 4 - Check the file that is output.
```
lmsurpre, line 30: Is there a way to automate the checking of the file that is output? (for future consideration)

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/data/CheckPointUserData.java
lmsurpre: I vote to update any classnames that have `CheckPoint` and change them to `Checkpoint`

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/fast/NopWriter.java
lmsurpre: private methods can have javadoc.  it makes it come up nice from your IDE.  it doesn't go into the published javadoc.  I'd vote to keep using javadoc-style comments for private methods.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/FileProvider.java
lmsurpre: sounds good

fhir-server-test/src/test/resources/test.properties
lmsurpre: I think it would be nice to exercise both providers (since we already did the work to enable minio).  Agree it can be a future thing.

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/client/BulkDataClient.java
lmsurpre: same comment as the other one...I think its totally acceptable to use javadoc-style comments on private methods

lmsurpre:
```suggestion
    /**
```

operation/fhir-operation-bulkdata/src/test/sh/export-group-1.sh
prb112: It really depends on the storage.

prb112: I gave an example in this one.

fhir-bulkdata-webapp/pom.xml
```diff
@@ -45,6 +45,23 @@
             <groupId>org.glassfish</groupId>
             <artifactId>jakarta.json</artifactId>
         </dependency>
+        <dependency>
+            <!-- This is provided, yet never used. -->
+            <groupId>${project.groupId}</groupId>
+            <artifactId>fhir-server</artifactId>
+            <version>${project.version}</version>
+            <scope>provided</scope>
+        </dependency>
```
lmsurpre, line 54: wondering what this is for... are we just using it for its transitive dependency on other things marked provided?

prb112: since we pick up operation-bulkdata, I don't want to pull in fhir-server.  I marked as provided, and it is not used.

fhir-server-test/src/test/resources/test.properties
prb112: both providers are in the IT tests

fhir-server/liberty-config/config/default/fhir-server-config-db2.json
```diff
@@ -117,30 +117,83 @@
...
-            "cosFileMaxResources": 200000,
-            "cosFileMaxSize": 209715200,
-            "validBaseUrls": [],
-            "useFhirServerTrustStore": true,
-            "enableParquet": false
+            "storageProviders": {
+                "default" : {
+                    "type": "file",
+                    "_type": "ibm-cos|aws-s3|file|https",
+                    "valid-base-urls": [],
```
lmsurpre, line 146: previously we've stuck to CamelCase for config props, but I think I'm OK with consistently using kabob-case on these culk properties...anyone disagree?

fhir-server/liberty-config/config/default/fhir-server-config-db2.json
```diff
@@ -117,30 +117,83 @@
...
-                "cos.srvinst.id": "change-password"
+            "enabled": true,
+            "core": {
+                "api": {
+                    "url": "https://localhost:9443/ibm/api/batch",
+                    "user": "fhiradmin",
+                    "password": "change-password",
+                    "truststore": "resources/security/fhirTrustStore.p12",
+                    "truststore-password": "change-password", 
+                    "trust-all": true
```
lmsurpre, line 128: what does this do?

prb112: sets the client ssl to trust the ca and hostname in the http client used in httpwrapper

fhir-server/liberty-config/config/default/fhir-server-config-db2.json
prb112: flipping to camelcase is quick  change.   just let me know

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/preflight/impl/FilePreflight.java
```diff
@@ -0,0 +1,82 @@
...
+                        // Skip out of the for loop
+                        return;
+                    }
+                }
+            }
+            if (!accessible) {
+                throw util.buildExceptionWithIssue("The location for the bulkdata operation is not found", IssueType.NOT_FOUND);
+            }
+        } else {
+            throw util.buildExceptionWithIssue("No File Base Configured for FHIR bulkdata operation", IssueType.INVALID);
```
lmsurpre, line 71: INVALID is used to indicate that the request was invalid, whereas I think this is a server-side configuration error.  We should be sure that these map to HTTP 500 errors instead, such as by using 
```suggestion
            throw util.buildExceptionWithIssue("No File Base Configured for FHIR bulkdata operation", IssueType.EXCEPTION);
```

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/preflight/impl/FilePreflight.java
```diff
@@ -0,0 +1,82 @@
...
+                    Path p1 = Paths.get(base, f.getAbsolutePath());
+                    accessible = Files.isReadable(p1);
+                    if (!accessible) {
+                        // Skip out of the for loop
+                        return;
+                    }
+                }
+            }
+            if (!accessible) {
+                throw util.buildExceptionWithIssue("The location for the bulkdata operation is not found", IssueType.NOT_FOUND);
```
lmsurpre, line 68: NOT_FOUND is typically mapped into a HTTP 404 error.  In this case, its actually the backend file storageProvider that is misconfigured and so I think we should make sure this maps to an HTTP 500 instead.
```suggestion
                throw util.buildExceptionWithIssue("The location for the bulkdata operation is not found", IssueType.NO_STORE);
```

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/preflight/impl/FilePreflight.java
```diff
@@ -0,0 +1,82 @@
...
+        } else {
+            throw util.buildExceptionWithIssue("No File Base Configured for FHIR bulkdata operation", IssueType.INVALID);
+        }
+    }
+
+    @Override
+    public void checkStorageAllowed(StorageDetail storageDetail) throws FHIROperationException {
+        if (storageDetail != null && !StorageType.FILE.value().equals(storageDetail.getType())){
+            CommonUtil util = new CommonUtil();
+            throw util.buildExceptionWithIssue("File: Configuration not set to import from storageDetail '" + getSource() + "'", IssueType.INVALID);
```
lmsurpre, line 79: INVALID maps to 400 Bad Request.  Is this scenario a bad request or a misconfigured server?

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/preflight/impl/S3Preflight.java
```diff
@@ -0,0 +1,159 @@
...
+            BucketHostCallable outcomeCallable = new BucketHostCallable(getOutcome());
+            callables.add(outcomeCallable);
+        }
+
+        ExecutorService executor = Executors.newFixedThreadPool(2);
+        try {
+            List<Future<Boolean>> futures = executor.invokeAll(callables, 60, TimeUnit.SECONDS);
+            for (Future<Boolean> future : futures) {
+                if (!future.get()) {
+                    throw export.buildOperationException("Unable to access s3 source or outcome during timeout", IssueType.INVALID);
```
lmsurpre, line 70:
```suggestion
                    throw export.buildOperationException("Unable to access s3 source or outcome during timeout", IssueType.EXCEPTION);
```

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/preflight/impl/S3Preflight.java
```diff
@@ -0,0 +1,159 @@
...
+                }
+            }
+        } catch (ExecutionException ee) {
+            throw export.buildOperationException("Failed to execute the s3 access, check the s3 configuration", IssueType.INVALID);
```
lmsurpre, line 74:
```suggestion
            throw export.buildOperationException("Failed to execute the s3 access, check the s3 configuration", IssueType.NO_STORE);
```

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/preflight/impl/S3Preflight.java
```diff
@@ -0,0 +1,159 @@
...
+        } catch (InterruptedException e) {
+            throw export.buildOperationException("Timeout hit trying to access s3, check the s3 configuration", IssueType.INVALID);
```
lmsurpre, line 76:
```suggestion
            throw export.buildOperationException("Timeout hit trying to access s3, check the s3 configuration", IssueType.NO_STORE);
```

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/preflight/impl/S3Preflight.java
```diff
@@ -0,0 +1,159 @@
...
+     * @param source
+     * @throws FHIROperationException
+     */
+    public void validate(String source) throws FHIROperationException {
+        ConfigurationAdapter adapter = ConfigurationFactory.getInstance();
+        if (adapter.isStorageProviderAuthTypeIam(source)) {
+            String apiKey = adapter.getStorageProviderAuthTypeIamApiKey(source);
+            String resourceId = adapter.getStorageProviderAuthTypeIamApiResourceInstanceId(source);
+            if (apiKey == null || resourceId == null || apiKey.isEmpty() || resourceId.isEmpty()) {
+                throw export.buildOperationException("bad configuration for the iam configuration", IssueType.INVALID);
```
lmsurpre, line 95:
```suggestion
                throw export.buildOperationException("bad configuration for the iam configuration", IssueType.EXCEPTION);
```

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/preflight/impl/S3Preflight.java
```diff
@@ -0,0 +1,159 @@
...
+            }
+        } else if (adapter.isStorageProviderAuthTypeHmac(source)) {
+            String accessKey = adapter.getStorageProviderAuthTypeHmacAccessKey(source);
+            String secretKey = adapter.getStorageProviderAuthTypeHmacSecretKey(source);
+            if (accessKey == null || secretKey == null || accessKey.isEmpty() || secretKey.isEmpty()) {
+                throw export.buildOperationException("bad configuration for the hmac configuration", IssueType.INVALID);
```
lmsurpre, line 101:
```suggestion
                throw export.buildOperationException("bad configuration for the hmac configuration", IssueType.EXCEPTION);
```

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/preflight/impl/S3Preflight.java
```diff
@@ -0,0 +1,159 @@
...
+            }
+        } else if (adapter.isStorageProviderAuthTypeBasic(source)) {
+            String user = adapter.getStorageProviderAuthTypeUsername(source);
+            String password = adapter.getStorageProviderAuthTypePassword(source);
+            if (user == null || password == null || user.isEmpty() || password.isEmpty()) {
+                throw export.buildOperationException("bad configuration for the basic configuration", IssueType.INVALID);
```
lmsurpre, line 107:
```suggestion
                throw export.buildOperationException("bad configuration for the basic configuration", IssueType.EXCEPTION);
```

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/preflight/impl/S3Preflight.java
```diff
@@ -0,0 +1,159 @@
...
+            }
+        } else {
+            throw export.buildOperationException("Failed to specify the source or outcome bucket's authentication mechanism", IssueType.INVALID);
```
lmsurpre, line 110:
```suggestion
            throw export.buildOperationException("Failed to specify the source or outcome bucket's authentication mechanism", IssueType.EXCEPTION);
```

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/preflight/impl/S3Preflight.java
```diff
@@ -0,0 +1,159 @@
...
+
+        private String source = null;
+        private String url = null;
+
+        public BucketHostCallable(String source) throws FHIROperationException {
+            this.source = source;
+            ConfigurationAdapter adapter = ConfigurationFactory.getInstance();
+            url = adapter.getStorageProviderEndpointExternal(source);
+            if (url == null || url.isEmpty()) {
+                throw export.buildOperationException("endpoint-internal is undefined.", IssueType.INVALID);
```
lmsurpre, line 127:
```suggestion
                throw export.buildOperationException("endpoint-internal is undefined.", IssueType.EXCEPTION);
```

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/preflight/impl/FilePreflight.java
prb112: it's actually both.  It means that  the request did not match the server settings. and that may mean the server has a bad configuration

fhir-bulkdata-webapp/src/main/resources/META-INF/batch-jobs/FhirBulkExportFastJob.xml
```diff
@@ -0,0 +1,59 @@
...
+                    <property name="fhir.search.fromdate" value="#{jobParameters['fhir.search.fromdate']}"/>
+                    <property name="fhir.search.todate" value="#{jobParameters['fhir.search.todate']}"/>
+                    <property name="fhir.typeFilters" value="#{jobParameters['fhir.typeFilters']}"/>
+                </properties>
+            </reader>
+            <writer ref="com.ibm.fhir.bulkdata.jbatch.export.fast.NopWriter">
+                <properties>
+                </properties>
+            </writer>
+            <checkpoint-algorithm ref="com.ibm.fhir.bulkdata.jbatch.export.fast.checkpoint.ResourceExportCheckpointAlgorithm">
```
lmsurpre, line 39: The class ref doesn't match its updated package name.
```suggestion
            <checkpoint-algorithm ref="com.ibm.fhir.bulkdata.export.fast.checkpoint.ResourceExportCheckpointAlgorithm">
```

lmsurpre:
One thing I'm not sure about is requiring the client to pass us `X-FHIR-BULKDATA-PROVIDER` to select between the various configurations.  Is it properly tenant-enanbled now so we can at least have a different `default` for each tenant?

prb112:
> One thing I'm not sure about is requiring the client to pass us `X-FHIR-BULKDATA-PROVIDER` to select between the various configurations.
> Because import and export storageProviders aren't distinct, won't this prevent us from using a different default storageProvider for import than we use for export?
> Is it properly tenant-enanbled now so we can at least have a different `default` for each tenant?

by default it is fully tenant aware.  Yes.  the import and  export share the same default.  This is intentional.  If a client wants to split them out they can specify.  Else, we could provide TWO configurations which  are duplicate for the same purposes.  I think this is the best option.  The second best option I  considered was using "export" and "import" for  the default names, once  again,  we duplicate the configuration.  

However, we could override the defaults for the system.  (slight addition where we point to the default in the configuration. 

Let me know, I can easily add it.

fhir-bulkdata-webapp/src/main/resources/META-INF/batch-jobs/FhirBulkExportFastJob.xml
prb112: merged locally

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/context/BatchContextAdapter.java
```diff
@@ -0,0 +1,189 @@
...
+        format(ctx);
+        return ctx;
+    }
+
+    public BulkDataContext getStepContextForFastResourceWriter() {
+        BulkDataContext ctx = new BulkDataContext();
+        context(ctx);
+        source(ctx);
+        audit(ctx);
+        search(ctx);
```
lmsurpre, line 104: I don't know if we need to set the search-related parameters or not (it doesn't support _typefilter), but we *definitely* need to set the systemWrite ones so that we get our cosPathPrefix.
```suggestion
        systemWrite(ctx);
        search(ctx);
```

lmsurpre: I'm also confused about why cosPathPrefix is associated with system export...shouldn't we be setting that for *all* exports?  Did you do something different to set it in the Group and Patient exports?

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/impl/AbstractSystemConfigurationImpl.java
```diff
@@ -0,0 +1,283 @@
...
+
+    @Override
+    public int getCoreCosMaxSize() {
+        return coreCosMaxSize;
+    }
+
+    private static final int defaultCoreCosMaxSize() {
+        final String PATH = "fhirServer/bulkdata/core/cos/max-size";
+        return FHIRConfigHelper.getIntProperty(PATH, DEFAULT_COSFILE_MAX_SIZE);
+    }
```
lmsurpre, line 136: I'm thinking that we should probably make this a `long`.
The default fits safely into an int, but--if we keep this in bytes--then its not too far off from Integer.MAX_VALUE:
`209715200`  (our default; 200MiB)
`2147483647` (Integer.MAX_VALUE; 2048 MiB)

Unfortunately, our fhir-config stuff doesn't currently support Long properties, so we'd either need to add that support or decide to represent the value in KB or MB (or MiB) in our config.

lmsurpre: I think my vote would be to just specify it in MiB in our config.  I don't think a max-size of <1 MiB makes any sense, and I don't think there's much value in allowing more granular values.

lmsurpre:
> However, we could override the defaults for the system. (slight addition where we point to the default in the configuration.
> Let me know, I can easily add it.

I opened https://github.com/IBM/FHIR/issues/2036 for it.

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/impl/AbstractSystemConfigurationImpl.java
prb112: I've changed to MIB in config

prb112: It's in the next commit

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/export/system/resource/SystemExportResourceHandler.java
```diff
@@ -0,0 +1,90 @@
...
+                    logger.log(Level.WARNING, "fillChunkDataBuffer: Error while writing resources with no id", e);
+                }
+            } catch (IOException e) {
+                logger.warning("fillChunkDataBuffer: chunkDataBuffer written error!");
+                throw e;
+            }
+        }
+        chunkData.setCurrentUploadResourceNum(chunkData.getCurrentUploadResourceNum() + resSubTotal);
+        chunkData.addCurrentUploadSize(chunkData.getBufferStream().size());
+        chunkData.setTotalResourcesNum(chunkData.getTotalResourcesNum() + resSubTotal);
```
lmsurpre, line 84: could use the addX flavor for all three (for consistency):
```suggestion
        chunkData.addCurrentUploadResourceNum(resSubTotal);
        chunkData.addCurrentUploadSize(chunkData.getBufferStream().size());
        chunkData.addTotalResourcesNum(resSubTotal);
```

lmsurpre, line 84: could use the addX flavor for all three (for consistency):
```suggestion
        chunkData.addCurrentUploadResourceNum(resSubTotal);
        chunkData.addCurrentUploadSize(chunkData.getBufferStream().size());
        chunkData.addTotalResourcesNum(resSubTotal);
```

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/fast/ResourcePayloadReader.java
```diff
@@ -491,9 +384,12 @@ public Boolean processPayload(ResourcePayload t) {
...
                 }
 
-                // Check if this data would cause us to exceed the max object size. If
+                // Check if this data would cause us to exceed max object size. If
                 // so, close the current upload and start a new one
-                if (this.uploadId != null && (currentObjectSize + this.outputStream.size() > maxObjectSize
+                //
+                // @implNote outputStream.size() is not set for the current resource yet set at this point of the current Resource, comparing it
+                //  can lead to unintended results with double counting.
+                if (this.uploadId != null && (currentObjectSize > maxObjectSize
```
prb112, line 392: Here is the point of discussion

prb112: In prior it was comparing against an already populated outputStream.size()
note the outputstream is not accumulated until line #405

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/impl/AbstractSystemConfigurationImpl.java
```diff
@@ -152,7 +152,7 @@ public long getCoreCosMaxResources() {
 
     private static final long defaultCoreCosMaxResources() {
         final String PATH = "fhirServer/bulkdata/core/cos/maxResourcesThreshold";
-        return 1000 * 1000 * FHIRConfigHelper.getIntProperty(PATH, DEFAULT_COSFILE_MAX_RESOURCESNUMBER);
+        return 1000 * FHIRConfigHelper.getIntProperty(PATH, DEFAULT_COSFILE_MAX_RESOURCESNUMBER);
```
lmsurpre, line 155: do we need to use `1000L` to force it to autobox to a long?  or just to make it more clear?

prb112: sure

fhir-server/liberty-config/config/default/fhir-server-config-db2.json
```diff
@@ -117,30 +117,81 @@
...
+                    "enableParquet": false,
+                    "disableBaseUrlValidation": true,
+                    "exportPublic": true,
+                    "disableOperationOutcomes": true,
+                    "duplicationCheck": false, 
+                    "validateResources": false
+                },
+                "minio" : {
+                    "type": "aws-s3",
+                    "bucket-name": "fhirbulkdata",
```
lmsurpre, line 176:
```suggestion
                    "bucketName": "fhirbulkdata",
```

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/impl/V2ConfigurationImpl.java
```diff
@@ -0,0 +1,178 @@
...
+        List<String> results = FHIRConfigHelper.getStringListProperty(PATH);
+        if (results == null) {
+            results = Collections.emptyList();
+        }
+        return results;
+    }
+
+    @Override
+    public String getStorageProviderBucketName(String provider) {
+        return FHIRConfigHelper.getStringProperty("fhirServer/bulkdata/storageProviders/" + provider + "/bucketBame", null);
```
lmsurpre, line 66:
```suggestion
        return FHIRConfigHelper.getStringProperty("fhirServer/bulkdata/storageProviders/" + provider + "/bucketName", null);
```

operation/fhir-operation-bulkdata/src/test/resources/config/default/fhir-server-config.json
```diff
@@ -0,0 +1,42 @@
...
+                "_comment": "max of 1000",
+                "batchIdEncryptionKey": "change-password",
+                "maxPartitions": 3,
+                "maxInputs": 5,
+                "systemExportImpl": "none"
+            },
+            "storageProviders": {
+                "default" : {
+                    "type": "ibm-cos",
+                    "validBaseRrls": [],
```
lmsurpre, line 32:
```suggestion
                    "validBaseUrls": [],
```

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
```diff
@@ -0,0 +1,441 @@
...
+            switch (mediaType) {
+            case FHIRMediaType.APPLICATION_PARQUET:
+                if (chunkData.getPageNum() > chunkData.getLastPageNum() || chunkData.isFinishCurrentUpload()) {
+                    pushFhirParquetToCos(dto.getResources());
+                    chunkData.setLastWritePageNum(chunkData.getPageNum());
+                }
+                break;
+            case FHIRMediaType.APPLICATION_NDJSON:
+            default:
+                if (chunkData.getPageNum() > chunkData.getLastPageNum() || chunkData.isFinishCurrentUpload()) {
```
lmsurpre, line 329: Looks like we accidentally lost our "do we have enough bytes to start a partUpload" check from here.
I believe what we wanted to do is remove the `chunkData.getPageNum() > chunkData.getLastPageNum()` check from here, but *leave* the `chunkData.getBufferStream().size() > COS_PART_MINIMALSIZE` one
(see https://github.com/IBM/FHIR/pull/1999/commits/61e8e40f7944aff4d8083f7e89e7b2c23b54e2e2#diff-83b573e778b2ec67186ea08aa4515c6d7e6b3583c592abf3fc44cbb3240a7f06L343).

PR #2003, opened 2021-03-01
issue 1955 use resource subset for all operations in schema tool

punktilious:

### DIFF
--- a/fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
+++ b/fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
@@ -64,7 +64,7 @@ import com.ibm.fhir.database.utils.transaction.SimpleTransactionProvider;
 import com.ibm.fhir.database.utils.transaction.TransactionFactory;
 import com.ibm.fhir.database.utils.version.CreateVersionHistory;
 import com.ibm.fhir.database.utils.version.VersionHistoryService;
-import com.ibm.fhir.model.type.code.FHIRResourceType;
+import com.ibm.fhir.model.util.ModelSupport;
 import com.ibm.fhir.schema.app.util.TenantKeyFileUtil;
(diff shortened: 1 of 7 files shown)


fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
```diff
@@ -1531,6 +1532,23 @@ protected void applyDataMigrationForV0010() {
...
     }
 
+    /**
+     * Get the list of resource types to drive resource-by-resource operations
+     * @return the full list of FHIR R4 resource types, or a subset of names if so configured
+     */
+    private Set<String> getResourceTypes() {
+        Set<String> result;
+        if (this.resourceTypeSubset == null || this.resourceTypeSubset.isEmpty()) {
+            result = Arrays.stream(FHIRResourceType.ValueSet.values())
```
lmsurpre, line 1542: this is going to get your abstract resource types too.
what I did in FhirSchemaGenerator was this:
```
    // TODO pass 'false' to getResourceTypes to avoid building tables for abstract resource types
    private static final Set<String> ALL_RESOURCE_TYPES = ModelSupport.getResourceTypes(true).stream()
            .map(t -> ModelSupport.getTypeName(t).toUpperCase())
            .collect(Collectors.toSet());
```

also, we should maybe remove that from there since you moved the processing up to Main

lmsurpre: since you removed the dependency on the RESOURCE_X_VALUES tables, we could probably get away with flipping the switch on that now if we want...though maybe we've jammed enough changes in for 4.6.0 already :-)

punktilious: Too risky right now. I've switched to using ModelSupport, but still using true to get all types. Added a comment to mention the desire to refactor including FhirSchemaGenerator.

PR #2005, opened 2021-03-02
issue-1993 use better default value for PostgreSQL query tuning options

punktilious:


lmsurpre (approved):
LGTM

PR #2008, opened 2021-03-02
Clarify that --update-schema is used for both initial deploy and migration

lmsurpre:
Also, use upper case for the postgresql schema names since the tool now upper-cases them anyway (minimize confusion)


punktilious (approved):
LGTM

PR #2010, opened 2021-03-02
share derby-bootstrap script for bootstrapping derby dbs

lmsurpre:
for preparing a non-docker local fhir-server-test in dev (e.g. for our local servers in eclipse)


build/derby-bootstrap.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
prb112, line 1:
```suggestion
#!/usr/bin/env bash
```

lmsurpre: done

prb112 (approved):

PR #2015, opened 2021-03-03
Use two COPY commands to be more selective

lmsurpre:
Previously we tried to get clever and copy the whole /opt directory,
but that really bloated our image. This change restores us to a more
reasonable size.


tbieste (approved):
LGTM

prb112 (approved):

Issue #1955, opened 2021-02-17
Support resource change tracking with custom operation to fetch list of changes since a known point


punktilious:
**Is your feature request related to a problem? Please describe.**
Systems need a way to reliably and efficiently track changes made to resources in the IBM FHIR Server.

**Describe the solution you'd like**
Add a change tracking table to the schema and insert records to the table each time a resource is changed. Record the Create/Update/Delete status and order the data with a timestamp and change sequence number. The change sequence number can be derived from the RESOURCE_ID because this comes from FHIR_SEQUENCE and will be unique across all resource types.

```
CREATE TABLE <fhir-data-schema>.RESOURCE_CHANGE_LOG (
  change_tstamp    TIMESTAMP NOT NULL,
  resource_id          BIGINT NOT NULL,
  change_type       CHAR(1) NOT NULL,
  resource_type_id INT NOT NULL,
  logical_id            VARCHAR(255) NOT NULL,
  version_id           INT NOT NULL,
  CONSTRAINT pk_resource_change_log PRIMARY KEY (change_tstamp, resource_id)
);

-- add FK to resource-types
```

The change-type is encoded as:

```
C - Resource CREATE
U - Resource UPDATE
D - Resource DELETE
```

The persistence stored procedure can be used to insert the change log record. This avoids another database round-trip. For Derby, the DAO will need to be updated.

A new custom operation `$changes` will be added to fetch the change log records from a known point. The custom operation supports the following parameters:

```
  limit - the number of changes to return in one call (default 1000)
  tstamp -  if given, return changes that have occurred on or after this time
  change_id - return changes with a change_id value greater than the given value
```

If no parameters are provided or the values are null, the scan starts from the first recorded change.

Take note of the inequality semantics relating to tstamp and change_id. These semantics allow a consumer of the API to efficiently retrieve changes that have occurred from a known point in time, but also to avoid receiving duplicates by providing a change_id if it is known. A consumer would implement the following loop (pseudo-code):

```
Date timestamp = chooseStartFrom();
Long changeId = null; // unless this is known already from an earlier checkpoint
do {
  // make the call the $changes custom operation
  List<ChangeRecord> changes = getChanges(timestamp, changeId);
  if (changes.size() > 0) {
    process(changes);

    // update the parameters so the next $changes call starts after the last record
    ChangeRecord last = changes.get(changes.size()-1);
    timestamp = last.getTimestamp();
    changeId = last.getChangeId();
} while (changeId != null);
```

[...]

PR #2022, opened 2021-03-04
issue #1955 - return http 400 error codes for invalid requests

lmsurpre:


punktilious (approved):
LGTM

PR #2030, opened 2021-03-06
issue 2029 - check PK constraint and ROW_ID exist before attempting t…

punktilious:
…o remove

### DIFF
--- a/docs/src/pages/Conformance.md
+++ b/docs/src/pages/Conformance.md
@@ -49,8 +49,8 @@ The `_format` parameter is supported and provides a useful mechanism for request
 
 The `_pretty` parameter is also supported.
 
-## System-level history
-The system-level history interaction can be used to obtain a list of changes (create, update, delete) to resources in the IBM FHIR Server. This may be useful for other systems to reliably track these changes and keep themselves in-sync.
+## Whole System History
+The whole system history interaction can be used to obtain a list of changes (create, update, delete) to resources in the IBM FHIR Server. This may be useful for other systems to reliably track these changes and keep themselves in-sync.
 
 ```
     curl -k -u '<username>:<password>' 'https://<host>:<port>/fhir-server/api/v4/_history'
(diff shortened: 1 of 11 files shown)


fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/common/DropColumn.java
```diff
@@ -7,49 +7,121 @@
 package com.ibm.fhir.database.utils.common;
 
 import java.sql.Connection;
+import java.sql.PreparedStatement;
```
prb112, line 10: 2021

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirResourceTableGroup.java
```diff
@@ -403,8 +403,8 @@ private void addCompositeMigrationStepsV0009(List<IDatabaseStatement> statements
         // Simplified composite value support. Eliminate the composites table and instead
         // use a new composite_id column to correlate parameters involved in a composite relationship
         // drop PK constraint; remove ROW_ID column after data migration
-        statements.add(new DropPrimaryKey(schemaName, tableName));
-        statements.add(new DropColumn(schemaName, tableName, ROW_ID));
+        statements.add(new DropPrimaryKey(schemaName, tableName, true));
+        statements.add(new DropColumn(schemaName, tableName, true, ROW_ID));
```
prb112, line 407: so this ignores any errors?

punktilious: Yes. Ideally we'd make it idempotent for all database types, but to do this properly means we should refactor Migration so that it's based on commands and adapters instead of raw statements. I'll create an issue.

punktilious: issue #2031

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/common/DropColumn.java
lmsurpre: looks to be addressed

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirSchemaVersion.java
```diff
@@ -26,6 +26,7 @@
     ,V0009(9, "issue-1683 refactor composite values")
     ,V0010(10, "issue-1958 add IS_DELETED flag to each xxx_LOGICAL_RESOURCES table")
     ,V0011(11, "issue-2011 add LAST_UPDATED to each xxx_LOGICAL_RESOURCES table")
+    ,V0012(12, "issue-2032 change column order for index on RESOURCE_CHANGE_LOG")
```
lmsurpre, line 29: Do we need to set the version of the RESOURCE_CHANGE_LOG table to v12 now?  Is that change included in this PR?

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -3282,7 +3295,17 @@ public Bundle doHistory(MultivaluedMap<String, String> queryParameters,
             nextRequest.append(serviceBase);
             nextRequest.append("?");
             nextRequest.append("_count=").append(count);
-            nextRequest.append("&_afterHistoryId=").append(lastChangeId);
+
+            if (historyContext.getSince() != null && historyContext.getSince().getValue() != null) {
+                // As _since was given, we need to go with time-based paging, and the client
+                // will be responsible for managing duplicates.
+                nextRequest.append("&_since=").append(lastChangeTime.atZone(UTC).format(DateTime.PARSER_FORMATTER));
```
lmsurpre, line 3302: not sure why we can't switch them from time-tased to resourceId-based paging, but I guess its probably fine.  in practice, i think it means that most people will just use the time-based paging.

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirSchemaVersion.java
lmsurpre: Maybe this change snuck in here but should have been saved for a future PR that addresses this issue?

punktilious: Good catch. I backed out the change after realizing the issue wasn't with the index column order, but with the sort columns in the query. I'll remove V0012 because it's not used.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
punktilious: I'll try to explain in the doc update, but there's an issue with concurrency making switching problematic. I imagine that downstream systems will use the afterHistoryId, real people will use since.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/common/CreateIndexStatement.java
```diff
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright IBM Corp. 2020
+ * (C) Copyright IBM Corp. 202, 2021
```
lmsurpre, line 2:
```suggestion
 * (C) Copyright IBM Corp. 2020, 2021
```

lmsurpre (approved):
LGTM

docs/src/pages/Conformance.md
```diff
@@ -69,17 +69,30 @@ By default, the returned bundle will contain up to 100 resource references. The
...
+
+To guarantee no data is skipped, clients should not process resources with a `lastModified` timestamp which is after `{current_time} - {transaction_timeout} - {max_cluster_clock_drift}`. By waiting for this time window to close, the client can be sure the data being returned is complete and in order, and new data will not appear. The default value for transaction timeout is 120s but this is configurable. A value of 2 seconds is a reasonable default to consider for `max_cluster_clock_drift` in lieu of specific information. Implementers should check with server administrators on the appropriate values to use.
```
lmsurpre, line 95: > clients should not process resources with a `lastModified` timestamp which is after `{current_time} - {transaction_timeout} - {max_cluster_clock_drift}`

can we be more specific with this recommendation.  how would they accomplish this?  just stop resolving the 'next' link once they get to a resource on this window, then re-issue the same query until the window is passed?

lmsurpre: that seems a bit complicated...maybe we could introduce a query parameter which will cause the server to only return resources with a lastModified instant that is less than `{current_time} - {transaction_timeout} - {max_cluster_clock_drift}`?

punktilious: Raised #2039 for this.

Issue #2040, opened 2021-03-09
Fast export implementation fails to close GZIPInputStream


lmsurpre:
**Describe the bug**
In our kubernetes deployment, our pod kept getting killed for going over its memory limit.  However, this was very strange because we have the max heap size set at 4GB.

**To Reproduce**
Do multiple large exports at the same time

**Expected behavior**
Export may slow down with load, but pods shouldn't be leaking memory like this.

**Additional context**
Looks like we're failing to close the GZipInputStream that is used in this fast export impl; GZipInputStream uses Inflater internally and native zlib is used and only freed on an explicit call to close.

lmsurpre:
After finding and fixing the leak, I've launched numerous exports in our test environment and the pods are properly staying within their limits.

PR #2042, opened 2021-03-09
issue #2040 - close input streams in finally blocks

lmsurpre:
The main one was in ResourcePayload.transferTo because, without this, we
never free the non-heap memory being held by the GZIPInputStream.


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -1751,9 +1752,12 @@ protected UserTransaction retrieveUserTransaction(String jndiName) {
                 } else {
                     resource = FHIRParser.parser(Format.JSON).parse(in);
                 }
-                in.close();
             }
         } finally {
+            if (in != null) {
+                in.close();
```
lmsurpre, line 1758: this is the key change

fhir-persistence/src/main/java/com/ibm/fhir/persistence/ResourcePayload.java
```diff
@@ -50,18 +50,24 @@ public long getResourceId() {
...
+                }
             }
-        }
 
-        // how many bytes did we transfer
-        return result;
+            // how many bytes did we transfer
+            return result;
+        } finally {
+            decompressedPayload.close();
```
lmsurpre, line 69: this is the key change

fhir-server/liberty-config/config/default/fhir-server-config-db2.json
```diff
@@ -128,7 +128,7 @@
                     "trustAll": true
                 },
                 "cos" : { 
-                    "maxResourcesThreshold": 200,
+                    "maxResourcesThreshold": 200000,
```
lmsurpre, line 131: @prb112 Robin and I discussed this one and we think measuring this in "thousands of resources" was a little odd.  we think it should be safe to stay an integer because we shouldn't have more than 2 billion resources in a single export.
side-note:  I think there is another spot in our persistence layer with a similar limitation.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/common/BulkDataUtils.java
```diff
@@ -112,6 +112,8 @@ public static PartETag multiPartUpload(AmazonS3 cosClient, String bucketName, St
             log("multiPartUpload", "Upload part Error - " + sdke.getMessage());
             cosClient.abortMultipartUpload(new AbortMultipartUploadRequest(bucketName, itemName, uploadID));
             throw sdke;
+        } finally {
+            dataStream.close();
```
punktilious, line 116: This feels like a side-effect to me...I don't think the method should close the stream...that should be the responsibility of the caller.

JohnTimm (commented):
Should we use try-with-resources instead? Also, should there always be a null check in the finally block?

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/common/BulkDataUtils.java
prb112: I agree with @punktilious

lmsurpre: @punktilious haha, thats what I said while we were on the call :-)
will create a separate close method and do it there

lmsurpre: ok, I'll take this back out.  presently we're always calling this method with a ByteArrayInputStream and so the close doesn't do anything either

lmsurpre:
> Should we use try-with-resources instead?

I do like that pattern, but in this case it was easier just to slap a finally on the exiting try blocks.

> Also, should there always be a null check in the finally block?

probably, although I reverted the one from BulkDataUtils based on the other feedback, so I don't think there's anything to change there now.

JohnTimm (approved):
LGTM

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
```diff
@@ -330,23 +331,22 @@ public void close() throws Exception {
...
+        case FHIRMediaType.APPLICATION_NDJSON:
+        default:
+            // Only if we're greater than zero, otherwise there is nothing to upload.
+            if (chunkData.getBufferStream().size() > 0) {
+                // TODO try PipedOutputStream -> PipedInputStream instead?  Or maybe a ByteBuffer with a flip instead?
+                pushFhirJsonsToCos(new ByteArrayInputStream(chunkData.getBufferStream().toByteArray()), chunkData.getBufferStream().size());
             }
+            break;
         }
+        chunkData.setLastWritePageNum(chunkData.getPageNum());
```
lmsurpre, line 349: this part of the change-set is more experimental in nature and should probably have been a separate PR

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
```diff
@@ -378,39 +378,36 @@ private void pushFhirJsonsToCos(InputStream in, int dataLength) throws Exception
...
-                    chunkData.setResourceTypeSummary(fhirResourceType + "[" + chunkData.getCurrentUploadResourceNum());
-                    if (chunkData.getPageNum() > chunkData.getLastPageNum()) {
-                        chunkData.setResourceTypeSummary(chunkData.getResourceTypeSummary() + "]");
-                    }
-                } else {
-                    chunkData.setResourceTypeSummary(chunkData.getResourceTypeSummary() + "," + chunkData.getCurrentUploadResourceNum());
-                    if (chunkData.getPageNum() > chunkData.getLastPageNum()) {
-                        chunkData.setResourceTypeSummary(chunkData.getResourceTypeSummary() + "]");
-                    }
+        chunkData.getCosDataPacks().add(BulkDataUtils.multiPartUpload(client, bucketName, itemName, uploadId, in, dataLength, chunkData.getPartNum()));
```
lmsurpre, line 381: looks like a big change, but I just removed the outer if...part of the "experimental" change mentioned above

lmsurpre: oh, and removed the extra check on line 387 since we're no longer looping through the DTOs and re-setting the page in the middle of the writeResources

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
lmsurpre: ok, I backed this part out

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
lmsurpre: backed these out

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -1738,9 +1738,10 @@ protected UserTransaction retrieveUserTransaction(String jndiName) {
         final String METHODNAME = "convertResourceDTO";
         log.entering(CLASSNAME, METHODNAME);
         T resource = null;
+        InputStream in = null;
         try {
```
prb112, line 1742: Why not a try with resources?

fhir-server/liberty-config/config/default/fhir-server-config.json
```diff
@@ -142,16 +142,16 @@
...
+                    "objectSizeThresholdMB": 200,
+                    "objectResourceCountThreshold": 200000,
                     "useServerTruststore": true
                 },
                 "pageSize": 100,
                 "batchIdEncryptionKey": "change-password",
                 "maxPartitions": 3, 
                 "maxInputs": 5,
-                "systemExportImpl": "none"
+                "systemExportImpl": "legacy"
```
prb112, line 154:
```suggestion
                "systemExportImpl": "fast"
```

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/impl/LegacyConfigurationImpl.java
```diff
@@ -49,15 +49,15 @@ public String getCoreApiBatchTrustStorePassword() {
...
-        return FHIRConfigHelper.getIntProperty(PATH, DEFAULT_COSFILE_MAX_RESOURCESNUMBER);
+        return FHIRConfigHelper.getIntProperty(PATH, DEFAULT_COS_OBJ_MAX_RESOURCE_COUNT);
     }
 
     @Override
-    public long getCoreCosThresholdSize() {
+    public long getCoreCosObjectSizeThreshold() {
         final String PATH = "fhirServer/bulkdata/cosFileMaxSize";
-        return FHIRConfigHelper.getIntProperty(PATH, DEFAULT_COSFILE_MAX_SIZE);
+        return FHIRConfigHelper.getIntProperty(PATH, DEFAULT_COS_OBJ_MAX_SIZE_MB);
```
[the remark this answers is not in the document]
lmsurpre: I just changed their names to match the newly proposed config names...happy to discuss the naming of those config parameters since this is our one chance to get it right.  The constant names obviously matter a lot less...just tried to keep them fairly consistent.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
lmsurpre: see https://github.com/IBM/FHIR/pull/2042#issuecomment-794759052

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/impl/LegacyConfigurationImpl.java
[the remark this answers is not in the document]
prb112: right, what I'm saying is by changing to 1000s we've actually changed the meaning of legacy.  
it's not what you did, it's that we really don't support legacy implementations out of the box.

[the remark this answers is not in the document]
lmsurpre: ah, ok.  I was wondering about that...did we even have these properties before?  I don't think I ever set them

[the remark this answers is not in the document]
prb112: We did have them before... 


fhirServer/bulkdata/cosFileMaxResources | int | The maximum number of FHIR resources per COS file, “-1” means no limit, the default value is 200000
-- | -- | --
fhirServer/bulkdata/cosFileMaxSize | int | The maximum COS file size in bytes, “-1” means no limit, the default value is 209715200 (200M)

punktilious (approved):
LGTM

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/impl/LegacyConfigurationImpl.java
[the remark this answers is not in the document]
lmsurpre: ok, confirmed we did have this before.  how about we just do this:
`return FHIRConfigHelper.getIntProperty(PATH, DEFAULT_COS_OBJ_MAX_SIZE_MB * 1024 * 1024);`

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/load/data/ImportTransientUserData.java
```diff
@@ -9,6 +9,9 @@
 import java.io.BufferedReader;
 import java.io.ByteArrayOutputStream;
 import java.io.InputStream;
+import java.util.List;
+
+import com.ibm.cloud.objectstorage.services.s3.model.PartETag;
```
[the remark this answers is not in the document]
lmsurpre: In the builder:
```   @Override
        public Builder dataPacksForOperationOutcomes(List<PartETag> dataPacksForOperationOutcomes) {
            return (Builder) super.dataPacksForOperationOutcomes(dataPacksForOperationOutcomes);
        }```
        
Previously this was in the parent builder only, but that meant that the ImportTransientUserData builder methods actually return the parent builder (forcing ugly casts from the builder's user).

prb112 (approved):
LGTM, waiting on the fix in #2043

PR #2043, opened 2021-03-10
Fix for Double Negation Problem

prb112:


operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/impl/V2ConfigurationImpl.java
```diff
@@ -1,5 +1,4 @@
 /*
- * (C) Copyright IBM Corp. 2021
```
lmsurpre, line 2: accidental?

prb112: not sure why

PR #2044, opened 2021-03-10
issue #2040 - cleanup in S3Provider.writeResources

lmsurpre:
1. Previously we were looping through all of the ReadResultDTOs
(presumably 1 per page of search results), and entering our logic for
each page.  However, the chunkData buffer stream constains the content
from all the results, so this doesn't make much sense to me.

2. Renamed lastWritePageNum to lastWrittenPageNum to make its purpose
more clear (more distinct from lastPageNum)

3. Update ExportOperationTest to properly look for an 'output' key in
the response json, not just anywhere in the response body

4. Add logging to ExportCheckpointAlgorithm

5. ExportTransientUserData.Builder methods now return
ExportTransientUserData instead of the super type (same change was made
to ImportTransientUserData previously)

6. Page number now matches the actual page number read; previously it
was always 1 higher

7. Set the checkpoint-policy of Group and Patient export to 'custom' to
indicate that the ExportCheckpointAlgorithm will be used to control
checkpointing

### DIFF
--- a/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/export/group/resource/GroupHandler.java
+++ b/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/export/group/resource/GroupHandler.java
@@ -62,17 +62,11 @@ public class GroupHandler {
     }
 
     /**
-     * get page of members
+     * resolve the groupId into a set of patients and add them to this handler
      *
-     * @param pageNum
-     * @param pageSize
-     * @return
+     * @param groupId
+     * @throws Exception
      */
-    public List<Member> getPageOfMembers(int pageNum, int pageSize){
-        return patientMembers.subList((pageNum - 1) * pageSize,
-            pageNum * pageSize <= patientMembers.size() ? pageNum * pageSize : patientMembers.size());
-    }
-
     public void process(String groupId) throws Exception {
         if (patientMembers == null) {
             Group group = findGroupByID(groupId);
@@ -83,10 +77,20 @@ public class GroupHandler {
         }
     }
 
-    /*
+    /**
+     * get a page of members from this handler
+     *
+     * @param pageNum
+     * @param pageSize
+     * @return
+     * @implNote {@link GroupHandler#process(String)} must be called first
+     */
+    public List<Member> getPageOfMembers(int pageNum, int pageSize){
+        return patientMembers.subList((pageNum - 1) * pageSize, Math.min(pageNum * pageSize, patientMembers.size()));
+    }
+
+    /**
      * recursively expands a group into a set of members
-     * @param fhirTenant
-     * @param fhirDatastoreId
      * @param group
      * @param patients
      * @param groupsInPath empty, or prior Groups scanned
--- a/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
+++ b/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
@@ -12,6 +12,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 
 import org.apache.http.conn.ssl.SSLConnectionSocketFactory;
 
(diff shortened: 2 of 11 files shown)


fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
```diff
@@ -379,39 +378,36 @@ private void pushFhirJsonsToCos(InputStream in, int dataLength) throws Exception
...
-                    chunkData.setResourceTypeSummary(fhirResourceType + "[" + chunkData.getCurrentUploadResourceNum());
-                    if (chunkData.getPageNum() > chunkData.getLastPageNum()) {
-                        chunkData.setResourceTypeSummary(chunkData.getResourceTypeSummary() + "]");
-                    }
-                } else {
-                    chunkData.setResourceTypeSummary(chunkData.getResourceTypeSummary() + "," + chunkData.getCurrentUploadResourceNum());
-                    if (chunkData.getPageNum() > chunkData.getLastPageNum()) {
-                        chunkData.setResourceTypeSummary(chunkData.getResourceTypeSummary() + "]");
-                    }
+        chunkData.getCosDataPacks().add(BulkDataUtils.multiPartUpload(client, bucketName, itemName, uploadId, in, dataLength, chunkData.getPartNum()));
```
lmsurpre, line 381: this change looks big, but i just removed the outer if and removed a single redundant check from line 387

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
```diff
@@ -330,25 +331,22 @@ public void close() throws Exception {
...
+        switch (mediaType) {
+        case FHIRMediaType.APPLICATION_PARQUET:
+            pushFhirParquetToCos(dtos.stream()
+                    .flatMap(dto -> dto.getResources().stream())
+                    .collect(Collectors.toList()));
+            break;
+        case FHIRMediaType.APPLICATION_NDJSON:
+        default:
+            // Only if we're greater than zero, otherwise there is nothing to upload.
+            if (chunkData.getBufferStream().size() > 0) {
```
prb112, line 343: you forgot about OperationOutcomes... we should leave the check here... OperationOutcomes should use this going forward in the future.

prb112: https://github.com/IBM/FHIR/issues/2046

fhir-server-test/src/test/java/com/ibm/fhir/server/test/bulkdata/ExportOperationTest.java
```diff
@@ -824,7 +836,8 @@ private void checkGroupExportStatus(boolean s3) throws Exception {
         // Export finished successfully, we should be about to find the "output" part in the message body
         // which includes all the COS objects download urls.
         String body = response.readEntity(String.class);
-        assertTrue(body.contains("output"));
+        JsonObject jsonObject = Json.createReaderFactory(null).createReader(new StringReader(body)).readObject();
```
prb112, line 839: I think this is fine, however there is already a factory in place on 111

lmsurpre: updated to use the factory constant

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/export/group/resource/GroupHandler.java
```diff
@@ -83,10 +77,20 @@ public void process(String groupId) throws Exception {
...
     }
 
-    /*
+    /**
+     * get a page of members from this handler
+     *
+     * @param pageNum
+     * @param pageSize
+     * @return
+     * @implNote {@link GroupHandler#process(String)} must be called first
```
lmsurpre, line 86: I think the design of this could be improved, but at this point I just tried making it clear that GroupHander is stateful and that process needs to be called before you can page through the members

prb112: Yes,  I tried to split  the prior implementation  so it can be further refactored.

prb112 (approved):
LGTM

Issue #2059, opened 2021-03-11
References in Patient Export are not honored


prb112:
**Describe the bug**
References in Patient Export are not honored

**To Reproduce**
References in Patient Export are not honored

**Expected behavior**

**Additional context**
Related to  Patient Export Results in duplicate data included in the Write. #2058

d0roppe:
verified this is now working as expected on the export

Issue #2058, opened 2021-03-11
Patient Export Results in duplicate data included in the Write.


prb112:
**Describe the bug**
Patient Export Results in duplicate data included in the Write.

**To Reproduce**
1. Execute integration test for ExportOperationTest
2. Note the duplicate Patient data the top. 

**Expected behavior**
From https://hl7.org/fhir/uv/bulkdata/export/index.html#response---complete-status:
Each file SHALL contain resources of only one type, but a server MAY create more than one file for each resource type returned.

Our intent is to include only the Resources in the file. Except Attachments and DocumentReferences.

**Additional context**
n/a

lmsurpre:
I wonder if its worth opening a separate issue for generating a better query for this one.
Specifically, we now have a special internal "compartment" search parameter that we should be able to leverage so that we don't need to aggregate all the different compartment inclusion criteria parameters separately.

For example, we might be able to get rid of this code in `PatientResourceHandler.fillChunkDataBuffer(List<String> patientIds, ReadResultDTO dto)`:
```
            List<String> compartmentSearchCriterias = CompartmentUtil.getCompartmentResourceTypeInclusionCriteria("Patient", resourceType.getSimpleName());
            if (compartmentSearchCriterias.size() > 1) {
                isDoDuplicationCheck = true;
            }
            for (String compartmentSearchCriteria : compartmentSearchCriterias) {
                HashMap<String, List<String>> queryTmpParameters = new HashMap<>();
                queryTmpParameters.putAll(queryParameters);
                queryTmpParameters.put(compartmentSearchCriteria, Arrays.asList(String.join(",", patientIds)));
                searchContext = SearchUtil.parseQueryParameters(resourceType, queryTmpParameters);
```

One thing that might make this difficult is that, under normal operation, a patient compartment search is only ever scoped to a single patient (whereas here we have a list of patients).  A couple ideas:
A. Perform a different search for each specific patient included in the export?  It would definitely be a lot more searches (i.e. more round trips to the DB), but maybe these simple searches will perform better than one giant OR query like we have now?
B. Add support (if we don't already support it) for having a logical OR for a compartment search parameter.

If we're able to do B, this would also benefit `user` scoped SMART patient access in our AuthzPolicyEnforcementPersistenceInterceptor, for cases where the user has access to a set of patient compartment resources and not just their own.

d0roppe:
verified this is now working correctly on the export.

PR #2060, opened 2021-03-11
Patient Export Results in duplicate data included in the Write #2058

prb112:
- Empty DTOs cause the write to skip
- PatientResourceHandler has confusing logic, and skips the resource
write.

References in Patient Export are not honored #2059

- We extract IDs, and should use the relative reference when searching not the id.


fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/dto/ReadResultDTO.java
```diff
@@ -54,6 +55,10 @@ public int size() {
     }
 
     public boolean empty() {
-        return resources.isEmpty();
+        return resources.isEmpty() || chunk;
+    }
+
+    public void makeChunk() {
```
lmsurpre, line 61: I don't understand what this means or when/why it would be called.  Maybe some javadoc is in order?

fhir-server-test/src/test/java/com/ibm/fhir/server/test/bulkdata/ExportOperationTest.java
```diff
@@ -349,8 +351,21 @@ public CloseableHttpClient getHttpClient() {
                 .build();
     }
 
-    public void verifyFileLines(String workItem, int count) throws IOException {
-        int actual = Files.readAllLines(new File(path + "/" + workItem).toPath()).size();
+    public void verifyFileLines(String workItem, int count, String resourceType) throws IOException {
+        List<String> lines = Files.readAllLines(new File(path + "/" + workItem).toPath());
```
lmsurpre, line 355: I wonder if we could just use Paths.get() instead
```suggestion
        List<String> lines = Files.readAllLines(Paths.get(path + "/" + workItem));
```

fhir-server-test/src/test/java/com/ibm/fhir/server/test/bulkdata/ExportOperationTest.java
```diff
@@ -349,8 +351,21 @@ public CloseableHttpClient getHttpClient() {
...
+            try (ByteArrayInputStream bais = new ByteArrayInputStream(line.getBytes())) {
+                try {
+                    // Checks to see if it's a single type.
+                    com.ibm.fhir.model.resource.Resource r = FHIRParser.parser(Format.JSON).parse(bais);
+                    assertEquals(resourceType, r.getClass().getSimpleName());
+                } catch (FHIRParserException e) {
+                    e.printStackTrace();
+                    Assert.fail();
+                }
+            }
```
lmsurpre, line 366: its just test code, but I think we can have a single try here:
```suggestion
            try (ByteArrayInputStream bais = new ByteArrayInputStream(line.getBytes())) {
                // Checks to see if it's a single type.
                com.ibm.fhir.model.resource.Resource r = FHIRParser.parser(Format.JSON).parse(bais);
                assertEquals(resourceType, r.getClass().getSimpleName());
            } catch (FHIRParserException e) {
                e.printStackTrace();
                Assert.fail();
            }
```

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/dto/ReadResultDTO.java
prb112, line 61: Skips over the ReadResultDTO being processed, and uses the chunk.

prb112, line 61:
```suggestion
    /**
     * Indicates the ChunkData bytestream should be used versus the results from the ReadResultDTO
     */
    public void makeChunk() {
```

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/patient/ChunkReader.java
```diff
@@ -215,19 +217,26 @@ public Object readItem() throws Exception {
...
+            if (resources != null && !resources.isEmpty()) {
                 if (logger.isLoggable(Level.FINE)) {
                     logger.fine("readItem[" + ctx.getPartitionResourceType() + "]: loaded " + dto.size() + " patients");
                 }
 
-                List<String> patientIds = dto.getResources().stream()
+                List<String> patientIds = resources.stream()
                             .filter(item -> item.getId() != null)
-                            .map(item -> item.getId())
+                            .map(item -> "Patient/" + item.getId())
```
lmsurpre, line 228: I'm planning to remove this in https://github.com/IBM/FHIR/pull/2102

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/patient/ChunkReader.java
```diff
@@ -192,6 +192,8 @@ public Object readItem() throws Exception {
                 auditLogger.logSearchOnExport(ctx.getPartitionResourceType(), queryParameters, resources.size(), startTime, endTime, Response.Status.OK, "StorageProvider@" + ctx.getSource(), "BulkDataOperator");
             }
 
+            pageNum++;
```
lmsurpre, line 195: why are we moving the page forward again here?

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/patient/ChunkReader.java
prb112: OK, please go ahead in 2012, I'd rather leave here since it resolves the issue and ensures a regression doesn't happen.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/patient/ChunkReader.java
prb112: We're passed the first page

lmsurpre: pageNum is incremented on line 148 for each call to readItem.  And I think we should only be reading one page per readItem, so I don't understand.

prb112: totally right.. not sure where it came from.. Just tested with a small page size of 100.

lmsurpre (approved):
LGTM

Issue #2072, opened 2021-03-12
Audit log entry for bulk data export should have a resource type associated with the total read in the log entry


d0roppe:
**Is your feature request related to a problem? Please describe.**
When doing a bulk data export with multiple resource types, the Read record entries should have a resource type specified for the count given in the content entry.

**Describe the solution you'd like**
If you export _type=Patient,Observation. the Read audit log entries currently show a count for a type in the list of types, but we should also state what type it is associated with. 

**Describe alternatives you've considered**

**Acceptance Criteria**
1.  
   GIVEN [a precondition]
    AND [another precondition]
   WHEN [test step]
    AND [test step]
   THEN [verification step]
    AND [verification step]

**Additional context**

prb112:
It'll output this... 
{
    "request_unique_id": "58429626-5637-4b91-8023-98a9053c2800",
    "action": "R",
    "start_time": "2021-03-12 01:56:49.439",
    "end_time": "2021-03-12 01:56:49.894",
    "api_parameters": {
        "request": "https://localhost:9443/fhir-server/api/v4/$export?_outputFormat=application/fhir%2Bndjson&_type=Patient",
        "request_status": 200
    },
    "query": "{_lastUpdated=[ge2019-01-01T08:21:26.94-04:00], _sort=[_lastUpdated]}",
    "batch": {
        "resources_read": 256
    },
    "event_type": "fhir-search",
    "description": "FHIR BulkData Search request",
    "location": "@source:default",
    "resource_name": "Patient"
}

d0roppe:
verified the audit record now has the resource type listed with the count read when there are multiple resources in the _type parm. or there are multiple resources exported from the export call.

PR #2074, opened 2021-03-12
Audit log entry for bulk data export should have a resource type

prb112:
associated with the total read in the log entry #2072

### DIFF
--- a/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/audit/BulkAuditLogger.java
+++ b/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/audit/BulkAuditLogger.java
@@ -61,7 +61,7 @@ public class BulkAuditLogger {
     }
 
     private void log(AuditLogEventType eventType, String action, String description, Resource oldResource, Resource newResource, Date startTime, Date endTime,
-        Response.Status responseStatus, String queryString, Long totalSearch, String location, String users) throws Exception {
+            Response.Status responseStatus, String queryString, Long totalSearch, String location, String users) throws Exception {
         log(eventType, action, description, oldResource, newResource, startTime, endTime, responseStatus, queryString, totalSearch, location, users, null);
     }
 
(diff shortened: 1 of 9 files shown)


fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/audit/BulkAuditLogger.java
```diff
@@ -60,27 +60,44 @@ public boolean shouldLog() {
         return svc.isEnabled();
     }
 
+    private void log(AuditLogEventType eventType, String action, String description, Resource oldResource, Resource newResource, Date startTime, Date endTime,
+        Response.Status responseStatus, String queryString, Long totalSearch, String location, String users) throws Exception {
```
lmsurpre, line 64:
```suggestion
            Response.Status responseStatus, String queryString, Long totalSearch, String location, String users) throws Exception {
```

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/load/ChunkWriter.java
```diff
@@ -195,7 +195,7 @@ public void writeItems(List<java.lang.Object> arg0) throws Exception {
                                 if (auditLogger.shouldLog()) {
                                     long endTime = System.currentTimeMillis();
                                     String location = "@source:" + ctx.getSource() + "/" + ctx.getImportPartitionWorkitem();
-                                    auditLogger.logUpdateOnImport(fhirResource, null, new Date(startTime), new Date(endTime), Response.Status.OK, location, "BulkDataOperator");
+                                    auditLogger.logUpdateOnImport(null, fhirResource, new Date(startTime), new Date(endTime), Response.Status.OK, location, "BulkDataOperator");
```
lmsurpre, line 198: Good catch, although its a little funny that this method takes the "oldResource" at all since we don't actually try to read that.  Assuming this is the only place its ever called from, maybe we should change the method signature?

prb112: I wanted to match as much as possible what's in the RestAuditLogger.

lmsurpre (approved):
LGTM

Issue #2077, opened 2021-03-12
Under certain conditions, bulk import fails to acquire a db connection


lmsurpre:
**Describe the bug**

**To Reproduce**
I've seen it twice in a row by roughly following these steps, but I'm not sure if its always repeatable or not.
1. dropin a datasource-postgres.xml similar to the following (in my case it points to an IBM Cloud databases for PostgreSQL instance):
```<server>
    <!-- ============================================================== -->
    <!-- TENANT: default; DSID: default; TYPE: read-write               -->
    <!-- ============================================================== -->
    <dataSource id="fhirDefaultDefault" jndiName="jdbc/fhir_default_default" type="javax.sql.XADataSource" statementCacheSize="200" syncQueryTimeoutWithTransactionTimeout="true">
        <jdbcDriver javax.sql.XADataSource="org.postgresql.xa.PGXADataSource" libraryRef="sharedLibPostgres"/>
        <properties.postgresql
             serverName="${env.FHIR_DB_HOSTNAME}"
             portNumber="${env.FHIR_DB_PORT}"
             ssl="true"
             sslmode="verify-full"
             sslrootcert="/opt/ol/wlp/usr/servers/fhir-server/resources/security/postgres.cert"
             databaseName="${env.FHIR_DB_NAME}"
             user="${env.FHIR_DB_USERNAME}"
             password="${env.FHIR_DB_PASSWORD}"
             currentSchema="${env.FHIR_DB_SCHEMA}"
         />
        <connectionManager maxPoolSize="200" minPoolSize="40"/>
    </dataSource>

    <dataSource id="fhirbatchDS" jndiName="jdbc/fhirbatchDB" type="javax.sql.XADataSource" statementCacheSize="200" syncQueryTimeoutWithTransactionTimeout="true">
        <jdbcDriver javax.sql.XADataSource="org.postgresql.xa.PGXADataSource" libraryRef="sharedLibPostgres"/>
        <properties.postgresql
            serverName="${env.BATCH_DB_HOSTNAME}"
            portNumber="${env.BATCH_DB_PORT}"
            ssl="true"
            sslmode="verify-full"
            sslrootcert="/opt/ol/wlp/usr/servers/fhir-server/resources/security/postgres.cert"
            databaseName="${env.BATCH_DB_NAME}"
            user="${env.BATCH_DB_USER}"
            password="${env.BATCH_DB_PASSWORD}" />
    </dataSource>
</server>
```
[...]

PR #2081, opened 2021-03-12
issue #2077 - add validationTimeout to datasource dropins

lmsurpre:
We used to have this attribute in our fhirProxyDatasource configuration, but we didn't carry it forward.
After adding it to my deploy, I followed the steps to reproduce the issue described in #2077 and was no longer able to produce it.

From the Liberty docs, here is what this setting does:
> Specifies a timeout for validation of pooled connections. When specified, pooled connections are validated before being reused from the connection pool. The validation timeout is also used when the connection manager validates connections in response to a stale connection for PurgePolicy=ValidateAllConnections. A value of 0 means that connections are validated without applying any timeout. Validation timeout requires a JDBC driver that complies with the JDBC 4.0 specification or higher. Specify a positive integer followed by a unit of time, which can be hours (h), minutes (m), or seconds (s). For example, specify 30 seconds as 30s. You can include multiple values in a single entry. For example, 1m30s is equivalent to 90 seconds.

### DIFF
--- a/docs/src/pages/guides/FHIRPerformanceGuide.md
+++ b/docs/src/pages/guides/FHIRPerformanceGuide.md
@@ -114,7 +114,7 @@ See [Managing PostgreSQL Connections](https://cloud.ibm.com/docs/databases-for-p
 The recommended approach for tenant datatstore configuration is to use individual JTA datasources, each with their own connection manager (connection pool):
 
 ```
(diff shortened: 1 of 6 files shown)


prb112 (approved):
LGTM

punktilious (approved):
LGTM

PR #2088, opened 2021-03-15
issue 2049 replace byte buffer streams with more efficient implementation

punktilious:


fhir-persistence/src/main/java/com/ibm/fhir/persistence/util/InputOutputByteStream.java
```diff
@@ -0,0 +1,285 @@
...
+                }
+            }
+            final long newEnd = System.nanoTime();
+
+            double origSeconds = (origEnd - origStart) / 1e9;
+            double newSeconds = (newEnd - newStart) / 1e9;
+
+            System.out.println(String.format("Original = %5.3fs, New = %5.3fs, Percent improvement = %5.1f%%", origSeconds, newSeconds, 100.0 * (origSeconds - newSeconds) / origSeconds));
+        } catch (IOException x) {
+            System.out.println("Test failed: " + x.getMessage());
```
prb112, line 282: let's move this into a test or tool under src/test/java

prb112 (approved):
One comment on the `public static void main(String... args)` should be in a test class

fhir-persistence/src/main/java/com/ibm/fhir/persistence/util/InputOutputByteStream.java
punktilious: OK I'll add a separate micro benchmark class for this in test. Agree this shouldn't be on the main class path.

punktilious: done

Issue #1277, opened 2020-06-24
Swagger Generator RegEx Error


CHFJr:
The Swagger Generator isn't creating the proper RegEx string for base64Binary

Within the profiles-types.json file the RegEx is
  "valueString" : "(\\s*([0-9a-zA-Z\\+/=]){4}\\s*)+"

However, the generated Swagger file is
  "pattern": "(\\s*([0-9a-zA-Z\\+\\=]){4}\\s*)+"

The problem is with the forward slash (/), for some reason it comes out as an escaped backslash (\\).

I've changed machines since generating the Swagger resources, so I'm not sure which version I ran with. I can update and test again if needed. 

Please note that this editor changes the RegEx. I'm not sure how to stop that behavior.

CHFJr:
Here is an image of what I see that includes all the characters.
![RegEx](https://user-images.githubusercontent.com/66623272/85607389-0fe48800-b622-11ea-8cc5-363d73434c77.png)

tbieste:
Looks like the Swagger file is using the definition from the FHIR spec for Base64, which does not include the forward slash:  https://www.hl7.org/fhir/datatypes.html#base64Binary
However, Base64 does include the forward slash, so looks like the right thing to do is update the Swagger file generator to match what is in profiles-types.json.

lmsurpre:
https://jira.hl7.org/browse/FHIR-28139 is the corresponding issue for updating the spec

lmsurpre:
I searched the code for instanced of the old regex and confirmed that they've all been switched over.
Unfortunately, HL7 applied the wrong regex in FHIR-28139 (at lest in my opinion) and so I opened https://jira.hl7.org/browse/FHIR-31746.
Based on the discussion there, they might go with the slightly longer (but slightly more correct) `/^(?:[A-Za-z\d+/]{4})*(?:[A-Za-z\d+/]{3}=|[A-Za-z\d+/]{2}==)?$/`

lmsurpre:
@tbieste think we should go with this slightly longer pattern instead?

tbieste:
I think at this point we go with what we have. If HL7 later decides to go with a new pattern, then we can change to be consistent with that. At least now it does accept all the valid characters, which it didn't before.

lmsurpre:
works for me

Issue #1203, opened 2020-06-08
JSON Schema Clarification on required and minItems Validation Keywords


CHFJr:
The Draft 4 version of the JSON Schema Specification (the version used with Swagger 2.0) is silent on the relationship between the "required" and "minItems" keywords. Is it possible to have a required property of type array that has a minItems value of 0, or should the minItems always be 1? In either case, can validators and generators enforce this situation?

lmsurpre:
Yes, I agree that we should add the "minItems" key (and keep the "required" one as well) to better reflect that a 1..* field in FHIR requires both the presence of the element and at least one entry.
My only real concern on this on is whether/how "minItems" is intended to work with null.  In FHIR, it is permitted to have an array element of null (with a corresponding non-null element in the extension array for this element).
However, since we took the approach of excluding primitives, I think I'm fine to include minItems either way.

lmsurpre:
I confirm that `minItems` now appears in the generated definitions.

For example, for swagger/OpenAPI 2.0:
```
        "OperationOutcome": {
            "allOf": [
                {
                    "$ref": "#/definitions/DomainResource"
                },
                {
                    "type": "object",
                    "properties": {
                        "issue": {
                            "type": "array",
                            "items": {
                                "$ref": "#/definitions/OperationOutcome_Issue",
                                "description": "An error, warning, or information message that results from a system action."
                            },
                            "minItems": 1
                        }
                    },
                    "required": [
                        "issue"
                    ]
                }
            ],
```

And for OpenAPI 3.0:
```
            "OperationOutcome": {
                "allOf": [
                    {
                        "$ref": "#/components/schemas/DomainResource"
                    },
                    {
                        "type": "object",
                        "properties": {
                            "issue": {
                                "type": "array",
                                "items": {
                                    "$ref": "#/components/schemas/OperationOutcome_Issue",
                                    "description": "An error, warning, or information message that results from a system action."
                                },
                                "minItems": 1
                            }
                        },
                        "required": [
                            "issue"
                        ]
                    }
                ],
```

In both cases, it seems to work just fine in the swagger editor / openapi ui including with liberty

PR #2091, opened 2021-03-15
Issue #1203 - Updated openapi/swagger generator

tbieste:
Issue #1203 - Added openapi/swagger minItems property to arrays
Issue #1277 - Update base64Binary pattern in openapi/swagger 

### DIFF
--- a/fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
+++ b/fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
@@ -1249,7 +1249,7 @@ public class FHIROpenApiGenerator {
             property.add("format", "int32");
         } else if (com.ibm.fhir.model.type.Base64Binary.class.equals(fieldClass)) {
             property.add("type", "string");
-            property.add("pattern","(\\s*([0-9a-zA-Z\\+\\=]){4}\\s*)+");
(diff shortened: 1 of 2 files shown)


fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
```diff
@@ -1247,7 +1249,7 @@ private static void generateProperty(StructureDefinition structureDefinition, Cl
             property.add("format", "int32");
         } else if (com.ibm.fhir.model.type.Base64Binary.class.equals(fieldClass)) {
             property.add("type", "string");
-            property.add("pattern","(\\s*([0-9a-zA-Z\\+\\=]){4}\\s*)+");
```
lmsurpre, line 1250: I think the regex we use here matches the one given in the spec at https://www.hl7.org/fhir/datatypes.html#base64Binary

I do think `/` is a legal character though, so we should add that.  Are you sue that `=` doesn't need to be escaped inside a regex "character class" (square brackets)?

lmsurpre: ok, I double-checked our regex and it looks fine.  the corresponding issue to get the regex fixed in the spec is https://jira.hl7.org/browse/FHIR-28139

lmsurpre (approved):
LGTM

PR #2096, opened 2021-03-15
Issue #2112 - checkpoint after each item for file-based export

lmsurpre:
Previously, the file-based export would use the COS properties for
checkpointing. This is a bit confusing and, additionally, I think we'd
want a different default for "when to start writing" (in COS there are
constraints on the size and count of the parts, while writing to local
disk has different concerns).

### DIFF
--- a/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/checkpoint/ExportCheckpointAlgorithm.java
+++ b/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/checkpoint/ExportCheckpointAlgorithm.java
@@ -20,6 +20,7 @@ import javax.inject.Inject;
 
 import com.ibm.fhir.bulkdata.jbatch.context.BatchContextAdapter;
 import com.ibm.fhir.bulkdata.jbatch.export.data.ExportTransientUserData;
+import com.ibm.fhir.operation.bulkdata.config.ConfigurationAdapter;
 import com.ibm.fhir.operation.bulkdata.config.ConfigurationFactory;
 import com.ibm.fhir.operation.bulkdata.model.type.BulkDataContext;
 
@@ -39,6 +40,8 @@ public class ExportCheckpointAlgorithm implements CheckpointAlgorithm {
 
     Boolean isFileExport;
 
+    ConfigurationAdapter config = ConfigurationFactory.getInstance();
+
     public ExportCheckpointAlgorithm() {
         // No Operation
     }
@@ -55,7 +58,7 @@ public class ExportCheckpointAlgorithm implements CheckpointAlgorithm {
             JobExecution jobExecution = jobOperator.getJobExecution(jobCtx.getExecutionId());
             BatchContextAdapter contextAdapter = new BatchContextAdapter(jobExecution.getJobParameters());
             BulkDataContext ctx = contextAdapter.getStepContextForSystemChunkWriter();
-            isFileExport = "file".equals(ConfigurationFactory.getInstance().getStorageProviderType(ctx.getSource()));
+            isFileExport = "file".equals(config.getStorageProviderType(ctx.getSource()));
         }
 
         if (logger.isLoggable(Level.FINE)) {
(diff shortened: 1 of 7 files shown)


fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/checkpoint/ExportCheckpointAlgorithm.java
```diff
@@ -25,11 +31,20 @@
...
 
+    final boolean isFileExport;
+
     public ExportCheckpointAlgorithm() {
-        // No Operation
+        JobOperator jobOperator = BatchRuntime.getJobOperator();
+        JobExecution jobExecution = jobOperator.getJobExecution(jobCtx.getExecutionId());
+        BatchContextAdapter contextAdapter = new BatchContextAdapter(jobExecution.getJobParameters());
+        BulkDataContext ctx = contextAdapter.getStepContextForSystemChunkWriter();
+        isFileExport = "file".equals(ConfigurationFactory.getInstance().getStorageProviderType(ctx.getSource()));
```
prb112, line 47: Don't do this here. It's unreliable with CDI

lmsurpre: ok, I moved it to beginCheckpoint() and hid it behind a null check (so it should only run the first time through).

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/checkpoint/ExportCheckpointAlgorithm.java
```diff
@@ -82,18 +101,18 @@ public boolean isReadyToCheckpoint() {
         long cosFileThresholdSize = ConfigurationFactory.getInstance().getCoreCosObjectSizeThreshold();
         long cosMultiPartMinSize = ConfigurationFactory.getInstance().getCoreCosPartUploadTriggerSize();
 
-        // Stop writing to the current COS object
+        // Set to true if we have enough bytes to write a part
+        boolean readyToWrite = isFileExport || (chunkData.getBufferStream().size() > cosMultiPartMinSize);
```
prb112, line 105: This causes a lot of writes to disk in potentially small fragments. Let's get Robin's input.  Writing after every page is what's happening above.

lmsurpre: I talked with Paul and we think it makes the most sense to introduce a new config parameter that works similar to the cosMultiPartMinSize except for writing to files instead of cos.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/checkpoint/ExportCheckpointAlgorithm.java
```diff
@@ -39,6 +53,18 @@ public int checkpointTimeout() {
...
+            JobOperator jobOperator = BatchRuntime.getJobOperator();
+            JobExecution jobExecution = jobOperator.getJobExecution(jobCtx.getExecutionId());
+            BatchContextAdapter contextAdapter = new BatchContextAdapter(jobExecution.getJobParameters());
+            BulkDataContext ctx = contextAdapter.getStepContextForSystemChunkWriter();
+            isFileExport = "file".equals(ConfigurationFactory.getInstance().getStorageProviderType(ctx.getSource()));
+        }
+
+        if (adapter == null) {
+            adapter = ConfigurationFactory.getInstance();
+        }
```
prb112, line 66: Two thoughts... put this ahead of line 56, and then call adapter in isFileExport. 

You can also inline it to line 43, it can be setup when the class is initialized as it is stateless

prb112 (approved):
one comment on checkpoint.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/checkpoint/ExportCheckpointAlgorithm.java
lmsurpre: moved to line 43 as suggested

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/checkpoint/ExportCheckpointAlgorithm.java
lmsurpre: the new properties are now in place

Issue #2018, opened 2021-03-04
Bulk Data ExitStatus


prb112:
**Describe the bug**
The Bulk Data Exit Status is limited to 512 characters. 

`.addVarcharColumn(EXITSTATUS, 512, NULL) // EXITSTATUS VARCHAR(512)`

The Bulk Data mutates the Exit Status and returns a single to the backend indicating success as an overloaded operation per type/input which leads to a large string that may exceed 512 and prunes 

**To Reproduce**
1. Create a very large export with a large number.

**Expected behavior**
Retrieve results in all circumstances.

**Additional context**
n/a

lmsurpre:
Question:  do we need backwards compatibility with previously-completed jobs?

Answer:  we think its probably not worth the effort.  Would be an extra couple hours to code up, but also more testing, etc.

PR #2101, opened 2021-03-17
Bulk Data aborts when setting the ExitStatus #2018

prb112:
Note: this may not be ideal... since resources which are Zero'd out, are not returned in a list, back to the end user. 

So we may not want to do this... and may want to raise 2018 to a P2


lmsurpre (approved):
I don't see much value in reporting on the resource types that we didn't export, so I think its a good change.  LGTM.

Issue #2094, opened 2021-03-15
Use the internal patient compartment search parameter during Patient/Group export


lmsurpre:
We now have a special internal "compartment" search parameter that we should be able to leverage so that we don't need to aggregate all the different compartment inclusion criteria parameters separately.

For example, we might be able to get rid of this code in `PatientResourceHandler.fillChunkDataBuffer(List<String> patientIds, ReadResultDTO dto)`:
```
            List<String> compartmentSearchCriterias = CompartmentUtil.getCompartmentResourceTypeInclusionCriteria("Patient", resourceType.getSimpleName());
            if (compartmentSearchCriterias.size() > 1) {
                isDoDuplicationCheck = true;
            }
            for (String compartmentSearchCriteria : compartmentSearchCriterias) {
                HashMap<String, List<String>> queryTmpParameters = new HashMap<>();
                queryTmpParameters.putAll(queryParameters);
                queryTmpParameters.put(compartmentSearchCriteria, Arrays.asList(String.join(",", patientIds)));
                searchContext = SearchUtil.parseQueryParameters(resourceType, queryTmpParameters);
```

One thing that might make this difficult is that, under normal operation, a patient compartment search is only ever scoped to a single patient (whereas here we have a list of patients).  A couple ideas:
A. Perform a different search for each specific patient included in the export.  It would definitely be a lot more searches (i.e. more round trips to the DB), but maybe these simple searches will perform better than one giant OR query like we have now?
B. Add support (if we don't already support it) for having a logical OR on the internal compartment search parameter.

If we're able to do B, this would also help us to support `user` scoped SMART App Launch scopes in our AuthzPolicyEnforcementPersistenceInterceptor, which are for cases where the user has access to a set of patient compartment resources and not just their own.

_Originally posted by @lmsurpre in https://github.com/IBM/FHIR/issues/2058#issuecomment-799022605_

lmsurpre:
Before this change, a compartment search like `Patient/123/Basic?string=teststring` (with `"useStoredCompartmentParam": false`) would result in SQL like:
```sql
SELECT COUNT(DISTINCT R.LOGICAL_RESOURCE_ID)
  FROM Basic_LOGICAL_RESOURCES LR
  JOIN Basic_RESOURCES R
    ON R.RESOURCE_ID = LR.CURRENT_RESOURCE_ID
   AND R.IS_DELETED = 'N'
  JOIN Basic_TOKEN_VALUES_V AS param0
[...]

Issue #1366, opened 2020-07-27
Storing references in their own parameter table may be more efficient


punktilious:
**Is your feature request related to a problem? Please describe.**
Currently reference values e.g. "Patient/some-id-value" are stored as string parameters (e.g. observation_str_values). This leads to repeated large entries in the str_values indexes, make worse by the fact that we store both upper and lower case even though the lower case value is never used. Because these are identifiers (which are essentially random), range scanning does not make sense (e.g. WHERE value > ?).

**Describe the solution you'd like**
1. Implement a new <resourceType>_reference_values table.
2. Update JDBC persistence layer to store reference parameters into this table
3. Update search queries to use this table for chained reference joins.

The reference parameter table should look something like this:

```
CREATE TABLE observation_reference_values (
    resource_type_id.       INT     NOT NULL,
    parameter_name_id       INT     NOT NULL,
    reference_value     VARCHAR(64) NOT NULL,
    logical_resource_id  BIGINT     NOT NULL,
    CONSTRAINT pk_observation_reference_values PRIMARY KEY (resource_type_id, parameter_name_id, reference_value, logical_resource_id)
)
```

An index on logical_resource_id is also required,  plus FK constraints to the reference and logical_resources tables.

**Describe alternatives you've considered**
Do not change the current implementation.

**Additional context**
Consider how the schema and search queries may be leveraged to resolve include/revinclude queries where we may want to deliver dependencies of a resource before the resource itself rather than forcing the client to have to page through an entire result set.

prb112:
Of interest - https://github.com/IBM/FHIR/issues/1329

PR #2102, opened 2021-03-17
issue #2094 - add support for 'or' within inclusion criteria

lmsurpre:
1. Introduces SearchUtil.buildInclusionCriteria() for building the
inclusion criteria parameter with multiple compartmentId values (maybe
this should go in CompartmentUtil instead?)
2. Updates JDBCQueryBuilder.processInclusionCriteria to use all of the
inclusionCriteria values; previously it used `=` against the first value
but now it uses `IN (val1,val2,val3,...)` instead.

And update 3 places to use it:
1. A new test in AbstractSearchCompartmentTest
2. com.ibm.fhir.bulkdata.export.patient.resource.PatientResourceHandler.fillChunkDataBuffer()
    This should be much simpler now and also much more efficient.
Instead of searching with a value of `id1,id2,id3,...` on each and every
inclusion criteria, we can do a single query with a whole page of
patient ids.
3. com.ibm.fhir.smart.AuthzPolicyEnforcementPersistenceInterceptor.beforeSearch()
    This lays the groundwork for supporting `user` scopes. Instead of
converting the request to a compartment search for a single patient, we
can now scope the search to the union of all the patient compartments passed via the patient_id
claim in access token.

Also, some bulk data refactoring.

### DIFF
--- a/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/export/patient/resource/PatientResourceHandler.java
+++ b/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/export/patient/resource/PatientResourceHandler.java
@@ -27,6 +27,7 @@ import com.ibm.fhir.core.FHIRMediaType;
 import com.ibm.fhir.model.format.Format;
 import com.ibm.fhir.model.generator.FHIRGenerator;
 import com.ibm.fhir.model.generator.exception.FHIRGeneratorException;
+import com.ibm.fhir.model.resource.Patient;
 import com.ibm.fhir.model.resource.Resource;
 import com.ibm.fhir.operation.bulkdata.config.ConfigurationAdapter;
 import com.ibm.fhir.operation.bulkdata.config.ConfigurationFactory;
@@ -35,6 +36,7 @@ import com.ibm.fhir.persistence.FHIRPersistence;
 import com.ibm.fhir.persistence.context.FHIRPersistenceContext;
 import com.ibm.fhir.persistence.context.FHIRPersistenceContextFactory;
 import com.ibm.fhir.search.SearchConstants;
+import com.ibm.fhir.search.SearchConstants.Prefix;
 import com.ibm.fhir.search.context.FHIRSearchContext;
 import com.ibm.fhir.search.parameters.QueryParameter;
 import com.ibm.fhir.search.util.SearchUtil;
@@ -104,10 +106,10 @@ public class PatientResourceHandler {
             List<String> searchCriteria = new ArrayList<>();
             if (ctx.getFhirSearchFromDate() != null) {
                 // https://www.hl7.org/fhir/r4/search.html#prefix
-                searchCriteria.add("ge" + ctx.getFhirSearchFromDate());
+                searchCriteria.add(Prefix.GE.value() + ctx.getFhirSearchFromDate());
             }
             if (ctx.getFhirSearchToDate() != null) {
-                searchCriteria.add("lt" + ctx.getFhirSearchToDate());
+                searchCriteria.add(Prefix.LT.value() + ctx.getFhirSearchToDate());
             }
 
             if (!searchCriteria.isEmpty()) {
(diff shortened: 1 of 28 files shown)


fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
```diff
@@ -138,15 +138,11 @@ public void beforeSearch(FHIRPersistenceEvent event) throws FHIRPersistenceInter
...
-                        FHIRSearchContext compartmentSearchContext = SearchUtil.parseCompartmentQueryParameters("Patient", patientIdFromToken.get(0),
-                                ModelSupport.getResourceType(event.getFhirResourceType()), Collections.emptyMap(), searchContext.isLenient());
-
-                        // Add compartment search parameters to front of search parameter list
-                        searchContext.getSearchParameters().addAll(0, compartmentSearchContext.getSearchParameters());
+                        // Build the Patient compartment inclusion criteria search parameter
+                        QueryParameter inclusionCriteria = SearchUtil.buildInclusionCriteria("Patient", patientIdFromToken, event.getFhirResourceType());
+
+                        // Add the inclusion criteria parameter to the front of the search parameter list
+                        searchContext.getSearchParameters().add(0, inclusionCriteria);
```
prb112, line 145: interesting change, seems pretty good

prb112 (approved):
I'd like to merge the PatientResourceHandler pr I have open, and then have you merge with main, and then we can merge.  Otherwise, looks great. 

Only comment seems reasonable ... should be a reasonable number under a 1000 variables. Maybe even far less 

LGTM

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1044,42 +1049,69 @@ protected SqlQueryData processInclusionCriteria(QueryParameter queryParm) throws
...
+                        commonTokenValueIds.add(commonTokenValueId.toString());
                     }
                 }
+
+                if (fallback) {
+                    // Can't use the common_token_value_id optimization, so do it the old way.
+                    // pX.token_value IN (val1, val2, ...) [ AND pX.code_system_id = n ]
+                    whereClauseSegment.append(PARAMETER_TABLE_ALIAS).append(DOT).append(TOKEN_VALUE)
+                            .append(IN).append(LEFT_PAREN)
+                            .append(String.join(", ", Collections.nCopies(currentParmValues.size(), BIND_VAR)))
```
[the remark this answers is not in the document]
lmsurpre: Good question.  The short answer is that I did it this way because thats what we were already doing:  originally we used parameterization here but when we added support for COMMON_TOKEN_VALUE_ID (#1366), we just used the literal directly (old line 1072).  Was that you or @punktilious?

Anyway, I debated this in my head but decided its fine to keep using the literal strings in the COMMON_TOKEN_VALUE_ID case because those ids are coming from the database (not user-provided).  Hopefully we're always in that case going forward, but I wasn't confident enough to remove the "old way" at this time (because I think use of the new way is still behind a config flag).

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/export/patient/resource/PatientResourceHandler.java
```diff
@@ -80,143 +72,83 @@ public void register(ExportTransientUserData chunkData, BulkDataContext ctx, FHI
...
                 queryParameters.putAll(searchParametersForResoureTypes.get(resourceType).get(indexOfCurrentTypeFilter));
-                if (searchParametersForResoureTypes.get(resourceType).size() > 1) {
-                    isDoDuplicationCheck = true;
-                }
             }
             List<String> searchCriteria = new ArrayList<>();
             if (ctx.getFhirSearchFromDate() != null) {
                 // https://www.hl7.org/fhir/r4/search.html#prefix
-                searchCriteria.add("ge" + ctx.getFhirSearchFromDate());
+                searchCriteria.add(Prefix.GE.value() + ctx.getFhirSearchFromDate());
```
prb112, line 103: nice cleanup

prb112 (approved):
Still looks good -

prb112:
A couple test failures to take care of in the e2e bulkdata.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/patient/ChunkReader.java
```diff
@@ -118,12 +130,16 @@ public void open(Serializable checkpoint) throws Exception {
         adapter.registerRequestContext(ctx.getTenantId(), ctx.getDatastoreId(), ctx.getIncomingUrl());
 
         searchParametersForResoureTypes = BulkDataUtils.getSearchParametersFromTypeFilters(ctx.getFhirTypeFilters());
-        resourceType = ModelSupport.getResourceType(ctx.getFhirResourceType());
```
lmsurpre, line 121: This one was tricky to find!   `ctx.getFhirResourceType()` actually returns a comma-delimited list of resource types.   If you export a single type, it worked just fine.  If you exported multiple, then `ModelSupport.getResourceType(ctx.getFhirResourceType())` came back null.  Nasty!

Issue #1620, opened 2020-10-22
Restrict access to reindex custom operation


punktilious:
**Is your feature request related to a problem? Please describe.**
The reindex custom operation being introduced in #1366 is not an end-user function and should be restricted to admin users only.

**Describe the solution you'd like**
Ensure that authenticated users with admin privileges can call the reindex custom operation.

**Describe alternatives you've considered**
Rely on external API management to hide the function, but security should be multi-layered and so this would be less than ideal.

prb112:
A couple things to know... 

1 - Operation.java is annotated with @RolesAllowed("FHIRUsers"), it should be annotated with the additional role that should  be allowed to perform this operation. 
2 - web.xml only specifies two mappings fhirusers fhiradmin (so fhiradmins should probably be used) 
3 - Update the method you want with a context
```
@POST
    @Path("${operationName}")
    public Response invoke(@PathParam("operationName") String operationName, Resource resource, @Context SecurityContext secContext) {
        log.entering(this.getClass().getName(), "invoke(String,Resource)");
        Date startTime = new Date();
        Response.Status status = null;

        try {
            checkInitComplete();

            FHIROperationContext operationContext =
                    FHIROperationContext.createSystemOperationContext();
            operationContext.setProperty(FHIROperationContext.PROPNAME_URI_INFO, uriInfo);
            operationContext.setProperty(FHIROperationContext.PROPNAME_HTTP_HEADERS, httpHeaders);
            operationContext.setProperty(FHIROperationContext.PROPNAME_METHOD_TYPE, HttpMethod.POST);
            operationContext.setProperty(FHIROperationContext.PROPNAME_SECURITY_CONTEXT, secContext);
```

pump it in the operationContext
4 - in the operation get the property and convert back to Security Context
5 - call secContext.isUserInRole(role)

prb112:
I'm doing the same thing in fhir-operation-erase

prb112:
After a review, we've decided to leave the current open behavior of the $reindex operation. We have marked this as invalid.

PR #2107, opened 2021-03-17
Restrict access to reindex custom operation #1620

prb112:


fhir-server/liberty-config/server.xml
```diff
@@ -50,7 +50,7 @@
         
         To send the trace messages to standard out, set TRACE_FILE to "stdout".
     -->
-    <variable name="TRACE_SPEC" defaultValue="*=info"/>
```
lmsurpre, line 53: we probably don't want finer by default, right?

prb112: Correct, I'm still working through the failure with reindex, and needed more detailed logs as I setup db2 locally.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/ReindexResourceDAO.java
```diff
@@ -294,7 +294,9 @@ public void updateParameters(String tablePrefix, List<ExtractedParameterValue> p
 
         if (parameters != null) {
             JDBCIdentityCache identityCache = new JDBCIdentityCacheImpl(getCache(), this, parameterDao, getResourceReferenceDAO());
-            try (ParameterVisitorBatchDAO pvd = new ParameterVisitorBatchDAO(connection, null, tablePrefix, false, logicalResourceId, 100,
+            // Check if this is multitenant
+            boolean mtId = this.getFlavor().isMultitenant();
```
lmsurpre, line 298: did this part already get committed as part of #2115 ?

fhir-server-test/src/test/java/testng.xml
```diff
@@ -64,4 +64,9 @@
             <package name="com.ibm.fhir.server.test.terminology" />
         </packages>
     </test>
+    <test name="OperationSpecificTests">
+        <packages>
+            <package name="com.ibm.fhir.server.test.operation" />
+        </packages>
+    </test>
```
lmsurpre, line 71: we have a whole bunch of operation tests under `com.ibm.fhir.server.test`...are you thinking to move those under here?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/ReindexResourceDAO.java
prb112: Yes, this is just an artifact of the merge (I solved the issue in this PR, and ended up having to merge)

fhir-server-test/src/test/java/testng.xml
prb112: I ended up doing this when I saw the EverythingOperation standalone

fhir-server/src/main/java/com/ibm/fhir/server/operation/spi/AbstractOperation.java
```diff
@@ -279,4 +284,40 @@ protected FHIROperationException buildExceptionWithIssue(String msg, IssueType i
...
+            Set<String> groups = jwtObj.getGroups();
+            if (groups != null) {
+                for (String group : groups) {
+                    if ("FHIROperationAdmin".equals(group)) {
+                        authorize = true;
+                        return;
+                    }
+                }
+            }
+        }
```
lmsurpre, line 312: I found out that liberty supports mapping JWT subjects to groups, so I don't think we actually need to do something special for JWT authorization.  I think we should just rely on `SecurityContext.isUserInRole` and call it good.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/ReindexResourceDAO.java
lmsurpre: maybe rebase onto main and see if this gets cleaned up?

fhir-server-test/src/test/java/testng.xml
lmsurpre: I do think it makes sense to have a separate package for operation-specific tests, I just think we should move the other ones over as well.

prb112: I also forgot one key thing with this operation. I wanted it to run last.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/ReindexResourceDAO.java
prb112: resolved, I checked out the file from the main branch and replaced directly

fhir-server/src/main/java/com/ibm/fhir/server/operation/spi/AbstractOperation.java
prb112: Removed

lmsurpre:
The code looks good to me, but I'm a tad concerned about making `$reindex` only available to admin users.

Current deployments will need to map users to this new access role in order to get `$reindex` to work (which is required for some migration scenarios).  Is it possible to document the steps for adding the fhiruser to this new security-role?  Or maybe add some fhir-server-config parameter that allows users mapped to the FHIRUsers role to continue invoking this operation?

fhir-server/src/main/java/com/ibm/fhir/server/operation/spi/FHIROperationContext.java
```diff
@@ -59,74 +59,79 @@
      * associated with the REST API request for which the interceptor is being invoked.
      */
     public static final String PROPNAME_SECURITY_CONTEXT = "SECURITY_CONTEXT";
-    
+
     /**
-     *This property is of type HttpMethod 
+     * This property is to inject the JWT into the Operation Context.
+     */
+    public static final String PROPNAME_JWT = "JWT";
```
lmsurpre, line 66: missed this before, but I don't think we need the JWT in the context.  we already have the pertinent info from that in the SecurityContext.  additionally, if someone really needed it, its available via the Authorization header in the request context as well.

docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -2005,6 +2005,7 @@ This section contains reference information about each of the configuration prop
 |`fhirServer/security/oauth/smart/enabled`|boolean|Whether or not the server is enabled for OAuth-based authentication/authorization|
 |`fhirServer/security/oauth/smart/scopes`|array|The list of SMART scopes to advertise in the `.well-known/smart-configuration endpoint|
 |`fhirServer/security/oauth/smart/capabilities`|array|The list of SMART capabilities to advertise in the `.well-known/smart-configuration endpoint|
+|`fhirServer/security/operation/authorize/reindex`|boolean|True, the reindex operation is made available only to fhiradmins, and users are denied access|
```
lmsurpre, line 2008: I understand how this could make sense under the `security` propertyGroup, but I think I'd prefer it to have its own `operations` securityGroup instead.  that would let us group it together with `fhirServer/core/disabledOperations` and would let us extend with additional properties in the future (if needed).

something like this:
```
    "fhirServer": {
        "operations": {
            "open": true,
            "reindex": {
                "enabled": false
            },
            "erase": {
                "enabled": true
                "rolesAllowed": ["FHIRUsers, FHIRAdmin"],
            }
        },
```

prb112: done

PR #2118, opened 2021-03-19
fix touchstone conformance link

lmsurpre:
and move the announcement under the blog header...its pretty old now and shouldn't be so prominent on the site.  i tried to preserve the URL though, so anything linking to it should be ok.


prb112 (approved):

JohnTimm (approved):
LGTM

Issue #1283, opened 2020-06-25
FHIRPath CLI


lmsurpre:
**Is your feature request related to a problem? Please describe.**
Currently, java project can use our fhirpath implementation by 
1. including a dependency like this:
```
<dependency>
    <groupId>com.ibm.fhir:fhir-path</groupId>
    <artifactId>fhir-path</artifactId>
    <version>4.3.0</version>
</dependency>
```
and
2. writing code like this:
```
FHIRPathEvaluator.evaluator().evaluate(resource, fhirpath);
```

This works ok, but I think would be cool to build an all-in-one shaded cli jar for our fhirpath module so that we can execute this more easily from the command line:
```sh
java -jar fhir-path-$VERSION.jar resource fhirpath
```

**Describe the solution you'd like**
Introduce a main that takes arguments for the resource content and the fhirpath string to execute.

Ideally, we could pass resource content from
* a file
* a string (raw command line arg); or
* stdin (so we can pipe to it)

One convention I've seen used is to interpret `-` as an indicator that we should read from stdin instead of a string or file, but I'm not tied to that.

Look at the `fhir-persistence-schema` and `fhir-swagger-generator` projects for examples on how to create the shaded all-in-one jar.

**Describe alternatives you've considered**
The alternative is to let clients do this themselves or use a different fhirpath engine for simple command-line stuff (e.g. the node fhirpath module).

**Additional context**

lmsurpre:
It would also be nice to:
1. allow some way to pass jar files to extend the registry for fhirpath operations like `memberOf`; and
2. provide some way to see supplemental issues that were generated during the evaluation, maybe by printing those to System.error by default (with an option to suppress that)

erodrago:
Would like to try out this issue @lmsurpre

prb112:
@erodrago please go right ahead.  if you have any questions, come find us on Zulip. https://chat.fhir.org/#narrow/stream/212434-ibm

erodrago:
Alright thanks @prb112

prb112:
PR https://github.com/IBM/FHIR/pull/2130

lmsurpre:
Java isn't the best for quick little cli utilities like this because the JVM is still a bit heavy to start.  I'd love to see if we could compile it to a native image using graalvm sometime...

PR #2130, opened 2021-03-23
FHIRPath CLI #1283

prb112:

### DIFF
--- a/docs/src/pages/guides/FHIRValidationGuide.md
+++ b/docs/src/pages/guides/FHIRValidationGuide.md
@@ -232,7 +232,6 @@ The fhir-path-cli enables commandline processing of a FHIR Path over a FHIR reso
 The options for the command line utility are:
 - `--path 'fhir-path'`
 - `--format ['json'|'xml']` Default is 'json'
-- `--type ['stdin'|'file'|'string']`
 - `--file path-to-file` The file that is accessible and read.
 - `--resource 'resource-payload'` The FHIR resource as a well formed string.
 - `--pretty` adds columns and start time and end time of the fhir path request
@@ -241,7 +240,7 @@ The options for the command line utility are:
 For instace, to run the fhir-path-cli on Bundle-2000002.json and extract the ids from the Bundle's resources (as stored in the file).
 
 ``` shell
-java -jar fhir-path-4.6.0-cli.jar --path entry.resource.id --type file --file Bundle-2000002.json --pretty
+java -jar fhir-path-4.6.0-cli.jar --path entry.resource.id --file Bundle-2000002.json --pretty
 ```
 
 ``` text
@@ -268,7 +267,7 @@ Without the `--pretty` option, the results are:
 For instance, to run the fhir-path-cli using a string.
 
 ``` shell
-java -jar fhir-path-4.6.0-cli.jar --path 'text.status.value' --type string --resource '{
+java -jar fhir-path-4.6.0-cli.jar --path 'text.status.value' --resource '{
   "resourceType": "Patient",
   "id": "pat1",
   "text": {
--- a/docs/src/pages/guides/FHIRValidationGuide.md
+++ b/docs/src/pages/guides/FHIRValidationGuide.md
@@ -265,7 +265,7 @@ Without the `--pretty` option, the results are:
 1000001
 ```
 
-For instace, to run the fhir-path-cli using a string.
+For instance, to run the fhir-path-cli using a string.
 
 ``` shell
 java -jar fhir-path-4.6.0-cli.jar --path 'text.status.value' --type string --resource '{
--- a/docs/src/pages/guides/FHIRValidationGuide.md
+++ b/docs/src/pages/guides/FHIRValidationGuide.md
@@ -238,7 +238,7 @@ The options for the command line utility are:
 - `--pretty` adds columns and start time and end time of the fhir path request
 - `--help`
 
-For instace, to run the fhir-path-cli on Bundle-2000002.json and extract the ids from the Bundle's resources.
+For instace, to run the fhir-path-cli on Bundle-2000002.json and extract the ids from the Bundle's resources (as stored in the file).
 
 ``` shell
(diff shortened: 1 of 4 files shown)


docs/src/pages/guides/FHIRValidationGuide.md
```diff
@@ -224,3 +224,124 @@ The IBM FHIR Server provides a basic implementation of the $validate operation t
...
+```
+
+Without the `--pretty` option, the results are: 
+
+```
+1
+1000001
+```
+
+For instace, to run the fhir-path-cli using a string.
```
lmsurpre, line 268:
```suggestion
For instance, to run the fhir-path-cli using a string.
```

docs/src/pages/guides/FHIRValidationGuide.md
```diff
@@ -224,3 +224,124 @@ The IBM FHIR Server provides a basic implementation of the $validate operation t
...
+
+``` shell
+java -jar fhir-path-4.6.0-cli.jar --path 'text.status.value' --type string --resource '{
```
lmsurpre, line 271: I wonder if we could get it to work over specific fhir data types.  they don't have an official serialization, but we do support `toString` on them in our model.  just food for thought on the `--resource` flag name.

docs/src/pages/guides/FHIRValidationGuide.md
```diff
@@ -224,3 +224,124 @@ The IBM FHIR Server provides a basic implementation of the $validate operation t
...
 Coded elements (code, Coding, CodeableConcept data types), maybe have a binding element that specifies a ValueSet that that element is bound to. This means that the coded element must have a value that comes from that value set. The FHIR profile component will expand value sets according to the ValueSet [expansion algorithm](http://hl7.org/fhir/valueset.html#expansion) for ValueSets that include CodeSystem resources available via the FHIR registry component.
+
+# fhir-path-cli - command line execution of fhirpath over a FHIR Resource
+
+The fhir-path-cli enables commandline processing of a FHIR Path over a FHIR resource and outputs the results. 
+
+The options for the command line utility are:
+- `--path 'fhir-path'`
+- `--format ['json'|'xml']` Default is 'json'
+- `--type ['stdin'|'file'|'string']`
```
lmsurpre, line 235: can `type` have a default too?  maybe `stdin` or `file`...whichever we think will be most common

lmsurpre: actually, maybe we can just remove this and deduce it from the other arguments?
* `stdin` can be the default
* if `--file` is present then its type=file
* if `--resource` is present then its type=string

prb112: I thought about that too. I decided to adhere to what was in the issue. 
Maybe we infer if it doesn't exist?

docs/src/pages/guides/FHIRValidationGuide.md
prb112: interesting

fhir-path/src/main/java/com/ibm/fhir/path/tools/Main.java
```diff
@@ -0,0 +1,386 @@
...
+import com.ibm.fhir.path.visitor.FHIRPathNodeVisitor;
+
+/**
+ * Processes the resource from one of the entry points, into an in memory resource
+ * a FHIR Path is executed over it, and the output is put out to the screen.
+ *
+ * The types of entry points are:
+ * a file
+ * a string (raw command line arg); or
+ * stdin (so we can pipe to it)
```
lmsurpre, line 53: javadoc will squash the whitespace here; i recommend putting it in a list
```suggestion
 * <ul>
 * <li>a file
 * <li>a string (raw command line arg); or
 * <li>stdin (so we can pipe to it)
 * </ul>
```

docs/src/pages/guides/FHIRValidationGuide.md
lmsurpre: I fine with that.  Although I can't imagine anyone wanting to specify it when they don't have to!

prb112: Fixed... interpolating the type from the command parameters.

docs/src/pages/guides/FHIRValidationGuide.md
prb112: maybe a separate issue? we could park it in a value in a Parameters object and call it.

prb112:
Rebuild depends on PR that fixes the snapshot

docs/src/pages/guides/FHIRValidationGuide.md
lmsurpre: I'd say don't even worry about this one right now.  I just know there were times where I wanted to execute FHIRPath against some specific backbone element and, while thats supported by our FHIREvaluator, it won't be supported from the cli with this design.

lmsurpre: For it to work the way I was sort of hoping for, we'd need to be able to parse a specific Element / Element sub-type (independent of its resource wrapper).  But I'm pretty sure we don't support that and its probably not even worth entertaining that with a separate issue at this point.

fhir-path/src/main/java/com/ibm/fhir/path/tools/Main.java
```diff
@@ -0,0 +1,372 @@
+/*
+ * (C) Copyright IBM Corp. 2021
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.ibm.fhir.path.tools;
```
lmsurpre, line 7: in other projects, I think I used `*.tool` in the package (singular vs plural).  Not sure how important it is to be consistent on this though

lmsurpre (approved):
LGTM

lmsurpre:
Anything we need to do from the release automation side to ensure that both the cli.jar and the normal jar get uploaded?

fhir-path/src/main/java/com/ibm/fhir/path/tools/Main.java
prb112: Interesting... I looked in another project and it had tools.  might be a good cleanup item

PR #2134, opened 2021-03-23
Everything Operation doesn't respond correctly to deleted resources

prb112:
…2133

### DIFF
--- a/operation/fhir-operation-everything/src/main/java/com/ibm/fhir/operation/everything/EverythingOperation.java
+++ b/operation/fhir-operation-everything/src/main/java/com/ibm/fhir/operation/everything/EverythingOperation.java
@@ -168,7 +168,6 @@ public class EverythingOperation extends AbstractOperation {
             patient = (Patient) resourceHelper.doRead(PATIENT, logicalId, false, false, null, null);
         } catch (FHIRPersistenceResourceDeletedException fde) {
             FHIROperationException exceptionWithIssue = buildExceptionWithIssue("Patient with ID '" + logicalId + "' does not exist.", IssueType.NOT_FOUND);
-            LOG.throwing(this.getClass().getName(), "doInvoke", exceptionWithIssue);
             throw exceptionWithIssue;
         } catch (Exception e) {
             FHIROperationException exceptionWithIssue = buildExceptionWithIssue("An unexpected error occurred while reading patient '" + logicalId + "'", IssueType.EXCEPTION);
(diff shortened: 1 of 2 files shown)


michaelwschroeder (approved):
LGTM - might want to add a testcase

prb112:
> LGTM - might want to add a testcase

WIll do...

prb112:
Added an Integration Test

operation/fhir-operation-everything/src/main/java/com/ibm/fhir/operation/everything/EverythingOperation.java
```diff
@@ -165,31 +166,33 @@ protected Parameters doInvoke(FHIROperationContext operationContext, Class<? ext
...
+            FHIROperationException exceptionWithIssue = buildExceptionWithIssue("Patient with ID '" + logicalId + "' does not exist.", IssueType.NOT_FOUND);
+            throw exceptionWithIssue;
         } catch (Exception e) {
             FHIROperationException exceptionWithIssue = buildExceptionWithIssue("An unexpected error occurred while reading patient '" + logicalId + "'", IssueType.EXCEPTION);
             LOG.throwing(this.getClass().getName(), "doInvoke", exceptionWithIssue);
             throw exceptionWithIssue;
         }
         if (patient == null) {
             FHIROperationException exceptionWithIssue = buildExceptionWithIssue("Patient with ID '" + logicalId + "' does not exist.", IssueType.NOT_FOUND);
-            LOG.throwing(this.getClass().getName(), "doInvoke", exceptionWithIssue);
```
lmsurpre, line 175: I looked up `Logger.throwing` and it only logs at log-level FINER.  I actually kind of like this because it makes the exit point clear when tracing.  I don't think we need to remove it and I wouldn't mind trying to use this pattern in more places that we rethrow.

prb112: It doesn't make sense though... it's not really an exception case.  It's just signaling here that the patient is deleted.  I don't think it makes sense at this point.

lmsurpre: That seems like an argument for not throwing here, rather than an argument for not logging that we're throwing (when level=FINER).  I guess we're expecting to never get here anyway (now that you catch and rethrow above)?

lmsurpre (approved):
LGTM

operation/fhir-operation-everything/src/main/java/com/ibm/fhir/operation/everything/EverythingOperation.java
prb112: right.

Issue #2145, opened 2021-03-25
Add maturity level annotation and/or javadoc to the fhir-model resource classes


lmsurpre:
**Is your feature request related to a problem? Please describe.**
HL7 FHIR defines a maturity model so that they can ship content at different levels of maturity in a single cohesive specification.

From https://confluence.hl7.org/display/FHIR/FHIR+Maturity+Model :
```
Draft

FMM0 = the artifact has been published on the current build. This level is synonymous with Draft.

STU

FMM1 = FMM0 + the artifact produces no warnings during the build process and the responsible WG has indicated that they consider the artifact substantially complete and ready for implementation. For resources, profiles and implementation guides, the FHIR Management Group has approved the underlying resource/profile/IG proposal.
FMM2 = FMM1 + the artifact has been tested and successfully supports interoperability among at least three independently developed systems leveraging most of the scope (e.g. at least 80% of the core data elements) using semi-realistic data and scenarios based on at least one of the declared scopes of the artifact (e.g. at a connectathon). These interoperability results must have been reported to and accepted by the FMG
FMM3 = FMM2 + the artifact has been verified by the work group as meeting the Conformance Resource Quality Guidelines; has been subject to a round of formal balloting; has at least 10 distinct implementer comments recorded in the tracker drawn from at least 3 organizations resulting in at least one substantive change
FMM4 = FMM3 + the artifact has been tested across its scope (see below), published in a formal publication (e.g. STU), and implemented in multiple prototype projects. As well, the responsible work group agrees the artifact is sufficiently stable to require implementer consultation for subsequent non-backward compatible changes.
FMM5 = FMM4 + the artifact has been published in two formal publication release cycles at FMM1+ (i.e. STU level) and has been implemented in at least 5 independent production systems in more than one country

Normative

FMM6 = FMM5 + the responsible work group and the FMG agree the material is ready to lock down and the artifact has passed HL7 normative ballot
```

Currently, the IBM FHIR Server supports all resource types, irregardless of maturity level.
However, users of the model should be mindful of the maturity level because depending on a normative resource is a lot different than depending on a draft one (in terms of stability and future-proofing your application).

**Describe the solution you'd like**
To make it easier for users to consider this information, we should put it right at their fingertips, in the javadoc for the resource class.

Additionally, I would like to define a new annotation to make it possible to programatically determine the maturity level of a given resource type at runtime.  The main driver behind this is that I'd like to be able to easily create "classes" of resources based on their maturity level for #819 

[...]

PR #2147, opened 2021-03-25
issue #2145 - add MaturityLevel javadoc and annotation to model classes

lmsurpre:
Also updated copyrights to 2021.

### DIFF
--- a/fhir-tools/src/main/java/com/ibm/fhir/tools/CodeGenerator.java
+++ b/fhir-tools/src/main/java/com/ibm/fhir/tools/CodeGenerator.java
@@ -913,11 +913,19 @@ public class CodeGenerator {
                 }
             }
 
-            List<String> javadocLines = new ArrayList<>(Arrays.asList(getElementDefinition(structureDefinition, path).getString("definition").split(System.lineSeparator())));
+            List<String> javadocLines = new ArrayList<>(Arrays.asList(
+                    getElementDefinition(structureDefinition, path).getString("definition").split(System.lineSeparator())));
             if (isDateTime(structureDefinition)) {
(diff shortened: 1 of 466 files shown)


JohnTimm (approved):
LGTM

fhir-tools/src/main/java/com/ibm/fhir/tools/CodeGenerator.java
```diff
@@ -913,12 +913,20 @@ private void generateClass(JsonObject structureDefinition, List<String> paths, C
...
             if (isDateTime(structureDefinition)) {
                 javadocLines.addAll(Arrays.asList("", "If seconds are specified, fractions of seconds may be specified up to nanosecond precision (9 digits). However, any fractions of seconds specified to greater than microsecond precision (6 digits) will be truncated to microsecond precision when stored."));
             } else if (isInstant(structureDefinition) || isTime(structureDefinition)) {
                 javadocLines.addAll(Arrays.asList("", "Fractions of seconds may be specified up to nanosecond precision (9 digits). However, any fractions of seconds specified to greater than microsecond precision (6 digits) will be truncated to microsecond precision when stored."));
             }
+            if (!nested && hasMaturityLevel(structureDefinition)) {
+                String maturityLevel = getMaturityLevel(structureDefinition);
+                if (maturityLevel != null) {
+                    String maturityClass = getMaturityClass(maturityLevel);
+                    javadocLines.addAll(Arrays.asList("", "Maturity level: FMM" + maturityLevel + " (" + maturityClass + ")"));
```
tbieste, line 927: What to do if getMaturityClass(...) returns null?

tbieste:
Left one minor comment, otherwise LGTM.

JohnTimm:
Do you think the @MaturityLevel annotation should be after the @Generated annotation indicating that "everything below the @Generated annotation is generated"?

fhir-tools/src/main/java/com/ibm/fhir/tools/CodeGenerator.java
lmsurpre: then it will print null in the javadoc, then go investigate why it was null (it should never be)

Issue #2092, opened 2021-03-15
Support for implicit value sets


JohnTimm:
Add support for implicit value sets per:
https://www.hl7.org/fhir/snomedct.html#implicit

Issue #1980, opened 2021-02-22
Introduce fhir-term-graph module


JohnTimm:
The `fhir-term-graph` module will have an implementation of the `FHIRTermProvider` interface built on top of a graph database (JanusGraph). This will enable the ability to efficiently support code systems with a large number of concepts (e.g. SNOMED-CT).

lmsurpre:
The graph loader definitely needs some better docs.

On my first attempt to load it, I passed the following args:
```
SNOMED
-base /Users/lmsurpre/SnomedCT_USEditionRF2_PRODUCTION_20210301T120000Z/Full
-concept Terminology/sct2_Concept_Full_US1000124_20210301.txt
-desc Terminology/sct2_Description_Full-en_US1000124_20210301.txt
-relation Terminology/sct2_Relationship_Full_US1000124_20210301.txt
-lang Refset/der2_cRefset_LanguageFull-en_US1000124_20210301.txt
-config src/test/resources/conf/janusgraph-berkeleyje-lucene.properties
```

It did a bunch of work and then bombed out with this:
> An error occurred: java.io.FileNotFoundException: /Users/lmsurpre/SnomedCT_USEditionRF2_PRODUCTION_20210301T120000Z/Full/../Refset/Language/Refset/der2_cRefset_LanguageFull-en_US1000124_20210301.txt (No such file or directory)


Apparently it wants something like this instead:
```
SNOMED
-base /Users/lmsurpre/SnomedCT_USEditionRF2_PRODUCTION_20210301T120000Z/Full/Terminology
-concept sct2_Concept_Full_US1000124_20210301.txt
-desc sct2_Description_Full-en_US1000124_20210301.txt
-relation sct2_Relationship_Full_US1000124_20210301.txt
-lang der2_cRefset_LanguageFull-en_US1000124_20210301.txt
-config src/test/resources/conf/janusgraph-berkeleyje-lucene.properties
```

lmsurpre:
I opened #2176 for the missing doc

lmsurpre:
With fhir-term-graph enabled as described above, I noticed the following output during startup:
```
[2021-03-30T21:25:17.347+0000] 0000002e SystemErr     R   SLF4J: Class path contains multiple SLF4J bindings.
[...]

Issue #1733, opened 2020-11-19
Support token modifiers ':above' and ':below'


lmsurpre:
from https://www.hl7.org/fhir/R4/search.html#token
|modifier|description|
| -- | -- | 
:above | The search parameter is a concept with the form [system]\|[code], and the search parameter tests whether the coding in a resource subsumes the specified search code. For example, the search concept has an is-a relationship with the coding in the resource, and this includes the coding itself.
:below | the search parameter is a concept with the form [system]\|[code], and the search parameter tests whether the coding in a resource is subsumed by the specified search code. For example, the coding in the resource has an is-a relationship with the search concept, and this includes the coding itself.

example:
> Search for any condition in the SNOMED CT value set "http://snomed.info/sct?fhir_vs=isa/126851005" that includes all descendants of "Neoplasm of liver"
    ```
    GET [base]/Condition?code:below=126851005
    ```

Note: I don't think that sample is actually right.  I think they need to include the system in there.  Should be:
```
GET [base]/Condition?code:below=http://snomed.info/sct|126851005
```

Also note: If we can support the modifier `:in` (#1734) then I think this one becomes less important because searching for anything in a valueset could be used for the same use cases and more.

lmsurpre:
To support, we'd need to create and maintain a closure table within the database and make that closure table part of the query mechanism.
Open questions on how to efficiently generate and maintain the table.

Getting it to work for a subset of scenarios with lots of assumptions/limitations might be a little easier...

JohnTimm:
A "lightweight" version of this feature could work as follows:
1. ask the terminology service for the closure of a given concept (the selection of whether the concept is coming from the search parameter or whether it is coming from the instance determines which concept to take)
2. convert the closure of the specified to a SQL IN clause with a literal enumeration IN ('a', 'b', c')

PR #2161, opened 2021-03-27
Issue #2092, Issue #1733, Issue #1980 - miscellaneous updates

JohnTimm:

1. Refactored common code for working with URLs into a utility class called `URLSupport`
2. Created abstract base class for implicit value set registry resource providers
3. Added `getAncestorsAndSelf` and `getDescedantsAndSelf` methods to `CodeSystemSupport` (for #1733)
4. Added unit tests for 1. and 3.
5. Added `hasLabel("Concept")` when `ValueSet` include filters list is empty

NOTE: I "sorted members" on `CodeSystemSupport`, therefore, it looks like more changes than are actually there. The only substantive changes are the addition of the new methods mentioned in 3. above.

### DIFF
--- a/fhir-core/src/main/java/com/ibm/fhir/core/util/URLSupport.java
+++ b/fhir-core/src/main/java/com/ibm/fhir/core/util/URLSupport.java
@@ -28,6 +28,14 @@ public class URLSupport {
 
     private URLSupport() { }
 
+    /**
+     * URL decode the input string
+     *
+     * @param s
+     *     the string to URL decode
+     * @return
+     *     the URL decoded string
+     */
     public static String decode(String s) {
         try {
             return URLDecoder.decode(s, "UTF-8");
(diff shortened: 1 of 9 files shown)


fhir-core/src/main/java/com/ibm/fhir/core/util/URLSupport.java
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * A utility class for working with URLs
+ */
+public class URLSupport {
+    private static final Map<String, URL> URL_CACHE = createLRUCache(128);
+
+    private URLSupport() { }
+
+    public static String decode(String s) {
```
michaelwschroeder, line 31: would be nice to have method prologues for these methods, possibly with simple example of an input string and what's returned for the less obvious ones

fhir-path/src/main/java/com/ibm/fhir/path/function/FHIRPathAbstractTermFunction.java
```diff
@@ -1,11 +1,12 @@
 /*
- * (C) Copyright IBM Corp. 2020
+ * (C) Copyright IBM Corp. 2021
```
michaelwschroeder, line 2: I believe the convention is to add the new year to the original year
```suggestion
 * (C) Copyright IBM Corp. 2020, 2021
```

michaelwschroeder (approved):
LGTM - just a couple minor comments

fhir-core/src/main/java/com/ibm/fhir/core/util/URLSupport.java
JohnTimm: Will add some Javadoc

fhir-path/src/main/java/com/ibm/fhir/path/function/FHIRPathAbstractTermFunction.java
JohnTimm: Good catch. Will update

fhir-core/src/main/java/com/ibm/fhir/core/util/URLSupport.java
JohnTimm: Javadoc added

fhir-path/src/main/java/com/ibm/fhir/path/function/FHIRPathAbstractTermFunction.java
JohnTimm: Updated copyright header

prb112 (approved):

Issue #2150, opened 2021-03-25
fast export impl 'txTimeoutMillis' should be set relative to the configured transaction timeout


lmsurpre:
**Describe the bug**
The fast export impl has a hardcoded "read window" of 90 seconds:
```
    // Assume for now we have a transaction timeout of 120s, but need to give the upload some time to work
    // TODO configuration
    private long txTimeoutMillis = 90000L;
```

If users have overridden the transaction timeout, this could result in failure.

**To Reproduce**
1. set the FHIR_TRANSACTION_MANAGER_TIMEOUT environment variable to something less than 90s (maybe 10s)
2. perform a system export

**Expected behavior**
the system export succeeds

**Alternative idea**
Check if it's possible to override the configured transaction timeout value (programatically or via bulk datasource config).
If so, we could set it to 120s for just the export transaction (without affecting anything else).

**Additional context**
Perhaps related to #2149

prb112:
@lmsurpre I added a configuration element for this, I just didn't plug it in.

prb112:
I spent some time looking at this... 

One, we could read from the variable FHIR_TRANSACTION_MANGER_TIMEOUT, however that's not a guarantee to be a number.  

We could interpolate the timeout using an SPI, I played with this a bit, it's not super clean, and requires reflections. 

Some alternatives that were mentioned on StackOverflow:
- https://stackoverflow.com/questions/36945450/how-do-i-configure-a-transaction-timeout-in-websphere-liberty-batch
- https://stackoverflow.com/questions/54312349/transaction-timeout-on-step-level-vs-server-xml

lmsurpre:
I set `maxChunkReadTime` to `"15000"` (15 seconds) and set `FHIR_TRANSACTION_MANGER_TIMEOUT` to `"20s"` and confirmed it worked as expected (based on the following output).

```
2021-03-31 14:27:53.140 INFO bulkexportfastjob[203] processed 190804 resources in 15.02 seconds (rate=12701.1 resources/second)
...
2021-03-31 14:32:05.778 INFO CWWKO1660I: The application fhir-openapi was processed and an OpenAPI document was produced.
2021-03-31 14:42:44.885 INFO bulkexportfastjob[204] processed 197040 resources in 15.34 seconds (rate=12846.6 resources/second)
2021-03-31 14:42:56.607 INFO bulkexportfastjob[204] processed 163336 resources in 11.60 seconds (rate=14082.4 resources/second)
```

[...]

PR #2171, opened 2021-03-30
configure the maxChunkReadTime for Fast Export and configurable timeout for presigned URL #2150

prb112:

### DIFF
--- a/docs/src/pages/guides/FHIRServerUsersGuide.md
+++ b/docs/src/pages/guides/FHIRServerUsersGuide.md
@@ -2032,6 +2032,7 @@ This section contains reference information about each of the configuration prop
 |`fhirServer/bulkdata/core/cos/requestTimeout`|number|The request timeout in second for the COS client|
 |`fhirServer/bulkdata/core/cos/socketTimeout`|number|The socket timeout in second for the COS client|
 |`fhirServer/bulkdata/core/cos/useServerTruststore`|boolean|If the COS Client should use the IBM FHIR Server's TrustStore to access S3/IBMCOS service |
+|`fhirServer/bulkdata/core/cos/presignedExpiry`|number|The time in seconds of the presigned download URL; must be using HMAC auth|
 |`fhirServer/bulkdata/core/file/writeTriggerSizeMB`|number|The size, in megabytes, at which to write the buffer to file.|
 |`fhirServer/bulkdata/core/file/sizeThresholdMB`|number|The size, in megabytes, at which to finish writing a given file. Use `0` to indicate that all resources of a given type should be written to a single file.|
 |`fhirServer/bulkdata/core/file/resourceCountThreshold`|number|The number of resources at which to finish writing a given file. The actual number of resources written to a single file may be slightly above this number, dependent on the configured page size. Use `0` to indicate that there is no limit to the number of resources to be written to a single file.|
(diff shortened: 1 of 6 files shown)


docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -2040,7 +2041,7 @@ This section contains reference information about each of the configuration prop
 |`fhirServer/bulkdata/core/maxPartitions`|number| The maximum number of simultaneous partitions that are processed per Export and Import |
 |`fhirServer/bulkdata/core/maxInputs`|number| The number of inputs allowed for $import |
 |`fhirServer/bulkdata/core/iamEndpoint`|string| Override the system's IAM endpoint |
-|`fhirServer/bulkdata/core/fastTxTimeout`|number| Time timeout for the fast implementations transaction |
+|`fhirServer/bulkdata/core/maxChunkReadTime`|string| Max time in milliseconds to read during a bulkdata export without type filters. The time should be three quarters of the transactionManager timeout (often the FHIR_TRANSACTION_MANAGER_TIMEOUT value). Note, this value is a string representation of a long value.|
```
lmsurpre, line 2044: I think my preference would have been to put this in seconds and use a number rather than a string (similar to what we did on other config params).  However, I can definitely live with this design as well if you prefer to keep it in ms.

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/impl/AbstractSystemConfigurationImpl.java
```diff
@@ -317,4 +317,10 @@ public int getImportNumberOfFhirResourcesPerRead(String provider) {
     public int getImportInflyRateNumberOfFhirResources(String provider) {
         return IMPORT_INFLY_RATE_NUMOFFHIRRESOURCES;
     }
+
+    @Override
+    public int getPresignedUrlExpiry() {
+        int pageSize = FHIRConfigHelper.getIntProperty("fhirServer/bulkdata/core/cos/presignedExpiry", 86400);
+        return Math.max(1, pageSize);
```
lmsurpre, line 324:
```suggestion
        int expirySeconds = FHIRConfigHelper.getIntProperty("fhirServer/bulkdata/core/cos/presignedExpiry", 86400);
        return Math.max(1, expirySeconds);
```

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/ConfigurationAdapter.java
```diff
@@ -216,13 +216,11 @@
     String getCoreIamEndpoint();
 
     /**
-     * get the tx for the fast endpoint
-     *
-     * @implNote System value.
+     * get the tx for the fast chunk read during an sexport
```
lmsurpre, line 219:
```suggestion
     * get the number ms to read payloads from the persistence layer before stopping to checkpoint
```
note: needs to be updated to say `seconds` if you take my recommendation below

docs/src/pages/guides/FHIRServerUsersGuide.md
prb112: I prefer the milliseconds use here.  FHIR_TRANSACTION_MANAGER_TIMEOUT can be set in fractional seconds.

prb112: Right, I was thinking more like 90.5 seconds or 3.5 seconds.

Issue #1734, opened 2020-11-19
Support token modifiers ':in' and ':not-in'


lmsurpre:
from https://www.hl7.org/fhir/R4/search.html#token
> The search parameter is a URI (relative or absolute) that identifies a value set, and the search parameter tests whether the coding is in the specified value set. The reference may be literal (to an address where the value set can be found) or logical (a reference to ValueSet.url). If the server can treat the reference as a literal URL, it does, else it tries to match known logical ValueSet.url values.

example:

> Search for any condition in the SNOMED CT value set "http://snomed.info/sct?fhir_vs=isa/126851005" that includes all descendants of "Neoplasm of liver"
    ```
    GET [base]/Condition?code:in=http%3A%2F%2Fsnomed.info%2Fsct%3Ffhir_vs%3Disa%2F126851005
    ```

lmsurpre:
The implementation to this one probably would vary a lot based on the size of the valuesets.

For "small" valuesets, we could probably piggy-back on our existing ValueSet expansion support.  For example, we expand the ValueSet if needed, then pass the list of systems+codes into the db as a list of literals via an IN clause. 

At some point, you run into limits on either the query size or on the IN list of literals.  Something like 2,000 or 20,000 on DB2 :-)

Therefor, for larger valuesets, you probably need something more sophisticated where you store cached representations of valuesets.

JohnTimm:
As written, this requires support for "implicit value sets".

csandersdev:
For the CQL Engine evaluation, we are assuming that ValueSet resources will be defined as extensional resources. We believe that support for the :in operator could greatly improve the overall performance of the system because we wouldn't have to move as much data over the wire and would be happy with a first cut that only supports extensional ValueSets.

lmsurpre:
Reduced points because bulk of the work was done with #1980

lmsurpre:
I uploaded two condition resources to my local server.  One with a code of `http://snomed.info/sct|30832001` (Rupture of patellar tendon) and one with `http://snomed.info/sct|10509002` (Acute bronchitis).

Then I configured the fhir-term-graph with SNOMED CT, which supports implicit valuesets.

Next I issued the following two queries:
* https://localhost:9443/fhir-server/api/v4/Condition?code:in=http://snomed.info/sct?fhir_vs=isa/32398004
* https://localhost:9443/fhir-server/api/v4/Condition?code:not-in=http://snomed.info/sct?fhir_vs=isa/32398004

[...]

Issue #1929, opened 2021-02-09
Search SQL needs to be looked at for this search.


d0roppe:
**Describe the bug**
All data referenced is Mock Data:
issuing a search like: {{local}}{{contextRoot}}Claim?patient=Patient/17456ca7efe-dfe09119-26e3-4f7f-939e-ece768ca35ad&use=claim&_include=Claim:*
on a very large data set times out.

**To Reproduce**
See Search above

**Expected behavior**
Expect the search to be run in a timely manor.
**Additional context**
[test-24.docx](https://github.com/IBM/FHIR/files/5953703/test-24.docx)

d0roppe:
verified this search and others speeded up. will report on slow queries from the performance bucket after this fix.

PR #2173, opened 2021-03-30
Issue #1734 - support :in and :not-in modifiers for Token search

michaelwschroeder:

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
@@ -66,6 +66,7 @@ import com.ibm.fhir.model.resource.Location;
 import com.ibm.fhir.model.util.ModelSupport;
 import com.ibm.fhir.persistence.exception.FHIRPersistenceException;
 import com.ibm.fhir.persistence.exception.FHIRPersistenceNotSupportedException;
+import com.ibm.fhir.persistence.jdbc.JDBCConstants;
 import com.ibm.fhir.persistence.jdbc.connection.QueryHints;
 import com.ibm.fhir.persistence.jdbc.dao.api.JDBCIdentityCache;
 import com.ibm.fhir.persistence.jdbc.dao.api.ParameterDAO;
@@ -90,6 +91,7 @@ import com.ibm.fhir.search.parameters.InclusionParameter;
 import com.ibm.fhir.search.parameters.QueryParameter;
 import com.ibm.fhir.search.parameters.QueryParameterValue;
 import com.ibm.fhir.search.util.SearchUtil;
+import com.ibm.fhir.term.util.ValueSetSupport;
 
 /**
  * This is the JDBC implementation of a query builder for the IBM FHIR Server
@@ -1224,36 +1226,42 @@ public class JDBCQueryBuilder extends AbstractQueryBuilder<SqlQueryData> {
                 }
 
                 whereClauseSegment.append(LEFT_PAREN);
-                // Include code
-                whereClauseSegment.append(tableAlias + DOT).append(TOKEN_VALUE).append(operator).append(BIND_VAR);
-                bindVariables.add(SqlParameterEncoder.encode(value.getValueCode()));
+                
+                if (Modifier.IN.equals(queryParm.getModifier()) || Modifier.NOT_IN.equals(queryParm.getModifier())) {
+                    populateValueSetCodesSubSegment(whereClauseSegment, value.getValueCode(), tableAlias);
+                } else {
+                    // Include code
+                    whereClauseSegment.append(tableAlias + DOT).append(TOKEN_VALUE).append(operator).append(BIND_VAR);
+                    bindVariables.add(SqlParameterEncoder.encode(value.getValueCode()));
 
-                // Include system if present.
-                if (value.getValueSystem() != null && !value.getValueSystem().isEmpty()) {
-                    Long commonTokenValueId = null;
-                    if (NE.equals(operator)) {
-                        whereClauseSegment.append(OR);
-                    } else {
-                        whereClauseSegment.append(AND);
+                    // Include system if present.
+                    if (value.getValueSystem() != null && !value.getValueSystem().isEmpty()) {
+                        Long commonTokenValueId = null;
+                        if (NE.equals(operator)) {
+                            whereClauseSegment.append(OR);
+                        } else {
+                            whereClauseSegment.append(AND);
 
-                        // use #1929 optimization if we can
-                        commonTokenValueId = getCommonTokenValueId(value.getValueSystem(), value.getValueCode());
-                    }
+                            // use #1929 optimization if we can
+                            commonTokenValueId = getCommonTokenValueId(value.getValueSystem(), value.getValueCode());
+                        }
 
-                    if (commonTokenValueId != null) {
-                        // #1929 improves cardinality estimation
-                        // resulting in far better execution plans for many search queries. Because COMMON_TOKEN_VALUE_ID
-                        // is the primary key for the common_token_values table, we don't need the CODE_SYSTEM_ID = ? predicate.
-                        whereClauseSegment.append(tableAlias).append(DOT).append(COMMON_TOKEN_VALUE_ID).append(EQ)
+                        if (commonTokenValueId != null) {
+                            // #1929 improves cardinality estimation
+                            // resulting in far better execution plans for many search queries. Because COMMON_TOKEN_VALUE_ID
+                            // is the primary key for the common_token_values table, we don't need the CODE_SYSTEM_ID = ? predicate.
+                            whereClauseSegment.append(tableAlias).append(DOT).append(COMMON_TOKEN_VALUE_ID).append(EQ)
                             .append(commonTokenValueId);
-                    } else {
-                        // common token value not found so we can't use the optimization. Filter the code-system-id
-                        // instead, which ends up being the logical equivalent.
-                        Integer codeSystemId = identityCache.getCodeSystemId(value.getValueSystem());
-                        whereClauseSegment.append(tableAlias).append(DOT).append(CODE_SYSTEM_ID).append(operator)
-                        .append(nullCheck(codeSystemId));
+                        } else {
+                            // common token value not found so we can't use the optimization. Filter the code-system-id
+                            // instead, which ends up being the logical equivalent.
+                            Integer codeSystemId = identityCache.getCodeSystemId(value.getValueSystem());
+                            whereClauseSegment.append(tableAlias).append(DOT).append(CODE_SYSTEM_ID).append(operator)
+                            .append(nullCheck(codeSystemId));
+                        }
                     }
                 }
+                
                 whereClauseSegment.append(RIGHT_PAREN);
                 parmValueProcessed = true;
             }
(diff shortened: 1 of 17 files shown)


michaelwschroeder:
This implementation is working under the following design assumptions:
1. search parameter value must be an absolute ValueSet URL
2. the referenced ValueSet must exist in the registry and be expandable
3. the query will use an IN clause with literal string values for all the codes in the ValueSet

Here's what the query will look like for a value set with codes from a single code system - note the code system IDs being checked are the code system specified in the value set and the default code system (for the case where a resource specifies only a code and not a system value):
```
SELECT 
  R.RESOURCE_ID,R.LOGICAL_RESOURCE_ID,R.VERSION_ID,R.LAST_UPDATED,R.IS_DELETED,R.DATA,LR.LOGICAL_ID 
FROM 
  ALLERGYINTOLERANCE_RESOURCES R 
  JOIN (
    SELECT 
      DISTINCT LR.LOGICAL_RESOURCE_ID,LR.LOGICAL_ID,LR.CURRENT_RESOURCE_ID 
    FROM 
      AllergyIntolerance_LOGICAL_RESOURCES LR 
      JOIN AllergyIntolerance_RESOURCES R ON R.RESOURCE_ID = LR.CURRENT_RESOURCE_ID 
      AND R.IS_DELETED = 'N' 
      JOIN AllergyIntolerance_TOKEN_VALUES_V AS param0 ON (
        param0.PARAMETER_NAME_ID = 1072 
        AND (
          (
            param0.TOKEN_VALUE IN ('active', 'inactive', 'resolved') 
            AND param0.CODE_SYSTEM_ID IN (20099, 20002)
          )
        )
      ) 
      AND LR.LOGICAL_RESOURCE_ID = param0.LOGICAL_RESOURCE_ID
  ) AS LR ON R.LOGICAL_RESOURCE_ID = LR.LOGICAL_RESOURCE_ID 
  AND R.RESOURCE_ID = LR.CURRENT_RESOURCE_ID 
  AND R.IS_DELETED = 'N' 
ORDER BY 
  LOGICAL_RESOURCE_ID ASC OFFSET 0 ROWS FETCH NEXT 10 ROWS ONLY
```

If there are codes from multiple code sets in a value set, then the TOKEN_VALUE/CODE_SYSTEM_ID clauses would be OR'd:
```
            param0.TOKEN_VALUE IN ('1002-5','1004-1','1006-6','1008-2','1010-8') 
            AND param0.CODE_SYSTEM_ID IN (24590, 20002) 
            OR
            param0.TOKEN_VALUE IN ('248153007','248152002','77386006') 
            AND param0.CODE_SYSTEM_ID IN (20057, 20002) 
```

lmsurpre:
I wonder if we'll want to support the same COMMON_TOKEN_VALUE optimization that we made for the normal case...
See #1929 for more detail.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1938,4 +1947,58 @@ public SqlQueryData buildIncludeQuery(Class<?> resourceType, FHIRSearchContext s
...
+        Map<String, Set<String>> codeSetMap = ValueSetSupport.getCodeSetMap(ValueSetSupport.getValueSet(parameterValue));
+        for (String codeSetUrl : codeSetMap.keySet()) {
+            Set<String> codes = codeSetMap.get(codeSetUrl);
+            if (codes != null) {
+                // Strip version from canonical codeSet URL. We don't store version in TOKEN_VALUES
+                // table so will just ignore it.
+                int index = codeSetUrl.lastIndexOf("|");
+                if (index != -1) {
+                    codeSetUrl = codeSetUrl.substring(0, index);
+                }
```
lmsurpre, line 1980: *should* we store version in token_values? 🤔

lmsurpre: under what circumstances would we get a versioned url here?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
```diff
@@ -601,7 +601,7 @@ protected void buildWhereClause(StringBuilder whereClause, String overrideType)
                                     .replaceAll(AS + PARAMETER_TABLE_ALIAS, AS + paramTableAlias)
                                     .replaceAll(PARAMETER_TABLE_NAME_PLACEHOLDER, valuesTable);
 
-                            if (Modifier.NOT.equals(param.getModifier())) {
+                            if (Modifier.NOT.equals(param.getModifier()) || Modifier.NOT_IN.equals(param.getModifier())) {
```
lmsurpre, line 604: almost seems to easy :-)

tbieste, line 604: Update comments on line 609 and 646.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -1293,32 +1295,48 @@ private static void checkSearchParameterCombinations(Class<?> resourceType, List
...
-                            null, null, Collections.singletonList(valueParameterValue));
+                        QueryParameter valueParameter = new QueryParameter(Type.TOKEN, SearchUtil.makeCompositeSubCode(ofTypeParmName,
+                            SearchConstants.OF_TYPE_MODIFIER_COMPONENT_VALUE), null, null, Collections.singletonList(valueParameterValue));
                         parameterValue.addComponent(valueParameter);
                     }
-              } else
-                if (parts.length == 2) {
+                } else if (Modifier.IN.equals(modifier) || Modifier.NOT_IN.equals(modifier)) {
+                    // Validate that the parameter value is a ValueSet URL that points to a registered ValueSet that is
+                    // expanded.
```
lmsurpre, line 1325: should this say "expandable"?

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -1293,32 +1295,48 @@ private static void checkSearchParameterCombinations(Class<?> resourceType, List
...
+                    ValueSet valueSet = ValueSetSupport.getValueSet(v);
+                    if (valueSet == null) {
+                        String msg = "ValueSet '" + v + "' specified for search parameter '" + searchParameter.getCode().getValue() +
+                                "' with modifier ':" + modifier.value() + "' could not be found";
+                        throw SearchExceptionUtil.buildNewInvalidSearchException(msg);
+                    }
+//                    ValueSetSupport.expand(valueSet);
```
lmsurpre, line 1332: delete this?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
michaelwschroeder: From what John told me, and looking at the code in ValueSetSupport.computeCodeSetMap() to verify, if a version is specified for the code system in the value set expansion, that version is appended. If no version is specified, a `|<version unknown>` is appended. So actually there should always be a version, meaning I probably don't need the check on line 1978, although I don't mind keeping it there as a fail-safe.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
michaelwschroeder: Done.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
michaelwschroeder: Yes - done.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
michaelwschroeder: Yes, thank you. Done.

michaelwschroeder:
> 
> 
> I wonder if we'll want to support the same COMMON_TOKEN_VALUE optimization that we made for the normal case...
> See #1929 for more detail.

Not sure we can do that here since we're dealing with a list of values.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1938,4 +1947,58 @@ public SqlQueryData buildIncludeQuery(Class<?> resourceType, FHIRSearchContext s
...
+                // <parameterTableAlias>.TOKEN_VALUE IN (...)
+                whereClauseSegment.append(tokenValuePredicateString)
+                    .append("'").append(String.join("','", codes)).append("'")
+                    .append(RIGHT_PAREN);
+
+                // AND <parameterTableAlias>.CODE_SYSTEM_ID IN ({n}, {default-code-system-id})
+                whereClauseSegment.append(AND).append(codeSystemIdPredicateString)
+                    .append(nullCheck(identityCache.getCodeSystemId(codeSetUrl)))
+                    .append(COMMA).append(defaultCodeSystemId)
+                    .append(RIGHT_PAREN);
```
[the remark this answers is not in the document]
michaelwschroeder: After discussion we decided that we will not match target resources that have code values with no system. This means `code` fields for which we can't find an implicit code system will never match using`:in`. Given that the number of that type of field is very small, it was felt that this was the better option vs. over-matching (allowing codes with no system to match).

michaelwschroeder:
> 
> 
> > I wonder if we'll want to support the same COMMON_TOKEN_VALUE optimization that we made for the normal case...
> > See #1929 for more detail.
> 
> Not sure we can do that here since we're dealing with a list of values.

Since issue #1733 will use very similar SQL constructs, use of COMMON_TOKEN_VALUE optimization will be investigated as part of that work.

fhir-persistence/src/test/java/com/ibm/fhir/persistence/search/test/AbstractSearchTokenTest.java
```diff
@@ -328,6 +368,30 @@ public void testSearchToken_Coding_chained_missing() throws Exception {
         assertSearchDoesntReturnComposition("subject:Basic.missing-Coding:missing", "false");
     }
 
+    @Test
+    public void testSearchToken_Coding_in() throws Exception {
+        assertSearchDoesntReturnSavedResource("Coding:in", "http://hl7.org/fhir/ValueSet/concept-property-type");
```
tbieste, line 373: For these Code/Coding/CodeableConcept tests. Maybe it would be good to have at least one ":in" test that does return the saved resource and at least one ":not-in" test that doesn't return the saved resource, to cover both sides.

lmsurpre:
> > > I wonder if we'll want to support the same COMMON_TOKEN_VALUE optimization that we made for the normal case...
> > > See #1929 for more detail.
> > 
> > 
> > Not sure we can do that here since we're dealing with a list of values.
> 
> Since issue #1733 will use very similar SQL constructs, use of COMMON_TOKEN_VALUE optimization will be investigated as part of that work.

Mike opened #2184 for making this possible, so I think this PR is good-to-go.

lmsurpre (approved):
LGTM

tbieste (approved):
LGTM

Issue #1186, opened 2020-06-02
Add javadoc and convenience methods for Base64Binary


lmsurpre:
**Is your feature request related to a problem? Please describe.**
I was trying to use Base64Binary.Builder and noticed that the Builder.value(String) method doesn't have javadoc.

**Describe the solution you'd like**
We need to add that so that we know whether to pass a string that has already been base64-encoded or if it will automatically base64-encoded string that was passed.

Also, I'd like there to be a `Base64Binary.of(byte[])` method

PR #2182, opened 2021-03-31
issue #1186 - minor javadoc update

lmsurpre:
super minor

### DIFF
--- a/fhir-model/src/main/java/com/ibm/fhir/model/type/Base64Binary.java
+++ b/fhir-model/src/main/java/com/ibm/fhir/model/type/Base64Binary.java
@@ -55,7 +55,7 @@ public class Base64Binary extends Element {
      * Factory method for creating Base64Binary objects from a byte array; this array should be the actual value.
      * 
      * @param value
-     *     The byte array of the previously encoded base64 content
+     *     The byte array of to-be-encoded content
      */
     public static Base64Binary of(byte[] value) {
         return Base64Binary.builder().value(value).build();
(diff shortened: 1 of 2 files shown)


fhir-model/src/main/java/com/ibm/fhir/model/type/Base64Binary.java
```diff
@@ -33,7 +33,7 @@ private Base64Binary(Builder builder) {
 
     /**
      * The actual value
-     * 
+     *
```
prb112, line 36: I'm happy to approve this... however... This is a generated class, perhaps the generator updates are missing?

lmsurpre: good call

prb112 (approved):

JohnTimm:
With the `Base64Binary` class we have this relationship:

```
Base 64 encoded string (java.lang.String) <-----> Unencoded binary data (byte[])
```

My preference would be to using a naming convention that captures that relationship. So for the `of` factory method that sets the binary data directly we would say something like:

```
Factory method for creating Base64Binary objects from a byte array of unencoded binary data

@param
    a byte array containing unencoded binary data
```

lmsurpre:
@JohnTimm you had previously approved https://github.com/IBM/FHIR/pull/2157 and so I assumed you were fine with what was there.  Based on that, I closed #1186.  Feel free to open a new issue and/or PR if you'd like to see further changes here.

Issue #1846, opened 2021-01-05
Add compartment variants to generated swagger / openapi


lmsurpre:
**Is your feature request related to a problem? Please describe.**
The IBM FHIR Server supports accessing data via compartment search as defined at https://www.hl7.org/fhir/http.html#vsearch

However, these variants are not listed in our generated swagger / openapi definitions

**Describe the solution you'd like**
Generated swagger / openapi definition should include these endpoints.

In the case that users have filtered the resource types to include, include only the passed resource types in the compartment.
In the case of the one-definition-per-resource-type case, I think we should generate 1 definition per compartment.
For example, `patient-swagger.json` would only have the current Patient-scoped operations (e.g. `GET Patient/[id]`) and we'd have a separate definition like `patient-compartment-swagger.json` which would have the compartment operations like `GET  Patient/[id]/Observation`

**Describe alternatives you've considered**
Instead of 1 definition per compartment, we could list the compartment apis underneath the generated api definition for the compartment resources.  For example, the patient-swagger.json would have both `GET Patient/[id]/Observation` and `GET Patient/[id]/Observation`.

**Additional context**

**Suggestion for how to perform QA**
1) For swagger: Go to https://editor.swagger.io/ and copy/paste generated swagger and verify it looks correct.
2) For openapi: Start up fhir-openapi server, (e.g. https://localhost:9443/openapi/ui/) and verify it looks correct.

lmsurpre:
I loaded the all-in-one-swagger that we publish with the server by navigating to https://localhost:9443/openapi/ui and I can see the compartment variants under the appropriate resources.

I discussed with @tbieste about whether it makes more sense to have all the compartment-related endpoints grouped under their compartment header, but we decided to keep it this way because otherwise those compartment sections are so long that its easy to get lost on the UI.

I also verified that, when a filter is passed, only the desired resource endpoints are added to the compartment definitions. Further, I confirmed that when all Patient compartment resources are filtered out, the tool skips generation of the Patient compartment swagger definition altogether.

Looks good!

PR #2192, opened 2021-04-01
Issue #1846 - Add compartment variants to generated swagger/openapi

tbieste:
Issue #1846 - Add compartment variants to generated swagger/openapi

### DIFF
--- a/fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
+++ b/fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
@@ -21,6 +21,7 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -102,6 +103,8 @@ import com.ibm.fhir.model.type.code.ResourceType;
 import com.ibm.fhir.model.util.FHIRUtil;
 import com.ibm.fhir.model.util.ModelSupport;
 import com.ibm.fhir.model.visitor.AbstractVisitable;
+import com.ibm.fhir.search.compartment.CompartmentUtil;
+import com.ibm.fhir.search.exception.FHIRSearchException;
 import com.ibm.fhir.search.util.SearchUtil;
 import com.ibm.fhir.swagger.generator.APIConnectAdapter;
 
@@ -112,7 +115,7 @@ import com.ibm.fhir.swagger.generator.APIConnectAdapter;
  * By default, this class will generate:
  * <ol>
  * <li> an "all-in-one" OpenAPI definition for the entire api
- * <li> a separate OpenAPI definition for each and every resource type; each with all HTTP interactions enabled
+ * <li> a separate OpenAPI definition for each resource type and compartment, with all HTTP interactions enabled
  * </ol>
  *
  * <p>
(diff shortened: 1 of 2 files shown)


fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
```diff
@@ -934,20 +1141,33 @@ private static void generateSearchViaPostPathItem(Class<?> modelClass, JsonObjec
...
-             * "<name>": { "type": "string" }
-             */
-            propertyValues.add("type", "string");
-            propertyValues.add("description", searchParameter.getDescription().getValue());
-
-            parameters.add(name, propertyValues);
-        }
+        //
+        // As of OpenAPI 3.0.0, empty form fields are still passed to the REST API, which is not valid
+        // for the FHIR search API.
```
lmsurpre, line 1146: very interesting.  do we allow it in lenient mode?
is there any discussion in the OpenAPI community about whether this is considered a tooling bug?

fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
```diff
@@ -612,6 +775,30 @@ private static void generatePaths(Class<?> modelClass, JsonObjectBuilder paths,
         // TODO: add patch
```
lmsurpre, line 775: forgot about this TODO; wonder if we should open a placeholder issue for this one too...

lmsurpre (commented):
My biggest concern with this one is that it makes the all-in-one openapi a lot longer.  Did you happen to try loading it from the liberty openapi ui?

lmsurpre (approved):
LGTM

fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
tbieste: I did find this: https://github.com/swagger-api/swagger-ui/issues/5303
It seems to indicate there should be a checkbox on the UI to choose whether parameters are passed on the REST call, but I wasn't sure about how to move to a newer version.
I can try and see what happens in lenient mode.

tbieste:
> My biggest concern with this one is that it makes the all-in-one openapi a lot longer. Did you happen to try loading it from the liberty openapi ui?
Yes, I did load it in the openapi ui.

fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
tbieste: With lenient, any parameters for which empty string is invalid are skipped over, but parameters such has _id, code, etc. which allow an empty string, cause the search to not find anything since there are no matches to those values.

tbieste: If I use https://editor.swagger.io/ there is a "Send empty values" checkbox on the form fields. However, with the openliberty openapi UI the IBM FHIR Server currently uses, the checkbox are not shown. So, for now, I'd prefer to just not show those fields. The user can still pass in the form fields on the UI in the request body.

lmsurpre: thanks for checking on it.

fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
lmsurpre: opened https://github.com/IBM/FHIR/issues/2202 for this...feel free to add.

I also realized we don't document our support for the "conditional" variants (conditional create, update, and delete), but I supposed we need to draw the line somewhere...

Issue #1385, opened 2020-07-30
Moving is_deleted flag to logical_resources will eliminate many joins to resources tables


punktilious:
**Is your feature request related to a problem? Please describe.**
Searches always need to join to the <resource-type>_RESOURCES table to filter on the value for IS_DELETED. This flag represents the current state of the logical resource and so should be defined at the LOGICAL_RESOURCE level, not the RESOURCE (which represents a version of LOGICAL_RESOURCE.

**Describe the solution you'd like**
Investigate if it should be added to the global table rather than the resource-specific table.
Add the column to the logical resources table
Migrate data
Remove column from the resources tables
reorg
Update the search query generation as required, removing joins to the resources table unless they are required. The only reason to join with resources should be when the lastUpdated value is needed, and the final join to fetch the data blob.

**Describe alternatives you've considered**
Status quo, and accept more expensive joins.

**Additional context**
N/A

lmsurpre:
> Team discussion - added a comment that we have to consider VERSION specific status which is also on the RESOURCES table.

I think what that means is that we would need IS_DELETED on *both* the RESOURCES and LOGICAL_RESOURCES tables.
The one on RESOURCES will continue to mean that this particular version of the resource is deleted.
The one on LOGICAL_RESOURCES will indicate with the "latest" version is deleted

lmsurpre:
This flag was added under #1958, but work remains to actually use it.

lmsurpre:
Now that we have https://github.com/IBM/FHIR/issues/2014 I'm not sure if we still need this one.  If so, how do they relate?

d0roppe:
Ran the performance bucket against this code and all searches were processed.  No odd behavior.  Closing issue as the new code seems to be working.

PR #2193, opened 2021-04-01
WIP:  issue #1385 - remove xx_RESOURCES join from the default FROM clause

lmsurpre:
Its not super clean, but I wanted to see if it was possible to get some
isolated benefits from #1385 until we are ready to adopt the new query
builder.

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
@@ -67,10 +67,10 @@ public class QuerySegmentAggregator {
     protected static final String SYSTEM_LEVEL_SELECT_ROOT =
             "SELECT RESOURCE_ID, LOGICAL_RESOURCE_ID, VERSION_ID, LAST_UPDATED, IS_DELETED, DATA, LOGICAL_ID ";
     protected static final String SYSTEM_LEVEL_SUBSELECT_ROOT = SELECT_ROOT;
-    protected static final String SELECT_COUNT_ROOT = "SELECT COUNT(DISTINCT R.LOGICAL_RESOURCE_ID) ";
+    protected static final String SELECT_COUNT_ROOT = "SELECT COUNT(DISTINCT LR.LOGICAL_RESOURCE_ID) ";
     protected static final String SYSTEM_LEVEL_SELECT_COUNT_ROOT = "SELECT SUM(CNT) ";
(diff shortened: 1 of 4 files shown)


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
```diff
@@ -633,19 +640,25 @@ protected void buildWhereClause(StringBuilder whereClause, String overrideType)
...
             } // end if SKIP_WHERE
         } // end for
 
         // If there were any query parameters with :missing, :not, or :not-in modifier, append the missingOrNotModifierWhereClause
         if (missingOrNotModifierWhereClause.length() > 0) {
             whereClause.append(missingOrNotModifierWhereClause.toString());
+            whereWordAdded = true;
+        }
+
+        if (!SortedQuerySegmentAggregator.class.equals(this.getClass())) {
```
lmsurpre, line 660: this was one of the hackiest lines.  we probably don't want to merge this...
or if we do merge it, it would just be a stopgap til we have a more elegant solution

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
```diff
@@ -416,10 +417,12 @@ protected void buildFromClause(StringBuilder fromClause, String simpleName) {
...
-        fromClause.append(JOIN);
-        processFromClauseForLastUpdated(fromClause, simpleName);
-        fromClause.append(" R ON R.RESOURCE_ID = LR.CURRENT_RESOURCE_ID ")
-                .append(" AND R.IS_DELETED = 'N' ");
+        if (!queryParmLastUpdateds.isEmpty()) {
+            fromClause.append(JOIN);
+            processFromClauseForLastUpdated(fromClause, simpleName);
+            fromClause.append(" R ON R.RESOURCE_ID = LR.CURRENT_RESOURCE_ID ")
+                    .append(" AND R.IS_DELETED = 'N' ");
+        }
```
lmsurpre, line 425: this is basically the only change I was trying to make...avoid the join with the xx_RESOURCES tables (unless we have a lastUpdated query because that was a bigger change than I wanted to take on)

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/QuerySegmentAggregator.java
lmsurpre: it was needed because the SortedQuerySegmentAggregator actually adds a JOIN clause *after* calling this buildWhereClause method...so in the case of SortedQuerySegmentAggregator I add this NOT_DELETED clause after that instead.

lmsurpre:
closed in favor of Robin's #2309

Issue #923, opened 2020-04-13
Improve the docker image


lmsurpre:
Follow-on from #837 

Currently, we install our own server.xml and our own server, then delete the defaultServer.
However, now that we're based on OpenLiberty, we should follow the guidance at https://github.com/OpenLiberty/ci.docker#building-an-application-image

Here's at least 3 things that I think we could clean up/take advantage of:

1.  The ibm-fhir-server container prints this warning during startup:
```
[AUDIT   ] CWWKG0102I: Found conflicting settings for defaultKeyStore instance of keyStore configuration.
  Property password has conflicting values:
    Secure value is set in file:/opt/ol/wlp/usr/servers/fhir-server/configDropins/defaults/keystore.xml.
    Secure value is set in file:/opt/ol/wlp/usr/servers/fhir-server/server.xml.
  Property password will be set to the value defined in file:/opt/ol/wlp/usr/servers/fhir-server/server.xml.
```

We might be able to workaround that by convincing Liberty not to mixin their own keystore.xml, but a better option would be to just let liberty generate its own keystore instead of pre-packaging ours.
For more info, see https://github.com/OpenLiberty/ci.docker/blob/master/SECURITY.md#providing-a-custom-keystore

2.  Annotation caching defined at https://developers.redhat.com/blog/2020/01/28/use-red-hat-openshifts-built-in-oauth-server-as-an-authentication-provider-in-open-liberty/

This should help us with the startup performance, which I feel is noticably slower since we moved to OpenLiberty 20.0.0.3 (not sure if its about OpenLiberty vs WebSphere Liberty or about 19.0.0.12 vs 20.0.0.3)

3.  Class caching (when using OpenJ9) as defined at https://github.com/OpenLiberty/ci.docker#openj9-shared-class-cache-scc

lmsurpre:
Currently we use the "full" image.
There's a open issue with startup time on this image:
https://github.com/OpenLiberty/ci.docker/issues/174

However, it looks like this should be improved in future versions: https://github.com/OpenLiberty/ci.docker/pull/175

To move to the "kernel" version, we need to move our server name from `fhir-server` to `defaultServer` (or ask liberty to make their `configure.sh` script more flexible).

lmsurpre:
#2196 addresses points number 2 and 3 from the description by extending `kernel-slim` instead of `full` and using the Liberty-managed `features.sh` and `configure.sh` to install the minimal set of required features and set up the shared class cache (respectively).

[...]

PR #2196, opened 2021-04-02
issue #923 - docker image updates

lmsurpre:
1. Dockerfile now extends liberty `kernel-slim` instead of `full`; this
required us to use the defaultServer server instead of fhir-server, but
let us take advantage of OpenJ9's Shared Class Cache for faster start
times (at the expense of longer docker build times)
2. Added logging environment variables to get basic log and trace
messages sent to console by default; everything is done via environment
variable so that it can be overridden (at first I used a
bootstrap.properties file but I found out that takes precedent over the
environment variables set at runtime)
3. Removed the configResources library from server.xml and bulkdata.xml;
it turns out we weren't loading our config files from the classpath and
so we didn't need this.


fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/control/JobControlContextListener.java
```diff
@@ -30,6 +32,9 @@
     @Override
     public void contextInitialized(ServletContextEvent sce) {
         log.info("Starting up the Bulk Data Web App");
+
+        FHIRConfiguration.setConfigHome(System.getenv("FHIR_CONFIG_HOME"));
```
prb112, line 36: I just double checked javadoc

``` 
Only throws if the name passed to getenv
    NullPointerException - if name is null**
```

fhir-install/Dockerfile
```diff
@@ -41,32 +43,20 @@ COPY --chown=1001:0 --from=base /opt/ibm-fhir-server /opt/ibm-fhir-server
...
-    yum clean all && \
-    rm -rf /var/cache/yum
-
-USER 1001
+RUN features.sh && \
+    VERBOSE=true configure.sh
 
 ENTRYPOINT ["/opt/ibm-fhir-server/bootstrap.sh"]
-CMD ["/opt/ol/wlp/bin/server", "run", "fhir-server"]
+CMD ["/opt/ol/wlp/bin/server", "run", "defaultServer"]
```
prb112, line 62: as I look at this... I realize that this really deserves a highlight of some sorts... maybe in the release notes or a quick video on getting started?

fhir-install/src/main/docker/ibm-fhir-server/bootstrap.properties
```diff
@@ -0,0 +1,3 @@
+# disable writing to trace.log by only sending trace data to console
+com.ibm.ws.logging.trace.format=BASIC
+com.ibm.ws.logging.trace.file.name=stdout
```
prb112, line 3: Question: bootstrap.properties is still overridden by the corresponding environment variables?  or does this take precedence?

prb112 (approved):
LGTM, left two comments (not necessarily for changes, rather to confirm/recommend)

fhir-install/Dockerfile
lmsurpre: we should *for sure* highlight this change in the release notes...anyone extending this image needs to know about this change

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/control/JobControlContextListener.java
lmsurpre: yeah, i checked it too....if the env var isn't set, it will pass null.  and our contract for setConfigHome treats that like "".

fhir-install/src/main/docker/ibm-fhir-server/bootstrap.properties
lmsurpre: strangely, there are no corresponding environment variables defined by liberty for these.  instead, we defined our own and using those in our server.xml.   and the server.xml settings override the bootstrap properties.

prb112: great

tbieste (approved):
LGTM

Issue #1732, opened 2020-11-19
Support token modifier ':text'


lmsurpre:
from https://www.hl7.org/fhir/R4/search.html#token:
> The search parameter is processed as a string that searches text associated with the code/value - either CodeableConcept.text, Coding.display, or Identifier.type.text. In this case, the search functions as a normal string search

Note: by "normal string search" I assume they mean it should be a "starts-with" and "case-insensitive" match.

example:
> Search for any Condition with a code that has a text "headache" associated with it (either in the text, or a display)
    ```
GET [base]/Condition?code:text=headache | 
    ```

lmsurpre:
I don't believe we even index these values today and it could be expensive (lots of new string values in our STR_VALUES tables).

There is a different issue for searching longer / more sophisticated text search:  https://github.com/IBM/FHIR/issues/481

michaelwschroeder:
Running against my local server with the latest codebase, I have performed token searches using the :text modifier against multiple fields and resources and have verified:
- CodeableConcept.text and Coding.display fields have been indexed properly
- search against TOKEN search parameters referencing those fields is case-insensitive and correct results are returned.

PR #2197, opened 2021-04-02
Issue #1732 - Support token modifier ':text'

tbieste:
Issue #1732 - Support token modifier ':text'

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
@@ -401,6 +401,14 @@ public class JDBCParameterBuildingVisitor extends DefaultVisitor {
         for (Coding coding : codeableConcept.getCoding()) {
             visit(elementName, elementIndex, coding);
         }
+        if (codeableConcept.getText() != null && codeableConcept.getText().hasValue()) {
+            // Extract as token as normalized string since :text modifier is simple string search
+            TokenParmVal p = new TokenParmVal();
+            p.setResourceType(resourceType);
+            p.setName(searchParamCode + SearchConstants.TEXT_MODIFIER_SUFFIX);
+            p.setValueCode(SearchUtil.normalizeForSearch(codeableConcept.getText().getValue()));
+            result.add(p);
+        }
         return false;
     }
 
@@ -418,6 +426,14 @@ public class JDBCParameterBuildingVisitor extends DefaultVisitor {
                 p.setValueSystem(coding.getSystem().getValue());
             }
             result.add(p);
+            if (coding.getDisplay() != null && coding.getDisplay().hasValue()) {
+                // Extract as token as normalized string since :text modifier is simple string search
+                p = new TokenParmVal();
+                p.setResourceType(resourceType);
+                p.setName(searchParamCode + SearchConstants.TEXT_MODIFIER_SUFFIX);
+                p.setValueCode(SearchUtil.normalizeForSearch(coding.getDisplay().getValue()));
+                result.add(p);
+            }
         }
         return false;
     }
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
@@ -1194,12 +1194,18 @@ public class JDBCQueryBuilder extends AbstractQueryBuilder<SqlQueryData> {
         StringBuilder whereClauseSegment = new StringBuilder();
         String operator = this.getOperator(queryParm, EQ);
         boolean parmValueProcessed = false;
+        boolean appendEscape;
         SqlQueryData queryData;
         List<Object> bindVariables = new ArrayList<>();
         String tableAlias = paramTableAlias;
+        String queryParmCode = queryParm.getCode();
 
-        String code = queryParm.getCode();
-        if (!QuerySegmentAggregator.ID.equals(code)) {
+        if (!QuerySegmentAggregator.ID.equals(queryParmCode)) {
+
+            // Append the suffix for :text modifier
+            if (Modifier.TEXT.equals(queryParm.getModifier())) {
+                queryParmCode += SearchConstants.TEXT_MODIFIER_SUFFIX;
+            }
 
             // Only generate NOT EXISTS subquery if :not modifier is within chained query;
             // when :not modifier is within non-chained query QuerySegmentAggregator.buildWhereClause generates the NOT EXISTS subquery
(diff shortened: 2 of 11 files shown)


fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -2102,18 +2105,18 @@ public static QueryParameter parseChainedInclusionCriteria(QueryParameter inclus
...
+     * @param value the string to normalize
+     * @return the normalized string
      */
     public static String normalizeForSearch(String value) {
 
         String normalizedValue = null;
         if (value != null) {
             normalizedValue = Normalizer.normalize(value, Form.NFD).replaceAll("\\p{InCombiningDiacriticalMarks}+", "");
-            normalizedValue = normalizedValue.toLowerCase();
+            normalizedValue = normalizedValue.replaceAll("\\s+", " ").toLowerCase();
```
michaelwschroeder, line 2119: Curious why the removal of consecutive whitespace characters - is that standard? I assume this affects string search as well as token :text search.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
```diff
@@ -418,6 +426,14 @@ public boolean visit(java.lang.String elementName, int elementIndex, Coding codi
...
             }
             result.add(p);
+            if (coding.getDisplay() != null && coding.getDisplay().hasValue()) {
+                // Extract as token as normalized string since :text modifier is simple string search
+                p = new TokenParmVal();
+                p.setResourceType(resourceType);
+                p.setName(searchParamCode + SearchConstants.TEXT_MODIFIER_SUFFIX);
+                p.setValueCode(SearchUtil.normalizeForSearch(coding.getDisplay().getValue()));
+                result.add(p);
+            }
```
michaelwschroeder, line 436: Don't we also want to index the Identifier.type.text field?

lmsurpre: > Don't we also want to index the Identifier.type.text field?

Interesting thought, but how would you search on that things display text?  I don't think you can layer modifiers like `reference:identifier:text` ...if we supported that I bet we'd be the only ones.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1215,23 +1221,34 @@ private SqlQueryData processTokenParm(Class<?> resourceType, QueryParameter quer
...
                 whereClauseSegment.append(LEFT_PAREN);
-                
+
                 if (Modifier.IN.equals(queryParm.getModifier()) || Modifier.NOT_IN.equals(queryParm.getModifier())) {
                     populateValueSetCodesSubSegment(whereClauseSegment, value.getValueCode(), tableAlias);
                 } else {
                     // Include code
                     whereClauseSegment.append(tableAlias + DOT).append(TOKEN_VALUE).append(operator).append(BIND_VAR);
-                    bindVariables.add(SqlParameterEncoder.encode(value.getValueCode()));
+                    if (LIKE.equals(operator)) {
```
lmsurpre, line 1242: Under what scenario would we have a LIKE here?

I was actually thinking line 1195 should always return EQ, so maybe we could just remove that variable altogether?  Maybe in the case of `:in` and `:not-in` (and soon `:above`/`:below`) it could be something different, but I think we handle those separately today.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
michaelwschroeder: I was assuming they were talking about the identifier search parameter - i.e. `identifier:text=abc`, but reading more closely, I guess the spec just says `Identifier.type.text`, which to your point could be any Identifier field, so I'm ok with not supporting that one.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
lmsurpre: never mind, now I get it.  we set LIKE because we want it to behave similar to our normal (no-modifier) string search behavior.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -1194,12 +1194,18 @@ private SqlQueryData processTokenParm(Class<?> resourceType, QueryParameter quer
...
+        String queryParmCode = queryParm.getCode();
 
-        String code = queryParm.getCode();
-        if (!QuerySegmentAggregator.ID.equals(code)) {
+        if (!QuerySegmentAggregator.ID.equals(queryParmCode)) {
+
+            // Append the suffix for :text modifier
+            if (Modifier.TEXT.equals(queryParm.getModifier())) {
+                queryParmCode += SearchConstants.TEXT_MODIFIER_SUFFIX;
+            }
```
lmsurpre, line 1208: I double-check and this is consistent with how we handled the `:identifier` modifier for reference search

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
lmsurpre: Oh, I was thinking of it in terms of being able to search on the text portion of an `of-type`.  But now I see that it does say `:text` is supposed to apply to `Identifier.type.text`.  Weird.  I'm not really sure when/how that would be used, but maybe we *should* support it :shrug:

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
tbieste: In the FHIR spec, https://www.hl7.org/fhir/search.html#string, it mentions that "Punctuation and non-significant whitespace (e.g. repeated space characters, tab vs space) should also be ignored."  This change was to normalize all whitespace to be handled the same.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
michaelwschroeder: I think it makes sense to support it for search parameters of type TOKEN where the search parameter field is of type Identifier, but not in other cases (like a Reference.identifier field). I also noticed :text is discussed in the spec here: https://www.hl7.org/fhir/search.html#modifiers, and it does not mention searching on the Identifier field, so I'm good with saying we're not supporting it as well:

`For token: :text (the match does a partial searches on the text portion of a CodeableConcept or the display portion of a Coding), instead of the default search which uses codes. Other defined modifiers are :in, :below, :above and :not-in which are described below.
`

tbieste: Yeah, the part of the spec that Mike just referenced is the part I prefer to follow, and feels the most consistent.

There's another section where it lists:

`:text | The search parameter is processed as a string that searches text associated with the code/value - either CodeableConcept.text, Coding.display, or Identifier.type.text.`

However, there's no examples of it and the other section doesn't mention Identifier.type.text. I'm wonder if it just lists it because `Identifier.type` is itself a CodeableConcept, so if a search parameter was defined for it, then it would make sense for :text to then do matches against Identifier.type.text.

lmsurpre: I support this interpretation.  However, in this case I think we should open a spec bug so we have something to point to if anyone ever asks about it.

michaelwschroeder (approved):

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
lmsurpre: I opened https://jira.hl7.org/browse/FHIR-31708 for it.

PR #2200, opened 2021-04-05
issue #923 - run features.sh earlier to enable layer caching

lmsurpre:
this change splits the COPY /opt/ol/wlp/usr /opt/ol/wlp/usr into two
steps:
* first we move just the server.xml and configDropins
* then, after installing the liberty features, we add the application
* finally, we add the bootstrap.sh script and related tools

### DIFF
--- a/fhir-install/Dockerfile
+++ b/fhir-install/Dockerfile
@@ -7,13 +7,10 @@
 
 FROM openliberty/open-liberty:21.0.0.3-kernel-slim-java8-openj9-ubi as base
 
-ENV LICENSE accept
-
 USER root
-
 RUN yum install -y unzip
-
-RUN mkdir -p /opt/ibm-fhir-server
+RUN install -d -o 1001 /opt/ibm-fhir-server
+USER 1001
 
 COPY target/fhir-server-distribution.zip /tmp/
 RUN unzip -qq /tmp/fhir-server-distribution.zip -d /tmp && \
@@ -22,13 +19,12 @@ RUN unzip -qq /tmp/fhir-server-distribution.zip -d /tmp && \
 COPY src/main/docker/ibm-fhir-server/bootstrap.properties /opt/ol/wlp/usr/servers/defaultServer/
 COPY src/main/docker/ibm-fhir-server/bootstrap.sh /opt/ibm-fhir-server/
 
-USER 1001
-
 # ----------------------------------------------------------------------------
 # Stage: Runnable
 
 FROM openliberty/open-liberty:21.0.0.3-kernel-slim-java8-openj9-ubi
 
+ARG VERBOSE=true
 ARG FHIR_VERSION=4.7.0-SNAPSHOT
 
 # The following labels are required: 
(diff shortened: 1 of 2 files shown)


fhir-install/Dockerfile
```diff
@@ -52,11 +48,21 @@ ENV FHIR_CONFIG_HOME=/opt/ol/wlp/usr/servers/defaultServer \
     TRACE_FILE=stdout \
     TRACE_FORMAT=BASIC
 
+COPY target/LICENSE /licenses/
```
prb112, line 51: isn't there a root /LICENSE ?

fhir-install/Dockerfile
```diff
@@ -52,11 +48,21 @@ ENV FHIR_CONFIG_HOME=/opt/ol/wlp/usr/servers/defaultServer \
...
+COPY --chown=1001:0 --from=base /opt/ol/wlp/usr/servers/defaultServer/configDropins /opt/ol/wlp/usr/servers/defaultServer/configDropins
+
+RUN features.sh
+
+COPY --chown=1001:0 --from=base /opt/ol/wlp/usr /opt/ol/wlp/usr
+
+RUN configure.sh
+
+COPY --chown=1001:0 --from=base /opt/ibm-fhir-server /opt/ibm-fhir-server
+
```
prb112, line 63: ah - the split we talked about... looks great

fhir-install/Dockerfile
```diff
@@ -29,6 +29,7 @@ USER 1001
 
 FROM openliberty/open-liberty:21.0.0.3-kernel-slim-java8-openj9-ubi
 
+ARG VERBOSE=true
```
prb112, line 32: maybe set build arg to false by default?

prb112 (approved):
LGTM

tbieste (approved):
LGTM

fhir-install/Dockerfile
lmsurpre: this wasn't really a change from before...I just moved it lower for some reason (I think I was grouping the COPYs).  that said, I don't really know that its needed.

fhir-install/Dockerfile
lmsurpre: I've found the VERBOSE output to be useful to see whats going on, and I don't see much downside to including it

fhir-install/Dockerfile
lmsurpre: It does look like the one under target here is a copy of our root LICENSE file:
```
            <plugin>
                <artifactId>maven-antrun-plugin</artifactId>
                <version>3.0.0</version>
                <executions>
                    <execution>
                        <phase>package</phase>
                        <configuration>
                            <target>
                                <copy
                                    todir="${project.build.directory}">
                                    <fileset
                                        dir="${project.build.directory}/../../">
                                        <include name="LICENSE" />
                                    </fileset>
                                </copy>
                            </target>
                        </configuration>
                        <goals>
                            <goal>run</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
            ```
            
I think we just need it here so that it exists under the context of the directory being passed to docker (because docker doesn't let you pass file from outside of this context)

Issue #2207, opened 2021-04-06
Patch Replaces Can Save Invalid Types


ssnowski:
**Describe the bug**
When saving with patch, it seems like it can corrupt existing resources, preventing them from being retrieved properly.  This seems to be possible by making a request to save an element as a different type than it was previously (EG: save a valueString in the Patient.extension field (type: Extension).

**Environment**
IBM FHIR Server 4.6.1 via docker image.   presumably 4.6.0 as well, but untested

**To Reproduce**
1. Setup:
```
POST to fhir-server/api/v4:
{
        "resourceType": "Bundle",
        "type": "transaction",
        "entry": [
        {
                "resource": {
                        "id": "1234",
                        "resourceType": "Patient",
                        "identifier": [{
                                "system": "ibm.com/mrn",
                                "value": "1234"
                        }],
                        "extension": [
                          {
                            "url": "https://ibm.com/ext1",
                            "valueString": "testExtension"
                          }
                        ]
                },
                "request": {
                        "url": "Patient/1234",
                        "method": "PUT"
                }
        }
        ]
}
```

2. Corrupt the data:
```
POST to fhir-server/api/v4:
{
  "resourceType": "Bundle",
  "type": "transaction",
  "entry": [
    {
      "request": {
        "url": "Patient/1234",
        "method": "PATCH"
      },
      "resource": {
        "resourceType": "Parameters",
        "parameter": [
          {
            "name": "operation",
            "part": [
              {
                "name": "type",
                "valueCode": "replace"
              },
              {
                "name": "value",
                "valueString": "this isn't an extension"
              },
              {
                "name": "path",
                "valueString": "Patient.extension[0]"
              }
            ]
          }
        ]
      }
    }
  ]
}
```
3. See the error:
```
GET to fhir-server/api/v4/Patient/1234
```

**Expected behavior**
I would have expected step 2 to fail, returning a 4xx error with some text about a type mismatch.

**Additional context**

tbieste:
@lmsurpre Attempted the steps to reproduce and it returns with a 500 Internal Server Error with the following response:

{
    "resourceType": "OperationOutcome",
    "issue": [
        {
            "severity": "fatal",
            "code": "exception",
            "details": {
                "text": "FHIRPatchException: An error occurred while replacing the value"
            }
        }
    ]
}

Instead of getting some type of 4xx error, is this the expected behavior?

tbieste:
[...]

PR #2209, opened 2021-04-07
issue #2207 - add type-checking for lists in fhir-model constructors

lmsurpre:
if a clever user adds objects of the wrong type to the builder (via
reflection), we were never catching this and so the data could become
corrupted


JohnTimm:
Per our conversation, I think that:
```java
if (Objects.isNull(element)) {
    throw new IllegalStateException(String.format("Repeating element: '%s' does not permit null elements", elementName));
}
Class<?> elementType = element.getClass();
if (!type.isAssignableFrom(elementType)) {
    throw new IllegalStateException(String.format("Invalid type: %s for list element: '%s' must be: %s", elementType.getSimpleName(), elementName, type.getSimpleName()));
}
```
or
```java
if (Objects.isNull(element)) {
    throw new IllegalStateException(String.format("Repeating element: '%s' does not permit null elements", elementName));
}
if (!type.isInstance(element)) {
    throw new IllegalStateException(String.format("Invalid type: %s for list element: '%s' must be: %s", element.getClass().getSimpleName(), elementName, type.getSimpleName()));
}
```
are both easier to read / understand than the current structure because it is painfully obvious that we are performing a null check followed by a type check on each list element. Combining things into an outer if and inner if-then-else structure seems less clear to me. Other than that, LGTM.

fhir-model/src/main/java/com/ibm/fhir/model/util/ValidationSupport.java
```diff
@@ -370,6 +374,35 @@ private static SchemaFactory createSchemaFactory() {
         return elements;
     }
 
+    /**
+     * @return the same list that was passed
+     * @throws IllegalStateException if the passed list contains any null objects or objects of an incompatible type
```
michaelwschroeder, line 379: also throws IllegalStateException if passed list is empty

michaelwschroeder (approved):
LGTM - one minor comment

fhir-model/src/main/java/com/ibm/fhir/model/util/ValidationSupport.java
lmsurpre: updated the javadoc to cover this

JohnTimm (approved):
LGTM

PR #2210, opened 2021-04-07
issue #2207 - consistently throw FHIRPatchException for bad patches

lmsurpre:
Previously, which exception you got was dependent on what was wrong with
the patch.

And prior to the model changes for this same issue, we actually allowed
invalid patches to be applied (in the case of adding, inserting, and
replacing on a list).


JohnTimm (approved):
LGTM (assuming that the changes to model classes work themselves out once the other PR is merged)

PR #2227, opened 2021-04-08
issue #2207 - throw a 400 for bad patches

lmsurpre:
previously we threw 500 Sever Error, even in cases where we knew it was
likely a bad patch.  now we favor 400 errors for those, but unexpected
errors should still result in 500 errors


tbieste (approved):
LGTM

JohnTimm (approved):
LGTM

PR #2228, opened 2021-04-09
add missing definitions to the metatadata swagger

lmsurpre:
specifically, we were missing the definitions for Resource and
DomainResource, which are referenced from the CapabilityStatement
definition


tbieste (approved):
LGTM

Issue #2241, opened 2021-04-13
Squelch opd-0 warnings during startup


lmsurpre:
**Is your feature request related to a problem? Please describe.**
Our startup log is getting cleaner and cleaner, but there's still a number of info messages that include the word "warning" on each start.
I believe they comes from validating OperationDefinition resources during initialization (or something like that) and it should be possible to get these cleaned up.

**Describe the solution you'd like**
Either 
A. add a name to the artifacts that are missing one, so that we don't have this issue any more; or
B. turn these log messages into trace messages instead

**Describe alternatives you've considered**

**Acceptance Criteria**
1.  
   WHEN you start the server with the default config
   THEN there are no log messages about `invariant:warning:opd-0`

**Additional context**
The issue may be rooted in published spec artifacts.  If so, we'll need to make a decision about whether to patch the spec artifacts or not.

prb112:
Confirmed - does not show on info, it does show with fine in the trace.log

```
[2021-04-13T19:54:20.922+0000] 00000038 id=00000000 com.ibm.fhir.server.operation.FHIROperationRegistry          1 Found FHIROperation implementation class: com.ibm.fhir.operation.document.DocumentOperation
[2021-04-13T19:54:21.216+0000] 00000038 id=00000000 com.ibm.fhir.server.operation.FHIROperationRegistry          1 Issue: invariant:warning:opd-0: Name should be usable as an identifier for the module by machine processing applications such as code generation
```

Issue #2057, opened 2021-03-11
Ensure all OperationDefinition resources are pulled from FHIRRegistry for specification-defined operations (e.g. $everything)


JohnTimm:
Previously, we would include a copy of the OperationDefinition resource in the implementation of that operation. The FHIRRegistry provides access to all spec artifacts and, as a best practice, should be used to access definitions for specification-defined operations.

lmsurpre:
Maybe we can tackle this one together with https://github.com/IBM/FHIR/issues/2057

prb112:
I tried with the $everything operation and $convert and $document and $validate.
All looks good

PR #2242, opened 2021-04-13
issues #2241 and #2057 - load operation from registry and skip warning

lmsurpre:
With this pull request, the spec-defined OperationDefinitions are now
pulled from the registry instead of having them copied to each
individual module.

Additionally, because so many of these spec-defined OperationDefinitions
violate opd-0 (`name.matches('[A-Z]([A-Za-z0-9_]){0,254}')`), I toned
down the logging so that only validation errors are logged by default
(and not validation warnings).


prb112 (approved):
LGTM

fhir-server/src/main/java/com/ibm/fhir/server/operation/FHIROperationRegistry.java
```diff
@@ -91,8 +91,10 @@ private boolean isValid(FHIROperation operation) throws FHIRValidationException,
         List<Issue> issues = FHIRValidator.validator().validate(opDef);
         if (!issues.isEmpty()) {
             for (Issue issue : issues) {
-                log.info("Issue: " + issue.getCode().getValue() + ":"
-            + issue.getSeverity().getValue() + ":" + issue.getDetails().getText().getValue());
+                if (log.isLoggable(Level.FINE)) {
+                    log.fine("Issue: " + issue.getCode().getValue() + ":"
```
prb112, line 95: this is where the warning is squashed for the opDef?

lmsurpre: yes, you got it

lmsurpre: I was hoping that getting the up-to-date operation definitiions would resolve the issue, but it didn't.  So rather than patch the spec artifacts, I just toned down the logging.

JohnTimm (approved):
LGTM

PR #2248, opened 2021-04-15
Avoid use of injected httpHeaders and uriInfo when in client mode

lmsurpre:
This change is needed to allow fhir-provider to execute in other
environments such as from within a standalone RestEasy client.

I also converted the static 'isPretty' helper to a member function to
allow access to the instance-level RuntimeType.

I also added javadoc to help distinguish FHIRProvider from
FHIRJsonProvider.


prb112 (approved):
LGTM - I had to read the isPretty method directly at https://github.com/IBM/FHIR/blob/d97ff4d16feafddd54dfc6025fdd49ead4a1b9a9/fhir-provider/src/main/java/com/ibm/fhir/provider/FHIRProvider.java#L140

Also, it makes sense given the use-case described.

JohnTimm (approved):
LGTM

PR #2252, opened 2021-04-15
Move FHIRUrlParser from fhir-provider to fhir-server

lmsurpre:
I actually like having this one in fhir-provider, but I was trying to
slim down `fhir-provider` and it was easier to move this than to remove
its unusual dependency on httpclient.

For those using just fhir-provider (and its runtime dependencies), its not a huge savings (9.6 MB to 8.1 MB) but its 4 fewer libs and it'll be nice not to pull in a whole http client for what should be a simple layer on top of fhir-model.


lmsurpre:
Side-by-side comparison (after the change is on the left):
image

tbieste (approved):
LGTM

prb112 (approved):
LGTM. - I'd like John to look at it

PR #2259, opened 2021-04-21
AuthzPolicyEnforcement changes

lmsurpre:
1. getPatientRefValue now uses ReferenceUtil to obtain the id of the
target Patient being referenced. This technique matches what we do
during the extraction of the compartment membership param and is more
robust to special reference values like versioned references or absolute
references that match the base server url.

2. getPatientIdFromToken will now split the patient_id claim on spaces.
This sets us up for scenarios where a single token can grant access to
multiple patient records. Additionally, I removed the log statement
about expecting patient_id to be a string and not a list.

3. Patient search is now allowed. Because we convert all searches into
compartment searches, there is no longer a risk of exposing whether a
given Patient resource by a given id exists or not. Therefore, we don't
need to prevent such searches any longer. That said, because a Patient
is not in its own compartment (unless it has a corresponding
Patient.link), Patient search behavior will likely to be confusing to
clients and so operators may want to continue to disallow Patient search 
from the REST layer in select deployments.

### DIFF
--- a/fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
+++ b/fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
@@ -8,13 +8,13 @@ package com.ibm.fhir.smart;
 
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import com.auth0.jwt.JWT;
 import com.auth0.jwt.interfaces.Claim;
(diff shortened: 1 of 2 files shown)


JohnTimm (approved):
LGTM

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
```diff
@@ -539,10 +537,12 @@ private String getAccessToken() throws FHIRPersistenceInterceptorException {
 
         String patientId = claim.asString();
         if (patientId == null) {
-            log.fine("Found patient_id claim was expected to be a string but is not; processing as a list");
             return claim.asList(String.class);
         }
 
-        return Collections.singletonList(patientId);
+        return Stream.of(patientId.split(" "))
```
JohnTimm, line 543: Should we split on \s+ here?

michaelwschroeder (approved):
LGTM

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
lmsurpre: Good question.  I think we're ok with supporting just space-delimited values for now, but do let me know if you feel we should change it.

d0roppe (approved):
LGTM as far as providing a fix for the issue.  I don't know much about how the code works.

Issue #2198, opened 2021-04-05
Implement a remote term service provider using a REST client


JohnTimm:
We now have a registry-based terminology service provider (`RegistryTermServiceProvider`) and a graph-based terminology service provider (`GraphTermServiceProvider`). If we would like to integrate an existing terminology service (e.g. SNOMED-CT endpoint, LOINC endpoint, ontoserver, etc.), then we will need to implement a remote term service provider (`RemoteTermServicdeProvider`) that uses a REST client to forward requests to another server.

kmbarton423:
Confirmed successful lookup on SNOMED-CT
---
---
added the following to fhir-server-config.json (same level as "core"):

        "term": {
             "remoteTermServiceProviders": [{
                "__comment": "Configuration for public SNOMED-CT endpoint",
                "enabled": true,
                "base": "https://snowstorm-fhir.snomedtools.org/fhir",
                "supports": [{
                    "system": "http://snomed.info/sct"
                }]
            }]
        },
		
added trustDefaultCerts="true" to server.xml:

<ssl clientAuthenticationSupported="true" id="defaultSSLConfig" keyStoreRef="defaultKeyStore" sslProtocol="TLSv1.2" trustStoreRef="defaultTrustStore" trustDefaultCerts="true"/>
		

GET  https://localhost:9443/fhir-server/api/v4/CodeSystem/$lookup?system=http://snomed.info/sct&code=195967001

PR #2261, opened 2021-04-21
Issue #2198 - getConcept should return null for responses other than 200

JohnTimm:

`RemoteTermServiceProvider.getConcept` throws an Exception for responses other than OK (200). However, in order to be aligned with the `getConcept` contract, we should return null instead.


lmsurpre (approved):
LGTM

Issue #2263, opened 2021-04-21
Support client opt-in for skipping the update if a resource hasn't changed


lmsurpre:
**Is your feature request related to a problem? Please describe.**
In scenarios where the FHIR Server is not the source of truth, some clients might want to reload/refresh the server with their data on some nightly basis.
For severs that don't supporting versioning, you can basically just replace the old data with the new data and be done.
IBM FHIR Sever doesn't support this mode...we always write a new version of a resource when we receive a PUT.

Currently, to avoid unneccessary PUTs, clients will need to either keep track of what has been uploaded or perform a read before they perform their updates.

What would be nice is if we could build that logic into the server and avoid performing the update when the resource contents havn't actually changed.  This would avoid the cost of historical versions filling up the database and obviate the need for some kind of historical version cleanup.

The other behavior of a PUT is that we re-extract all of the search parameters.
Because skipping the update would also skip this part, I think it would be best to force users to opt-in to this special no-op "fast update".

**Describe the solution you'd like**
Allow users to pass a special header, maybe X-FHIR-UPDATE-IF-MODIFIED, to opt into this special optimization.  When present (and with a value of "true"?), perform a comparison of the prior version of this resource with the new version that was just sent in.

Unfortunately, we can't do a normal equals, because we want to explicitly ignore the server-assigned Resource.meta.lastUpdated and Resource.meta.versionId fields.

**Describe alternatives you've considered**
At some point in the future (probably after we get https://github.com/IBM/FHIR/issues/2155 in place), it would be neat to automatically convert a PUT with no content change into a reindex.  Then, based on the proposal in #2155 we would skip the reindex if the extracted parameters havn't changed either.

**Acceptance Criteria**
1.  
   GIVEN an existing resource in the server
   WHEN we get an update request
    AND the X-FHIR-UPDATE-IF-MODIFIED header is set to true
   THEN avoid performing the update
    AND return the existing resource as appropriate (per the `Prefer: return` request)

**Additional context**
Similar, but different, from #2050

JohnTimm:
Verified this with @lmsurpre using Postman.

PR #2265, opened 2021-04-22
issue #2263 - support skippable updates

lmsurpre:
1. add a skippableUpdate flag to the FHIRResourceHelpers interface
2. set this flag from a new header X-FHIR-UPDATE-IF-MODIFIED
3. if skippableUpdate is true, use the resource fingerprinter to compare
the resource from the update request to the resource in the database and
skip the update when they match


fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -456,6 +458,34 @@ private FHIRRestOperationResponse doPatchOrUpdate(String type, String id, FHIRPa
...
+                                    .code(IssueType.INFORMATIONAL)
+                                    .details(CodeableConcept.builder()
+                                        .text(string("Update resource matches the existing resource; skipping the update"))
+                                        .build())
+                                    .build())
+                                .build());
+
+                        return ior; // early exit
+                    }
+                }
```
lmsurpre, line 488: this is the meat of the change

prb112:
Checked to see if build needed a rebuild.  FYI, Build fails with Error 

```
Error:  Failures: 
Error:    UpdateTest.testUpdateIfModified:134 expected:<2> but was:<5>
```

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
prb112: sorry, I know I'm not a reviewer, what happens in the Audit Path?  what data is sent on the Audit? is it consistent?

lmsurpre: great question.  presently it will emit an audit log just like any other update, except that the prevResource will match the newResource.
```
                RestAuditLogger.logUpdate(httpServletRequest,
                        ior != null ? ior.getPrevResource() : null,
                        ior != null ? ior.getResource() : null,
                        startTime, new Date(), status);
```

lmsurpre: let me know if you think we need to do something different

JohnTimm:
Now that the `ResourceFingerprintVisitor` is being used in the REST layer should we move it into the `fhir-model` visitor or util package?

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
prb112: then I think we're good... it was good to confirm.  thanks!

lmsurpre:
> Now that the ResourceFingerprintVisitor is being used in the REST layer should we move it into the fhir-model visitor or util package?
Sure.  I can do that.

prb112 (approved):

docs/src/pages/guides/FHIRPerformanceGuide.md
```diff
@@ -31,15 +30,18 @@ permalink: /FHIRServerPerformanceGuide/
...
+    - [5.2 Conditional Update](#52-conditional-update)
+- [6 Client Access Scenarios](#5-client-access-scenarios)
+    - [6.1 Read](#61-read)
+    - [6.2 Version Read](#62-version-read)
+    - [6.3 History](#63-history)
+    - [6.4 Search Performance](#64-search-performance)
+    - [6.5 Search Examples](#65-search-examples)
+    - [6.6 Tools](#66-tools)
+    - [6.7 Making FHIR Requests With curl](#67-making-fhir-requests-with-curl)
+    - [6.8 Making FHIR Requests with IBM FHIR Server Client](#68-making-fhir-requests-with-ibm-fhir-server-client)
```
lmsurpre, line 44:
```suggestion
- [6.8 Making FHIR Requests with IBM FHIR Server Client](#68-making-fhir-requests-with-the-ibm-fhir-server-client)
```

Issue #2281, opened 2021-04-23
Populate Bundle.entry.response.outcome with validation warnings


lmsurpre:
**Is your feature request related to a problem? Please describe.**
While reviewing our bundle processing code, I noticed that we pass validation warnings back through Bundle.entry.resource (at least in some cases).

This is the proper location for errors, but for warnings the spec defines a supplemental location: `Bundle.entry.response.outcome`

From the spec:
```
For a POST/PUT operation, this is the equivalent outcome that would be returned for prefer = operationoutcome - except that the resource is always returned whether or not the outcome is returned.
This outcome is not used for error responses in batch/transaction, only for hints and warnings. In a batch operation, the error will be in Bundle.entry.response, and for transaction, there will be a single OperationOutcome instead of a bundle in the case of an error.
```

**Describe the solution you'd like**
Lets start adding our validation outcomes to `Bundle.entry.response.outcome`.

However, I still like having clients be able to control whether the resource content comes back in the Bundle.entry.resource or not, so I propose we do this:
* always populate `Bundle.entry.response.outcome` with validation hints / warnings
* when return=minimal (default), omit the resource contents from `Bundle.entry.resource`
* when return=OperationOutcome, write an interaction-related (not validation-related) OperationOutcome in the `Bundle.entry.resource`
* when return=representation, write the resulting resource contents to the response `Bundle.entry.resource`

**Describe alternatives you've considered**

**Acceptance Criteria**

**Additional context**
There is plenty of room for cleanup here as well.
1. We loop through the entire bundle for each method, whereas we really only need to iterate this list once or twice. 
2. We create a responseBundle up front and then we insert the entries one-at-a-time while we process them.  But that was designed when our model is immutable...each "insert" is actually copying the bundle into a new builder and then building a brand new bundle.  All of this copying is unnecessary and could slow things down / bloat the memory usage for large bundle requests.

Issue #864, opened 2020-03-29
DELETE non existing resource return 200 code with "fatal" issue


smhgit:
I don't know if this can be consider as a bug or not, but I would expect different code return if resource wasn't found (although technically, the command was successful)

![image](https://user-images.githubusercontent.com/10852598/77844137-79343680-71ac-11ea-94dd-d32e63440877.png)

lmsurpre:
Usually best to turn to the spec for this sort of thing:
 https://www.hl7.org/fhir/r4/http.html#delete

In this case, I don't see much guidance on the "proper" behavior for deleting a resource that never existed, but there is this guidance on deleting a previously-deleted resource:
> Performing this interaction on a resource that is already deleted has no effect, and the server should return either a 200 OK if the response contains a payload, or a 204 No Content with no response payload.

In our case, the response contains a payload that explains that the resource doesn't exist, so I think we're OK to return a 200.

Also, we use the AEGIS TouchStone conformance-testing tool and we're passing all their delete tests, so thats a good sign.

However, one thing that I think we *should* clean up is the `OperationOutcome.issue.severity`...IMHO that should be `informational` (or `warning` at worst).

Some other options I'd also be ok with (if they pass touchstone tests):
A. return 404 for deleting a resource that doesn't exist and ever existed.  the downside here is that deleting a resource that doesn't exist behaves different from deleting a resource that did exist but was deleted
B. return 204 with no body by default.  This has the advantage of making the deletion of a non-existent resource consistent with the behavior of deleting a previously-deleted resource, but there there is a downside to that too (client might think they succeeded in delete when they actually performed the delete on the wrong id).

But even in option B, I think we'd want to return a 200 OK + an OperationOutcome warning if the user has set their return preference to `OperationOutcome` (https://www.hl7.org/fhir/http.html#ops)

kmbarton423:
don't write an exception to the log

lmsurpre:
It looks to me like we already dropped the severity of this OO to "warning":
```
{
    "resourceType": "OperationOutcome",
    "issue": [
        {
            "severity": "warning",
            "code": "not-found",
            "details": {
                "text": "Cannot find Patient with id 'Patient00002'."
            }
        }
    ]
}
```

Still need to check the logs to ensure we're not logging this like a server error.
Additionally, I think there is work to make this behave as expected when the delete is part of a batch/transaction bundle.

lmsurpre:
[...]

PR #2282, opened 2021-04-23
issues #2281 and #864 - batch/transaction bundle processing changes

lmsurpre:
1. FHIRRestHelper.validateBundle now returns a sparse array (modeled as
a map) of bundle response entries with errors / warnings. It used to construct an
entire response bundle.

2. FHIRRestHelper.processEntriesByMethod replaces
FHIRRestHelper.processEntriesForMethod and iterates the request entries
a single time, instead of once-per-method.

3. Entries are collected into an array instead of "modifying" the
responseBundle with each one (which forced lots of needless copying
since our model is now immutable).

4. Stop duplicating custom operation results in the
`Bundle.entry.response.outcome` field of the response. We already copy
the operation's result in the `Bundle.entry.resource` and that should be
sufficient.

I also made significant formatting changes in FHIRRestHelper.

### DIFF
--- a/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
@@ -1183,7 +1183,7 @@ public class FHIRRestHelper implements FHIRResourceHelpers {
         FHIRRequestContext requestContext = FHIRRequestContext.get();
 
         try {
-            // First, validate the bundle and create the response bundle.
+            // First, validate the bundle and save the error / warning responses by index entry.
             Map<Integer, Entry> validationResponseEntries = validateBundle(inputBundle);
 
             // Next, process each of the entries in the bundle.
@@ -1621,7 +1621,7 @@ public class FHIRRestHelper implements FHIRResourceHelpers {
                     + bundleRequestCorrelationId);
             }
 
-            // Clear both correlation id fields since we're done processing the bundle.
+            // Clear the request correlation id field since we're done processing the bundle.
             bundleRequestCorrelationId = null;
 
             if (txn != null) {
--- a/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
@@ -8,6 +8,7 @@ package com.ibm.fhir.server.util;
 
 import static com.ibm.fhir.model.type.String.string;
 import static com.ibm.fhir.model.util.ModelSupport.getResourceType;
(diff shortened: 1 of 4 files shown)


fhir-server/src/test/java/com/ibm/fhir/server/util/FHIRRestHelperTest.java
```diff
@@ -141,10 +123,10 @@ public void testTransactionBundlePostSingle() throws Exception {
         assertNotNull(responseBundle);
         assertEquals(1, responseBundle.getEntry().size());
         Bundle.Entry entry = responseBundle.getEntry().get(0);
-        assertEquals(ALL_OK, entry.getResource());
```
lmsurpre, line 144: @michaelwschroeder the order of the assert arguments in this class are all messed up.  I fixed the ones that I touched, but I *did not* wholistically fix them in the file (nor check for it anywhere else in the project).
I think its because we use JUnit style asserts in some places and testng ones in others.  Something to watch for going forward.

JohnTimm (approved):
LGTM

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -1639,7 +1623,6 @@ private Bundle processBundleEntries(Bundle requestBundle, Bundle responseBundle)
 
             // Clear both correlation id fields since we're done processing the bundle.
```
michaelwschroeder, line 1624: Update comment: "Clear request correlation id field..."

michaelwschroeder (approved):
LGTM - just a couple minor comments

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -1181,12 +1184,10 @@ public Bundle doBundle(Bundle inputBundle) throws Exception {
 
         try {
             // First, validate the bundle and create the response bundle.
```
lmsurpre, line 1186:
```suggestion
            // First, validate the bundle and save the error / warning responses by index entry.
```

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -1181,12 +1184,10 @@ public Bundle doBundle(Bundle inputBundle) throws Exception {
...
-            Bundle responseBundle = validateBundle(inputBundle);
+            Map<Integer, Entry> validationResponseEntries = validateBundle(inputBundle);
```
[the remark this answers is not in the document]
lmsurpre: updated the comment

Issue #1814, opened 2020-12-11
FHIR primitive static factory helpers should be more resilient to nulls


lmsurpre:
**Is your feature request related to a problem? Please describe.**
When populating an object from a different source, its convenient to use the static factory methods like Code.of() like the following
```
Coding.builder()
    .code(Code.of(myObj.getMyCode()))
    .display(string("Pregnant"))
    .build()
```

However, if that source data can contain nulls, then today this breaks on the `Code.of(null)` with the following:
```
java.lang.IllegalStateException: ele-1: All FHIR elements must have a @value or children
```

This despite the following code working with no issue:
```
Coding.builder()
    .code(null)
    .display(string("Pregnant"))
    .build()
```

**Describe the solution you'd like**
`Code.of(null)` and similar convience functions should return null if null is passed in.  Instead of trying to create a privitive FHIR object with no value which is guaranteed to fail with `ele-1`.

**Describe alternatives you've considered**
Add javadoc to explain it will throw if you pass a null.

lmsurpre:
Team decision:
1. update javadoc to say that the arguments to the factory methods must be non-null
2. add Object.requireNonNull to all the generated factory methods (with the name of the argument so we know what was null)

michaelwschroeder:
I have validated the javadoc and the addition of Object.requireNotNull() method calls via visual inspection of classes in the fhir-model project and have validated that when attempting to pass a null value to these factory methods a NullPointerException is thrown indicating the name of the argument containing the null value.

PR #2287, opened 2021-04-25
issue #1814 - add javadoc and requireNonNull in factory methods

lmsurpre:


prb112 (approved):
LGTM, the CodeGenerator seems pretty straightforward

JohnTimm:
This seems redundant as we already check for null in the constructor of the object being built.

JohnTimm (approved):
LGTM

lmsurpre:
> This seems redundant as we already check for null in the constructor of the object being built.

Specifically this was the documented team decision at https://github.com/IBM/FHIR/issues/1814#issuecomment-783491677

I think the biggest benefit is that we'll get a better error message for these cases.
Additionally, I think the generated NPE is more applicable for a factory method than the previous IllegalStateException (which is more appropriate for `Builder.build()`).

Issue #2169, opened 2021-03-29
Rename fhir-model code subtype nested classes from ValueSet to Value


lmsurpre:
**Is your feature request related to a problem? Please describe.**

When we first designed the `fhir-model` classes for R4, we decided to model Code types with a required binding in the base spec as their own classes that extend Code and the bound valueset as a nested class called ValueSet.

However, as I've been working with these enums more and more, the name `ValueSet` has started to bug me.
Each enum value is like an instance of the Enum and so there is some dissonance when we define a member like this:
`MyCode.ValueSet myVal;`

Because myVal is actually a specific value for MyCode, not its ValueSet.

**Describe the solution you'd like**
1. Add a peer to the nested ValueSet classes and override functions as necessary to make the values equivalent to the existing ValueSet values.
2. Deprecate the old ValueSet classes and remove all references to these throughout the codebase.
3. After a couple releases, remove the old ValueSet classes; probably in combination with the introduce of support for #2068 

**Describe alternatives you've considered**
live with the ValueSet name for these nested classes

**Acceptance Criteria**

**Additional context**

PR #2293, opened 2021-04-27
issue #2169 - change getValueConstant to getValueAsEnum and update project code

lmsurpre:
The first commit has the model change (getValueConstant  -> getValueAsEnum) and not much else.
I suggest focusing on the second commit for the review.


fhir-model/src/main/java/com/ibm/fhir/model/util/FHIRUtil.java
```diff
@@ -131,9 +131,9 @@ public static String toString(Visitable visitable) {
     private static Pattern buildReferencePattern() {
         StringBuilder sb = new StringBuilder();
         sb.append("((http|https)://([A-Za-z0-9\\\\\\/\\.\\:\\%\\$\\-])*)?(");
-        sb.append(Arrays.asList(ResourceType.ValueSet.values()).stream()
-            .map(v -> v.value())
-            .collect(Collectors.joining("|")));
+        sb.append(ModelSupport.getResourceTypes(false).stream()
+                .map(t -> ModelSupport.getTypeName(t))
+                .collect(Collectors.joining("|")));
```
prb112, line 136: Interesting... so we allowed Resource/1 and DomainResource/1 as valid references before. 
This is a good catch.

fhir-search/src/main/java/com/ibm/fhir/search/util/ReferenceUtil.java
```diff
@@ -30,8 +30,8 @@
     private static final String URN = "urn:";
 
     // A set of resource type names to assist computing the base URL string
-    private static final Set<String> resourceTypes = Arrays.stream(FHIRResourceType.ValueSet.values())
-            .map(FHIRResourceType.ValueSet::value)
+    private static final Set<String> resourceTypes = Arrays.stream(FHIRResourceType.Value.values())
+            .map(FHIRResourceType.Value::value)
```
prb112, line 34: I think there was an early piece of code that did a similar thing and used ModelSupport with the false.  Do we want that here too?

fhir-model/src/main/java/com/ibm/fhir/model/util/FHIRUtil.java
lmsurpre: yes, very astute

prb112 (approved):
LGTM - left two comments... one about a nice change, and the other about applying the change in a different spot potentially.

fhir-search/src/main/java/com/ibm/fhir/search/util/ReferenceUtil.java
lmsurpre: sure, i can make that change

lmsurpre: done

JohnTimm (approved):
LGTM

Issue #2284, opened 2021-04-24
Expand support of X-FHIR-UPDATE-IF-MODIFIED to batch/transaction entries


lmsurpre:
**Is your feature request related to a problem? Please describe.**
In https://github.com/IBM/FHIR/issues/2263 we added support for a server-side update optimization.
However, this optmization was only implemented for PUT requests.
Now I'd like to expand that to apply to batch/transaction interactions that contain PUTs as well.

**Describe the solution you'd like**
When users set X-FHIR-UPDATE-IF-MODIFIED to true, each PUT request within the bundle will be processed according to the "skippableUpdate" processing.

**Describe alternatives you've considered**

**Acceptance Criteria**
1.  
   GIVEN an existing resource in the server
   WHEN we get an update request as part of a batch request
    AND the X-FHIR-UPDATE-IF-MODIFIED header is set to true
   THEN avoid performing the update
    AND return the existing resource as appropriate (per the `Prefer: return` request)

**Additional context**

kmbarton423:
Ran a variety of bundles with PUT requests (both type=batch and type=transaction) with a mix of resources that were and were not equivalent to what already persisted.  Confirmed the following OperationOutcome was received for those resources where equivalency was true and X-FHIR-UPDATE-IF-MODIFIED was test to true:

'''        {
            "resource": {
                "resourceType": "OperationOutcome",
                "issue": [
                    {
                        "severity": "information",
                        "code": "informational",
                        "details": {
                            "text": "Update resource matches the existing resource; skipping the update"
                        }
                    }
                ]
            },
'''

PR #2295, opened 2021-04-28
issue #2284 - support skippable updates for batch/transaction methods

lmsurpre:
If `X-FHIR-UPDATE-IF-MODIFIED` is set to true on the request, then all
updates (including patches) in the bundle will be processed in accordance 
with this update optimization.

### DIFF
--- a/fhir-path/src/main/java/com/ibm/fhir/path/util/FHIRPathUtil.java
+++ b/fhir-path/src/main/java/com/ibm/fhir/path/util/FHIRPathUtil.java
@@ -905,6 +905,11 @@ public final class FHIRPathUtil {
 
         FHIRPathTree tree = evaluator.getEvaluationContext().getTree();
         FHIRPathNode parentNode = tree.getParent(node);
+
+        if (parentNode == null) {
+            throw new FHIRPatchException("Unable to compute the parent for '" + elementName + "';" +
+                    " a FHIRPathPatch replace FHIRPath must select a node with a parent", fhirPath);
+        }
         Visitable parent = parentNode.isResourceNode() ?
                 parentNode.asResourceNode().resource() : parentNode.asElementNode().element();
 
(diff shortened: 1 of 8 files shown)


fhir-path/src/main/java/com/ibm/fhir/path/util/FHIRPathUtil.java
```diff
@@ -905,6 +905,11 @@ public static String unescape(String s) {
 
         FHIRPathTree tree = evaluator.getEvaluationContext().getTree();
         FHIRPathNode parentNode = tree.getParent(node);
+
+        if (parentNode == null) {
+            throw new FHIRPatchException("Unable to compute the parent for '" + elementName + "';" +
+                    " a FHIRPathPatch replace FHIRPath must select a node with a parent", fhirPath);
+        }
```
lmsurpre, line 912: this is a somewhat unrelated change that might merit its own separate issue, but i found this NPE bug when I added a patch test and got the patch wrong

fhir-server-test/src/test/java/com/ibm/fhir/server/test/BundleTest.java
```diff
@@ -2331,6 +2334,94 @@ public void testTransactionConditionalDeletes() throws Exception {
         assertGoodGetResponse(responseBundle.getEntry().get(1), Status.OK.getStatusCode(), HTTPReturnPreference.MINIMAL);
     }
 
+    /**
+     * Test transaction bundle put with post local reference dependency.
```
michaelwschroeder, line 2338: Update testcase description to match what is being tested.

michaelwschroeder (approved):
LGTM

fhir-server-test/src/test/java/com/ibm/fhir/server/test/BundleTest.java
lmsurpre:
```suggestion
     * Sets UPDATE_IF_MODIFIED_HEADER_NAME  and posts a transaction bundle with an update and a patch; both should be skipped on the server
```

JohnTimm (approved):
LGTM

PR #2278, opened 2021-04-23
Issue #1551 - perform token search based on codeSystem case-sensitivity

michaelwschroeder:
Signed-off-by: Mike Schroeder <mschroed@us.ibm.com>

PR #2298, opened 2021-04-28
bump project modules to 4.8.0-SNAPSHOT

lmsurpre:
I also tried changing our modules to reference fhir-examples 4.7.1 (instead of the snapshot), but it turns out we have already merged https://github.com/IBM/FHIR/pull/2278/files which relies on the latest snapshot.

### DIFF
--- a/fhir-parent/pom.xml
+++ b/fhir-parent/pom.xml
@@ -5,7 +5,7 @@
 
     <groupId>com.ibm.fhir</groupId>
     <artifactId>fhir-parent</artifactId>
-    <version>4.7.0-SNAPSHOT</version>
+    <version>4.8.0-SNAPSHOT</version>
     <packaging>pom</packaging>
     <name>IBM FHIR Server</name>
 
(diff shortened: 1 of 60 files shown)


fhir-parent/pom.xml
```diff
@@ -62,8 +62,8 @@
         <fhir-persistence-jdbc.index>MINIMAL_JSON</fhir-persistence-jdbc.index>
         <fhir-server-test.index>MINIMAL_JSON</fhir-server-test.index>
         <java.version>1.8</java.version>
-        <fhir-examples.version>4.7.0-SNAPSHOT</fhir-examples.version>
-        <fhir-tools.version>4.7.0-SNAPSHOT</fhir-tools.version>
+        <fhir-examples.version>4.7.1</fhir-examples.version>
```
lmsurpre, line 65: if any modules rely on unreleased fhir-examples, we will need to flip this back to the latest snapshot

prb112: In my erase work, I'll flip back to snapshot

prb112 (approved):
LGTM

fhir-parent/pom.xml
lmsurpre: it turns out we already merged a commit that relied on snapshot, so i reverted this part of the change

PR #2302, opened 2021-04-29
Issue 1627

michaelwschroeder:


fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -1685,6 +1658,24 @@ private Bundle processBundleEntries(Bundle requestBundle, Map<Integer, Entry> va
                         "HEAD" + requestEntriesByMethod.get(HTTPVerb.Value.HEAD));
             }
 
+            // If we're working on a 'transaction' type interaction, or an interaction
+            // where we're processing only GET requests, then start a new transaction now.
+            BundleType.Value bundleType = requestBundle.getType().getValueAsEnum();
+            if (bundleType == BundleType.Value.TRANSACTION ||
```
punktilious, line 1664: Perhaps encapsulate this logic in a utility function? Would make it easier to unit-test just this logic.

lmsurpre: > Also, this requires several passes through the bundle. If the bundle is large, this is inefficient. Why not make one pass through the bundle and collect the different HTTPVerb values? Maybe a visitor if that fits?

I didn't understand this part of the comment. We iterate the entry list and group them by method above.
The added code here is just checking the results of that grouping.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -1685,6 +1658,24 @@ private Bundle processBundleEntries(Bundle requestBundle, Map<Integer, Entry> va
...
+                    (!requestEntriesByMethod.get(HTTPVerb.Value.GET).isEmpty() &&
+                            requestEntriesByMethod.get(HTTPVerb.Value.DELETE).isEmpty() &&
+                            requestEntriesByMethod.get(HTTPVerb.Value.POST).isEmpty() &&
+                            requestEntriesByMethod.get(HTTPVerb.Value.PUT).isEmpty() &&
+                            requestEntriesByMethod.get(HTTPVerb.Value.PATCH).isEmpty() &&
+                            requestEntriesByMethod.get(HTTPVerb.Value.HEAD).isEmpty())) {
```
lmsurpre, line 1670: I'm really not sure where we stand in terms of HEAD requests from within a bundle, but technically that should be safe to do within the same transaction as the GETs.  Would it have any benefit?

lmsurpre (approved):
LGTM

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -1559,56 +1559,31 @@ private String getVersionIdFromETagValue(String ifMatchValue) {
...
+            // TODO add support for posting history bundles
             Bundle.Builder bundleResponseBuilder = Bundle.builder().entry(responseEntries);
             if (bundleType == BundleType.Value.BATCH) {
                 bundleResponseBuilder.type(BundleType.BATCH_RESPONSE);
             } else if (bundleType == BundleType.Value.TRANSACTION) {
                 bundleResponseBuilder.type(BundleType.TRANSACTION_RESPONSE);
-            } else {
-                // TODO add support for posting history bundles
-                String msg = "Bundle.type must be either 'batch' or 'transaction'.";
-                throw buildRestException(msg, IssueType.VALUE);
```
lmsurpre, line 1598: did you confirm that we do this validation somewhere else now?  i can't remember if we do it in the validate(Bundle) method or not...

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
punktilious: Ah, requestEntriesByMethod is already a map, so there's no multiple scanning involved. I misread and thought that the requestEntriesByMethod would be a list of everything which get(..) would then filter on. Ignore.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: yes, we do it in validateBundle() on line 1260

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
michaelwschroeder: Makes sense to me to treat GET and HEAD the same - have updated the code to do so.

Issue #2307, opened 2021-04-30
Null pointer on $validate of a Bundle 


d0roppe:
**Describe the bug**
I was testing with a bundle and got a null pointer error.
```
{
  "resourceType": "OperationOutcome",
  "issue": [
    {
      "severity": "fatal",
      "code": "exception",
      "details": {
        "text": "NullPointerException: &lt;null message&gt;"
      }
    }
  ]
}
```
**Environment**
Which version of IBM FHIR Server?

**To Reproduce**
This was in postman, but here is the curl:
Data is mock data.

```
curl --location --request POST 'https://localhost:9443/fhir-server/api/v4/bundle/$validate' \
--header 'X-FHIR-TENANT-ID: default' \
--header 'Content-Type: application/json' \
--header 'X-FHIR-FORWARDED-URL: https://minions-r.us' \
--header 'prefer: return=OperationOutcome' \
--header 'Authorization: Basic ...' \
--data-raw '{
  "resourceType":"Bundle",
  "type":"batch",
  "entry":[
    {
      "resource":{
        "resourceType":"Claim",

        "id":"my-Sniff-Test-R1",
        "status":"active",
        "type":{
          "coding":[
            {
              "system":"http://terminology.hl7.org/CodeSystem/claim-type",
              "code":"professional"
            }
          ]
        },
        "use":"claim",
        "patient":{
          "reference":"Patient/my-Sniff-Test-P1"
        },
        "created":"2020-02-28T00:00:00+00:00",
        "insurer":{
          "reference":"Organization/1601660375-0ef3e098-3af0-456b-8330-4bb5885d14e7-Payer",
          "display":"XXX Health Plan"
        },
        "insurance":[
          {
            "sequence":2,
            "focal":true,
            "coverage":{
              "reference":"Coverage/1601660375-6e1cfe4e-6eca-46dd-adad-e75ff9466ea6-Coverage"
            }
          }
        ],
        "provider":{
          "reference":"Organization/1601660375-4961efdc-1404-4fe7-8b80-4a32849a3d59-Provider",
          "display":"XXX Health Plan"
        },
        "priority":{
          "coding":[
            {
              "system":"http://terminology.hl7.org/CodeSystem/process-priority",
              "code":"normaly"
            }
          ]
        }
      },
      "request":{
        "method":"PUT",
        "url":"Claim/my-Sniff-Test-R1"
      }
    },
    {
      "resource":{
        "resourceType":"ExplanationOfBenefit",
[...]

PR #2310, opened 2021-05-03
Null pointer on $validate of a Bundle #2307

prb112:
- Add check for the URI Path
- Add Integration Test

### DIFF
--- a/fhir-server/src/main/java/com/ibm/fhir/server/operation/spi/AbstractOperation.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/operation/spi/AbstractOperation.java
@@ -6,20 +6,30 @@
 
 package com.ibm.fhir.server.operation.spi;
 
+import static com.ibm.fhir.model.type.String.string;
+
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 
+import org.owasp.encoder.Encode;
+
 import com.ibm.fhir.exception.FHIROperationException;
 import com.ibm.fhir.model.resource.OperationDefinition;
 import com.ibm.fhir.model.resource.OperationOutcome;
+import com.ibm.fhir.model.resource.OperationOutcome.Issue;
 import com.ibm.fhir.model.resource.Parameters;
 import com.ibm.fhir.model.resource.Resource;
+import com.ibm.fhir.model.type.Code;
+import com.ibm.fhir.model.type.CodeableConcept;
+import com.ibm.fhir.model.type.Extension;
+import com.ibm.fhir.model.type.code.IssueSeverity;
 import com.ibm.fhir.model.type.code.IssueType;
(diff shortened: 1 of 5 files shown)


fhir-server/src/main/java/com/ibm/fhir/server/operation/spi/AbstractOperation.java
```diff
@@ -143,6 +143,10 @@ protected void validateOperationContext(FHIROperationContext operationContext, C
                 String msg = "Operation context RESOURCE_TYPE is not allowed for operation: '" + getName() + "'";
                 throw buildExceptionWithIssue(msg, IssueType.INVALID);
             } else {
+                if (resourceType == null) {
+                    String msg = "Operation context RESOURCE_TYPE must include a valid ResourceType in the URL Path";
+                    throw buildExceptionWithIssue(msg, IssueType.INVALID);
```
lmsurpre, line 148: I'm thinking it should maybe be a not-supported (mapped to a 404) instead of an invalid (mapped to 400).

That would be more consistent with what we return if you try to read/create/update/delete a resource for an invalid resource type.  For example, this is our 404 response for a `GET [base]/Fake/1`
```
{
    "id": "ac-1d-0-3-fda86c56-02dc-4ce9-970b-080203c030e4",
    "issue": [
        {
            "severity": "fatal",
            "code": "not-supported",
            "_code": {
                "extension": [
                    {
                        "url": "http://ibm.com/fhir/extension/not-supported-detail",
                        "valueCode": "resource"
                    }
                ]
            },
            "details": {
                "text": "'Fake' is not a valid resource type."
            }
        }
    ]
}
```

lmsurpre: actually, i take it back.  we do a 404 for READ, but not for create.  interesting...

prb112: I'd prefer not to pass back the resource type since that'd mutate the signature for invoke pretty heavily, unless I pack it in the FHIROperationContext.

prb112:
```

lmsurpre (approved):
Yes, this is just what I had in mind wrt generating a 404.  LGTM.

prb112:
> Yes, this is just what I had in mind wrt generating a 404. LGTM.

great - it's just not ideal since I have to pack it into the context.

lmsurpre:
> > Yes, this is just what I had in mind wrt generating a 404. LGTM.
> 
> great - it's just not ideal since I have to pack it into the context.

Yeah, I would have been happy with just a generic message to go along with the 404.

Issue #2313, opened 2021-05-03
equivalency check when X-FHIR-UPDATE-IF-MODIFIED to true is not detecting that a field changed


kmbarton423:
Working with conditional update functionality (issues 2263 and 2284).  See Performance Guide section 5.2:
   https://ibm.github.io/FHIR/guides/FHIRPerformanceGuide/#5-ingestion-scenarios

Using FHIR main branch.

To reproduce:
-Create a Patient (PUT with predefined resource.id)
-Set header X-FHIR-UPDATE-IF-MODIFIED to true
-Set header Prefer to return=OperationOutcome
-PUT the same patient again
-Confirm history stays at "1" and OperationOutcome returns the following:
     "Update resource matches the existing resource; skipping the update"
-Modify a field within "identifier" in the Patient resource and PUT
-Again received history at "1" and OperationOutcome returns the following:
     "Update resource matches the existing resource; skipping the update"

Expected the resource to be saved as version "2"

kmbarton423:
Fix looks good.

PR #2314, opened 2021-05-04
issue #2313 - refactor FingerprintVisitor to avoid startsWith checks

lmsurpre:
Previously, we constructed the excludePaths on each leaf node, then
checked whether the current path started with any of those values.

This was faulty because the exclude path Resource.id accidentally
resulted in us ignoring fields like Resource.identifier as well.

Now we construct the excludePaths as a set during the initial
doVisitStart and we don't need "startsWith" because we the visitor
supports returning a boolean to control whether child paths should be
visited or not.


prb112 (approved):
LGTM

fhir-model/src/main/java/com/ibm/fhir/model/visitor/ResourceFingerprintVisitor.java
```diff
@@ -188,11 +186,6 @@ protected void updateDigest(String name, String value) {
      * @return
      */
     protected boolean includePath() {
-        String idName = currentResourceName + ".id";
-        String versionIdName = currentResourceName + ".meta.versionId";
-        String lastUpdatedName = currentResourceName + ".meta.lastUpdated";
-        String path = getPath();
-        return !path.startsWith(idName) && !path.startsWith(versionIdName) && !path.startsWith(lastUpdatedName);
-
+        return !excludePaths.contains(getPath());
```
punktilious, line 189: Minor point that being protected, includePaths() could be called before exludePaths is initialized, resulting in a NPE. But it's not really a valid use-case, so not sure it's worth addressing.

fhir-model/src/test/java/com/ibm/fhir/model/util/test/ResourceFingerprintVisitorTest.java
```diff
@@ -0,0 +1,177 @@
...
+                .id("test")
+                .meta(patient.getMeta().toBuilder()
+                    .versionId(Id.of("ignoreMe"))
+                    .lastUpdated(Instant.now())
+                    .build())
+                .build();
+        resourceFingerprintVisitor = new ResourceFingerprintVisitor(baseline);
+        patient.accept(resourceFingerprintVisitor);
+
+        System.out.println(patient);
```
lmsurpre, line 147:
```suggestion
```

fhir-model/src/main/java/com/ibm/fhir/model/visitor/ResourceFingerprintVisitor.java
lmsurpre:
```suggestion
        return excludePaths == null ? null : !excludePaths.contains(getPath());
```

punktilious (approved):
LGTM...assuming you are going to commit the System.out.println changes?

lmsurpre:
> assuming you are going to commit the System.out.println changes?
I think I already merged those.  I think this one is ready.

PR #2315, opened 2021-05-04
revert to minimal examples for the OpenAPI and use subset for demo env

lmsurpre:
1. Package the minimal examples in fhir-openapi. The Liberty OpenAPI UI
doesn't seem to like our complete-mock examples. For example, instead of
including the datetime values it just uses an empty object for these
fields (which fails FHIR validation). It seems to do a little better
with our minimal examples (although these are a tad less useful in terms
of documenting the allowed fields). Perhaps more importantly, switching
back to the minimal examples saves about 50,000 lines in the all-in-one
openapi definition, which should make it a bit more managable.

2. Overwrite the default openapi.json in our demo environment with one
that matches the subset of resources that are actually exposed in that
environment. This makes viewing and playing with the API surface area a
bit more managable and prevents users from tryinig to use endpoints
which aren't actually supported in this environment.

3. A small, unrelated doc fix in the BringYourOwnPersistence guide.


docs/src/pages/guides/BringYourOwnPersistence.md
```diff
@@ -19,7 +19,10 @@ Persistence layer interfaces are defined in the `fhir-persistence` project.
 ### Config
```
tbieste, line 19: Typo on line 4:  "Lean how to..."

docs/src/pages/guides/BringYourOwnPersistence.md
```diff
@@ -19,7 +19,10 @@ Persistence layer interfaces are defined in the `fhir-persistence` project.
...
 Which persistence layer is used by the server is determined by the `/fhirServer/persistence/factoryClassname` property in `fhir-server-config.json`.
 
-When the default `com.ibm.fhir.persistence.jdbc.FHIRPersistenceJDBCFactory` is used, the returned FHIRPersistenceJDBCImpl instance will look up a corresponding datasource name using the value of the `fhirServer/persistence/jdbc/dataSourceJndiName` property in the tenant-specific configuration (`jdbc/fhirProxyDataSource` by default) and establish the connection.
+When the default `com.ibm.fhir.persistence.jdbc.FHIRPersistenceJDBCFactory` is used, the returned FHIRPersistenceJDBCImpl instance will look up a corresponding datasource, by JNDI name, using the combination of the tenant id and datastore id from the request context. Specifically, the connection strategy will use the `fhirServer/persistence/<datastore_id>/jndiName` property in the tenant config, or--if the property is omitted--it will construct the name via the following pattern:
+```
+jdbc/fhir_<tenant_id>_<datastore_id>[_ro]
```
tbieste, line 24: What is the "_ro" for?

tbieste (approved):
A couple minor documentation comments, otherwise LGTM.

docs/src/pages/guides/BringYourOwnPersistence.md
lmsurpre: read-only.  we don't really use it yet, but the code is there for it.

docs/src/pages/guides/BringYourOwnPersistence.md
lmsurpre: fixed

PR #2316, opened 2021-05-05
upgrade liberty features

lmsurpre:
Note: merging this will mean that everyone will need to upgrade their version of Liberty to one that supports these newer features.


prb112 (approved):
LGTM

Issue #2041, opened 2021-03-09
make quantity values searchable even without unit or code


d0roppe:
**Is your feature request related to a problem? Please describe.**
Adding observation data component with valueQuantity, but only adding a value, fhir should index and allow searches on the value.
**Describe the solution you'd like**
When adding Observation data like this (mock data)
      "component": [{
        "code": {
          "coding": [{
            "system": "http://loinc.org",
            "code": "2339-0"
          }]
        },
        "valueQuantity": {
          "value": 58.7
        }
      }]
Should be able to search on this Observation with searches like:
[base]/Observation?component-code-value-quantity=http://loinc.org|2339-0$58.7

**Describe alternatives you've considered**
Continue to require valueQuantity units or code allong with value to index the data and allow searches.

**Acceptance Criteria**
At least one acceptance criteria is included.

1.  
  GIVEN [a precondition]
    AND [another precondition] WHEN [test step]
    AND [test step]
    THEN [verification step]
    AND [verification step]

**Additional context**

lmsurpre:
The system+code columns are nullable, so we could either store without one or with some kind of placeholder values.

lmsurpre:
Hopefully it just requires changes to extraction and no corresponding query changes.  Estimated accordingly.

d0roppe:
Verified that this search capability now works.

Issue #1921, opened 2021-02-08
whole-system-search: populate global search parameter tables


punktilious:
To address #264 new global-level tables were introduced to the schema but are currently not populated.

Update the search parameter extraction and persistence to populate these parameter tables.

Migration will require a reindex...consider a reindex which can be parameterized to only update these tables, which should be much faster than a full reindex.

Part of epic #264.

d0roppe:
verified new tables have values added on ingestion. Also reindexed a schema and the values were then populated. Closing issue.

PR #2324, opened 2021-05-10
Issue #2041 - Make quantity values searchable even without unit or code

tbieste:
Issue #2041 - Make quantity values searchable even without unit or code

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ParameterVisitorBatchDAO.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ParameterVisitorBatchDAO.java
@@ -277,7 +277,12 @@ public class ParameterVisitorBatchDAO implements ExtractedParameterValueVisitor,
         BigDecimal valueLow = param.getValueNumberLow();
         BigDecimal valueHigh = param.getValueNumberHigh();
 
-        // No isBase check because there are no system-level number search parameters
+        // System-level number search parameters are not supported
+        if (isBase(param)) {
+            String msg = "System-level number search parameters are not supported: " + parameterName;
+            logger.warning(msg);
+            throw new IllegalArgumentException(msg);
+        }
 
         try {
             int parameterNameId = getParameterNameId(parameterName);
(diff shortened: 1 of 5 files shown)


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ParameterVisitorBatchDAO.java
```diff
@@ -399,10 +399,11 @@ public void visit(QuantityParmVal param) throws FHIRPersistenceException {
         BigDecimal quantityLow = param.getValueNumberLow();
         BigDecimal quantityHigh = param.getValueNumberHigh();
 
-        // XXX why no check for isBase on this one?
+        // XX why no check for isBase on this one?
```
lmsurpre, line 402: I sort of get the `XXX` to `XX` change, but `XXX` is a built-in marker for eclipse (sort of like a TODO).
Anyway, what would be nice is if we could just replace the question with the answer.

My guess is that we skip the isBase check just because we happen to know there are no cross-resource/global search parameters of type quantity.
@punktilious does that sound right to you?

lmsurpre: Options:
A. just replace the existing question like "No isBase check because there are no system-level quantity search parameters"
B. add a check for isBase and throw an UnsupportedOperationException (or similar)
C. add proper support for system-level quantity parameters

I think maybe B is the best for now because I don't think its important to add, but I think it might be better to error out than to store it as a resource-type-specific param...at least when we actually start using the global search param tables (via #1921)

punktilious: I think B too...as long as we're absolutely confident we can't inadvertently hit this. It seems a little over-the-top to make it a configurable (fhir-server-config flag) error vs. warning.

I'm just starting to look at properly supporting system-level parameters.

tbieste: There's a couple other parameter types that skip the isBase check, so I added a comment "No isBase check because there are no system-level xxxxxxx search parameters".  This was option A.  If option B is preferred, I can throw an exception instead.

tbieste: Updated to option B.

PR #2328, opened 2021-05-10
enable liberty dev mode for fhir-server-webapp

lmsurpre:


fhir-persistence-scout/pom.xml
```diff
@@ -18,6 +18,26 @@
 
     <build>
         <plugins>
+            <plugin>
```
prb112, line 21: Bad Merge?

fhir-server-webapp/src/main/liberty/config/bootstrap.properties
```diff
@@ -0,0 +1,3 @@
+javax.xml.stream.XMLInputFactory=com.sun.xml.internal.stream.XMLInputFactoryImpl
+javax.xml.stream.XMLOutputFactory=com.sun.xml.internal.stream.XMLOutputFactoryImpl
+javax.xml.stream.XMLTransformerFactory=com.sun.org.apache.xalan.internal.xsltc.trax.TransformerFactoryImpl
```
prb112, line 3: Can we add a README in the prior location that tells people where to look now?

prb112 (approved):
LGTM - left one comment.... README.md to mark the fact the folder changed. 
2. it'd be great if there is some how-to docs updated in the dev wiki

lmsurpre:
> 2. it'd be great if there is some how-to docs updated in the dev wiki

Yes, if we merge this then we'll definitely need to make a few dev wiki updates.  I will plan on it.

fhir-persistence-scout/pom.xml
lmsurpre: this is actually here on purpose, but its only tangentially related to the rest of the PR.
basically, I was trying to use liberty dev mode in vscode.  without this plugin, our experimental fhir-persistence-scout project fails to compile in vscode (because the classes in there depend on this one generated class that winds up in generated-sources).  the config here tells vscode to add the folder as a source folder.

lmsurpre: see https://github.com/redhat-developer/vscode-java/issues/177 for a long sorted history..

fhir-server-webapp/src/main/liberty/config/bootstrap.properties
lmsurpre: done

michaelwschroeder (approved):
LGTM

PR #2333, opened 2021-05-10
issue-1631 PostgreSQL drop schema fails with shared memory error

punktilious:


lmsurpre (approved):
LGTM

Issue #1885, opened 2021-01-25
conflicting search parameters messages should note the full URL of the conflict code


d0roppe:
**Is your feature request related to a problem? Please describe.**
When a message of conflicting search parameters is issued, it would be good to know the full URL of the code so you can more easily code up the filter to disambiguate the conflict.

Example of current message:
00000035 com.ibm.fhir.search.util.SearchUtil                          W Skipping search parameter with id='location-address'. Found multiple search parameters for code 'address' on resource type 'Location'; use search parameter filtering to disambiguate.

**Describe the solution you'd like**
would like to know full URL of the conflicting searchparameters.  Not just the one conflicting with what is already loaded, but all of them.

**Acceptance Criteria**
1.  
  GIVEN: Starting the FHIR server with conflicting search parameters from the base and optional IG's loaded
    THEN look in the log file to verify the messages have the FULL URL's of the options listed.

tbieste:
During QA testing, Dag noticed another related code path:

[5/10/21, 17:51:28:054 UTC] 0000004c com.ibm.fhir.search.util.SearchUtil                          W Found multiple resource-specific search parameters for code ‘name’ on resource type Patient; use search parameter filtering to disambiguate. Using ‘http://hl7.org/fhir/SearchParameter/Patient-name’.

d0roppe:
Verified new messages
```
[5/13/21, 16:27:15:784 UTC] 00000039 com.ibm.fhir.search.util.SearchUtil                          W Skipping search parameter with id='explanationofbenefit-identifier'. Found multiple search parameters, 'http://hl7.org/fhir/SearchParameter/ExplanationOfBenefit-identifier' and 'http://hl7.org/fhir/us/carin-bb/SearchParameter/explanationofbenefit-identifier', for code 'identifier' on resource type 'ExplanationOfBenefit'; use search parameter filtering to disambiguate.
[5/13/21, 16:27:15:784 UTC] 00000039 com.ibm.fhir.search.util.SearchUtil                          W Skipping search parameter with id='explanationofbenefit-provider'. Found multiple search parameters, 'http://hl7.org/fhir/SearchParameter/ExplanationOfBenefit-provider' and 'http://hl7.org/fhir/us/carin-bb/SearchParameter/explanationofbenefit-provider', for code 'provider' on resource type 'ExplanationOfBenefit'; use search parameter filtering to disambiguate.
[...]

Issue #1625, opened 2020-10-26
Search configuration cache


lmsurpre:
**Is your feature request related to a problem? Please describe.**
`SearchUtil.getFilteredBuiltinSearchParameters()` reads in tenant-specific configuration and uses that config to filter the built-in search parameters down to a list of enabled search parameters on a per-tenant basis.

The fhir-config module already caches fhir-server-config.json and so reading the config should be relatively fast (no file read if the file hasn't changed), but we're still doing some extra work with each request.

Namely, if the config hasn't changed then there should be no need to recompute the set of search parameters.

**Describe the solution you'd like**
During server startup, we could read the fhir-server-config.json for all the tenants and pre-construct a ParametersMap for each one.

Then, we should only ever rebuild the ParametersMap if/when the fhir-server-config.json is updated for this tenant (and only if the dynamic behavior is required).

**Describe alternatives you've considered**
Instead of server startup, we could build the tenant-specific ParametersMap on the first request from the tenant.

**Additional context**
This change is highly related to https://github.com/IBM/FHIR/issues/1596 because, if we get the "built-in" search parameters from the registry, then the caching here would be complicated by the fact that the registry could return different SearchParameters over time (at least when `serverRegistryResourceProviderEnabled` is set to true).

punktilious:
Try to avoid concurrency issues, particularly with locking. Probably worth considering a ThreadLocal cache implementation for this stuff as long as the cached content isn't big (which seems to be the case here).

lmsurpre:
The "cache" was implemented as a revision to the existing ParameterUtils class.  Now, upon initialization, that will loop over all configured tenants (i.e. all directories under the fhir-server's config dir) and apply the search filtering rules therein to the search parameters loaded from the registry.

prb112:
QA Complete

Looks good

# Test 1. Server Startup with a single Parameter
[11/12/21, 15:52:14:076 UTC] 00000022 FHIRConfigura 1   Returning list of tenant ids: [default]
[11/12/21, 15:52:14:076 UTC] 00000022 ParametersUti 1   Computing search parameters for tenant default
[11/12/21, 15:52:14:077 UTC] 00000022 TenantSpecifi 1   Loading config file from /wlp/usr/servers/fhir-server/config/default/extension-search-parameters.json
[...]

PR #2335, opened 2021-05-10
Issue #1885 - Include search parameter URLs in conflict warning message

tbieste:
Issue #1885 - Include search parameter URLs in conflict warning message (in additional code path found during QA)

### DIFF
--- a/fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
+++ b/fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
@@ -18,6 +18,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -451,20 +452,27 @@ public static SearchParameter getSearchParameter(String resourceType, String cod
...
-                if (params.size() > 1) {
-                    log.warning("Found multiple cross-resource search parameters for code '" + code + "';"
-                            + " use search parameter filtering to disambiguate. Using '" + result.getUrl().getValue() + "'.");
+                Iterator<SearchParameter> iterator = params.iterator();
+                result = iterator.next();
+                String configuredUrl = result.getUrl().getValue();
+                while (iterator.hasNext()) {
+                    SearchParameter conflict = iterator.next();
+                    log.warning("Found multiple cross-resource search parameters, '" + configuredUrl + "' and '" + conflict.getUrl().getValue()
+                        + "', for code '" + code + "'; use search parameter filtering to disambiguate. Using '" + configuredUrl + "'.");
```
prb112, line 475: it feels like this message should have an option to silence it.  It could be rather chatty if one has a decent sized Search Bundle.

lmsurpre: The original idea behind the warning was that operators *should* use our fhir-server-config support to indicate which search parameter they want us to use in case of conflicts.  So the way to silence it is to "use search parameter filtering to disambiguate".

What this task is doing is putting enough information in the message so that the config change is as easy as possible.

Aside:  because search parameter filtering is done on a per-tenant basis, we can't easily move these warnings to startup unless we start looping through the various tenant configurations during startup.  I had put some related thoughts for that into https://github.com/IBM/FHIR/issues/1625

lmsurpre: i don't think we have more than 2 conflicting params  too often, but maybe it would help to collect the various conflicts into a single message instead of logging a different warning for each one?

for example:
```
log.warning("Found multiple cross-resource search parameters for code '" + code + "';"
                            + " use search parameter filtering to disambiguate. Using '" + result.getUrl().getValue() + "'. "
                            + result.stream().map(sp -> sp.getUrl().getValue()).collect(Collectors.joining(", ") );
```

lmsurpre (approved):
LGTM

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
tbieste: @prb112 Are you ok with this as is?

Issue #1674, opened 2020-11-06
Implicit base not picked up in FHIR chained search


lmsurpre:
**Describe the bug**
I found a case that we missed when we implemented #300 for adding the implicit resource type of a reference:  chained parameters.

Specifically, note this line from https://build.fhir.org/search.html#chaining
> Because the Diagnostic Report subject can be one of a set of different resources, it's necessary to limit the search to a particular type:
`GET [base]/DiagnosticReport?subject:Patient.name=peter`

To me, that implies that if the type was NOT ambiguous, then the type modifier should be optional (just like in the normal case).  However, my query for `Observation?patient.gender=female` returns the following issue:
```
'issue': [{'severity': 'fatal', 'code': 'invalid', 'details': {'text': "Search parameter: 'patient' must have resource type name modifier"}
```

**Expected behavior**
The type modifier is not required if the reference search parameter only targets a single type.

**Additional context**
We should open an HL7 JIRA ticket on this section of the spec as well, because it contradicts itself in a bad way:
```
For instance, given that the resource DiagnosticReport has a search parameter named subject, which is usually a reference to a Patient resource, and the Patient resource includes a parameter name which searches on patient name, then the search
 GET [base]/DiagnosticReport?subject.name=peter
is a request to return all the lab reports that have a subject whose name includes "peter".
Because the Diagnostic Report subject can be one of a set of different resources, it's necessary to limit the search to a particular type:
 GET [base]/DiagnosticReport?subject:Patient.name=peter
```

If it's "necessary to limit the search to a particular type" then the first query they give isn't actually valid, but they don't make this very clear.

michaelwschroeder:
@lmsurpre I believe this is working as designed. When parsing a chained search parameter, we check the `SearchParameter.target` contents. If the target contains more than one resource type and a type modifier is not specified on the search parameter, then an error (as shown above) is returned.

In the example above, the `Observation?patient.gender=female` search correctly returns an error because the `patient` search parameter specifies both the `Patient` and `Group` resource types as targets. Here's the full search parameter definition:

```
   {
      "resourceType" : "SearchParameter",
      "id" : "clinical-patient",
      "extension" : [{
        "url" : "http://hl7.org/fhir/StructureDefinition/structuredefinition-standards-status",
        "valueCode" : "trial-use"
      }],
      "url" : "http://hl7.org/fhir/SearchParameter/clinical-patient",
      "version" : "4.0.1",
      "name" : "patient",
      "status" : "draft",
[...]

PR #2355, opened 2021-05-12
issue #1674 - workaround for HL7 FHIR-13601

lmsurpre:
The built-in SearchParameter-clinical-patient search parameter should
have a single target of type "Patient" but instead it has a target of
both "Patient" and "Group".

In this commit, I move the fhir-registry IndexGenerator from
src/main/java to src/test/java and add a step to modify this one spec
resource so that users can search on the `patient` search parameter
without needing a type modifier.


tbieste (approved):
LGTM

michaelwschroeder (approved):
LGTM

JohnTimm:
Why did you move it to the test package?

lmsurpre:
> Why did you move it to the test package?

we discussed this previously.  its not needed at runtime so its a little better not to have it under `src/main` because that ends up in the jar whereas `src/test` packages do not.

JohnTimm (approved):
LGTM

Issue #2365, opened 2021-05-14
Conditional delete operation outcome has nulls in issue details


JohnTimm:
I recently created 3 `Patient` resources each using a standard create interaction (POST) all with the same `Patient.identifier` element. I then did a search:

`GET <base>/fhir-server/api/v4/Patient?identifier=12345`

and I get a Bundle resource containing 3 patients. Then I tried using the "conditional delete" interaction:

`DELETE <base>/fhir-server/api/v4/Patient?identifier=12345`

and I got the following response:

```json
{
    "resourceType": "OperationOutcome",
    "issue": [
        {
            "severity": "information",
            "code": "informational",
            "details": {
                "text": "Deleted 3 Patient resource(s) with the following id(s): null,null,null"
            }
        }
    ]
}
```
Then I confirmed that the resource(s) have been deleted by performing another search:

`GET <base>/fhir-server/api/v4/Patient?identifier=12345`

```json
{
    "resourceType": "Bundle",
    "id": "4196a09f-c220-4ca2-91c4-9d8e3bbcdd60",
    "type": "searchset",
    "total": 0,
    "link": [
        {
            "relation": "self",
            "url": "https://localhost:9443/fhir-server/api/v4/Patient?_count=10&identifier=12345&_page=1"
        }
    ]
}
```
The "conditional delete" behavior works as advertised but there should be actual resource ids in the operation outcome instead of nulls.

JohnTimm:
After this was merged, I ran a the following test to verify it:

`DELETE <base>/fhir-server/api/v4/Patient?identifier=12345`

and this was the result:

```json
{
    "resourceType": "OperationOutcome",
    "issue": [
        {
            "severity": "information",
            "code": "informational",
            "details": {
                "text": "Deleted 1 Patient resource(s) with the following id(s): 12345"
            }
        }
    ]
}
```

PR #2367, opened 2021-05-14
Issue #2365 - Show resource ids in conditional delete operation outcome

tbieste:
Issue #2365 - Show resource ids in conditional delete operation outcome

### DIFF
--- a/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
@@ -1810,7 +1810,7 @@ public class FHIRRestHelper implements FHIRResourceHelpers {
                 txn.commit();
                 txn = null;
             }
-            
+
             return Arrays.asList(responseEntries);
 
         } finally {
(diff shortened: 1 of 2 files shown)


michaelwschroeder (approved):
LGTM

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -2563,6 +2563,7 @@ Bundle createSearchBundle(List<Resource> resources, FHIRSearchContext searchCont
                 Entry.Builder entryBuilder = Entry.builder();
                 if (resource != null) {
                     if (resource.getId() != null) {
+                        entryBuilder.id(resource.getId());
```
lmsurpre, line 2566: why are we using the Entry.id for this?

lmsurpre: I guess its because there is no resource in the response entries, and so we can't use the normal `Entry.resource.id`.  Still, I think we might want a comment for this because using Entry.id for the id of the deleted resource is a bit unique.

lmsurpre: side-note:  you definitely didn't introduce this idea of using the entry id for this.  that comes from 
FHIRRestHelper.doDelete near line 722:
```
responseBundle.getEntry().stream().map(Bundle.Entry::getId).collect(Collectors.joining(","))
```

Issue #2296, opened 2021-04-28
resolve function in FHIRPATH expression


dipanjanmukherjeeibm:
**Is your feature request related to a problem? Please describe.**
resolve() function is a function in FHIR that allows to reference elements  across FHIR resources. The resolve() function currently doesn't work in IBM FHIR server.

**Describe the solution you'd like**
resolve() function is a function in FHIR that allows to reference elements  across FHIR resources. The resolve() function is integral to build business rules for FHIR validation requireing cross-resource validation. The documentation for resolve () function in FHIR server states:
 ```
 /**
     * For each item in the collection, if it is a string that is a uri (or canonical or url), locate the target of the
     * reference, and add it to the resulting collection. If the item does not resolve to a resource, the item is ignored
     * and nothing is added to the output collection. The items in the collection may also represent a Reference, in which
     * case the Reference.reference is resolved.
     *
     * <p>This method creates a resource node that is a placeholder for the actual resource, thus allowing for the FHIRPath
     * evaluator to perform type checking on the result of the resolve function. For example:
     *
     * <pre>Observation.subject.where(resolve() is Patient)</pre>
     *
     * <p>If the resource type cannot be inferred from the reference URL or type, then {@code FHIR_UNKNOWN_RESOURCE_TYPE} is used
     * so that we index these references by default.
     *
     * @param evaluationContext
     *     the evaluation environment
     * @param context
     *     the current evaluation context
     * @param arguments
     *     the arguments for this function
     * @return
     *     the result of the function applied to the context and arguments
     */
```
**Describe alternatives you've considered**
example of a FHIR path expression with resolve function that would work in the IBM FHIR Server on coverage profile
`period.end <= beneficiary.resolve().deceased as deceasedDateTime`

**Acceptance Criteria**
Any FHIR PATH expression with resolve() function should work when the profile on the IG is invoked by IBM FHIR Server during validation.

**Additional context**

lmsurpre:
@dipanjanmukherjeeibm this feature was added and released in IBM FHIR Server 4.8.1.

To enable it, you must set `fhirServer/core/serverResolveFunctionEnabled` to `true` in your fhir-server-config.json

Please give it a try and confirm that this addresses your request.  If we don't hear back in a day or so, I will plan to close this one out.

dipanjanmukherjeeibm:
I have tried to add the above in the config file and the problem is now resolved. I am happy for the dev team  to close this request as resolved

PR #2375, opened 2021-05-17
Issue #2296 - implement server resolve function

JohnTimm:


fhir-model/src/test/java/com/ibm/fhir/model/util/test/ReferencePatternTest.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+package com.ibm.fhir.model.util.test;
+
+import static com.ibm.fhir.model.util.FHIRUtil.REFERENCE_PATTERN;
+
+import java.util.regex.Matcher;
+
+public class ReferencePatternTest {
+    public static void main(String[] args) throws Exception {
```
prb112, line 14: Worthwhile making a Test since it'd be a good confirmation.

prb112 (approved):
LGTM (Pending the docs)

fhir-model/src/test/java/com/ibm/fhir/model/util/test/ReferencePatternTest.java
JohnTimm: Done

lmsurpre:
should we consider resolving absolute references as well?

how about absolute references with a base that matches the configured server base url?

JohnTimm:
> should we consider resolving absolute references as well?

> how about absolute references with a base that matches the configured server base url?

Yes we could consider absolute references but only in the context of the ServerResolveFunction. The base ResolveFunction in the fhir-path module has no way of knowing the server base URL.

> should we try to consolidate any of the reference parsing logic with what we have in ReferenceUtil?

Perhaps, however, ReferenceUtil is in fhir-search which is not a dependency of fhir-path. I’m good keeping them separate for now and leaving reconciliation for future work.

fhir-server/src/test/java/com/ibm/fhir/server/test/ServerResolveFunctionTest.java
```diff
@@ -0,0 +1,374 @@
...
+            return persistence;
+        }
+
+        @Override
+        public FHIRPersistence getFHIRPersistenceImplementation(String factoryPropertyName) throws FHIRPersistenceException {
+            throw new UnsupportedOperationException();
+        }
+    }
+
+    public static class PersistenceImpl implements FHIRPersistence {
```
lmsurpre, line 208: This is like the 4th mock / stubbed out persistence impl we have across our various modules.  I wonder if its worth trying to consolidate.

fhir-config/src/main/java/com/ibm/fhir/config/FHIRConfiguration.java
```diff
@@ -33,6 +33,7 @@
     public static final String PROPERTY_CHECK_REFERENCE_TYPES = "fhirServer/core/checkReferenceTypes";
     public static final String PROPERTY_CONDITIONAL_DELETE_MAX_NUMBER = "fhirServer/core/conditionalDeleteMaxNumber";
     public static final String PROPERTY_SERVER_REGISTRY_RESOURCE_PROVIDER_ENABLED = "fhirServer/core/serverRegistryResourceProviderEnabled";
+    public static final String PROPERTY_SERVER_RESOLVE_FUNCTION_ENABLED = "fhirServer/core/serverResolveFunctionEnabled";
```
lmsurpre, line 36: as paul mentioned, we need to document this on in the user's guide

fhir-server/src/test/java/com/ibm/fhir/server/test/ServerResolveFunctionTest.java
JohnTimm: Yes, the one I made is map-based and could be generalized and perhaps reused elsewhere. That said, such consolidation is outside the scope of this PR.

fhir-config/src/main/java/com/ibm/fhir/config/FHIRConfiguration.java
JohnTimm: Users Guide updates are forthcoming.

JohnTimm: Updated FHIR Server User's Guide with information on the new configuration property.

lmsurpre (approved):
LGTM

Issue #2322, opened 2021-05-07
Error on bulk data status call where import did not import anything. 


d0roppe:
**Describe the bug**
after issuing an $import on a location where no data can be imported, Checking the status of the $import gives 500 error:
```
{
  "resourceType": "OperationOutcome",
  "id": "a-82-2-c-d34cbd42-578c-4c10-b183-e60d1f856bbd",
  "issue": [
    {
      "severity": "error",
      "code": "transient",
      "diagnostics": "An unexpected error has ocurred while checking the status - Array index out of range: 1"
    }
  ]
}
```

attaching log
[error.log](https://github.com/IBM/FHIR/files/6443086/error.log)

**Environment**
Which version of IBM FHIR Server?

**To Reproduce**

**Expected behavior**

**Additional context**
The exitStatus just comes back as "COMPLETED", so when the code attempts to parse it as an array of values, it causes the exception.

prb112:
Can you grab the joblog as well?

prb112:
Worked with Dag on this.  
There is a first issue that I didn't cover - File Does not Exist on S3. 
There is a second issue that I did cover - File Exists but is Empty.

prb112:
PR https://github.com/IBM/FHIR/pull/2394/files

d0roppe:
now gives nice error

```
{
  "resourceType": "OperationOutcome",
  "id": "c0-a8-1-79-1adbaa92-7a0f-4c27-8b91-4f83d1c67819",
  "issue": [
    {
      "severity": "error",
      "code": "invalid",
      "diagnostics": "A bad source input was used during a call to $import"
    }
  ]
}
```

PR #2376, opened 2021-05-18
Error on bulk data status call where import did not import anything #2322

prb112:
- We pruned zero sized files, where we should keep them to align with
the output needed
- Updated the corresponding automation


lmsurpre:
Just wanted to double-check one thing:  this change won't start writing empty COS objects / empty export files, right?

prb112:
> Just wanted to double-check one thing: this change won't start writing empty COS objects / empty export files, right?

It should not.  This is on the $import side only.

lmsurpre (approved):
LGTM

Issue #2054, opened 2021-03-11
Deprecate and remove the exportPublic + ACL-based approach to exported object expiry


lmsurpre:
**Is your feature request related to a problem? Please describe.**
When the export is configured to create public objects (fhirServer/bulkdata/storageProviders/(source)/exportPublic = true), we attempt to set an ACL policy for objects to make them temporarily public. I believe that is currently hard-coded to 2 hours.

Unfortunately, neither minio nor IBM COS actually honor this setting:
> Please note that IBM COS does not support expiration time for each single COS object, so please configure retention policy (e.g, 1 day) for the bucket if IBM COS is used.

Although its documented, this seems like it could end badly (exports staying public too long and hacked/stolen because of it).

In #882 we added support for using presigned URLs for clients to fetch the resources and that seems like a better approach.

**Describe the solution you'd like**
Deprecate and eventually remove support for `fhirServer/bulkdata/storageProviders/(source)/exportPublic`.
Guide users toward using HMAC auth and presigned URLs instead.

**Describe alternatives you've considered**
Continue to support setting ACLs.

**Acceptance Criteria**
Documentation is clear that HMAC auth and presigned URLs is the preferred approach.

**Additional context**
Originally discussed at https://github.com/IBM/FHIR/pull/2052/files#r592359447

lmsurpre:
two locations:  config and bulkdata util
mostly just testing

lmsurpre:
The deprecation was done in 4.8.1.
The removal will be done via #2380 for 4.9.0.

lmsurpre:
I reviewed the results of #2380 and I think we should go even further...I opened https://github.com/IBM/FHIR/pull/2543 with a couple additional removals including a proposal to remove the legacy bulkdata config that has been deprecated since 4.6.0.

lmsurpre:
#2543 is merged and I'm calling this one done.

PR #2380, opened 2021-05-18
Remove the exportPublic + ACL-based approach to exported object

prb112:
expiry #2054


docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -2068,7 +2066,6 @@ This section contains reference information about each of the configuration prop
 |`fhirServer/bulkdata/storageProviders/<source>/fileBase`|string| The absolute path of the output directory |
 |`fhirServer/bulkdata/storageProviders/<source>/validBaseUrls`|list|The list of supported urls which are approved for the fhir server to access|
 |`fhirServer/bulkdata/storageProviders/<source>/disableBaseUrlValidation`|boolean|Disables the URL checking feature, allowing all URLs to be imported|
-|`fhirServer/bulkdata/storageProviders/<source>/exportPublic`|boolean|Whether or not the server is configured to support export to parquet; to properly enable it the administrator must first make spark and stocator available to the fhir-bulkdata-webapp (e.g through the shared lib at `wlp/user/shared/resources/lib`)|
 |`fhirServer/bulkdata/storageProviders/<source>/enableParquet`|boolean|If give public read only access to the exported files|
```
lmsurpre, line 2069: looks like we had a documentation bug here.  the enableParquet descriptiion was on the exportPublic line and vice-versa.
mind copying the current `exportPublic` description down to the enableParquet line?

prb112, line 2069:
```suggestion
|`fhirServer/bulkdata/storageProviders/<source>/enableParquet`|boolean|Whether or not the server is configured to support export to parquet; to properly enable it the administrator must first make spark and stocator available to the fhir-bulkdata-webapp (e.g through the shared lib at `wlp/user/shared/resources/lib`)|
```

lmsurpre (approved):
A bit nervous about deleting this function in a patch release, without any real warning for folks that might be relying on it.  However, I think we can cover it in the "special migration considerations" and presigned urls should be the way to go.

prb112:
I think we can also hold this until after 4.8.1

lmsurpre:
Maybe for 4.8.1 we do a docs-only change that says this setting is deprecated.  Then we merge this after 4.8.1?

prb112:
> Maybe for 4.8.1 we do a docs-only change that says this setting is deprecated. Then we merge this after 4.8.1?

https://github.com/IBM/FHIR/pull/2383 updates the docs

Issue #2377, opened 2021-05-18
Add Support for Pattern Value Constraint with $this in the Path


prb112:
**Is your feature request related to a problem? Please describe.**
Add Support for Pattern Value Constraint with $this in the Path

support canonical pattern value constraints

We generate a constraint =  "generated-C4BB-Organization-1: Constraint violation: meta.where(lastUpdated.exists() and profile = 'http://hl7.org/fhir/us/carin-bb/StructureDefinition/C4BB-Organization%7C1.0.0').exists()"
should this be = or an in http://hl7.org/fhirpath/#in-membership
Currently, we say the following is an error, however profile is an array...
```json
{
  "resourceType" : "Organization",
  "id" : "OrganizationProvider1",
  "meta" : {
    "lastUpdated" : "2019-12-12T09:14:11+00:00",
    "profile" : [
      "http://hl7.org/fhir/us/carin-bb/StructureDefinition/C4BB-Organization",
      "http://hl7.org/fhir/us/carin-bb/StructureDefinition/C4BB-Organization|1.0.0"
    ]
  }
}
```
https://build.fhir.org/ig/HL7/carin-bb/Organization-OrganizationProvider1.html

they had a constraint on the number of profiles, I'm opening a JIRA issue for it

lmsurpre:
> they had a constraint on the number of profiles, I'm opening a JIRA issue for it

![image](https://user-images.githubusercontent.com/5478798/118639612-1897fc80-b7a6-11eb-81a1-e10567b6eb10.png)

What they're trying to say here is that at least one of the profiles should be `http://hl7.org/fhir/us/carin-bb/StructureDefinition/C4BB-Organization%7C1.0.0`.
The profile element itself is `1..*`, its just the slice with this pattern that is `1..1`.

Net:  I think this is solely our issue and I think the IG is just fine.

prb112:
hum... I thought this asserted that the canonical pattern has to be...

prb112:
@JohnTimm I think you've already implemented this?

prb112:
I've already tried this out. 
QA done

PR #2389, opened 2021-05-19
Issue #2377 - update ConstraintGenerator

JohnTimm:

Changes made:
- Updated `ConstraintGenerator.generatePatternValueConstraint` to support the pattern value constraints that are also slices with a discriminator path of `$this`
- Updated `FHIRPathEvaluator` to remove "workaround" for `in` operator. Previously we turned an `in` operator expression into a `memberOf` function invocation in some circumstances.
- disabled test in `MemberOfFunctionTest`
- added unit test using latest version of C4BB Patient structure definition

### DIFF
--- a/conformance/fhir-ig-carin-bb/src/test/resources/JSON/StructureDefinition-C4BB-Patient.json
+++ b/conformance/fhir-ig-carin-bb/src/test/resources/JSON/StructureDefinition-C4BB-Patient.json
@@ -1,0 +1,5908 @@
+{
+    "resourceType": "StructureDefinition",
+    "id": "C4BB-Patient",
+    "text": {
+        "status": "extensions",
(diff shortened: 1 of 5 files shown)


prb112 (approved):
LGTM - based on our discussion and screen share this makes sense.

conformance/fhir-ig-carin-bb/src/test/resources/JSON/StructureDefinition-C4BB-Patient.json
```diff
@@ -0,0 +1,5908 @@
+{
```
lmsurpre, line 1: I think I'd have preferred a minimal example that shows the issue.  Will we remove it once we package the next version of C4BB?

fhir-path/src/test/java/com/ibm/fhir/path/test/MemberOfFunctionTest.java
```diff
@@ -118,7 +118,7 @@ public void testMemberOfFunction6() throws Exception {
         Assert.assertEquals(result, SINGLETON_FALSE);
     }
 
-    @Test
+    @Test(enabled = false)
```
lmsurpre, line 121: why is it disabled?  Maybe add a comment to explain the why?

lmsurpre: if its because "`in` just doesn't work like this any more" then maybe it should just be deleted?

conformance/fhir-ig-carin-bb/src/test/resources/JSON/StructureDefinition-C4BB-Patient.json
JohnTimm: A minimal example is hard to get right so I decided to use one that I knew allowed us to exercise what we needed. Yes, we will remove it when we upgrade.

fhir-path/src/test/java/com/ibm/fhir/path/test/MemberOfFunctionTest.java
JohnTimm: I'll add the comment in my next PR.

Issue #2391, opened 2021-05-19
FetchResourcePayloadsDAO hits an error condition under debug without _since parameter


prb112:
**Describe the bug**
FetchResourcePayloadsDAO hits an error condition under debug without _since parameter

```
[5/19/21, 19:43:47:460 UTC] 0000005e com.ibm.fhir.persistence.jdbc.impl.FHIRPersistenceJDBCImpl   E Unexpected error while processing token value records.
java.lang.NullPointerException
    at java.sql.Timestamp.from(Timestamp.java:545)
    at com.ibm.fhir.persistence.jdbc.dao.impl.FetchResourcePayloadsDAO.count(FetchResourcePayloadsDAO.java:184)
    at com.ibm.fhir.persistence.jdbc.impl.FHIRPersistenceJDBCImpl.fetchResourcePayloads(FHIRPersistenceJDBCImpl.java:2652)
    at com.ibm.fhir.bulkdata.jbatch.export.fast.ResourcePayloadReader.readItem(ResourcePayloadReader.java:307)
    at com.ibm.jbatch.container.artifact.proxy.ItemReaderProxy.readItem(ItemReaderProxy.java:77)
    at com.ibm.jbatch.container.controller.impl.ChunkStepControllerImpl.readItem(ChunkStepControllerImpl.java:329)
    at com.ibm.jbatch.container.controller.impl.ChunkStepControllerImpl.readAndProcess(ChunkStepControllerImpl.java:245)
    at com.ibm.jbatch.container.controller.impl.ChunkStepControllerImpl.invokeChunk(ChunkStepControllerImpl.java:626)
    at com.ibm.jbatch.container.controller.impl.ChunkStepControllerImpl.invokeCoreStep(ChunkStepControllerImpl.java:795)
    at com.ibm.jbatch.container.controller.impl.BaseStepControllerImpl.execute(BaseStepControllerImpl.java:295)
    at com.ibm.jbatch.container.controller.impl.ExecutionTransitioner.doExecutionLoop(ExecutionTransitioner.java:118)
    at com.ibm.jbatch.container.controller.impl.WorkUnitThreadControllerImpl.executeCoreTransitionLoop(WorkUnitThreadControllerImpl.java:96)
    at com.ibm.jbatch.container.controller.impl.WorkUnitThreadControllerImpl.executeWorkUnit(WorkUnitThreadControllerImpl.java:178)
[...]

PR #2392, opened 2021-05-19
FetchResourcePayloadsDAO hits an error condition under debug without _since parameter #2391

prb112:


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/FetchResourcePayloadsDAO.java
```diff
@@ -181,14 +181,12 @@ public int count(Connection c) throws FHIRPersistenceException {
         final String select = query.toString();
 
         if (logger.isLoggable(Level.FINE)) {
-            Timestamp from = Timestamp.from(fromLastUpdated);
```
prb112, line 184: This line caused a NPE when logging is turned on. 
We tested it with logging on/logging off.

PR #2393, opened 2021-05-19
Issue #2296 - miscellaneous updates

JohnTimm:

### DIFF
--- a/fhir-validation/src/main/java/com/ibm/fhir/validation/FHIRValidator.java
+++ b/fhir-validation/src/main/java/com/ibm/fhir/validation/FHIRValidator.java
@@ -35,12 +35,14 @@ import com.ibm.fhir.model.type.Extension;
 import com.ibm.fhir.model.type.code.IssueSeverity;
 import com.ibm.fhir.model.type.code.IssueType;
 import com.ibm.fhir.model.util.ModelSupport;
+import com.ibm.fhir.path.FHIRPathBooleanValue;
 import com.ibm.fhir.path.FHIRPathElementNode;
 import com.ibm.fhir.path.FHIRPathNode;
 import com.ibm.fhir.path.FHIRPathResourceNode;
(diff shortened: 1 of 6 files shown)


prb112 (approved):

fhir-validation/src/main/java/com/ibm/fhir/validation/FHIRValidator.java
```diff
@@ -168,6 +170,7 @@ private FHIRValidator() { }
             throw new IllegalArgumentException("Root must be resource node");
         }
         try {
+            evaluationContext.setExternalConstant(ResolveFunction.RESOLVE_RELATIVE_REFERENCES, FHIRPathBooleanValue.TRUE);
```
lmsurpre, line 173: are you sure you want to set this to true?

lmsurpre: could it be an unexpected side-effect of callers that have passed an evaluation context?

JohnTimm: Yes.

lmsurpre: > are you sure you want to set this to true?
john explained to me that we have 2 layers to this opt-in.  setting RESOLVE_RELATIVE_REFERENCES to true is just the first layer.  it still won't actually resolve those references unless the corresponding fhir-server-config property is set such that we use the ServerResolveFunction and not the default ResolveFunction.

lmsurpre: as for the side-effect concern, john informs me that we already use this pattern in a few other spots.
setting an external constant more than once shouldn't cause any issues, so there's no concern for re-using the same context for additional validations.

Issue #2285, opened 2021-04-24
Use the "skippableUpdate" optimization during bulk import


lmsurpre:
**Is your feature request related to a problem? Please describe.**
In #2263 we added support for a server-side update optimization.
This feature request is to start using that optimization when we process a bulk import.
My theory is that this will be much faster for "repeated import" scenarios.

**Describe the solution you'd like**
We should use the server optimization by default.
Optionally, we might want to introduce a config parameter to opt out of it.

**Describe alternatives you've considered**

**Acceptance Criteria**
1.  
   GIVEN an existing resource in the server
   WHEN we get an import request that contains a copy of that same resource (same type, id, and contents)
   THEN avoid performing the update when the contents match

**Additional context**

d0roppe:
This does not seem to be quite right. I set the fhirServer/bulkdata/core/enableSkippableUpdates to false and it did not import the data, I set it to true and it imported the data.
With set to false:
```
[INFO    ] Operation Type: $import
[INFO    ] Resource Type         failures              success               processed             totalRead             totalSkip             totalValidation       totalWrite            fileSize              Resource Size         
[INFO    ] Patient               200                   0                     0                     2031                  0                     0                     412                   145284                726.0                 
[INFO    ]  ---- Total: 0 ImportRate: 0.00 ----
```

With set to true:
```
[INFO    ] Operation Type: $import
[INFO    ] Resource Type         failures              success               processed             totalRead             totalSkip             totalValidation       totalWrite            fileSize              Resource Size         
[INFO    ] Patient               0                     200                   200                   2279                  0                     0                     18218                 145284                726.0                 
[INFO    ]  ---- Total: 200 ImportRate: 8.19 ----
```

d0roppe:
Also note I did a second import with the value set to True right after the first import with the value set to True, and this was the output:
```
[INFO    ] Operation Type: $import
[INFO    ] Resource Type         failures              success               processed             totalRead             totalSkip             totalValidation       totalWrite            fileSize              Resource Size         
[INFO    ] Patient               0                     200                   200                   310                   0                     0                     9071                  145284                726.0                 
[INFO    ]  ---- Total: 200 ImportRate: 17.12 ----
```
was wondering if the total skip should still be 0

prb112:
I had flipped the logic in a downstream, and fixed the logic in current azure provider

d0roppe:
Moving this back to in progress, as importing a patient that was deleted was working earlier. 
```
[...]

PR #2398, opened 2021-05-20
Use the "skippableUpdate" optimization during bulk import #2285

prb112:
- Adds skippable updates to Bulk Import 
- Adds report to number of skipped 


operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/impl/AbstractSystemConfigurationImpl.java
```diff
@@ -352,4 +352,9 @@ public String getOperationOutcomeProvider(String provider) {
     public boolean hasStorageProvider(String storageProvider) {
         return FHIRConfigHelper.getStringProperty("fhirServer/bulkdata/storageProviders/" + storageProvider + "/type", null) != null;
     }
+
+    @Override
+    public boolean enableSkippableUpdates() {
+        return FHIRConfigHelper.getBooleanProperty("fhirServer/bulkdata/core/enableSkippableUpdates", Boolean.TRUE);
```
punktilious, line 358: Do you think it's worth making`fhirServer/bulkdata` prefix a constant? Sometimes I find the resulting concatenation harder to read...so not sure.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/load/ChunkWriter.java
```diff
@@ -248,4 +261,68 @@ public void writeItems(List<java.lang.Object> arg0) throws Exception {
...
+     * @param skip should skip the resource if it matches
+     * @param localCache map containing the key-saltHash
+     * @param persistence used to facilitate the calls to the underlying db
+     * @param context used in db calls
+     * @param logicalId the logical id of the FHIR resource (e.g. 1-2-3-4)
+     * @param resource the FHIR Resource
+     * @return outcomes including information or warnings
+     * @throws FHIRPersistenceException
+     */
+    public OperationOutcome conditionalFingerprintUpdate(ImportTransientUserData chunkData, boolean skip, Map<String, SaltHash> localCache, FHIRPersistence persistence, FHIRPersistenceContext context, String logicalId, Resource resource) throws FHIRPersistenceException {
```
lmsurpre, line 284: We should probably try to avoid the extra read + fingerprinting in cases where we're not even going to skip the updates when the fingerprints match.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/load/ChunkWriter.java
```diff
@@ -248,4 +261,68 @@ public void writeItems(List<java.lang.Object> arg0) throws Exception {
...
+                    .severity(IssueSeverity.INFORMATION)
+                    .code(IssueType.INFORMATIONAL)
+                    .details(CodeableConcept.builder()
+                        .text(string("Update resource matches the existing resource; skipping the update for '" + key + "'"))
+                        .build())
+                    .build())
+                .build();
+        } else {
+            // We need to update the db and update the local cache
+            oo = persistence.update(context, logicalId, resource).getOutcome();
```
lmsurpre, line 322: I'm wondering if we should add the previously-read "oldResource" to the PersistenceEvent within context.
If the event is missing this prevResource then the JDBC impl will do yet another read to get it:
```
            // If the "previous resource" is set in the persistence event, then get the
            // existing version # from that.
            if (context.getPersistenceEvent() != null && context.getPersistenceEvent().isPrevFhirResourceSet()) {
                Resource existingResource = context.getPersistenceEvent().getPrevFhirResource();
                if (existingResource != null) {
                    log.fine("Using pre-fetched 'previous' resource.");
                    String version = existingResource.getMeta().getVersionId().getValue();
                    existingVersion = Integer.valueOf(version);
                }
            }

            // Otherwise, go ahead and read the resource from the datastore and get the
            // existing version # from it.
            else {
                log.fine("Fetching 'previous' resource for update.");
                existingResourceDTO = resourceDao.read(logicalId, resourceType.getSimpleName());
                if (existingResourceDTO != null) {
                    existingVersion = existingResourceDTO.getVersionId();
                }
            }
```

prb112:
> When I opened #2285, my hunch was that skipping the update would be more performant. But I don't think I realized that we were doing a straight update (with no bulkdata-layer read first). Because the current PR seems to always do an extra read, I added a couple comments that could potentially reduce the number of reads we make before performing the update.
> 
> However, even with those tweaks, its probably worth testing my perf assumption before merging this one.

I could run a small workload to test, but I'm unsure if it adds performance.  I did add the check to see if we're skipping an update.

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/impl/AbstractSystemConfigurationImpl.java
prb112: it's subtle, but it's a big change in that class.  if we change this one, we change them all in the class. I'm not sure if it is a benefit in this case.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/load/ChunkWriter.java
```diff
@@ -248,4 +261,68 @@ public void writeItems(List<java.lang.Object> arg0) throws Exception {
...
+     * @return outcomes including information or warnings
+     * @throws FHIRPersistenceException
+     */
+    public OperationOutcome conditionalFingerprintUpdate(ImportTransientUserData chunkData, boolean skip, Map<String, SaltHash> localCache, FHIRPersistence persistence, FHIRPersistenceContext context, String logicalId, Resource resource) throws FHIRPersistenceException {
+        // Key is scoped to the ResourceType.
+        String key = resourceType + "/" + logicalId;
+        SaltHash oldBaseLine = localCache.get(key);
+
+        ResourceFingerprintVisitor fp = new ResourceFingerprintVisitor();
+        if (oldBaseLine == null) {
```
prb112, line 290:
```suggestion
        if (oldBaseLine == null && skip) {
```

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/load/ChunkWriter.java
prb112: Good catch, I had intended to add a skip on 290, when I add it it'll skip the extra read.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/load/ChunkWriter.java
prb112, line 322:
```suggestion
            if (oldResource != null) {
                // Old Resource is set so we avoid an extra read
                context.getPersistenceEvent().setPrevFhirResource(oldResource);
            }
            oo = persistence.update(context, logicalId, resource).getOutcome();
```

prb112: I've added with an if-else

lmsurpre: looks good

lmsurpre (approved):
LGTM.  I'd still like to see some performance comparison but I think with the requested changes to avoid unneeded reads, it should be a win.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/load/ChunkWriter.java
lmsurpre: This addresses my concern for the extra read, but its still computing and caching the resource fingerprint, even though that isn't used.  Not sure how much that costs, but I think we should skip that as well.

prb112: I added a wrap to the skip.  thanks for the catch

lmsurpre (approved):
LGTM

prb112:
> LGTM

I'll wait until post 4.8.3

PR #2404, opened 2021-05-20
bump project modules to 4.9.0-SNAPSHOT

lmsurpre:

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -17,6 +17,7 @@ A clear and concise description of what you want to happen.
 A clear and concise description of any alternative solutions or features you've considered.
 
 **Acceptance Criteria**
+
 1.  
    GIVEN [a precondition]
     AND [another precondition]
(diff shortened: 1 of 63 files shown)


prb112 (approved):
LGTM,  should we push a small change to cause a rebuild?

lmsurpre:
> LGTM, should we push a small change to cause a rebuild?

how would you feel about changing the version of fhir-examples in the parent pom so that our modules depend on 4.8.1 instead of 4.9.0-SNAPSHOT? :-)

prb112:
> > LGTM, should we push a small change to cause a rebuild?
> 
> how would you feel about changing the version of fhir-examples in the parent pom so that our modules depend on 4.8.1 instead of 4.9.0-SNAPSHOT? :-)

I opted not to mention that; however it is my preference. We'd want to set it outside of the properties (so the automation doesn't try to reference 4.9.0-SNAPSHOT). We might need to update the build somewhat to take advantage of it.

lmsurpre:
> > > LGTM, should we push a small change to cause a rebuild?
> > 
> > 
> > how would you feel about changing the version of fhir-examples in the parent pom so that our modules depend on 4.8.1 instead of 4.9.0-SNAPSHOT? :-)
> 
> I opted not to mention that; however it is my preference. We'd want to set it outside of the properties (so the automation doesn't try to reference 4.9.0-SNAPSHOT). We might need to update the build somewhat to take advantage of it.

I was just going to set it in the fhir-parent/pom.xml property.  I'm not sure that I understand the concern of with that, so for now I just pushed a minor change to the ISSUE_TEMPLATE to force a rebuild.

PR #2408, opened 2021-05-21
issue 2405 fixed concurrent updates to correctly return 409 on version conflict

punktilious:
Tagged as a schema-change because the Db2 and PostgreSQL stored procedures have been updated. There were no other structural changes to the schema (no table/column/index changes).


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/derby/DerbyResourceDAO.java
```diff
@@ -304,7 +303,7 @@ public long storeResource(String tablePrefix, List<ExtractedParameterValue> para
                     stmt.setString(2, p_logical_id);
                     stmt.setString(3, p_is_deleted ? "Y" : "N");
                     stmt.setTimestamp(4, p_last_updated, UTC);
-                    stmt.setInt(5, 1); // initial version
+                    stmt.setInt(5, p_version); // initial version
```
lmsurpre, line 306: will p_version always be 1 on a create?

punktilious: That's business logic decided by the Java persistence code...so I wanted to improve separation of concerns and just trust the Java code will pass the value it wants to use for the initial value (because it will have also injected that value into the JSON). It's subtle, but it means we now only have one bit of code deciding what the initial value should be (and that's in Java).

punktilious: That's also why I decided to not add a check that p_version == 1...the stored proc shouldn't care - it just needs to do as it's told.

lmsurpre: sounds good

lmsurpre (approved):
LGTM

lmsurpre:
I'd love to put a plan in place for getting our ConcurrentUpdateTest fixed up and running as part of each CI (at least against PostgreSQL and Db2 if not Derby). If we don't do it under this PR, we should open a separate issue for it.

lmsurpre:
> I'd love to put a plan in place for getting our ConcurrentUpdateTest fixed up and running as part of each CI (at least against PostgreSQL and Db2 if not Derby). If we don't do it under this PR, we should open a separate issue for it.

this was done under #2420

Issue #2411, opened 2021-05-21
Vanity URLs with AWS result in bucket NOT_FOUND on $import


prb112:
**Describe the bug**
Vanity URLs with AWS result in bucket NOT_FOUND on $import

**Environment**
Which version of IBM FHIR Server? 4.8.1

**To Reproduce**
1. On AWS, create a bucket such as my-bucket
2. Configure bulkdata with a external and internal endpoint with the bucketized url https://my-bucket.s3.notmyhost.host.com
Note: the api endpoint works fine.
3. Execute an $import with a file that exist

**Expected behavior**
The File should be imported

**Additional context**
@d0roppe found this while testing.

prb112:
PR  https://github.com/IBM/FHIR/pull/2465/files

d0roppe:
with the documented change of only allowing the vanity URL on the endpointExternal URL it is working
            "endpointInternal": "https://s3.region.host-name.com",
            "endpointExternal": "https://bucket-name.s3.region.host-name.com",

PR #2412, opened 2021-05-21
Vanity URLs with AWS result in bucket NOT_FOUND on  #2411

prb112:

### DIFF
--- a/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
+++ b/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
@@ -177,7 +177,7 @@ public class S3Provider implements Provider {
         // In certain circumstances, a direct url to the bucket can be used. https://mybucketdemo123.s3.us.east-2.amazonaws.com
         // versus an API enabled url e.g. https://s3.us.east-2.amazonaws.com
         // These end up with TWO different responses, the former is false, and the latter is true.
-        if (!client.doesBucketExistV2(bucketName)) {
+        if (ex || !client.doesBucketExistV2(bucketName)) {
             logger.warning("Bucket '" + bucketName + "' not found! Client [" + (client != null) + "]");
         }
 
--- a/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
+++ b/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
@@ -80,7 +80,7 @@ public class S3Provider implements Provider {
 
         String cosLocation = adapter.getStorageProviderLocation(source);
         String cosEndpointUrl = adapter.getStorageProviderEndpointInternal(source);
-        String bucketName = adapter.getStorageProviderBucketName(source);
+        bucketName = adapter.getStorageProviderBucketName(source);
 
         boolean iam = adapter.isStorageProviderAuthTypeIam(source);
         isExportPublic = adapter.isStorageProviderExportPublic(source);
@@ -171,10 +171,16 @@ public class S3Provider implements Provider {
      * @return
      */
     public boolean exists() {
-        boolean ex = client != null && client.doesBucketExistV2(bucketName);
-        if (ex == false) {
+        boolean ex = client != null;
+
+        // We only want to log a warning here, and assume it's true if the client exists
+        // in certain circumstances, a direct url to the bucket can be used. https://mybucketdemo123.s3.us.east-2.amazonaws.com
+        // versus an API enabled url e.g. https://s3.us.east-2.amazonaws.com
+        // These end up with TWO different responses, the former is false, and the latter is true.
+        if (!client.doesBucketExistV2(bucketName)) {
             logger.warning("Bucket '" + bucketName + "' not found! Client [" + (client != null) + "]");
         }
+
         return ex;
     }
 
--- a/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
+++ b/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
@@ -169,6 +169,7 @@ public class S3Provider implements Provider {
      * checks to see if the bucket exists.
      *
      * @return
+     * @implNote to maintain the exists method behavior, we're only checking the client exists or does not  (true  or false), and warning if the bucket is not found as the S3 client does not do an exists properly when using a doesBucketExistV2 when using a vanity url to access the service on AWS.
      */
     public boolean exists() {
         boolean ex = client != null;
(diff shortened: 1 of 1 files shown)


fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
```diff
@@ -171,10 +171,16 @@ private static AmazonS3 getClient(boolean iam, String cosApiKeyProperty, String
      * @return
      */
     public boolean exists() {
-        boolean ex = client != null && client.doesBucketExistV2(bucketName);
-        if (ex == false) {
+        boolean ex = client != null;
+
+        // We only want to log a warning here, and assume it's true if the client exists
+        // in certain circumstances, a direct url to the bucket can be used. https://mybucketdemo123.s3.us.east-2.amazonaws.com
```
lmsurpre, line 177:
```suggestion
        // We only want to log a warning here, and assume it's true if the client exists.
        // In certain circumstances, a direct url to the bucket can be used. https://mybucketdemo123.s3.us.east-2.amazonaws.com
```

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
```diff
@@ -171,10 +171,16 @@ private static AmazonS3 getClient(boolean iam, String cosApiKeyProperty, String
      * @return
      */
     public boolean exists() {
-        boolean ex = client != null && client.doesBucketExistV2(bucketName);
-        if (ex == false) {
+        boolean ex = client != null;
+
+        // We only want to log a warning here, and assume it's true if the client exists
```
lmsurpre, line 176: So client will always be null if the bucket doesn't exist?

lmsurpre: Or, if not, can you help me understand how will the end-user-facing error will change from this commit (in the case of a non-existent bucket)?

prb112: previously we were checking if the client was valid, and then doing the check.  In this case, we're only warning if the client doesn't exist.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
```diff
@@ -171,10 +171,16 @@ private static AmazonS3 getClient(boolean iam, String cosApiKeyProperty, String
...
+        // in certain circumstances, a direct url to the bucket can be used. https://mybucketdemo123.s3.us.east-2.amazonaws.com
+        // versus an API enabled url e.g. https://s3.us.east-2.amazonaws.com
+        // These end up with TWO different responses, the former is false, and the latter is true.
+        if (!client.doesBucketExistV2(bucketName)) {
```
prb112, line 180:
```suggestion
        if (ex || !client.doesBucketExistV2(bucketName)) {
```

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
prb112: I fixed line 180 to include ex, it was suppose to be !ex

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
```diff
@@ -171,10 +171,16 @@ private static AmazonS3 getClient(boolean iam, String cosApiKeyProperty, String
...
     public boolean exists() {
-        boolean ex = client != null && client.doesBucketExistV2(bucketName);
-        if (ex == false) {
+        boolean ex = client != null;
+
+        // We only want to log a warning here, and assume it's true if the client exists.
+        // In certain circumstances, a direct url to the bucket can be used. https://mybucketdemo123.s3.us.east-2.amazonaws.com
+        // versus an API enabled url e.g. https://s3.us.east-2.amazonaws.com
+        // These end up with TWO different responses, the former is false, and the latter is true.
+        if (ex || !client.doesBucketExistV2(bucketName)) {
```
prb112, line 180:
```suggestion
        if (!ex || !client.doesBucketExistV2(bucketName)) {
```

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
lmsurpre: I'm sorry, I'm still a bit confused.  The logic seems the same as the original logic (outside this PR) now.
I think maybe you were purposefully no longer relying on `client.doesBucketExistV2(bucketName)` and I was trying to understand why (and how that would bubble up in the errors we generate).

lmsurpre: ok, i had another look and i understand how its different now.

lmsurpre (approved):
Paul informed me that the system will still generate a reasonable error message for cases where the bucket doesn't actually exist.

However, I'd still like to better understand why the S3 SDK doesn't work with the bucket "vanity url" when the other methods do.  I wonder if its worth opening on stack overflow (or similar)...

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
```diff
@@ -171,10 +171,16 @@ private static AmazonS3 getClient(boolean iam, String cosApiKeyProperty, String
      * @return
      */
     public boolean exists() {
```
lmsurpre, line 173: I wonder if we should update the javadoc to explain that it doesn't actually return false when the bucket doesn't exist now...

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
```diff
@@ -171,10 +171,16 @@ private static AmazonS3 getClient(boolean iam, String cosApiKeyProperty, String
      * @return
```
prb112, line 171:
```suggestion
     * @return
     * @implNote to maintain the exists method behavior, we're only checking the client exists or does not  (true  or false), and warning if the bucket is not found as the S3 client does not do an exists properly when using a doesBucketExistV2 when using a vanity url to access the service on AWS.
```

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
prb112: Updated

PR #2415, opened 2021-05-23
issue 2420 set Derby isolationLevel to avoid deadlock during ingestion

punktilious:
Also aligned the stored procedures for Db2 and PostgreSQL so they match the logic of the Derby DAO. This ended up eliminating an update statement which we were executing for new resources which could be avoided. This should help PostgreSQL in particular, because it avoids tombstones which would need to be vacuumed.

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/derby/DerbyResourceDAO.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/derby/DerbyResourceDAO.java
@@ -22,6 +22,7 @@ import java.util.logging.Logger;
 
 import javax.transaction.TransactionSynchronizationRegistry;
 
+import com.ibm.fhir.database.utils.derby.DerbyMaster;
 import com.ibm.fhir.database.utils.derby.DerbyTranslator;
 import com.ibm.fhir.persistence.exception.FHIRPersistenceException;
 import com.ibm.fhir.persistence.exception.FHIRPersistenceVersionIdMismatchException;
@@ -204,7 +205,6 @@ public class DerbyResourceDAO extends ResourceDAOImpl {
         logger.entering(CLASSNAME, METHODNAME);
 
         Long v_logical_resource_id = null;
-        Long v_current_resource_id = null;
         Long v_resource_id = null;
         Integer v_resource_type_id = null;
         boolean v_new_resource = false;
@@ -221,16 +221,45 @@ public class DerbyResourceDAO extends ResourceDAOImpl {
             throw new IllegalStateException("resource type not found: " + v_resource_type);
         }
 
+        // Get a value for the new resource_id we'll be storing later
+        final String sql2 = "VALUES (NEXT VALUE FOR fhir_sequence)";
+        try (PreparedStatement stmt = conn.prepareStatement(sql2)) {
+            try (ResultSet res = stmt.executeQuery()) {
+                if (res.next()) {
+                    v_resource_id = res.getLong(1); //Assign result of the above query
+                    res.next();
+                }
+                else {
+                    // unlikely
+                    throw new IllegalStateException("no row returned: " + sql2);
+                }
+            }
+        }
+
+        if (logger.isLoggable(Level.FINEST)) {
+            // Debug locking...
+            DerbyMaster.dumpLockInfo(conn);
+        }
+
         // Get a lock at the system-wide logical resource level. Note the Derby-specific syntax
+        if (logger.isLoggable(Level.FINEST)) {
+            logger.finest("Getting LOGICAL_RESOURCES row lock for: " + v_resource_type + "/" + p_logical_id);
+        }
         final String SELECT_FOR_UPDATE = "SELECT logical_resource_id FROM logical_resources WHERE resource_type_id = ? AND logical_id = ? FOR UPDATE WITH RS";
         try (PreparedStatement stmt = conn.prepareStatement(SELECT_FOR_UPDATE)) {
             stmt.setInt(1, v_resource_type_id);
             stmt.setString(2, p_logical_id);
             ResultSet rs = stmt.executeQuery();
             if (rs.next()) {
+                if (logger.isLoggable(Level.FINEST)) {
+                    logger.finest("Resource locked: " + v_resource_type + "/" + p_logical_id);
+                }
                 v_logical_resource_id = rs.getLong(1);
             }
             else {
+                if (logger.isLoggable(Level.FINEST)) {
+                    logger.finest("Resource not found: " + v_resource_type + "/" + p_logical_id);
+                }
                 v_not_found = true;
                 v_logical_resource_id = -1L; // just to be careful
             }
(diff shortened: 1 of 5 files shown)


fhir-persistence-schema/src/main/resources/db2/add_any_resource.sql
```diff
@@ -93,9 +96,9 @@ BEGIN
       -- we created the logical resource and therefore we already own the lock. So now we can
       -- safely create the corresponding record in the resource-type-specific logical_resources table
       PREPARE stmt FROM
-         'INSERT INTO ' || v_schema_name || '.' || p_resource_type || '_logical_resources (mt_id, logical_resource_id, logical_id, is_deleted, last_updated, version_id) '
-      || '     VALUES (?, ?, ?, ?, ?, ?)';
-      EXECUTE stmt USING {{ADMIN_SCHEMA_NAME}}.sv_tenant_id, v_logical_resource_id, p_logical_id, p_is_deleted, p_last_updated, p_version;
+         'INSERT INTO ' || v_schema_name || '.' || p_resource_type || '_logical_resources (mt_id, logical_resource_id, logical_id, is_deleted, last_updated, version_id, current_resource_id) '
+      || '     VALUES (?, ?, ?, ?, ?, ?, ?)';
+      EXECUTE stmt USING {{ADMIN_SCHEMA_NAME}}.sv_tenant_id, v_logical_resource_id, p_logical_id, p_is_deleted, p_last_updated, p_version, v_resource_id;
```
prb112, line 101: interesting that we didn't set current_resource_id before... good find.

fhir-server/liberty-config/configDropins/defaults/datasource.xml
```diff
@@ -7,7 +7,7 @@
     <!-- ============================================================== -->
     <!-- TENANT: default; DSID: default; TYPE: read-write               -->
     <!-- ============================================================== -->
-    <dataSource id="bootstrapDefaultDefault" jndiName="jdbc/bootstrap_default_default" type="javax.sql.XADataSource" statementCacheSize="200" syncQueryTimeoutWithTransactionTimeout="true" validationTimeout="30s">
+    <dataSource id="bootstrapDefaultDefault" jndiName="jdbc/bootstrap_default_default" type="javax.sql.XADataSource" statementCacheSize="200" syncQueryTimeoutWithTransactionTimeout="true" validationTimeout="30s" isolationLevel="TRANSACTION_READ_COMMITTED">
```
prb112, line 10: do we need to reflect the same with postgres/db2? based on the conversation, it sounds like no since the locks are the way you want.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/derby/DerbyResourceDAO.java
```diff
@@ -504,4 +568,18 @@ public int getOrCreateResourceType(String resourceTypeName, Connection conn) thr
...
+     */
+    private void safeClose(ResultSet rs) {
+        try {
+            if (rs != null) {
+                rs.close();
+            }
+        } catch (SQLException x) {
+            // NOP
+        }
+    }
```
prb112, line 584: I thought ResultSets autoclose out of scope?

prb112: actually, where is this used?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/derby/DerbyResourceDAO.java
```diff
@@ -296,34 +345,49 @@ public long storeResource(String tablePrefix, List<ExtractedParameterValue> para
 
                 // Insert the resource-specific logical resource record. Remember that logical_id is denormalized
                 // so it gets stored again here for convenience
-                final String sql3 = "INSERT INTO " + tablePrefix + "_logical_resources (logical_resource_id, logical_id, is_deleted, last_updated, version_id) VALUES (?, ?, ?, ?, ?)";
-                try (PreparedStatement stmt = conn.prepareStatement(sql3)) {
+                if (logger.isLoggable(Level.FINEST)) {
+                    logger.finest("Creating " + tablePrefix + "_logical_resources row: " + v_resource_type + "/" + p_logical_id);
+                }
+                final String sql5 = "INSERT INTO " + tablePrefix + "_logical_resources (logical_resource_id, logical_id, is_deleted, last_updated, version_id, current_resource_id) VALUES (?, ?, ?, ?, ?, ?)";
```
prb112, line 351: Looks consistent/good

prb112 (approved):
LGTM - changes make sense based on thread

fhir-persistence-schema/src/main/resources/db2/add_any_resource.sql
punktilious: Yeah, we weren't setting it because we didn't establish the resource_id value until later in the procedure and it just felt wrong creating a relationship to something which didn't actually exist yet (because we haven't inserted the xx_resources row). But because the id value is obtained from a sequence, we can just grab the value early and use it right away when we do the very first insert into xx_logical_resources.

fhir-server/liberty-config/configDropins/defaults/datasource.xml
punktilious: For PostgreSQL default isolation level is Read Committed, which is what we want. For Db2, it's an open question because allegedly the default is Repeatable Read which is more restrictive than necessary, but it doesn't impact us in this case because of the better concurrency control implementation within Db2.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/derby/DerbyResourceDAO.java
punktilious: Yeah, it's not. I added this as part of debugging to see if actively closing the ResultSet would make a difference (thinking there might be a bug lurking underneath even though closing the statement should close any associated ResultSet). I'll remove.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/derby/DerbyResourceDAO.java
```diff
@@ -204,7 +205,6 @@ public long storeResource(String tablePrefix, List<ExtractedParameterValue> para
         logger.entering(CLASSNAME, METHODNAME);
 
         Long v_logical_resource_id = null;
-        Long v_current_resource_id = null;
```
lmsurpre, line 207: not related to this line and also very minor, but we could update the javadoc of this method for correctness.

specifically the path "fhir-persistence-schema/src/main/resources/add_any_resource.sql" should be updated to "fhir-persistence-schema/src/main/resources/*db2*/add_any_resource.sql"

lmsurpre (approved):
assuming we remove that unused safeClose method, this LGTM

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/derby/DerbyResourceDAO.java
punktilious: done

PR #2418, opened 2021-05-24
Issue #2296, Issue #2377 - miscellaneous FHIRPath updates

JohnTimm:
Miscellaneous Updates

- Fold `resolveRelativeReferences` into the `EvaluationContext` class. Typically FHIRPath "external constants" are treated as data. In this case, however, we are using an external constant as configuration parameter. It makes more sense to make this a field in the `EvaluationContext` class itself.
- Created `hasResourceAssertedProfile` method in `ProfileSupport`
- Added validation logic to the `FHIRPathUtil.compile` method to prevent partial expressions from being evaluated without throwing an exception


prb112 (approved):
LGTM

Issue #2421, opened 2021-05-24
Add operations for clearing terminology caches


csandersdev:
**Is your feature request related to a problem? Please describe.**
The terminology subsystem caches terminology resources (e.g. CodeSystems and ValueSets) such that when a resource change is made to the same URL and version, the resource changes are not immediately reflected in all operation results (e.g. validate-code).

**Describe the solution you'd like**
I would like for the server to either clear the cache automatically on resource update (difficult to do right now because the cache is not shared between clustered nodes) or provide mechanism through which a client application can clear the cache on demand.

**Describe alternatives you've considered**
This isn't an issue if the client application does not try to reuse the same URL and version, but that is common at dev time. There is a server flag ``fhirServer/term/disableCaching`` that was introduced that will disable caching in the terminology subsystem, but is not desirable for a production application. We have a production use case that wants caching, but also wants to be able to clear the cache after updating a single resource url + version.

**Acceptance Criteria**

1.  
   GIVEN a valueset resource exists
    AND the valueset cache has been populated (e.g. $validate-code operation called)
   WHEN the valueset resource is updated
    AND the clear-cache operation is called on that resource
   THEN the next valueset operation returns the correct result for the new resource.

2.  
   GIVEN a codesystem resource exists
    AND the codesystem cache has been populated (e.g. $validate-code operation called)
   WHEN the codesystem resource is updated
    AND the clear-cache operation is called on that resource
   THEN the next codesystem operation returns the correct result for the new resource.

3.  
   GIVEN an intensional valueset resource exists
    AND the codesystem cache has been populated (e.g. ValueSet/xxx/$expand operation called)
   WHEN the codesystem resource is updated
    AND the clear-cache operation is called on that resource
   THEN the next valuset expand operation returns the correct result for the new resource.

**Additional Context**
It is a known constraint that the FHIR server caches are JVM-local. In the event that a clear-cache operation is needed, the client application will need to call clear-cache on each FHIR server in a cluster.

lmsurpre:
A potential workaround for this is to do a rolling restart of the fhir-server pods if you really need to clear the caches.

JohnTimm:
> A potential workaround for this is to do a rolling restart of the fhir-server pods if you really need to clear the caches.

@lmsurpre the other option we discussed is using the `fhirServer/term/disableCaching` configuration option during `CodeSystem` and `ValueSet` development. The clear cache operations give finer grained control over what gets invalidated down to specific value sets.

[...]

PR #2422, opened 2021-05-24
#2421 - Add terminology clear-cache operations

csandersdev:
The terminology subsystem caches terminology resources by URL and/or URL + version. If the user tries to update a CodeSystem or ValueSet resource keeping the same URL + version, the cache does not get cleared and some operations (e.g. validate-code) do not return the expected results after update. This change adds an operation ($clear-cache) that can be called on either terminology resource and will clear the appropriate caches.

Signed-off-by: Corey Sanders <corey.thecolonel@gmail.com>


fhir-server-test/src/test/java/com/ibm/fhir/server/test/terminology/CodeSystemClearCacheOperationTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+        // After cache is cleared, subsumes outcome should be updated
+        clearCache();
+        
+        parameters = validateCode(SNOMED_CT, "K");
+        assertEquals(getBooleanParameterValue(parameters, "result"), Boolean.FALSE);
+    }
+	
+	private Parameters validateCode(String system, String code) throws Exception {
+		Response response = doGet(BASE_VALID_URL + "/" + CODE_SYSTEM_ID + "/$validate-code", "system", system, "code", code);
```
JohnTimm, line 94: I see several horizontal spacing issues. In some cases it looks like more than 4 spaces are present. In other cases it looks like less than 4 spaces are present. I also what looks to be tab characters. Will need to pull down the raw file to verify.

csandersdev: Yeah, sorry about that. My Eclipse had some config issues. I'll do a sweep through again and make sure I've got the styles correct.

JohnTimm:
@csandersdev Per our discussion and given the general feedback that I received from the team, this is a low-level function that addresses a very specific requirement and should not be packaged with the FHIR server by default. Since it is currently part of the `fhir-operation-term` module it will need to be broken out into a separate project so that it can be built and packaged as an "add-on".

csandersdev:
@JohnTimm How about ``/add-ons/fhir-operation-term-cache``? Do you have an opinion on what we should do with the integration tests in fhir-server-test after the move? There isn't a ton of code that I'm using there, but I would hate to duplicate.

JohnTimm:
> @JohnTimm How about `/add-ons/fhir-operation-term-cache`? Do you have an opinion on what we should do with the integration tests in fhir-server-test after the move? There isn't a ton of code that I'm using there, but I would hate to duplicate.

I assume that the /add-ons folder would live under the existing operation folder. As for the integration tests I think we can include it in the build automation so they get executed but exclude it from the default FHIR server packaging. @prb112 any thoughts on this?

prb112:
> > @JohnTimm How about `/add-ons/fhir-operation-term-cache`? Do you have an opinion on what we should do with the integration tests in fhir-server-test after the move? There isn't a ton of code that I'm using there, but I would hate to duplicate.
> 
> I assume that the /add-ons folder would live under the existing operation folder. As for the integration tests I think we can include it in the build automation so they get executed but exclude it from the default FHIR server packaging. @prb112 any thoughts on this?

Let's leave in /operations 
Let's update /README.md to indicate that this is an addon and the description. 
Let's update https://github.com/IBM/FHIR/blob/main/build/docker/copy-test-operations.sh to copy the test operation
Let's update automation to copy the operation
-  https://github.com/IBM/FHIR/blob/main/build/persistence/postgres/pre-integration-test.sh#L66 
- https://github.com/IBM/FHIR/blob/main/build/notifications/kafka/pre-integration-test.sh#L40
- https://github.com/IBM/FHIR/blob/main/build/audit/kafka/pre-integration-test.sh#L41
- https://github.com/IBM/FHIR/blob/main/build/pre-integration-test.sh#L76
Let's update

csandersdev:
@prb112 Is your expectation that the operation classes for the new module all live under src/test/java? If that is the case, is there any reason why we wouldn't just keep them in ``fhir-operation-term`` and move them from main to test and then make the rest of the script changes you recommend above for the fhir-operation-term module?

prb112:
> @prb112 Is your expectation that the operation classes for the new module all live under src/test/java? If that is the case, is there any reason why we wouldn't just keep them in `fhir-operation-term` and move them from main to test and then make the rest of the script changes you recommend above for the fhir-operation-term module?

Is the purpose test/dev only? If so, we'll mark it as test/dev only, and still treat this separately. 
We'll also add fhir-operation-test-term-clear-cache as the name.

csandersdev:
> Is the purpose test/dev only? If so, we'll mark it as test/dev only, and still treat this separately.
> We'll also add fhir-operation-test-term-clear-cache as the name.

No. Given current constraints, I expect IHE to use these operations in a production setting. I was just trying to reconcile your suggestion of copying the test artifacts. I don't have anything under src/test/java right now. I'm not clear on why I would need those steps of copying the <module>-tests.jar unless my stuff lived under the test tree.

JohnTimm:
> > Is the purpose test/dev only? If so, we'll mark it as test/dev only, and still treat this separately.
> > We'll also add fhir-operation-test-term-clear-cache as the name.
> 
> No. Given current constraints, I expect IHE to use these operations in a production setting. I was just trying to reconcile your suggestion of copying the test artifacts. I don't have anything under src/test/java right now. I'm not clear on why I would need those steps of copying the -tests.jar unless my stuff lived under the test tree.

I believe that he is suggesting test jars so that they get deployed with the server such that you can still run integration tests. This is exactly what we do with fhir-operation test.

prb112:
I've started a direct message to clarify.

operation/fhir-operation-term-cache/README.md
```diff
@@ -0,0 +1,7 @@
...
+
+If you are in development mode and always want to reload data, the simplest mechanism is to use the ``fhirServer/term/disableCaching``  configuration option. If you are in a hybrid or production mode and require both caching and dynamic content updates, the operations in this module can be used to clear specific cached resources.
+
+This module introduces a $clear-cache operation on the ValueSet and CodeSystem resources. Depending on the type of resource being changed, one or the other or both may need to be called in order to completely clear the cache. For instance, an intensional ValueSet that depends on a server CodeSystem resource, will require a call to both ValueSet/$clear-cache and CodeSystem/$clear-cache.
+
+FHIR terminology caching is done on a per-JVM basis, so the clear-cache operations must be called on every member of a FHIR server cluster after data changes are made.
```
prb112, line 7: It's also per tenant.  You are using CacheManager which is tenant aware.

prb112 (approved):
LGTM - @JohnTimm @lmsurpre . I think this is OK with the automation change.

operation/fhir-operation-term-cache/README.md
```diff
@@ -0,0 +1,7 @@
+The FHIR terminology subsystem assumes that artifacts with the same canonical URL and version are stable and thus can be stored in long-living cache space. If there is a need to iterate the content of a CodeSystem or ValueSet resource without updating the business version, additional operations are needed to clear the cache.
+
+If you are in development mode and always want to reload data, the simplest mechanism is to use the ``fhirServer/term/disableCaching``  configuration option. If you are in a hybrid or production mode and require both caching and dynamic content updates, the operations in this module can be used to clear specific cached resources.
```
lmsurpre, line 3: Not sure what putting two backticks around a word does, but we usually use just one:
```suggestion
If you are in development mode and always want to reload data, the simplest mechanism is to use the `fhirServer/term/disableCaching`  configuration option. If you are in a hybrid or production mode and require both caching and dynamic content updates, the operations in this module can be used to clear specific cached resources.
```

operation/fhir-operation-term-cache/src/main/resources/operation-codesystem-clear-cache.json
```diff
@@ -0,0 +1,52 @@
...
+    "resource": ["CodeSystem"],
+    "system": false,
+    "type": true,
+    "instance": true,
+    "parameter": [    {
+      "name": "url",
+      "use": "in",
+      "min": 0,
+      "max": "1",
+      "documentation": "CodeSystem URL. The server must know the code system (e.g. it is defined explicitly in the server\u0027scode systems, or it is known implicitly by the server",
```
lmsurpre, line 25: missing space.  also, any reason to encode the apostrophe?
```suggestion
      "documentation": "CodeSystem URL. The server must know the code system (e.g. it is defined explicitly in the server's code systems, or it is known implicitly by the server",
```

lmsurpre (approved):
Left a couple minor wordsmith things, but LGTM.

prb112:
I resolved the DCO manually.

JohnTimm:
@csandersdev The latest build failure is due to the fact that we upgraded to v2.0.1 of the JSONP reference implementation and they changed their package name prefix from `javax.json` to `jakarta.json`.

Issue #2420, opened 2021-05-24
Deadlock during concurrent updates on Derby


lmsurpre:
The check to catch the version conflict was a relatively simple fix. Cleaned things up a bit because for this implementation there's no need to support an asserted version id (which might not be the latest version id - useful for logical replication). However the fix did expose a deadlock scenario for Derby caused by the default isolation level being TRANSACTION_REPEATABLE_READ. This resulted in shared locks lingering on rows which were selected when reading the current version of the resource earlier in the transaction. In Derby, these locks are not compatible with the exclusive lock requested by the UPDATE xx_logical_resources statement. The resulting deadlock graph looks something like this:
```
A lock could not be obtained due to a deadlock, cycle of locks and waiters is:
Lock : ROW, LOGICAL_RESOURCES, (115,11)
  Waiting XID : {74421, U} , APP, SELECT logical_resource_id FROM logical_resources WHERE resource_type_id = ? AND logical_id = ? FOR UPDATE WITH RS
  Granted XID : {74409, U}
Lock : ROW, PATIENT_LOGICAL_RESOURCES, (8,29)
  Waiting XID : {74409, X} , APP, UPDATE Patient_logical_resources SET current_resource_id = ?, is_deleted = ?, last_updated = ?, version_id = ? WHERE logical_resource_id = ?
  Granted XID : {74401, S} , {74416, S} , {74409, S} , {74421, S}
. The selected victim is XID : 74421.
```

The solution is to set the isolation level for Derby datasources in openliberty:

```
     <dataSource id="fhirDatasourceBootstrapDefaultDefault" jndiName="jdbc/fhir_default_default" type="javax.sql.XADataSource" statementCacheSize="200" syncQueryTimeoutWithTransactionTimeout="true" isolationLevel="TRANSACTION_READ_COMMITTED">
        <jdbcDriver javax.sql.XADataSource="org.apache.derby.jdbc.EmbeddedXADataSource" libraryRef="fhirSharedLib"/>
            <properties.derby.embedded createDatabase="create" databaseName="derby/fhirDB"/>
        <connectionManager maxPoolSize="50" minPoolSize="10"/>
    </dataSource>
```

[...]

PR #2423, opened 2021-05-24
issue #2420 - enable the ConcurrentUpdateTest for all db types

lmsurpre:
Previously this was flagged as a db2-only test, which helped contribute
to us missing this issue.

Additionally, this test was written to highlight a potential issue but
didn't yet take into account our solution which is to return a 409 for
one of the updates (in the case of a conflict).

The updated test will retry on a tight loop, ensuring that we can
continue to make progress in the face of contentious PUT requests to the
same resource.


prb112 (approved):

fhir-server-test/src/test/java/com/ibm/fhir/server/test/ConcurrentUpdateTest.java
```diff
@@ -0,0 +1,126 @@
...
+import com.ibm.fhir.model.resource.Patient;
+import com.ibm.fhir.model.test.TestUtil;
+
+/**
+ * Tests specifically the concurrent creation and updating of a single patient, with the updateCreate feature enabled.
+ * The idea is to ensure that no patients with duplicate logical ids are created.
+ *
+ * NOTE: This test is NOT included in the fhir-server testng suite because of an internal problem with Derby that
+ * surfaces after the first execution of this test against a clean Derby DB. Subsequent test runs of this class will
+ * fail with an XAException that has an unknown error code. It should only be run against a DB2 database.
```
prb112, line 37:
```suggestion
 *
 * NOTE: This test is NOT included in the fhir-server testng suite because of an internal problem with Derby that
 * surfaces after the first execution of this test against a clean Derby DB. Subsequent test runs of this class will
 * fail with an XAException that has an unknown error code. It should only be run against a DB2 database.
```

Do we need this comment?

lmsurpre: yes, good call

lmsurpre: I updated the javadoc accordingly

PR #2426, opened 2021-05-25
issue 2283 - eliminate redundant reads during ingestion

punktilious:

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
@@ -1496,7 +1496,8 @@ public class FHIRPersistenceJDBCImpl implements FHIRPersistence, SchemaNameSuppl
             ResourceDAO resourceDao = makeResourceDAO(connection);
 
             resourceDTO = resourceDao.read(logicalId, resourceType.getSimpleName());
-            if (resourceDTO != null && resourceDTO.isDeleted() && !context.includeDeleted()) {
+            boolean resourceIsDeleted = resourceDTO != null && resourceDTO.isDeleted();
+            if (resourceIsDeleted && !context.includeDeleted()) {
                 throw new FHIRPersistenceResourceDeletedException("Resource '" +
                         resourceType.getSimpleName() + "/" + logicalId + "' is deleted.");
             }
@@ -1505,19 +1506,17 @@ public class FHIRPersistenceJDBCImpl implements FHIRPersistence, SchemaNameSuppl
             SingleResourceResult<T> result = new SingleResourceResult.Builder<T>()
                     .success(true)
                     .resource(resource)
+                    .deleted(resourceIsDeleted)
                     .build();
 
             return result;
-        }
-        catch(FHIRPersistenceResourceDeletedException e) {
+        } catch(FHIRPersistenceResourceDeletedException e) {
             throw e;
-        }
-        catch(Throwable e) {
+        } catch(Throwable e) {
             FHIRPersistenceException fx = new FHIRPersistenceException("Unexpected error while performing a read operation.");
             log.log(Level.SEVERE, fx.getMessage(), e);
             throw fx;
-        }
-        finally {
+        } finally {
             log.exiting(CLASSNAME, METHODNAME);
         }
     }
--- a/fhir-server/src/main/java/com/ibm/fhir/server/operation/spi/FHIRResourceHelpers.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/operation/spi/FHIRResourceHelpers.java
@@ -17,6 +17,7 @@ import com.ibm.fhir.model.resource.OperationOutcome;
 import com.ibm.fhir.model.resource.Resource;
 import com.ibm.fhir.persistence.FHIRPersistenceTransaction;
 import com.ibm.fhir.persistence.ResourceEraseRecord;
+import com.ibm.fhir.persistence.SingleResourceResult;
 import com.ibm.fhir.persistence.erase.EraseDTO;
 
 /**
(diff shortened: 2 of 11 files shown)


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -1505,19 +1506,17 @@ private FHIRPersistenceNotSupportedException buildNotSupportedException(String m
             SingleResourceResult<T> result = new SingleResourceResult.Builder<T>()
                     .success(true)
                     .resource(resource)
+                    .deleted(resourceIsDeleted)
                     .build();
 
             return result;
-        }
-        catch(FHIRPersistenceResourceDeletedException e) {
+        } catch(FHIRPersistenceResourceDeletedException e) {
```
lmsurpre, line 1513: I wonder if it would be possible to remove this exception now that we have the idDeleted marker in the response object...

punktilious: We still need to throw it: 
```
            if (resourceIsDeleted && !context.includeDeleted()) {
                throw new FHIRPersistenceResourceDeletedException("Resource '" +
                        resourceType.getSimpleName() + "/" + logicalId + "' is deleted.");
            }
```

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -753,13 +757,13 @@ public FHIRRestOperationResponse doDelete(String type, String id, String searchQ
...
     @Override
-    public Resource doRead(String type, String id, boolean throwExcOnNull, boolean includeDeleted,
+    public SingleResourceResult<? extends Resource> doRead(String type, String id, boolean throwExcOnNull, boolean includeDeleted,
             Resource contextResource) throws Exception {
         return doRead(type, id, throwExcOnNull, includeDeleted, contextResource, null);
     }
 
     @Override
-    public Resource doRead(String type, String id, boolean throwExcOnNull, boolean includeDeleted,
+    public SingleResourceResult<? extends Resource> doRead(String type, String id, boolean throwExcOnNull, boolean includeDeleted,
```
lmsurpre, line 766: Do we still need to pass `includeDeleted` now that we have the `isDeleted` marker in the response?

fhir-server/src/main/java/com/ibm/fhir/server/operation/spi/FHIRResourceHelpers.java
```diff
@@ -162,7 +163,7 @@ FHIRRestOperationResponse doPatch(String type, String id, FHIRPatch patch, Strin
      * @return the Resource
```
lmsurpre, line 163: update this to explain the wrapper a bit?

tbieste (approved):

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -392,6 +393,9 @@ private FHIRRestOperationResponse doPatchOrUpdate(String type, String id, FHIRPa
                     } else {
                         id = newResource.getId();
                     }
+
+                    // No match, so deletion status doesn't matter
+                    isDeleted = false;
```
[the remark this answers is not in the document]
lmsurpre: yeah, i actually prefer explicitly initializing it (even though it is redundant)...just one less thing to remember :-)

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
lmsurpre: OK.  I'm definitely fine with this...I've just been looking out for chances to remove exception-handling from the normal request processing path where possible.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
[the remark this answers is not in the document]
punktilious: Actually it's not initialized on purpose:
```
        boolean isDeleted; // stash the deleted status of the resource when we first read it
```

I do this intentionally to make sure I set it in each of the if/then/else paths. If I miss a path, the compiler will catch it.

fhir-server/src/main/java/com/ibm/fhir/server/operation/spi/FHIRResourceHelpers.java
```diff
@@ -185,7 +186,7 @@ default Resource doRead(String type, String id, boolean throwExcOnNull, boolean
...
```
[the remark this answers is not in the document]
punktilious: done

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
punktilious: It would be possible to refactor the entire handling of deleted status, but this is a much bigger task. Currently the persistence layer will throw an exception to signal the resource has been deleted. Handling this without using an exception is non-trivial.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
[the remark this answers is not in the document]
lmsurpre: I think that is true for Boolean fields, but is it true for booleans?

lmsurpre (approved):
LGTM

prb112 (approved):

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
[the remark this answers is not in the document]
punktilious: Default is false if it's a member of the class. But for local variables in the method, default is unassigned (Eclipse will complain too if you try to use it without assigning a value).

PR #2430, opened 2021-05-26
Update the Users Guide to correct issue with auth and update README.md with the Erase operation

prb112:


lmsurpre (approved):
LGTM

Issue #896, opened 2020-04-06
Support for MIME-type parameter fhirVersion


lmsurpre:
**Is your feature request related to a problem? Please describe.**
Conformance.md reads:
> The IBM FHIR Server supports only version 4.0.1 of the specification and presently has no support for the MIME-type parameter fhirVersion.

Currently, we just drop the parameter from all requests.  However, a better behavior would be to reject requests that include the fhirVersion parameter with the wrong fhirVersion.

**Describe the solution you'd like**
Since the server only supports version 4.0.1, it should reject requests for other fhirVersion values other than `4.0.1` and `4.0`.

> 406 Not Acceptable is the appropriate response when the Accept header requests a format that the server does not support, and 415 Unsupported Media Type when the client posts a format that is not supported to the server.

**Describe alternatives you've considered**

**Additional context**
A description of the parameter and its meaning:  https://www.hl7.org/fhir/versioning.html#mt-version

I would be great if we can get JAX-RS to handle this for us, but given that we'd also like to support `application/fhir+json` and `application/fhir+json` with no parameter, it might not be possible.

We should use this flavor of the MediaType constructor:
https://javaee.github.io/javaee-spec/javadocs/javax/ws/rs/core/MediaType.html#MediaType-java.lang.String-java.lang.String-java.util.Map-

**QA Notes**
Call an IBM FHIR Server API, passing the following HTTP header values:
1) Accept: application/fhir+json; fhirVersion=4.0  [Valid]
2) Content-Type: application/fhir+json; fhirVersion=4.0  [Valid]
3) Accept: application/fhir+json; fhirVersion=3.0  [406 Not Acceptable]
4) Content-Type: application/fhir+json; fhirVersion=3.0  [415 Unsupported Media Type]

lmsurpre:
Confirmed that using PUT or POST with a fhirVersion other than 4.0 or 4.0.1 results in a 415 Unsuported Media Type with an error like:
```
{
    "resourceType": "OperationOutcome",
    "issue": [
        {
            "severity": "fatal",
            "code": "invalid",
            "details": {
                "text": "FHIRRestServletRequestException: Invalid 'fhirVersion' parameter value in 'Content-Type' header; the following FHIR versions are supported: [4.0, 4.0.1]"
            }
        }
    ]
}
```

Also confirmed that a GET request with an Accept header with fhirVersion other than 4.0 or 4.0.1 results in a 406 Not Acceptable with a response like:
```
{
[...]

PR #2434, opened 2021-05-26
Issue #896 - Add checking for MIME-type parameter fhirVersion

tbieste:
Issue #896 - Add checking for MIME-type parameter fhirVersion

### DIFF
--- a/fhir-core/src/main/java/com/ibm/fhir/core/FHIRMediaType.java
+++ b/fhir-core/src/main/java/com/ibm/fhir/core/FHIRMediaType.java
@@ -1,10 +1,15 @@
 /*
- * (C) Copyright IBM Corp. 2019
+ * (C) Copyright IBM Corp. 2019, 2021
  *
  * SPDX-License-Identifier: Apache-2.0
  */
 
 package com.ibm.fhir.core;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
 
 import javax.ws.rs.core.MediaType;
 
@@ -12,6 +17,7 @@ import javax.ws.rs.core.MediaType;
  * This class contains definitions of some non-standard media types.
  */
 public class FHIRMediaType extends MediaType {
+
     public final static String SUBTYPE_FHIR_JSON = "fhir+json";
     public final static String APPLICATION_FHIR_JSON = "application/" + SUBTYPE_FHIR_JSON;
     public final static MediaType APPLICATION_FHIR_JSON_TYPE = new MediaType("application", SUBTYPE_FHIR_JSON);
@@ -29,6 +35,13 @@ public class FHIRMediaType extends MediaType {
     public final static MediaType APPLICATION_FHIR_NDJSON_TYPE = new MediaType("application", SUBTYPE_FHIR_NDJSON);
 
     public final static String SUBTYPE_FHIR_PARQUET = "fhir+parquet";
-    public static final String APPLICATION_PARQUET = "application/"  + SUBTYPE_FHIR_PARQUET;
+    public static final String APPLICATION_PARQUET = "application/" + SUBTYPE_FHIR_PARQUET;
     public final static MediaType APPLICATION_FHIR_PARQUET_TYPE = new MediaType("application", SUBTYPE_FHIR_PARQUET);
+
+    // Supported values for the MIME-type parameter fhirVersion.
+    // https://www.hl7.org/fhir/http.html#version-parameter
+    // The value of this parameter is the publication and major version number for the specification.
+    public static final String FHIR_VERSION_PARAMETER = "fhirVersion";
+    public static final Set<String> SUPPORTED_FHIR_VERSIONS =
+            Collections.unmodifiableSet(new HashSet<>(Arrays.asList("4.0")));
 }
--- a/fhir-core/src/main/java/com/ibm/fhir/core/FHIRMediaType.java
+++ b/fhir-core/src/main/java/com/ibm/fhir/core/FHIRMediaType.java
@@ -40,8 +40,7 @@ public class FHIRMediaType extends MediaType {
 
     // Supported values for the MIME-type parameter fhirVersion.
     // https://www.hl7.org/fhir/http.html#version-parameter
-    // The value of this parameter is the publication and major version number for the specification.
     public static final String FHIR_VERSION_PARAMETER = "fhirVersion";
     public static final Set<String> SUPPORTED_FHIR_VERSIONS =
-            Collections.unmodifiableSet(new HashSet<>(Arrays.asList("4.0")));
+            Collections.unmodifiableSet(new HashSet<>(Arrays.asList("4.0","4.0.1")));
 }
--- a/fhir-server-test/src/test/java/com/ibm/fhir/server/test/BasicServerTest.java
+++ b/fhir-server-test/src/test/java/com/ibm/fhir/server/test/BasicServerTest.java
@@ -15,12 +15,13 @@ import static org.testng.AssertJUnit.fail;
 import java.io.StringWriter;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 
-import jakarta.json.JsonObject;
 import javax.ws.rs.client.Entity;
 import javax.ws.rs.client.WebTarget;
+import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 
 import org.testng.annotations.Test;
@@ -48,6 +49,8 @@ import com.ibm.fhir.path.FHIRPathNode;
 import com.ibm.fhir.path.evaluator.FHIRPathEvaluator;
 import com.ibm.fhir.path.evaluator.FHIRPathEvaluator.EvaluationContext;
 import com.ibm.fhir.path.exception.FHIRPathException;
+
+import jakarta.json.JsonObject;
 
 /**
  * Basic sniff test of the FHIR Server.
--- a/fhir-server-test/src/test/java/com/ibm/fhir/server/test/BasicServerTest.java
+++ b/fhir-server-test/src/test/java/com/ibm/fhir/server/test/BasicServerTest.java
@@ -126,6 +126,38 @@ public class BasicServerTest extends FHIRServerTestBase {
     }
 
     /**
+     * Verify the 'metadata' API with valid fhirVersion in Accept header.
+     */
+    @Test(groups = { "server-basic" })
+    public void testMetadataAPI_validFhirVersion() {
+        WebTarget target = getWebTarget();
+        MediaType mediaType = new MediaType("application", "fhir+json",
+            Collections.singletonMap(FHIRMediaType.FHIR_VERSION_PARAMETER, "4.0"));
+        Response response = target.path("metadata").request(mediaType).get();
+        assertResponse(response, Response.Status.OK.getStatusCode());
+        assertEquals(mediaType, response.getMediaType());
+
+        CapabilityStatement conf = response.readEntity(CapabilityStatement.class);
+        assertNotNull(conf);
+        assertNotNull(conf.getFormat());
+        assertEquals(6, conf.getFormat().size());
+        assertNotNull(conf.getVersion());
+        assertNotNull(conf.getName());
+    }
+
+    /**
+     * Verify the 'metadata' API with invalid fhirVersion in Accept header.
+     */
+    @Test(groups = { "server-basic" })
+    public void testMetadataAPI_invalidFhirVersion() {
+        WebTarget target = getWebTarget();
+        MediaType mediaType = new MediaType("application", "fhir+json",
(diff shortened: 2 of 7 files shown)


JohnTimm:
Would we want to use 4.0.0 and 4.0.1 as supported versions or is 4.0 sufficient?

fhir-core/src/main/java/com/ibm/fhir/core/FHIRMediaType.java
```diff
@@ -29,6 +35,13 @@
...
-    public static final String APPLICATION_PARQUET = "application/"  + SUBTYPE_FHIR_PARQUET;
+    public static final String APPLICATION_PARQUET = "application/" + SUBTYPE_FHIR_PARQUET;
     public final static MediaType APPLICATION_FHIR_PARQUET_TYPE = new MediaType("application", SUBTYPE_FHIR_PARQUET);
+
+    // Supported values for the MIME-type parameter fhirVersion.
+    // https://www.hl7.org/fhir/http.html#version-parameter
+    // The value of this parameter is the publication and major version number for the specification.
+    public static final String FHIR_VERSION_PARAMETER = "fhirVersion";
+    public static final Set<String> SUPPORTED_FHIR_VERSIONS =
+            Collections.unmodifiableSet(new HashSet<>(Arrays.asList("4.0")));
```
prb112, line 46: I think 4.0.1 is valid too? Right?

fhir-server-test/src/test/java/com/ibm/fhir/server/test/BasicServerTest.java
```diff
@@ -151,13 +229,60 @@ public void testCreatePatientMinimalWithFormat() throws Exception {
...
+        Entity<Patient> entity = Entity.entity(patient, FHIRMediaType.APPLICATION_FHIR_JSON);
+        Response response = target.path("Patient").request().post(entity, Response.class);
+        assertResponse(response, Response.Status.CREATED.getStatusCode());
+
+        // Get the patient's logical id value.
+        String patientId = getLocationLogicalId(response);
+
+        // Next, call the 'read' API to attempt to retrieve the new patient
+        response = target.path("Patient/" + patientId).queryParam("_format", "application/fhir+json;fhirVersion=3.0").request(FHIRMediaType.APPLICATION_FHIR_JSON).get();
+        assertResponse(response, Response.Status.NOT_ACCEPTABLE.getStatusCode());
```
prb112, line 283: makes me think  we should have a lenient mode.

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
```diff
@@ -261,6 +279,30 @@ private HTTPReturnPreference computeReturnPref(ServletRequest request, HTTPHandl
...
     }
 
+    /**
+     * Checks if the FHIR version parameter in the specified HTTP header is valid.
+     * @param headerName the name of the header
+     * @param requestHeaders the headers
+     * @return the error message if FHIR version parameter in not valid, otherwise null
+     */
+    private String checkFhirVersionParameter(String headerName, Map<String, List<String>> requestHeaders) throws FHIRException {
+        for (String headerValue : requestHeaders.getOrDefault(headerName, Collections.emptyList())) {
```
prb112, line 289: I think we should ignore multiple headers, and only check  the first, reject multiples.

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
```diff
@@ -261,6 +279,30 @@ private HTTPReturnPreference computeReturnPref(ServletRequest request, HTTPHandl
         return returnPref;
     }
 
+    /**
+     * Checks if the FHIR version parameter in the specified HTTP header is valid.
+     * @param headerName the name of the header
+     * @param requestHeaders the headers
+     * @return the error message if FHIR version parameter in not valid, otherwise null
+     */
+    private String checkFhirVersionParameter(String headerName, Map<String, List<String>> requestHeaders) throws FHIRException {
```
prb112, line 288: You could convert this to a void method, that  only throws exceptions if invalid, also control the message in this code, and then make it testable/self-contained. the outer catch should catch  and  mark as invalid. you could also have a special  exception for this.

fhir-core/src/main/java/com/ibm/fhir/core/FHIRMediaType.java
tbieste: > Would we want to use 4.0.0 and 4.0.1 as supported versions or is 4.0 sufficient?

Yeah, I wasn't sure about that, based on https://www.hl7.org/fhir/http.html#version-parameter, it says the value of this parameter is the publication and major version number for the specification. So it sounds like it is not supposed to include the minor version.  However, I'm ok with adding 4.0.1 to the list if that makes more sense.

tbieste: Added 4.0.1.

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
tbieste: Updated to use special exception for cleaner code.

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
tbieste: Updated to reject multiple different versions for Content-Type, but still checks fhirVersion on each header.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/BasicServerTest.java
tbieste: Since invalid Content-Type and Accept headers result in HTTP 406/415 as of today, I'm not going to add a lenient mode for this. The `fhirVersion` parameter value is optional.

tbieste:
> Would we want to use 4.0.0 and 4.0.1 as supported versions or is 4.0 sufficient?

Yeah, I wasn't sure about that, based on https://www.hl7.org/fhir/http.html#version-parameter, it says the value of this parameter is the publication and major version number for the specification. So it sounds like it is not supposed to include the minor version. However, I'll add 4.0.1 to the list since it is consistent with the version shown in the Capabilities statement returned on the `metadata` API.

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
```diff
@@ -261,6 +274,50 @@ private HTTPReturnPreference computeReturnPref(ServletRequest request, HTTPHandl
...
+                    for (String headerValueElement : headerValue.split(",")) {
+                        Map<String, String> parameters = MediaType.valueOf(headerValueElement).getParameters();
+                        if (parameters != null) {
+                            for (Map.Entry<String, String> parameter : parameters.entrySet()) {
+                                if (FHIRMediaType.FHIR_VERSION_PARAMETER.equalsIgnoreCase(parameter.getKey())) {
+                                    String curFhirVersion = parameter.getValue();
+                                    if (curFhirVersion != null && !FHIRMediaType.SUPPORTED_FHIR_VERSIONS.contains(curFhirVersion)) {
+                                        throw new FHIRRestServletRequestException("Invalid '" + FHIRMediaType.FHIR_VERSION_PARAMETER
+                                            + "' parameter value in '" + headerName + "' header; the following FHIR versions are supported: "
+                                                + FHIRMediaType.SUPPORTED_FHIR_VERSIONS, headerStatusMap.get(headerName));
```
lmsurpre, line 302: thats a lot of nesting!

lmsurpre: do we need the null checks on every level?  if yes, then maybe there's not much we can do about it.

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
```diff
@@ -261,6 +274,50 @@ private HTTPReturnPreference computeReturnPref(ServletRequest request, HTTPHandl
...
+        Map<String, Integer> headerStatusMap = new LinkedHashMap<>();
+        // 415 Unsupported Media Type is the appropriate response when the client posts a format that is not supported to the server.
+        headerStatusMap.put(HttpHeaders.CONTENT_TYPE, HttpServletResponse.SC_UNSUPPORTED_MEDIA_TYPE);
+        // 406 Not Acceptable is the appropriate response when the Accept header requests a format that the server does not support.
+        headerStatusMap.put(HttpHeaders.ACCEPT, HttpServletResponse.SC_NOT_ACCEPTABLE);
+
+        for (String headerName : headerStatusMap.keySet()) {
+            String fhirVersion = null;
+            for (String headerValue : requestHeaders.getOrDefault(headerName, Collections.emptyList())) {
+                if (headerValue != null) {
```
lmsurpre, line 292: I'd hope that the `getOrDefault` contract would prevent the need for this null check

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
```diff
@@ -261,6 +274,50 @@ private HTTPReturnPreference computeReturnPref(ServletRequest request, HTTPHandl
...
+                    for (String headerValueElement : headerValue.split(",")) {
+                        Map<String, String> parameters = MediaType.valueOf(headerValueElement).getParameters();
+                        if (parameters != null) {
```
lmsurpre, line 295: can parameters actually be null.  from https://docs.oracle.com/javaee/7/api/javax/ws/rs/core/MediaType.html#getParameters-- I think not, but probably needs a quick test to be sure

lmsurpre: at least for the version in my workspace, I still think not:
```
    private MediaType(String type, String subtype, String charset, Map<String, String> parameterMap) {

        this.type = type == null ? MEDIA_TYPE_WILDCARD : type;
        this.subtype = subtype == null ? MEDIA_TYPE_WILDCARD : subtype;

        if (parameterMap == null) {
            parameterMap = new TreeMap<String, String>(new Comparator<String>() {

                @Override
                public int compare(String o1, String o2) {
                    return o1.compareToIgnoreCase(o2);
                }
            });
        }

        if (charset != null && !charset.isEmpty()) {
            parameterMap.put(CHARSET_PARAMETER, charset);
        }
        this.parameters = Collections.unmodifiableMap(parameterMap);
    }
```

fhir-core/src/main/java/com/ibm/fhir/core/FHIRMediaType.java
```diff
@@ -40,8 +40,7 @@
 
     // Supported values for the MIME-type parameter fhirVersion.
     // https://www.hl7.org/fhir/http.html#version-parameter
-    // The value of this parameter is the publication and major version number for the specification.
     public static final String FHIR_VERSION_PARAMETER = "fhirVersion";
     public static final Set<String> SUPPORTED_FHIR_VERSIONS =
-            Collections.unmodifiableSet(new HashSet<>(Arrays.asList("4.0")));
+            Collections.unmodifiableSet(new HashSet<>(Arrays.asList("4.0","4.0.1")));
```
prb112, line 45: Should it have 4 listed?

tbieste: I haven't found any example of just "4" being a version number, and since the spec, https://www.hl7.org/fhir/http.html#version-parameter, says the value of this parameter is the publication and major version number for the specification, I thought it should just be "4.0", but I since other API operations return "4.0.1" as well, I thought including that would be good to include as well.

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
tbieste: Ah, right. I didn't verify that all constructors indeed set the parameters member variable. Looks like they all do.

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
tbieste: I was able to remove a couple levels of nesting.

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
tbieste: I was able to remove a couple levels of nesting.

lmsurpre (approved):
LGTM

Issue #2143, opened 2021-03-24
Support customizable compartment definitions


lmsurpre:
**Is your feature request related to a problem? Please describe.**
Compartment search provides a way to efficient scope a search request to resources that have a single parameter value of some kind (e.g. a specific patient id).

It might be nice for users (i.e. operators of a given fhir server deployment) to be able to both:
1. customize existing compartments (like the patient compartment) with their own "inclusion criteria" for what to include/exclude; and
2. define new compartment definitions that can be used to scope search requests by a common criteria

**Describe the solution you'd like**
1. Update CompartmentUtil to pull compartment definitions from the registry (instead of a packaged compartments.json).
2. Document how users/operators can override the default compartment definitions.
3. Document how users/operators can supply new compartment definitions.

**Describe alternatives you've considered**
Do not support customizable compartment definitions.

**Acceptance Criteria**

**Additional context**
This feature becomes more powerful with #2371

lmsurpre:
I found that the CompartmentDefinition.code has a required binding of `Patient | Encounter | RelatedPerson | Practitioner | Device`.
Therefor, to enable custom compartment definitions, I introduced a special extension with a url of `http://ibm.com/fhir/extension/custom-compartment-type`.

Using this extension, I registered a new compartment definition for "DrugPlan" as follows:
```
[{
    "resourceType": "CompartmentDefinition",
    "url": "http://ibm.com/health-interop/fhir/CompartmentDefinition/drugPlanCompartment",
    "name": "DrugPlan",
    "status": "draft",
    "_code": {
        "extension": [
            {
                "url": "http://terminology.hl7.org/CodeSystem/data-absent-reason",
                "valueCode": "not-permitted"
            },
            {
                "url": "http://ibm.com/fhir/extension/custom-compartment-type",
                "valueCode": "DrugPlan"
            }
        ]
    },
    "search": false,
    "resource": [
        {
            "code": "List",
            "param": [
                "planid"
            ]
        },
        {
            "code": "MedicationKnowledge",
            "param": [
                "DrugPlan"
            ]
        }
    ]
}]
```

With this CompartmentDefinition and the corresponding search parameter available in the registry, I can now perform compartment searches like `GET [base]/DrugPlan/123/List` or `GET [base]/DrugPlan/123/MedicationKnowledge` and the results appear as expected.
Additionally, it becomes possible to scope a request to more than one drugplan via persistence interceptor.

lmsurpre:
[...]

PR #2440, opened 2021-05-27
issue #2143 - support custom compartment types via extension

lmsurpre:
The CompartmentDefinition resource has a required binding to a fixed
list of compartment types. To support a custom compartments, we will now
look for the `http://ibm.com/fhir/extension/custom-compartment-type`
extension on this element (when it has no value).

Additionally, I created a constant in fhir-core FHIRConstants for our
base extension url and I updated all the different places we have
extension urls to use it.

### DIFF
--- a/fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
+++ b/fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
@@ -6,6 +6,8 @@
 
 package com.ibm.fhir.search.util;
 
+import static com.ibm.fhir.model.util.ModelSupport.FHIR_STRING;
+
 import java.io.FileNotFoundException;
 import java.math.BigDecimal;
 import java.net.URISyntaxException;
@@ -43,6 +45,7 @@ import com.ibm.fhir.model.resource.SearchParameter.Component;
 import com.ibm.fhir.model.resource.ValueSet;
 import com.ibm.fhir.model.type.Canonical;
 import com.ibm.fhir.model.type.Code;
+import com.ibm.fhir.model.type.Element;
 import com.ibm.fhir.model.type.Reference;
 import com.ibm.fhir.model.type.Uri;
 import com.ibm.fhir.model.type.code.IssueSeverity;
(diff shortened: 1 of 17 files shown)


fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -2587,24 +2590,46 @@ private static void manageException(String message, IssueType issueType, FHIRSea
...
+                                        continue;
+                                    }
+                                }
+                            } else if (element.is(FHIR_STRING)) {
+                                if (paramEntry.getValue().size() != 1) {
+                                    log.warning("CompartmentDefinition inclusion criteria must be of type Reference unless they have 1 and only 1 resource target");
+                                    continue;
                                 }
+                                compartmentName = paramEntry.getValue().iterator().next();
+                                compartmentId = element.as(FHIR_STRING).getValue();
```
lmsurpre, line 2620: previously, we supported only inclusion criteria that selected elements of type Reference.  now we will support elements of type String as well, but only in cases where the compartment type is unambiguous (so that we do not need to check the target resourceType from the reference to determine which compartment the value is for).

fhir-registry/src/main/java/com/ibm/fhir/registry/FHIRRegistry.java
```diff
@@ -177,13 +178,10 @@ public String getDefaultVersion(String url, Class<? extends Resource> resourceTy
...
-        for (FHIRRegistryResourceProvider provider : providers) {
-            for (FHIRRegistryResource registryResource : provider.getRegistryResources(resourceType)) {
-                resources.add(resourceType.cast(registryResource.getResource()));
-            }
-        }
-        return Collections.unmodifiableList(resources);
+        return providers.parallelStream()
+                .flatMap(p -> p.getRegistryResources(resourceType).stream())
+                .map(rr -> resourceType.cast(rr.getResource()))
+                .collect(Collectors.collectingAndThen(Collectors.toList(), Collections::unmodifiableList));
```
prb112, line 184: @JohnTimm didn't you find this was less efficient? is this OK?  (semantically correct)

prb112 (approved):
LGTM - left one comment (just clarifying code history/decisions with John)

fhir-registry/src/main/java/com/ibm/fhir/registry/resource/FHIRRegistryResource.java
```diff
@@ -111,8 +111,8 @@ public boolean isDefaultVersion() {
         return defaultVersion;
     }
 
-    public Resource getResource() {
-        return resource;
+    public <T extends Resource> T getResource() {
+        return (T) resource;
```
lmsurpre, line 115: didn't actually mean to include this one... @JohnTimm still think its better to force the users into an explicit cast?

fhir-registry/src/main/java/com/ibm/fhir/registry/FHIRRegistry.java
lmsurpre: i think John found that streams were a tad slower than loops in general.  however, in this particular case, the parallelStream should let the registry as all the RegistryResourceProviders for all their resources in parallel.  I think it should be faster when multiple providers are in play (especially if each one needs to read its own index file like in the case of a server start).

JohnTimm: I'm not opposed to this change but would like to use benchmarking to determine if it helps. When I switched out the streams to standard loops it was driven by some benchmarking results and some hot spots I found with VisualVM

fhir-registry/src/main/java/com/ibm/fhir/registry/resource/FHIRRegistryResource.java
JohnTimm: I think this change seems somewhat unnecessary. `FHIRRegistry.getResource()` already does casting and we also have the Resource.as(Class<? extends Resource>) method.

lmsurpre: yeah, here is the difference for users of this method.

old:
```
SearchParameter sp = registryResource.getResource().as(SearchParameter.class);
```

new:
```
SearchParameter sp = registryResource.getResource();
```

Not a huge win, but seems like a slight improvement to me.

lmsurpre: I backed it out since it doesn't really belong with this PR anyway

fhir-registry/src/main/java/com/ibm/fhir/registry/FHIRRegistry.java
lmsurpre: ok, I backed it out since it doesn't really belong with this PR anyway

fhir-registry/src/main/java/com/ibm/fhir/registry/resource/FHIRRegistryResource.java
JohnTimm: > yeah, here is the difference for users of this method.
> 
> old:
> 
> ```
> SearchParameter sp = registryResource.getResource().as(SearchParameter.class);
> ```
> 
> new:
> 
> ```
> SearchParameter sp = registryResource.getResource();
> ```
> 
> Not a huge win, but seems like a slight improvement to me.

In most cases I think clients will use `FHIRRegistry.getInstance().getResource(url, SearchParameter.class)` and won't be working with `FHIRRegistryResource` instance directly.

JohnTimm (approved):
LGTM

PR #2475, opened 2021-06-05
attempt to free disk space before our docker build

lmsurpre:
using the same approach that we use in our integration tests


punktilious (approved):
LGTM

Issue #2473, opened 2021-06-04
Audit record uses the exact value of: X-FHIR-FORWARDED-URL header as the URL called for the log entry


d0roppe:
**Describe the bug**
when using header X-FHIR-FORWARDED-URL, it is used un altered as the api-parameters/request and address/url in the audit entry.
This was for a Update patient:
```
    "target": {
        "id": "sniff-test--BDE-check-P1",
        "typeURI": "data/database",
        "addresses": [
            {
                "url": "https://dag.minions-r.us/fhir-server/api/v4/Group/sniff-test--BDE-check-R1",
                "name": "",
                "port": ""
            }
        ],
```
Same in the attachment:
```
    "api_parameters": {
        "request": "https://dag.minions-r.us/fhir-server/api/v4/Group/sniff-test--BDE-check-R1",
        "request_status": 200
    },
```

**Environment**
Which version of IBM FHIR Server?

**To Reproduce**

**Expected behavior**
Instead of using the originalRequestUriHeader, we should use the exact request that came in (from the HttpServletRequest).

**Additional context**

prb112:
Also BulkDataAuditLogger
```
        // Uses the FHIRRestServletFilter to pass the OriginalRequestUri to the backend.
        entry.getContext().setApiParameters(ApiParameters.builder()
                                                .request(FHIRRequestContext.get().getOriginalRequestUri())
                                                .status(responseStatus.getStatusCode())
                                                .build());
```

prb112:
EraseOperationAuditLogger
        // Uses the FHIRRestServletFilter to pass the OriginalRequestUri to the backend.
        requestUrl = new StringBuilder(FHIRRequestContext.get().getOriginalRequestUri());

prb112:
Only one location to change, others are already accounted for

d0roppe:
This is much better now. here is the update patient info with action of update:
```
    "target": {
        "id": "sniff-test--BDE-check-P1",
        "typeURI": "data/database",
        "addresses": [
            {
                "url": "/fhir-server/api/v4/Patient/sniff-test--BDE-check-P1",
                "name": "",
                "port": ""
            }
        ],
```

and for the entry:

```
    "action": "U",
    "start_time": "2021-06-22 00:12:12.694",
    "end_time": "2021-06-22 00:12:13.017",
    "api_parameters": {
        "request": "/fhir-server/api/v4/Patient/sniff-test--BDE-check-P1",
        "request_status": 200
    },
```

PR #2481, opened 2021-06-07
Audit record uses the exact value of: X-FHIR-FORWARDED-URL header as the URL called for the log entry #2473

prb112:

### DIFF
--- a/docs/src/pages/guides/FHIRServerUsersGuide.md
+++ b/docs/src/pages/guides/FHIRServerUsersGuide.md
@@ -1429,6 +1429,12 @@ The following table describes the JSON fields of the CADF audit log entries logg
 |`observer/geolocation/city`                           |Value is determined by "fhirServer/audit/serviceProperties/geoCity" configuration property.|
 |`observer/geolocation/state`                          |Value is determined by "fhirServer/audit/serviceProperties/geoState" configuration property.|
 |`observer/geolocation/region`                         |Value is determined by "fhirServer/audit/serviceProperties/geoCounty" configuration property.|
+|`target/id`                                           |Value is the UUID from the resource|
+|`target/typeUri`                                      |Value is always "compute/node".|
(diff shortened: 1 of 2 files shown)


lmsurpre:
I was going to suggest that we document that we'll use the request from the servlet context (not the original request url), but it looks like we never actually added the `target` to our CADF documentation.  Maybe it would be good to fix that omission as part of this PR?

prb112:
> I was going to suggest that we document that we'll use the request from the servlet context (and not the original request url), but it looks like we never actually added the `target` to our CADF documentation at https://ibm.github.io/FHIR/guides/FHIRServerUsersGuide#4111-cadf-audit-log-entry. Maybe it would be good to fix that omission as part of this PR?

Sure, once I have a clean workspace, I'll update.

prb112:
All updated

docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -1429,6 +1429,12 @@ The following table describes the JSON fields of the CADF audit log entries logg
 |`observer/geolocation/city`                           |Value is determined by "fhirServer/audit/serviceProperties/geoCity" configuration property.|
 |`observer/geolocation/state`                          |Value is determined by "fhirServer/audit/serviceProperties/geoState" configuration property.|
 |`observer/geolocation/region`                         |Value is determined by "fhirServer/audit/serviceProperties/geoCounty" configuration property.|
+|`target/id`                                           |Value is the UUID from the resource|
```
lmsurpre, line 1432: is this an internal UUID or is this the logical id?  if its the logical id, then its not necessarily a UUID.

prb112: Fixed

lmsurpre (approved):
LGTM

PR #2486, opened 2021-06-08
modify Provenance-patient parameter during index generation

lmsurpre:
Per https://jira.hl7.org/browse/FHIR-32876 this search parameter is
wrong in the latest draft of the spec.  I thought it was wrong in our
packaged version (4.0.1) as well, but it turns out I was mistaken.

Still, I thought it would be good to keep this logic in place for when
we move to the R5 artifacts.


JohnTimm (approved):
LGTM

Issue #2354, opened 2021-05-12
When multiple tenants exist, Drop Tenant Results in an incorrect FHIRDATA.PARAMETER_NAMES association and drop. 


prb112:
**Describe the bug**
When multiple tenants exist, Drop Tenant Results in an incorrect FHIRDATA.PARAMETER_NAMES association and drop. 

```
2021-05-12 12:13:59.397 00000001    INFO   com.ibm.fhir.schema.app.Main Dropping tablespace for tenant 2/test1d
2021-05-12 12:14:08.973 00000001 WARNING ls.pool.PoolConnectionProvider Rolling back transaction
2021-05-12 12:14:08.977 00000001  SEVERE   com.ibm.fhir.schema.app.Main schema tool failed

com.ibm.db2.jcc.am.SqlException: DB2 SQL Error: SQLCODE=-282, SQLSTATE=55024, SQLERRMC=TS_TENANT2;FHIRDATA.PARAMETER_NAMES, DRIVER=4.28.11
com.ibm.fhir.database.utils.api.DataAccessException: com.ibm.db2.jcc.am.SqlException: DB2 SQL Error: SQLCODE=-282, SQLSTATE=55024, SQLERRMC=TS_TENANT2;FHIRDATA.PARAMETER_NAMES, DRIVER=4.28.11
        at com.ibm.fhir.database.utils.db2.Db2Translator.translate(Db2Translator.java:115)
        at com.ibm.fhir.database.utils.db2.Db2DropTablespace.run(Db2DropTablespace.java:46)
        at com.ibm.fhir.database.utils.common.CommonDatabaseAdapter.runStatement(CommonDatabaseAdapter.java:459)
        at com.ibm.fhir.database.utils.db2.Db2Adapter.runStatement(Db2Adapter.java:546)
        at com.ibm.fhir.database.utils.db2.Db2Adapter.dropTablespace(Db2Adapter.java:430)
        at com.ibm.fhir.database.utils.db2.Db2Adapter.dropTenantTablespace(Db2Adapter.java:580)
        at com.ibm.fhir.database.utils.model.PhysicalDataModel.dropTenantTablespace(PhysicalDataModel.java:519)
        at com.ibm.fhir.schema.app.Main.dropDetachedPartitionTables(Main.java:1152)
        at com.ibm.fhir.schema.app.Main.dropTenant(Main.java:1106)
        at com.ibm.fhir.schema.app.Main.process(Main.java:1803)
[...]

PR #2487, opened 2021-06-08
Cleanly handle tablespace removal error (fix for  When multiple tenants exist, Drop Tenant Results in an incorrect FHIRDATA.PARAMETER_NAMES association and drop. #2354 )

prb112:
The core issue was a timing problem with lots of partitions dettaching. 
This adds a delay to the dropTablespace so the async operation can complete, and cleanly exit with a specific error code so downstream consumers can work around the issue. 


lmsurpre (approved):
LGTM

Issue #1922, opened 2021-02-08
whole-system-search: use global parameter tables in whole system search join


punktilious:
Related to #264. Once resource ingestion populates the global parameter tables (parameter tables which are tied directly to the global LOGICAL_RESOURCES table, the whole system search query can be rewritten to greatly simplify the join.

The best implementation to retrieve the data payload for each resource type needs to be investigated. Options include a WITH+UNION or perhaps an outer join with a large coalesce statement. To simplify the join as much as possible, it may be necessary to hold the payload at the global layer, which would help to align the design with the ability to offload to the payload to external storage. These options should be evaluated for their impact on performance before any changes are made to the search query build code.

For example, the following structure may perform reasonably well:

```WITH driver AS (SELECT rt.resource_type, lr.logical_resource_id
        FROM fhirdata.resource_types rt,
             fhirdata.logical_resources lr,
             fhirdata.str_values AS param0
       WHERE rt.resource_type_id = lr.resource_type_id
         AND param0.parameter_name_id = 2582160
         AND param0.str_value = 'abc123'
         AND lr.logical_resource_id = param0.logical_resource_id
     ) 
         SELECT lres.logical_resource_id
           FROM driver
LEFT OUTER JOIN fhirdata.patient_logical_resources AS lres
             ON (lres.logical_resource_id = driver.logical_resource_id
            AND driver.resource_type = 'Patient')
UNION ALL
         SELECT lres.logical_resource_id
           FROM driver
LEFT OUTER JOIN fhirdata.observation_logical_resources AS lres
             ON (lres.logical_resource_id = driver.logical_resource_id
            AND driver.resource_type = 'Observation')
UNION ALL
         SELECT lres.logical_resource_id
           FROM driver
LEFT OUTER JOIN fhirdata.claim_logical_resources AS lres
             ON (lres.logical_resource_id = driver.logical_resource_id
            AND driver.resource_type = 'Claim')
;
```

Although this form requires the common table expression (CTE) result to be scanned multiple times even though it is evaluated only once. The 'resource_type = 'Claim' predicate in the outer join helps to avoid any logical reads for rows not matching the particular resource type for that sub-query which should keep CPU usage lower.

An alternative is to outer-join each of the resource-specific tables to the main whole-system-search subquery:

```
SELECT COALESCE(patients.logical_resource_id,
                claims.logical_resource_id,
                observations.logical_resource_id) AS logical_resource_id
[...]

PR #2490, opened 2021-06-09
Issue #1922 - implement whole-system search in new query builder

michaelwschroeder:

### DIFF
--- a/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/query/Select.java
+++ b/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/query/Select.java
@@ -246,13 +246,7 @@ public class Select {
             // Generate a pretty-printed string using the renderer so we
             // get an accurate version of the string.
             StringStatementRenderer renderer = new StringStatementRenderer(TRANSLATOR, null, true);
-            if (union != null) {
-                StringBuilder result = new StringBuilder();
-                result.append(this.render(renderer)).append(union.toDebugString());
-                return result.toString();
-            } else {
-                return this.render(renderer);
-            }
+            return this.render(renderer);
         } catch (Exception x) {
             // If we can't render, it's very likely this debug is being used already
             // in a catch clause from an earlier exception. So rather than propagating
@@ -268,18 +262,9 @@ public class Select {
      * @param renderer
      * @return
      */
-    @SuppressWarnings("unchecked")
     public <T> T render(StatementRenderer<T> renderer) {
-        if (union != null) {
-            StringBuilder result = new StringBuilder();
-            T rendered = renderer.select(distinct, selectList, fromClause, whereClause,
-                groupByClause, havingClause, orderByClause, paginationClause, unionAll, union);
-            result.append(rendered).append(union.render(renderer));
-            return (T) result.toString();
-        } else {
-            return renderer.select(distinct, selectList, fromClause, whereClause, groupByClause, havingClause,
-                orderByClause, paginationClause, unionAll, union);
-        }
+        return renderer.select(distinct, selectList, fromClause, whereClause, groupByClause, havingClause,
+            orderByClause, paginationClause, unionAll, union);
     }
 
     /**
--- a/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/query/Select.java
+++ b/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/query/Select.java
@@ -335,17 +335,17 @@ public class Select {
     }
 
     /**
-     * Add a select to UNION with this query.
+     * Set a select to UNION with this query.
      */
-    public void addUnion(Select union) {
+    public void setUnion(Select union) {
         this.union = union;
         this.unionAll = false;
     }
 
     /**
-     * Add a select to UNION ALL with this query.
+     * Set a select to UNION ALL with this query.
      */
-    public void addUnionAll(Select unionAll) {
+    public void setUnionAll(Select unionAll) {
         this.union = unionAll;
         this.unionAll = true;
     }
--- a/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/query/Select.java
+++ b/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/query/Select.java
@@ -9,6 +9,8 @@ package com.ibm.fhir.database.utils.query;
 import static com.ibm.fhir.database.utils.query.SqlConstants.FROM;
 import static com.ibm.fhir.database.utils.query.SqlConstants.SELECT;
 import static com.ibm.fhir.database.utils.query.SqlConstants.SPACE;
+import static com.ibm.fhir.database.utils.query.SqlConstants.UNION;
+import static com.ibm.fhir.database.utils.query.SqlConstants.UNION_ALL;
 
 import com.ibm.fhir.database.utils.api.IDatabaseTranslator;
 import com.ibm.fhir.database.utils.derby.DerbyTranslator;
@@ -47,6 +49,12 @@ public class Select {
 
     // offset/limit for pagination
     private PaginationClause paginationClause;
+
+    // Another Select to UNION with this select. Optional
+    private Select union;
+
+    // If true, the specified UNION is a UNION ALL
+    private boolean unionAll = false;
 
     /**
      * Default constructor. Not a DISTINCT select.
@@ -218,6 +226,11 @@ public class Select {
             result.append(SPACE).append(this.paginationClause.toString());
         }
 
+        if (this.union != null) {
+            result.append(SPACE).append(unionAll ? UNION_ALL : UNION).append(SPACE)
+                    .append(this.union.toString());
+        }
+
         return result.toString();
     }
 
@@ -233,7 +246,13 @@ public class Select {
             // Generate a pretty-printed string using the renderer so we
             // get an accurate version of the string.
             StringStatementRenderer renderer = new StringStatementRenderer(TRANSLATOR, null, true);
-            return this.render(renderer);
+            if (union != null) {
+                StringBuilder result = new StringBuilder();
+                result.append(this.render(renderer)).append(union.toDebugString());
+                return result.toString();
+            } else {
+                return this.render(renderer);
+            }
         } catch (Exception x) {
             // If we can't render, it's very likely this debug is being used already
             // in a catch clause from an earlier exception. So rather than propagating
@@ -249,8 +268,18 @@ public class Select {
      * @param renderer
      * @return
      */
+    @SuppressWarnings("unchecked")
     public <T> T render(StatementRenderer<T> renderer) {
-        return renderer.select(distinct, selectList, fromClause, whereClause, groupByClause, havingClause, orderByClause, paginationClause);
+        if (union != null) {
+            StringBuilder result = new StringBuilder();
+            T rendered = renderer.select(distinct, selectList, fromClause, whereClause,
+                groupByClause, havingClause, orderByClause, paginationClause, unionAll, union);
+            result.append(rendered).append(union.render(renderer));
+            return (T) result.toString();
+        } else {
+            return renderer.select(distinct, selectList, fromClause, whereClause, groupByClause, havingClause,
+                orderByClause, paginationClause, unionAll, union);
+        }
     }
 
     /**
@@ -318,5 +347,21 @@ public class Select {
      */
     public OrderByClause getOrderByClause() {
         return this.orderByClause;
+    }
+
+    /**
+     * Add a select to UNION with this query.
+     */
+    public void addUnion(Select union) {
+        this.union = union;
+        this.unionAll = false;
+    }
+
+    /**
+     * Add a select to UNION ALL with this query.
+     */
+    public void addUnionAll(Select unionAll) {
+        this.union = unionAll;
+        this.unionAll = true;
     }
 }
--- a/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/query/expression/StatementRenderer.java
+++ b/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/query/expression/StatementRenderer.java
@@ -14,6 +14,7 @@ import com.ibm.fhir.database.utils.query.GroupByClause;
 import com.ibm.fhir.database.utils.query.HavingClause;
 import com.ibm.fhir.database.utils.query.OrderByClause;
 import com.ibm.fhir.database.utils.query.PaginationClause;
+import com.ibm.fhir.database.utils.query.Select;
 import com.ibm.fhir.database.utils.query.SelectList;
 import com.ibm.fhir.database.utils.query.WhereClause;
 import com.ibm.fhir.database.utils.query.node.ExpNode;
@@ -36,10 +37,12 @@ public interface StatementRenderer<T> {
      * @param havingClause
      * @param orderByClause
      * @param paginationClause
+     * @param unionAll
+     * @param union
      * @return
      */
     T select(boolean distinct, SelectList selectList, FromClause fromClause, WhereClause whereClause, GroupByClause groupByClause, HavingClause havingClause,
-        OrderByClause orderByClause, PaginationClause paginationClause);
+        OrderByClause orderByClause, PaginationClause paginationClause, boolean unionAll, Select union);
 
     /**
      * @param items
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/FHIRPersistenceJDBCCache.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/FHIRPersistenceJDBCCache.java
@@ -7,6 +7,7 @@
 package com.ibm.fhir.persistence.jdbc;
 
 import com.ibm.fhir.persistence.jdbc.dao.api.ICommonTokenValuesCache;
+import com.ibm.fhir.persistence.jdbc.dao.api.IIdNameCache;
 import com.ibm.fhir.persistence.jdbc.dao.api.INameIdCache;
 
 /**
@@ -33,6 +34,12 @@ public interface FHIRPersistenceJDBCCache {
      * @return
      */
     INameIdCache<Integer> getResourceTypeCache();
+
+    /**
+     * Getter for the cache of resource type ids used to look up resource type name
+     * @return
+     */
+    IIdNameCache<Integer> getResourceTypeNameCache();
 
     /**
      * Getter for the cache of parameter names
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/JDBCConstants.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/JDBCConstants.java
@@ -48,6 +48,8 @@ public class JDBCConstants {
     public static final String IS_DELETED = "IS_DELETED";
 
     // Special parameter names
+    public static final String PARAM_NAME_ID = "_id";
+    public static final String PARAM_NAME_LAST_UPDATED = "_lastUpdated";
     public static final String PARAM_NAME_PROFILE = "_profile";
     public static final String PARAM_NAME_SECURITY = "_security";
     public static final String PARAM_NAME_TAG = "_tag";
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/JDBCConstants.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/JDBCConstants.java
@@ -47,13 +47,6 @@ public class JDBCConstants {
     public static final String IS_DELETED_NO = "IS_DELETED = 'N'";
     public static final String IS_DELETED = "IS_DELETED";
 
-    // Special parameter names
-    public static final String PARAM_NAME_ID = "_id";
-    public static final String PARAM_NAME_LAST_UPDATED = "_lastUpdated";
-    public static final String PARAM_NAME_PROFILE = "_profile";
-    public static final String PARAM_NAME_SECURITY = "_security";
-    public static final String PARAM_NAME_TAG = "_tag";
-
     // Generic SQL query string constants
     public static final String DOT = ".";
     public static final char DOT_CHAR = '.';
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/cache/IdNameCache.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/cache/IdNameCache.java
@@ -1,0 +1,89 @@
+/*
+ * (C) Copyright IBM Corp. 2021
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.ibm.fhir.persistence.jdbc.cache;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import com.ibm.fhir.persistence.jdbc.dao.api.IIdNameCache;
+
+
+/**
+ * @param <T> the type of the key value held by the cache
+ */
+public class IdNameCache<T> implements IIdNameCache<T> {
+
+    // We use LinkedHashMap for the local map because we also need to maintain order
+    // of insertion to make sure we have correct LRU behavior when updating the shared cache
+    private final ThreadLocal<Map<T,String>> local = new ThreadLocal<>();
+    
+    // The cache shared at the server level
+    private final ConcurrentHashMap<T,String> shared = new ConcurrentHashMap<>();
+    
+    /**
+     * Public constructor
+     */
+    public IdNameCache() {
+    }
+
+    @Override
+    public void updateSharedMaps() {
+        
+        Map<T,String> localMap = local.get();
+        if (localMap != null) {
+            // no need to synchronize because we can use a ConcurrentHashMap
+            shared.putAll(localMap);
+            localMap.clear();
+        }
+    }
+    
+    @Override
+    public String getName(T key) {
+        String result = null;
+        Map<T,String> localMap = local.get();
+        if (localMap != null) {
+            result = localMap.get(key);
+        }
+
+        if (result == null) {
+            result = shared.get(key);
+        }
+        return result;
+    }
+
+    @Override
+    public void addEntry(T id, String name) {
+        Map<T,String> localMap = local.get();
+        if (localMap == null) {
+            localMap = new HashMap<>();
+            local.set(localMap);
+        }
+        localMap.put(id, name);
+    }
+
+    @Override
+    public void reset() {
+        local.remove();
+        shared.clear();
+    }
+
+    @Override
+    public void clearLocalMaps() {
+        Map<T,String> map = local.get();
+        if (map != null) {
+            map.clear();
+        }
+    }
+
+    @Override
+    public void prefill(Map<T,String> content) {
+        // as the given content is supposed to be already committed in the database,
+        // we can add it directly to the shared map
+        this.shared.putAll(content);
+    }
+}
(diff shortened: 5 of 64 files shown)


michaelwschroeder:
Included in this PR:
1. whole-system search excluding the _type parameter in new query builder

Excluded - will address in subsequent PRs:
1. old query builder updates to use new _profiles and _tags tables
2. whole-system search with _type parameter in new query builder (including UNION support)

michaelwschroeder:
Second commit includes code to allow old query builder to use new _profiles and _tags tables.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/cache/IdNameCache.java
```diff
@@ -0,0 +1,89 @@
...
+import com.ibm.fhir.persistence.jdbc.dao.api.IIdNameCache;
+
+
+/**
+ * @param <T> the type of the key value held by the cache
+ */
+public class IdNameCache<T> implements IIdNameCache<T> {
+
+    // We use LinkedHashMap for the local map because we also need to maintain order
+    // of insertion to make sure we have correct LRU behavior when updating the shared cache
```
punktilious, line 22: I don't think this comment still applies to this implementation, which caches everything and doesn't need to be LRU

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/EraseResourceDAO.java
```diff
@@ -305,9 +305,13 @@ public void deleteFromAllParametersTables(String tablePrefix, long logicalResour
         deleteFromParameterTable(tablePrefix + "_latlng_values", logicalResourceId);
         deleteFromParameterTable(tablePrefix + "_resource_token_refs", logicalResourceId);
         deleteFromParameterTable(tablePrefix + "_quantity_values", logicalResourceId);
+        deleteFromParameterTable(tablePrefix + "_profiles", logicalResourceId);
+        deleteFromParameterTable(tablePrefix + "_tags", logicalResourceId);
```
punktilious, line 309: Ouch. Thanks for catching this. This really makes we want to factor these deletes out into their own procedure.

michaelwschroeder:
Third commit includes code to:
1. handle _sort with whole-system search
2. implement whole-system search with _type in new query builder
3. add UNION support to new query model

Remaining:
 - support text searches of _tag parameter (i.e. with modifiers like :text, :in, :not-in, :above, :below)

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/query/Select.java
```diff
@@ -249,8 +268,18 @@ public String toDebugString() {
      * @param renderer
      * @return
      */
+    @SuppressWarnings("unchecked")
     public <T> T render(StatementRenderer<T> renderer) {
```
lmsurpre, line 272: Is T ever NOT a string?

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/query/Select.java
```diff
@@ -48,6 +50,12 @@
     // offset/limit for pagination
     private PaginationClause paginationClause;
 
+    // Another Select to UNION with this select. Optional
+    private Select union;
+
+    // If true, the specified UNION is a UNION ALL
+    private boolean unionAll = false;
```
lmsurpre, line 57: is UNION ALL just an optimization or we need duplicates after the union somewhere?

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/query/expression/StatementRenderer.java
```diff
@@ -36,10 +37,12 @@
      * @param havingClause
      * @param orderByClause
      * @param paginationClause
+     * @param unionAll
+     * @param union
      * @return
      */
     T select(boolean distinct, SelectList selectList, FromClause fromClause, WhereClause whereClause, GroupByClause groupByClause, HavingClause havingClause,
-        OrderByClause orderByClause, PaginationClause paginationClause);
+        OrderByClause orderByClause, PaginationClause paginationClause, boolean unionAll, Select union);
```
lmsurpre, line 45: would it make sense to overload this so that we have a simpler option to call?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/FHIRPersistenceJDBCCache.java
```diff
@@ -7,6 +7,7 @@
 package com.ibm.fhir.persistence.jdbc;
 
 import com.ibm.fhir.persistence.jdbc.dao.api.ICommonTokenValuesCache;
+import com.ibm.fhir.persistence.jdbc.dao.api.IIdNameCache;
 import com.ibm.fhir.persistence.jdbc.dao.api.INameIdCache;
```
lmsurpre, line 11: I guess its just a reverse of the same info...is it worth trying to combine these?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/JDBCConstants.java
```diff
@@ -48,6 +48,8 @@
     public static final String IS_DELETED = "IS_DELETED";
 
     // Special parameter names
+    public static final String PARAM_NAME_ID = "_id";
+    public static final String PARAM_NAME_LAST_UPDATED = "_lastUpdated";
     public static final String PARAM_NAME_PROFILE = "_profile";
     public static final String PARAM_NAME_TAG = "_tag";
```
lmsurpre, line 54: wondering if these should be in `fhir-persistence` since they aren't specific to our JDBC impl.  We probably have them already in some more common place as well (just a guess)?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/FHIRPersistenceJDBCCache.java
lmsurpre: something like http://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/BidiMap.html ...

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/api/IIdNameCache.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.util.Map;
+
+/**
+ * Interface to a cache mapping an id of type T to a string. Supports
+ * thread-local caching to support temporary staging of values pending
+ * successful completion of a transaction.
+ * @param<T>
+ */
+public interface IIdNameCache<T> {
```
lmsurpre, line 17: I think you're just following the pattern of the INameIdCache but I don't really understand why we need interfaces for these...will they ever have more than one implementation?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
```diff
@@ -814,4 +814,41 @@ public int searchCount(Select countQuery) throws FHIRPersistenceDataAccessExcept
         }
         return resourceIds;
    }
+
+    @Override
+    public Map<Integer, List<Long>> searchWholeSystem(Select wholeSystemQuery) throws FHIRPersistenceDataAccessException,
+            FHIRPersistenceDBConnectException {
+        final String METHODNAME = "searchWholeSystem";
+        log.entering(CLASSNAME, METHODNAME);
```
lmsurpre, line 822: worth adding the argument to this?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/domain/SearchQueryRenderer.java
```diff
@@ -270,8 +281,8 @@ public QueryData getParameterBaseQuery(QueryData parent) {
         //           AND LRn.LOGICAL_RESOURCE_ID = LRp.LOGICAL_RESOURCE_ID
         SelectAdapter exists = Select.select("1");
         exists.from(xxLogicalResources, alias(lrAlias))
-         .where(lrAlias, "IS_DELETED").eq().literal("N") // TODO remove either from here or parent
-         .and(lrAlias, "LOGICAL_RESOURCE_ID").eq(parentLRAlias, "LOGICAL_RESOURCE_ID"); // correlate to parent query
+           .where(lrAlias, "IS_DELETED").eq().literal("N") // TODO remove either from here or parent
```
lmsurpre, line 284: looks like we only touched this line to fix the formatting, but now I'm wondering about this TODO...

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/domain/SearchQueryRenderer.java
```diff
@@ -710,6 +831,26 @@ public QueryData addSorting(QueryData queryData, String lrAlias) {
         return queryData;
     }
 
+    @Override
+    public QueryData addWholeSystemSorting(QueryData queryData, List<DomainSortParameter> sortParms, String lrAlias) {
+        if (sortParms == null || sortParms.isEmpty()) {
+            return addSorting(queryData, lrAlias);
+        } else {
+            for (DomainSortParameter sortParm : sortParms) {
+                // for whole-system searches, code has been validated to only be _id or _lastUpdated
```
lmsurpre, line 840: is it not possible for a tenant to define their own whole-system search param of type code?
we should be up front about whether this is supported or not.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/domain/SearchQueryRenderer.java
```diff
@@ -2088,4 +2361,28 @@ private void addAggregateAndOrderByExpressions(QueryData queryData, String code,
...
+        return Resource.class.getSimpleName().equals(resourceType);
+    }
+
+    /**
+     * Check if query parameter is a _tag parameter whose modifier requires
+     * search of the tag's value.
+     * @param queryParm
+     * @return true if tag value search, false otherwise
+     */
+    private boolean isTagValueSearch(QueryParameter queryParm) {
```
lmsurpre, line 2380: I don't totally understand this one.  Why is a tag search with no modifier not a "Tag Value Search"?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/JDBCIdentityCacheImpl.java
```diff
@@ -70,6 +72,30 @@ public Integer getResourceTypeId(String resourceType) throws FHIRPersistenceExce
...
+                }
+            }
+
+            if (result == null) {
+                // likely a configuration error, caused by the schema being generated
+                // for a subset of all possible resource types
+                throw new FHIRPersistenceDataAccessException("Resource type ID not registered in database: '" + resourceTypeId + "'");
+            }
+
+            cache.getResourceTypeNameCache().addEntry(resourceTypeId, result);
```
tbieste, line 94: Can we populate the cache with all the entries we get back from the DB, instead of just this one?  And even populate the getResourceTypeCache as well.  Same thing with getResourceTypeId method, can that populate this cache with at least the entry it gets back?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -2074,4 +2098,203 @@ private void populateCodesSubSegment(StringBuilder whereClauseSegment, Modifier
...
+                whereClauseSegment.append(OR);
+            }
+
+            // Reuse the same CanonicalSupport code used for param extraction to parse the search value
+            ResourceProfileRec rpc = CanonicalSupport.makeResourceProfileRec(-1, resourceType.getSimpleName(), -1, -1,
+                    value.getValueString(), false);
+            int canonicalId = identityCache.getCanonicalId(rpc.getCanonicalValue());
+            whereClauseSegment.append(tableAlias).append(DOT).append("CANONICAL_ID").append(EQ).append(canonicalId);
+            
+            // TODO double-check semantics of ABOVE and BELOW in this context
```
tbieste, line 2127: Is this TODO still needed?

fhir-persistence-jdbc/src/test/java/com/ibm/fhir/persistence/jdbc/test/spec/Main.java
```diff
@@ -53,6 +53,7 @@
 import com.ibm.fhir.persistence.jdbc.FHIRPersistenceJDBCCache;
```
tbieste, line 53: Update copyright date.

tbieste (approved):

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/domain/SearchWholeSystemQuery.java
```diff
@@ -0,0 +1,83 @@
...
+    // Sort parameters
+    final List<DomainSortParameter> sortParameters = new ArrayList<>();
+
+    /**
+     * Public constructor
+     * @param domainModels
+     * @param isCountQuery
+     * @param addPagination
+     */
+    public SearchWholeSystemQuery(List<SearchQuery> domainModels, boolean isCountQuery, boolean addPagination) {
```
lmsurpre, line 39: should we prohibit `isCountQuery=true` if `addPagination=true`?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -638,14 +638,11 @@ private ParameterDAO makeParameterDAO(Connection connection) throws FHIRPersiste
...
             throws FHIRPersistenceException {
 
-        // Fall back to the old search code for whole-system searches which are not yet supported
-        // by the new code.
-        if (isSystemLevelSearch(resourceType) || !this.optQueryBuilderEnabled) {
-            // New query builder doesn't support system-level search at this point, so route
-            // to the old way.
+        // Fall back to the old search code if new query builder has been disabled.
+        if (!this.optQueryBuilderEnabled) {
             return oldSearch(context, resourceType);
```
lmsurpre, line 643: opened #2522 for deprecating this

fhir-persistence-schema/src/main/resources/db2/erase_resource.sql
```diff
@@ -80,13 +80,25 @@ BEGIN
     PREPARE d_stmt FROM 'DELETE FROM {{SCHEMA_NAME}}.' || p_resource_type || '_quantity_values     WHERE logical_resource_id = ?';
     EXECUTE d_stmt USING v_logical_resource_id;
 
-    PREPARE d_stmt FROM 'DELETE FROM {{SCHEMA_NAME}}.' || 'str_values          WHERE logical_resource_id = ?';
+    PREPARE d_stmt FROM 'DELETE FROM {{SCHEMA_NAME}}.' || p_resource_type || '_profiles            WHERE logical_resource_id = ?';
```
lmsurpre, line 83: are there "global" tables for profiles and tags that we need to delete from as well, or just these resource-type-specific ones?

lmsurpre: ah, now i see them down below.  `logical_resource_profiles` and `logical_resource_tags` respectively.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/cache/IdNameCache.java
michaelwschroeder: comment removed

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/query/Select.java
michaelwschroeder: UNION ALL will not remove duplicates, while UNION will. This flag indicates if the Select statement in the 'union' field should be a UNION ALL or a UNION. Just gives the caller the option to choose which they want to use.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/FHIRPersistenceJDBCCache.java
michaelwschroeder: I looked at combining vs. separate. I thought keeping them separate was a little more straight-forward.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/JDBCConstants.java
michaelwschroeder: I removed them from fhir-persistence. There are existing ones in SearchConstants, so added the new ones there as well.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceDAOImpl.java
michaelwschroeder: The SQL actually gets logged by QueryUtil.prepareSelect(), so I chose not to add here

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/domain/SearchQueryRenderer.java
michaelwschroeder: This method gets called by multiple parents, so I'm not comfortable pulling it out. Certainly worth looking at though.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/domain/SearchQueryRenderer.java
michaelwschroeder: The code referred to here is actually just the sort parameter name. For whole-system search we only support sort parameters of _id and _lastUpdated. As we've since discussed, we will allow user-defined whole-system search parameters of types for which we have global values tables.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/domain/SearchQueryRenderer.java
michaelwschroeder: I've removed this method since it only got called in one place after I re-factored the code. What I meant by this is that for the :text, :in, :not-in, :above, and :below modifiers, we need access to the token value. In the new global tables for _tag and _security, we only have a common_token_value_id, so that means we have to join to the common_token_values table to access the token value.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/JDBCIdentityCacheImpl.java
michaelwschroeder: done

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
michaelwschroeder: I think it needs further discussion. I pulled this code over from some existing NewQueryBuilder code, so same issue there. The spec isn't real clear on the semantics of :above and :below for canonical values.

fhir-persistence-jdbc/src/test/java/com/ibm/fhir/persistence/jdbc/test/spec/Main.java
michaelwschroeder: done

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/domain/SearchWholeSystemQuery.java
michaelwschroeder: Since pagination (and sort as well) is not applicable for count queries I just updated the code to ignore that flag if isCountQuery=true

michaelwschroeder:
Note: some of the _security testcases will fail until Robin's schema changes go in.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/query/Select.java
michaelwschroeder: T is going to be whatever the renderer returns (currently only have a String implementation of the renderer, but I guess could potentially have others). I rewrote this so the renderer handles the union - so not casting to String. I removed the SuppressWarnings.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/query/expression/StatementRenderer.java
michaelwschroeder: This is only called in one place, and if we simplified the call, we'd have to create getters for the attributes in the Select object, which don't currently exist. I think we can defer until we need a simpler option.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/api/IIdNameCache.java
michaelwschroeder: Yeah, just wanted to be consistent with the other caches.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/domain/SearchQueryRenderer.java
michaelwschroeder: I re-looked at this and verified that we could remove it, and went ahead and removed it and left a comment that it's the caller's responsibility to check IS_DELETED if necessary.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/query/Select.java
```diff
@@ -319,4 +348,20 @@ public void addPagination(int offset, int rowsPerPage) {
...
+     */
+    public void addUnion(Select union) {
+        this.union = union;
+        this.unionAll = false;
+    }
+
+    /**
+     * Add a select to UNION ALL with this query.
+     */
+    public void addUnionAll(Select unionAll) {
```
punktilious, line 363: unionAll is a single value, so I think this should be set, not add.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/query/expression/StatementRenderer.java
punktilious: This is the renderer interface, so I think it's OK, because it's really just a framework call, not something used by consumers of the framework.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/FHIRPersistenceJDBCCache.java
punktilious: I'm OK with keeping them separate for now. We're probably going to move to a distribute cache at some point for better scalability, so we look at options then.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/EraseResourceDAO.java
```diff
@@ -305,9 +305,13 @@ public void deleteFromAllParametersTables(String tablePrefix, long logicalResour
...
         deleteFromParameterTable("str_values", logicalResourceId);
         deleteFromParameterTable("date_values", logicalResourceId);
         deleteFromParameterTable("resource_token_refs", logicalResourceId);
+        deleteFromParameterTable("logical_resource_profiles", logicalResourceId);
+        deleteFromParameterTable("logical_resource_tags", logicalResourceId);
```
punktilious, line 314: @michaelwschroeder you mentioned you'd add the new security table deletes here too.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/api/IIdNameCache.java
punktilious: I like to hide implementation details where possible so we can change the underlying implementation in the future if we need to.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
```diff
@@ -2074,4 +2101,194 @@ private void populateCodesSubSegment(StringBuilder whereClauseSegment, Modifier
         log.exiting(CLASSNAME, METHODNAME);
     }
 
+    private SqlQueryData processProfileParm(Class<?> resourceType, QueryParameter queryParm, String tableAlias)
```
punktilious, line 2104: Isn't the old query builder going away completely with this change?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/NewQueryBuilder.java
```diff
@@ -181,8 +225,48 @@ public Select buildQuery(Class<?> resourceType, FHIRSearchContext searchContext)
...
+                    // query by resource type. Add an extension to the domain model
+                    // with the specified resource type IDs.
+                    addResourceTypeExtension(wholeSystemFilterQuery, searchContext.getSearchResourceTypes());
+                }
+                domainModel = wholeSystemFilterQuery;
+            } else {
+                // Not all search parameters are global (values indexed into global values tables).
+                // Need to do old-style UNION'd query against all resource types.
+                List<String> resourceTypes = searchContext.getSearchResourceTypes();
+                if (resourceTypes == null) {
```
punktilious, line 247: I guess at the moment we don't fully support working with a limited set of resources in the schema. I think Lee has an issue open for that, which I guess will be important here.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/NewQueryBuilder.java
```diff
@@ -181,8 +225,48 @@ public Select buildQuery(Class<?> resourceType, FHIRSearchContext searchContext)
...
+                    // The _type parameter was not specified, so we need to generate a list
+                    // of all supported resource types for our UNION query.
+                    resourceTypes = new ArrayList<>(this.identityCache.getResourceTypeNames());
+                    resourceTypes.remove("Resource");
```
punktilious, line 251: Remove will be a linear scan each time. Although perhaps making this a set would be slower. I guess we can create this list at bootstrap time (in the servlet listener) when we add proper support for limited resource types in the schema.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCQueryBuilder.java
michaelwschroeder: Lee opened issue #2522 for this.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/query/Select.java
michaelwschroeder: Done.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/EraseResourceDAO.java
michaelwschroeder: Done

punktilious (approved):
LGTM - will fully review when this hits main

PR #2498, opened 2021-06-10
Added new parser benchmark that works against an examples index

lmsurpre:
Instead of parsing a single file, this variant parses a set of files.
I included benchmark variants for single threaded parsing vs a
parallelStream to parse the examples in parallel.


JohnTimm (approved):
LGTM

Issue #2195, opened 2021-04-02
Drive reindex using client side checkpointing for improved throughput


punktilious:
**Is your feature request related to a problem? Please describe.**
The current `$reindex` operation uses the `reindex_tstamp` column in LOGICAL_RESOURCES for selecting which resources to process for a particular thread. This selection process involves updating the column and using database-specific techniques to avoid concurrency issues from the resulting row-locks.

In PostgreSQL, this update leaves "tombstone" markers in the blocks which only get cleaned when the table is next vacuumed. If vacuuming is not aggressive enough, `$reindex` slows significantly due to the extra index blocks being scanned every time the request processor attempts to acquire a new resource to process. 

**Describe the solution you'd like**
Although using the `reindex_tstamp` simplifies the client needed to drive the reindex operation (it can be as simple as a shell script running curl in a loop), better throughput could be achieved by avoid the update statement and instead tracking progress (checkpointing) with a more sophisticated client implementation.

1. Implement a new API endpoint to provide a list of logical_resource_ids in increasing order (similar to the whole system history operation, but ignoring multiple versions). Each request would include an "afterLogicalResourceId=xxx" parameter, and `_count` can be used to limit the number of resources selected each time. Only the logical_resource_id values would need to be returned, not the resources.
2. The client can package up a list of logical resources and submit to a thread pool for reindexing.
3. Support new reindex parameters allowing the client to specify which logical_resource_ids (or possibly resourceType/logicalId). The current parameter block allows only a single resource to be specified.
4. The client keeps track of the max contiguous logical_resource_id which has been successfully processed (non-trivial, but implementable with care)
5. Client can checkpoint this value by writing a file locally so it can be restarted from a known point if necessary.
6. As soon as the batch of logical_resource_ids is submitted to the thread pool, the main thread can make another request for the next batch
7. The client can ask for a large number of ids in one go and break them into smaller batches for processing in parallel.

Check that the resulting throughput is greater than the current reindex operation (which can be kept), and that its throughput doesn't slow over time due to delayed vacuum of the logical_resources table

tbieste:
**Sub-task #1:**
-- Get resource IDs (in order), so a client can call the $reindex operation passing in the resource IDs of resources to reindex.

New Operation:  $list-index
-- Similar to whole system history, but only current version of resources

Input parameters:
  -- _afterLogicalResourceId (optional): similar to _afterHistoryId for history API
  -- _count (optional): similar to _count for history API

Output:
[...]

PR #2489, opened 2021-06-08
End-to-end automation test for reindexing #1751

prb112:
This PR addresses: 

1 - Hardening the Reindex Operation to bad input and reflective responses
2 - Update the documentation to include an omitted input parameter 
3 - Add Reindex workflow with three jobs (in the matrix) - db2, postgres, derby
4 - Add multiple phase tests which require restarts between them
5 - BasicDate.json had a small error using valueString instead of code
6 - Added LogicalResource and Type level for the ReindexOperation
7 - Cleaned up the FHIROperationRegistry to register at system, type and instance level

What's not included: 
1 - Reindex after Migration (a separate PR, based on these integration tests, will address the tests)

Signed-off-by: Paul Bastide <pbastide@us.ibm.com>

PR #2502, opened 2021-06-11
Issue #2195 - Enable client checkpoint driven reindex

tbieste:
Issue #2195 - Enable client checkpoint driven reindex


docs/src/pages/guides/FHIRPerformanceGuide.md
```diff
@@ -200,7 +200,7 @@ The following table summarizes how the transaction timeout is used for different
...
 | Transaction Bundle | Single transaction scope for entire request |
-| $reindex           | One HTTP call can request multiple resources to be reindexed. Each resource is reindexed in the scope of its own transaction. Reindexing is a relatively quick operation per resource - usually well under 1s - so transaction timeouts are unlikely. Reduce the number of resources processed per reindex operation to avoid read timeouts. Use concurrent requests to increase overall throughput. |
+| $reindex           | One HTTP call can request multiple resources to be reindexed. By default, each resource is reindexed in the scope of its own transaction. Reindexing is a relatively quick operation per resource - usually well under 1s - so transaction timeouts are unlikely. However, if a list of logical resource IDs is specified, all those resources will be reindexed within a single tranaction, so reduce the number logical resource IDs specified if transaction timeouts occur. Reduce the number of resources processed per reindex operation to avoid read timeouts. Use concurrent requests to increase overall throughput. |
```
lmsurpre, line 203: seems a bit redundant now :-)

lmsurpre:
```suggestion
| $reindex           | One HTTP call can request multiple resources to be reindexed. By default, each resource is reindexed in the scope of its own transaction. Reindexing is a relatively quick operation per resource - usually well under 1s - so transaction timeouts are unlikely. However, if a list of logical resource IDs is specified, all those resources will be reindexed within a single tranaction, so reduce the number logical resource IDs specified if transaction timeouts occur. Use concurrent requests to increase overall throughput. |
```

tbieste: Updated.

docs/src/pages/guides/FHIRSearchConfiguration.md
```diff
@@ -263,20 +263,43 @@ In order to avoid this issue, inclusion criteria search parameters should not be
...
+|`tstamp`|string|Reindex only resources not previously reindexed since this timestamp. Format as a date YYYY-MM-DD or time YYYY-MM-DDTHH:MM:SSZ.|
 |`resourceCount`|integer|The maximum number of resources to reindex in this call. If this number is too large, the processing time might exceed the transaction timeout and fail.|
 
 The IBM FHIR Server tracks when a resource was last reindexed and only resources with a reindex_tstamp value less than the given tstamp parameter will be processed. When a resource is reindexed, its reindex_tstamp is set to the given tstamp value. In most cases, using the current date (for example "2020-10-27") is the best option for this value.
 
-To aid in the re-indexing process, the IBM FHIR Server team has expanded the fhir-bucket resource-loading tool to support driving the reindex. The fhir-bucket tool uses a thread-pool to make concurrent POST requests to the IBM FHIR Server `$reindex` custom operation.
+### 2.2 Client-side-driven approach
+Another option is to have the client utilize the `$retrieve-index` and ``$reindex` in parallel to drive the processing.
```
lmsurpre, line 281:
```suggestion
Another option is to have the client utilize the `$retrieve-index` and `$reindex` in parallel to drive the processing.
```

docs/src/pages/guides/FHIRSearchConfiguration.md
```diff
@@ -263,20 +263,43 @@ In order to avoid this issue, inclusion criteria search parameters should not be
...
+|name|type|description|
+|----|----|-----------|
+|`logicalResourceIds`|string|Reindex only resources with a logical resource ID in the specified list, formatted as a comma-delimited list of strings. If number of logical resource IDs in the list is too large, the processing time might exceed the transaction timeout and fail.|
+
+By specifying the logical resource IDs on the `$reindex` operation, the IBM FHIR Server avoids the database overhead of choosing the next resource to reindex and updating the reindex_tstamp. Though it requires the client side to tracking the reindex progress, it should allow for an overall faster reindex.
+
+### 2.3 fhir-bucket
+To aid in the re-indexing process, the IBM FHIR Server team has expanded the fhir-bucket resource-loading tool to support driving the reindex, with the option of using either the server-side-driven or client-side-driven approach. The fhir-bucket tool uses a thread-pool to make concurrent POST requests to the IBM FHIR Server `$retrieve-index` and ``$reindex` custom operations.
```
lmsurpre, line 300:
```suggestion
To aid in the re-indexing process, the IBM FHIR Server team has expanded the fhir-bucket resource-loading tool to support driving the reindex, with the option of using either the server-side-driven or client-side-driven approach. The fhir-bucket tool uses a thread-pool to make concurrent POST requests to the IBM FHIR Server `$retrieve-index` and `$reindex` custom operations.
```

fhir-bucket/src/main/java/com/ibm/fhir/bucket/app/Main.java
```diff
@@ -842,9 +851,13 @@ protected void scanAndLoad() {
...
         if (this.reindexTstampParam != null) {
-            this.driveReindexOperation = new DriveReindexOperation(fhirClient, reindexConcurrentRequests, reindexTstampParam, reindexResourceCount);
+            if (this.clientSideDrivenReindex) {
+                this.driveReindexOperation = new ClientDrivenReindexOperation(fhirClient, reindexConcurrentRequests, reindexTstampParam, reindexResourceCount);
+            } else {
+                this.driveReindexOperation = new ServerDrivenReindexOperation(fhirClient, reindexConcurrentRequests, reindexTstampParam, reindexResourceCount);
+            }
             this.driveReindexOperation.init();
-        }
+       }
```
lmsurpre, line 860:
```suggestion
        }
```

docs/src/pages/guides/FHIRSearchConfiguration.md
```diff
@@ -263,20 +263,43 @@ In order to avoid this issue, inclusion criteria search parameters should not be
...
+|`notModifiedAfter`|string|Only retrieve logical resource IDs for resources not last updated after this timestamp. Format as a date YYYY-MM-DD or time YYYY-MM-DDTHH:MM:SSZ.|
+|`afterLogicalResourceId`|string|Retrieve logical resource IDs starting with the first logical resource ID after this logical resource ID. If this parameter is not specified, the retrieved logical resource IDs start with the first logical resource ID.|
+
+The `$reindex` operation is called to reindex the resources with logical resource IDs in the specified list. The operation supports the following parameters to control the behavior:
 
-Reindexing is resource-intensive and can take several hours or even days to complete depending on the number of resources currently in the system and the capability of the hosting platform.
+|name|type|description|
+|----|----|-----------|
+|`logicalResourceIds`|string|Reindex only resources with a logical resource ID in the specified list, formatted as a comma-delimited list of strings. If number of logical resource IDs in the list is too large, the processing time might exceed the transaction timeout and fail.|
```
lmsurpre, line 295: is this our database-internal ids or the resource ids?
if the former, we need to change the name... "logical resource id" has a specific meaning to FHIR users and this aint it.

docs/src/pages/guides/FHIRSearchConfiguration.md
```diff
@@ -263,20 +263,43 @@ In order to avoid this issue, inclusion criteria search parameters should not be
...
+Reindexing is resource-intensive and can take several hours or even days to complete depending on the approach used, the number of resources currently in the system, and the capability of the hosting platform.
+
+### 2.1 Server-side-driven approach
+By default, the operation will select 10 resources and re-extract their search parameters values based on the current configuration of the server. The operation supports the following parameters to control the behavior:
 
 |name|type|description|
 |----|----|-----------|
-|`tstamp`|string|Reindex any resource not previously reindexed before this timestamp. Format as a date YYYY-MM-DD or time YYYY-MM-DDTHH:MM:DDZ.|
+|`tstamp`|string|Reindex only resources not previously reindexed since this timestamp. Format as a date YYYY-MM-DD or time YYYY-MM-DDTHH:MM:SSZ.|
 |`resourceCount`|integer|The maximum number of resources to reindex in this call. If this number is too large, the processing time might exceed the transaction timeout and fail.|
```
lmsurpre, line 276: will need to get reconciled with https://github.com/IBM/FHIR/pull/2489

fhir-bucket/src/main/java/com/ibm/fhir/bucket/reindex/ClientDrivenReindexOperation.java
```diff
@@ -0,0 +1,359 @@
...
+    private BlockingQueue<String> blockingQueue;
+
+    // Last logical resource ID found by monitor thread
+    private String lastLogicalResourceId;
+
+    // Monitor thread
+    private Thread monitorThread;
+
+    // Thread pool for processing requests
+    private final ExecutorService pool = Executors.newCachedThreadPool();
```
lmsurpre, line 74: Does it use `the number of cpu cores on the system - 1` in the pool?
Doe it relate to the passed `--reindex-concurrent-requests` value at all?

docs/src/pages/guides/FHIRSearchConfiguration.md
lmsurpre: I do realize that this is what its called in our database (as documented at https://github.com/IBM/FHIR/blob/main/fhir-persistence-schema/docs/SchemaMigration.md#logical_resources-and-resources) but I really don't want to expose this to our users because it is really confusing.  Happy to brainstorm alternative names as needed.

docs/src/pages/guides/FHIRSearchConfiguration.md
tbieste: Fixed.

docs/src/pages/guides/FHIRSearchConfiguration.md
tbieste: Fixed.

fhir-bucket/src/main/java/com/ibm/fhir/bucket/app/Main.java
tbieste: Fixed.

fhir-bucket/src/main/java/com/ibm/fhir/bucket/reindex/ClientDrivenReindexOperation.java
tbieste: Yes, the passed `--reindex-concurrent-requests` value determines the number of worker threads created in the thread pool in the monitorLoop method.  This was same logic as used in the ServerDrivenReindexOperation, so I followed that.

docs/src/pages/guides/FHIRSearchConfiguration.md
tbieste: I could just change the name to something like "indexId", so users just see it as just some ID used in reindex processing.

tbieste: Updated to use the name "indexIds" for the parameter, instead of "logicalResourceIds".  This should avoid confusing user into thinking it is the same as resource IDs.

docs/src/pages/guides/FHIRSearchConfiguration.md
tbieste: Yeah, those merge conflicts will need to be resolved.  Since the upstream branch `robin-perf-eval` is pretty far behind main right now; I wonder if that may be easy to do once `robin-perf-eval` is brought up-to-date with those commits.

docs/src/pages/guides/FHIRSearchConfiguration.md
lmsurpre: I am happy with `indexIds`...this is WAY better than `logicalResourceIds` :-)

lmsurpre (approved):
LGTM.  The test failures are expected until #2490 is merged.

operation/fhir-operation-reindex/src/main/resources/retrieve-index.json
```diff
@@ -0,0 +1,53 @@
...
+	"url": "http://ibm.com/fhir/OperationDefinition/retrieve-index",
+	"name": "RetrieveIndex",
+	"status": "draft",
+	"kind": "operation",
+	"publisher": "IBM FHIR Server",
+	"date": "2021-06-11",
+	"description": "The retrieve-index operation retrieves the index IDs from the database.",
+	"code": "retrieve-index",
+	"system": true,
+	"type": false,
```
lmsurpre, line 17: If we support this one at the type level, that would allow us to add selective reindexing to the fhir-bucket client:  https://github.com/IBM/FHIR/issues/2525

In #2489 we did the same thing for the `$reindex` operation (i.e. made it work on the type level in addition to the system level).

tbieste: Yeah, that makes sense.  I can add `$retrieve-index` support at the type level.

tbieste: Added.

operation/fhir-operation-reindex/src/main/java/com/ibm/fhir/operation/reindex/ReindexOperation.java
```diff
@@ -148,8 +148,9 @@ protected Parameters doInvoke(FHIROperationContext operationContext, Class<? ext
                             String[] parts = resourceLogicalId.split("/");
                             rt = parts[0];
                         }
+                        // Check resource type
                         if (!ModelSupport.isConcreteResourceType(rt)) {
-                            throw FHIROperationUtil.buildExceptionWithIssue("ResourceType is not valid", IssueType.INVALID);
+                            throw FHIROperationUtil.buildExceptionWithIssue("Resource type '" + rt + "' is not valid", IssueType.INVALID);
                         }
```
lmsurpre, line 154: is it safe to remove this now that we're checking it in AbstractOperation?

lmsurpre: never mind, forgot that we still allow users to pass it in via that logicalResourceId param

docs/src/pages/guides/FHIRSearchConfiguration.md
```diff
@@ -265,62 +265,43 @@ Reindexing is implemented as a custom operation that tells the IBM FHIR Server t
...
-            "diagnostics": "Processed Patient/1795df2b501-04f88a35-9f2f-4871-a05e-ba8090fa18f5"
-        }
-    ]
-}
-```
+### 2.2 Client-side-driven approach
+Another option is to have the client utilize the `$retrieve-index` and `$reindex` in parallel to drive the processing.
 
-An example response when processing is complete: 
+The `$retrieve-index` operation is called to retrieve index IDs of resources available for reindexing. This can be done repeated by using the `_count` and `afterIndexId` parameters for pagination. The operation supports the following parameters to control the behavior:
```
punktilious, line 283: repeatedly

tbieste: Fixed.

fhir-bucket/README.md
```diff
@@ -355,13 +355,15 @@ java.util.logging.FileHandler.pattern=fhirbucket-%u-%g.log
...
 
+With client-side-driven, the fhir-bucket will repeatedly call two operations repeatedly in parallel; the `$retrieve-index` operation to determine the list of resources available to reindex, and the `$reindex` operation with a list of resources to reindex. Driving the reindex this way avoids database contention associated with updating the reindex timestamp of each resource with reindex "marker", which is used by the server-side-driven approach to keep track of the next resource to reindex.
```
punktilious, line 362: repeatedly is repeated

fhir-bucket/README.md
```diff
@@ -355,13 +355,15 @@ java.util.logging.FileHandler.pattern=fhirbucket-%u-%g.log
...
 
-The fhir-bucket main app has been extended to support driving a reindex operation with high concurrency. Processing will stop once the FHIR server returns the OperationOutcome with an issue diagnostic value "Reindex complete".
+To avoid read timeouts, the number of resources processed in a single reindex call can be limited. Reindex calls can be made in parallel to increase throughput. The best number for concurrent requests depends on the capabilities of the underlying platform and any desire to balance load with other users. Concurrency up to 200 threads have been tested. Monitor the IBM FHIR Server response times when increasing concurrency. Also, make sure that the connection pool configured in the FHIR server cluster can support the required number of threads. This also means that the database needs to be configured to support this number of connections (sessions) plus any overhead.
 
+The fhir-bucket main app has been extended to support driving a reindex operation with high concurrency.
```
punktilious, line 366: Do we want to mention when to stop client-driven reindexing? This is only relevant if ingestion is started again after an upgrade/config update but before reindex has completed. The newly ingested resources won't need to be reindexed, so we could stop earlier. Perhaps there should be an option for an "end-stop" timestamp with fhir-bucket. It's perhaps not critical, unless the new ingestion workload is high and you are trying to get reindex to complete as quickly as possible.

fhir-bucket/src/main/java/com/ibm/fhir/bucket/reindex/ClientDrivenReindexOperation.java
```diff
@@ -0,0 +1,359 @@
...
+            logger.severe("FHIR Server retrieve-index operation returned an error: " + response.getStatusCode() + " " + response.getStatusMessage());
+        }
+
+        return result;
+    }
+
+    /**
+     * Extract the index IDs from the retrieve-index operation output.
+     * @param output the retrieve-index operation output
+     * @return true if index IDs were found, otherwise false
```
punktilious, line 252: Not sure this return comment lines up with the logic in the method. Not sure I understand how queued and active are working together here. I'd like to walk through this. I'd like to understand what happens when ids are rejected by the blockingQueue.

fhir-bucket/src/main/java/com/ibm/fhir/bucket/reindex/ClientDrivenReindexOperation.java
```diff
@@ -0,0 +1,359 @@
...
+                    // So probably this will be caught first time monitorLoop calls callOnce and not here.
+                    logger.severe("IllegalStateException caught. FHIR client thread will exit." + x.toString() );
+                }
+                if (!ok) {
+                    // stop everything on the first failure
+                    this.active = false;
+                }
+            } else {
+                // queue is empty, wait a partial second and try again
+                safeSleep(100);
```
punktilious, line 313: This could limit throughput if threads are frequently sleeping waiting for new work. Better to use locks and conditions.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/ReindexResourceDAO.java
```diff
@@ -51,6 +51,13 @@
     // Note that currently the global logical_resources table does not carry
```
punktilious, line 51: The is_deleted flag is now available on the LOGICAL_RESOURCES table.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/RetrieveIndexDAO.java
```diff
@@ -0,0 +1,118 @@
...
+        this.notModifiedAfter = notModifiedAfter;
+    }
+
+    /**
+     * Run the DAO command on the database connection.
+     * @param c connection
+     * @return list of logical resource IDs
+     * @throws FHIRPersistenceException
+     */
+    public List<Long> run(Connection c) throws FHIRPersistenceException {
```
punktilious, line 63: To avoid any odd behavior by the optimizer, we should use the cached resourceTypeId value and give it as a literal instead of joining against the resource_types table. It will be slightly quicker.

punktilious (approved):
LGTM if we either address the performance comments or create issues

fhir-bucket/README.md
tbieste: Fixed.

fhir-bucket/README.md
tbieste: Since the "notModifiedAfter" parameter is set to the timestamp the reindex is started, any newly ingested resource won't get reindex; so that should already be handled.

fhir-bucket/src/main/java/com/ibm/fhir/bucket/reindex/ClientDrivenReindexOperation.java
tbieste: If the IDs are rejected by the blockingQueue after waiting up to 10 seconds, then "queued = false", so `lastIndexId` is not updated.  So the next time retrieveIndex is called it calls right where it last left off.  I can change OFFER_TIMEOUT_IN_SEC to a bigger number (i.e. 30 seconds), I just didn't want it too high so if signalStop() is called while the queue is full, it doesn't take so long to shut everything it down.

fhir-bucket/src/main/java/com/ibm/fhir/bucket/reindex/ClientDrivenReindexOperation.java
tbieste: Replaced this with a call to blockingQueue.poll(...), so it doesn't use this arbitrary sleep to wait to check if the queue is still empty.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/ReindexResourceDAO.java
tbieste: Updated.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/RetrieveIndexDAO.java
tbieste: Updated.

fhir-bucket/src/main/java/com/ibm/fhir/bucket/reindex/ClientDrivenReindexOperation.java
tbieste: The return true, is just for if "active" is false (i.e. the worker threads are not active and consuming work), we don't want to just keep trying to enqueue more work, but rather return back to the monitorLoop() to start up the worker threads (or shutdown if running is false).

docs/src/pages/guides/FHIRSearchConfiguration.md
tbieste: Changed merged.

tbieste:
I think I have all the comments addressed.  Ready to merge when we have a clean build.

Issue #2155, opened 2021-03-26
Improve reindex performance by skipping resources with no search parameter changes


punktilious:
The `$reindex` operation reads every resource, extracts the search parameters, deletes the current search parameters from the database and inserts the new set.

This is expensive and unnecessary if the new search parameter set is identical to the search parameters currently stored. However, reading the existing search parameters for comparison would also be expensive and non-trivial. A possible solution is to store a fingerprint (hash) of the parameter set and use this as a comparison. If the stored hash matches the hash computed from the newly extracted parameter set then the delete/insert step can be skipped.

This optimization is also applicable to new versions of a resource.

Care is needed when computing the hash to make sure it is deterministic - meaning that the parameters must be sorted first before the hash is computed.

A hash such as SHA-256 is statistically reliable for such a feature.

A new column is required on the xx_logical_resources table. No data migration is needed because a default value of null is suitable. The next time a reindex is performed, the hash will be computed and stored. Future reindex operations will then benefit. New resources will be stored with the hash and therefore immediately benefit.

lmsurpre:
Could be combined with a similar schema-change needed for #2417

lmsurpre:
Need to make sure we consider the impact of global search parameters as well.
Just using the name + value may not be enough in the hash (if we move the target of where they are stored)

lmsurpre:
should be done in conjunction with https://github.com/IBM/FHIR/issues/1751 to ensure good e2e test coverage.

tbieste:
After looking over the code involved, here's an overall design came up with:

DB changes
- Add a new column to the LOGICAL_RESOURCES table (32-byte VARBINARY column) for a SHA256 value.
- Update add_any_resource stored procedure to populate new column.

Class changes
- Add an "existing search parameters hash" field to the ResourceIndexRecord class; which is populated from the value of a new column in the LOGICAL_RESOURCES table.
- Add a toSearchParameterHash method to ExtractedParameterValue.  This would get the SHA256 hash of the "searchParameterCode|searchParameterUri|searchParameterVersion|extractedValue", with extractedValue implemented appropriately for each ParameterType (token, composite, string, etc).

Logic changes after calling extractSearchParameters(...)
- Call toSearchParameterHash on each ExtractedParameterValue to get a list of hashes.
- Sort the list of hashes using normal string sorting.
- Hash the toString of the ExtractedParameterValue list to obtain the "new search parameters hash" for the resource.
- If reindex:
[...]

PR #2524, opened 2021-06-16
Issue #2155 - Use hash to determine if search parameters need update

tbieste:
Issue #2155 - Use hash to determine if search parameters need update


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/ReindexResourceDAO.java
```diff
@@ -294,6 +296,33 @@ public void updateParameters(String tablePrefix, List<ExtractedParameterValue> p
...
+     */
+    protected void updateParameterHash(Connection conn, long logicalResourceId, String parameterHashB64) throws SQLException {
+        final String SQL = "UPDATE logical_resources SET parameter_hash = ? WHERE logical_resource_id = ?";
+        try (PreparedStatement stmt = conn.prepareStatement(SQL)) {
+            // bind parameters
+            stmt.setString(1, parameterHashB64);
+            stmt.setLong(2, logicalResourceId);
+            stmt.executeUpdate();
+            if (logger.isLoggable(Level.FINEST)) {
+                logger.finest("Update parameter_hash [" + parameterHashB64 + "] for logicalResourceId [" + logicalResourceId + "]");
```
punktilious, line 321: Might be useful to add [took = xx ms] timing information for this update. This would make it easier to see if this update slows down over time (due to the effect of tombstones in PostgreSQL, until vacuum cleans things up). I don't think it will (because there are no indexes being changed here)...but individual timing in the finest message would be helpful.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dto/DateParmVal.java
```diff
@@ -50,13 +52,22 @@ public void setValueDateEnd(Timestamp valueDateEnd) {
...
      * We know our type, so we can call the correct method on the visitor
      */
+    @Override
     public void accept(ExtractedParameterValueVisitor visitor) throws FHIRPersistenceException {
         visitor.visit(this);
     }
 
+    @Override
+    public String getHash(ParameterHashUtil parameterHashUtil) {
+        StringBuilder sb = new StringBuilder();
```
punktilious, line 62: I don't think we should be using strings to build the hash. Rendering these values as strings is going to consume more CPU than necessary. It should be possible to accumulate the hash directly from the raw Java types.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dto/LocationParmVal.java
```diff
@@ -42,7 +45,16 @@ public void setValueLatitude(Double valueLatitude) {
...
      * We know our type, so we can call the correct method on the visitor
      */
+    @Override
     public void accept(ExtractedParameterValueVisitor visitor) throws FHIRPersistenceException {
         visitor.visit(this);
     }
+
+    @Override
+    public String getHash(ParameterHashUtil parameterHashUtil) {
+        StringBuilder sb = new StringBuilder();
```
punktilious, line 55: Same as previous comment. Compute hash from raw values instead of rendering as a string first.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -2031,79 +2101,34 @@ private TransactionSynchronizationRegistry getTrxSynchRegistry() throws FHIRPers
...
+                addCompartmentParams(extractedParameters.getParameters(), fhirResource);
             }
 
-            // Augment the extracted parameter list with special values we use to represent compartment relationships.
-            // These references are stored as tokens and are used by the search query builder
-            // for compartment-based searches
-            addCompartmentParams(allParameters, fhirResource);
+            // Generate the hash which is used to quickly determine whether the extracted parameters
+            // are different than the extracted parameters that currently exist in the database
+            extractedParameters.generateHash(parameterHashUtil);
```
punktilious, line 2126: Perhaps ExtractedSearchParameters should be immutable like a DTO. Just construct one at the end with the list of parameters and the computed hash. I don't think ExtractedSearchParameters should be in the business of computing the hash itself, just holding it.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/ParameterHashUtil.java
```diff
@@ -0,0 +1,87 @@
...
+     * @param parameters extracted search parameters
+     * @return the Base64-encoded SHA-256 hash
+     */
+    public String getParametersHash(List<ExtractedParameterValue> parameterValues) {
+        // Sort hashes to make it deterministic
+        List<String> sortedList = new ArrayList<>(parameterValues.size());
+        for (ExtractedParameterValue parameterValue : parameterValues) {
+            sortedList.add(Objects.toString(parameterValue.getHash(this), ""));
+        }
+        sortedList.sort(Comparator.comparing(String::toString));
```
punktilious, line 55: I think this algorithm is incorrect. This is sorting on the values of each parameter. The list should be sorted on the parameter names first, and then compute the hash. The current implementation may compute the same hash for parameter sets which are different.

fhir-persistence-jdbc/src/test/java/com/ibm/fhir/persistence/jdbc/test/util/ParameterHashUtilTest.java
```diff
@@ -0,0 +1,189 @@
...
+import com.ibm.fhir.persistence.jdbc.util.ParameterHashUtil;
+import com.ibm.fhir.persistence.jdbc.util.type.NumberParmBehaviorUtil;
+import com.ibm.fhir.search.date.DateTimeHandler;
+import com.ibm.fhir.search.util.ReferenceValue;
+import com.ibm.fhir.search.util.ReferenceValue.ReferenceType;
+
+/**
+ * Utility to do testing of the parameter hash utility.
+ */
+public class ParameterHashUtilTest {
```
punktilious, line 39: Need to add a test case where a pair of values is swapped between two parameters.

param1: value1
param2: value2

compare with:

param1: value2
param2: value1

The computed hash should be different (but with the current implementation I believe the hashes will come out the same).

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/ReindexResourceDAO.java
tbieste: Updated.

fhir-persistence-jdbc/src/test/java/com/ibm/fhir/persistence/jdbc/test/util/ParameterHashUtilTest.java
tbieste: I added testcase testExtractedParametersSwappedValues(), which shows that the values do come out different, since the parameter (name/code) is part of the hash.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/ParameterHashUtil.java
tbieste: The hash is computed from the name/version/url + value, so I thought that would be ok.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dto/DateParmVal.java
tbieste: I think I need more details on exactly how you'd like me to generate the hash.  Can walk through it, so I can understand better.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
tbieste: Updated to be immutable.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dto/LocationParmVal.java
tbieste: Updated to compute hash from raw values.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dto/DateParmVal.java
tbieste: No longer using strings to build the hash.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/ParameterHashUtil.java
tbieste: Updated to generate the hash faster by using a visitor.

fhir-persistence-jdbc/src/test/java/com/ibm/fhir/persistence/jdbc/test/util/ParameterHashUtilTest.java
tbieste: Testcase updated after sorting and using a visitor to generate the hash.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dto/ExtractedParameterValue.java
```diff
@@ -93,4 +80,99 @@ public String getName() {
...
+            if (retVal != 0) {
+                return retVal;
+            }
+        }
+        String thisName = this.getName();
+        String otherName = o.getName();
+        if (thisName != null || otherName != null) {
+            if (thisName == null) {
+                return -1;
+            } else if (otherName == null) {
```
punktilious, line 133: names can never be null (because we need the name in order to store the value in the database. The null checks are therefore a little redundant. If we want to assert them as non-null, we should do so in the constructor.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/ExtractedSearchParameters.java
```diff
@@ -0,0 +1,44 @@
...
+
+import com.ibm.fhir.persistence.jdbc.dto.ExtractedParameterValue;
+
+/**
+ * Contains a list of extracted search parameters and a Base64-encoded hash.
+ */
+public class ExtractedSearchParameters {
+
+    private List<ExtractedParameterValue> parameters = new ArrayList<>();
+    private String parameterHashB64 = null;
```
punktilious, line 22: final

punktilious (approved):
LGTM, just a couple of minor comments/suggestions.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/ExtractedSearchParameters.java
tbieste: Updated.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dto/ExtractedParameterValue.java
tbieste: That is true.  Since this is a compareTo method, I just wanted to keep it generic and use the same logic for each field for consistency, and not have to care which fields are nullable, and which are not.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/ParameterHashVisitor.java
```diff
@@ -0,0 +1,216 @@
...
+     * Updates the digest with the BigDecimal.
+     * @param value the BigDecimal
+     */
+    private void updateDigestWithBigDecimal(BigDecimal value) {
+        if (value != null) {
+            ByteBuffer bb = ByteBuffer.allocate(Double.SIZE);
+            bb.putDouble(value.doubleValue());
+            digest.update(bb);
+        }
+        digest.update(DELIMITER);
```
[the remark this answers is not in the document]
tbieste: I added a testcase to cover a situation where I felt they were needed.

Issue #2528, opened 2021-06-18
Sum of age groups doesn't match with the result from Count


MeloTheMelon:
Hey,
I just made several search queries on my FHIR server:
[base]\Patient?_has:Condition:patient:code=C49.0&_summary=count 
=> Result: 1500

[base]\Patient?_has:Condition:patient:code=C49.0&birthdate=gt1961&_summary=count
=> Result: 369
[base]\Patient?_has:Condition:patient:code=C49.0&birthdate=le1961&_summary=count
=> Result: 1255
=> Total: 1624

The total of both queries should result in the same as the general count, right?
If I split the same search by gender the numbers work out again.
Is there something I'm doing wrong with the search?

I'm using version 4.7.1

lmsurpre:
> The total of both queries should result in the same as the general count, right?

Not necessarily.  We use the range definitioins of `gt` and `le` as defined at https://www.hl7.org/fhir/search.html#prefix

Try changing your `gt1961` to `sa1961`.  The `sa` is for "starts after" and I believe this one has the semantics that your after.  Let us know how it goes.

lmsurpre:
Actually, let me have a look deeper look at this to see if we have something wrong here.  The only real overlap should be from target values that have a range that spans across the upper bound two search and in the case of a birthday we really shouldn't have that.

MeloTheMelon:
I did another set of tests on a different subsection of my patients.

general count to have a reference value
[base]\Patient?_has:Condition:patient:code=C49.0&_summary=count
Result: 1495

gt1800 to check if it counts any patients that might not have a birthdate set.
[base]\Patient?_has:Condition:patient:code=C49.0&birthdate=gt1800&_summary=count
Result 1495 => All patients seem to have a birth year

Now I did one lt1961, le1961, gt1961, ge1961 and =1961.

[base]\Patient?_has:Condition:patient:code=C49.0&birthdate=lt1961&_summary=count
Results: 1167

[base]\Patient?_has:Condition:patient:code=C49.0&birthdate=le1961&_summary=count
Results: 1167

[base]\Patient?_has:Condition:patient:code=C49.0&birthdate=gt1961&_summary=count
Results: 364

[base]\Patient?_has:Condition:patient:code=C49.0&birthdate=ge1961&_summary=count
Results: 364

[base]\Patient?_has:Condition:patient:code=C49.0&birthdate=1961&_summary=count
Results: 36

[...]

PR #2531, opened 2021-06-18
issue #2528 - change interpretation of lt and gt for date searches

lmsurpre:
We use the range interpretation of the search prefixes for both date and
number/quantity searches.  The spec says:
* `lt` means "the range below the search value intersects (i.e. overlaps) with the
range of the target value"
* `gt` means "the range above the search value intersects (i.e. overlaps) with the
range of the target value"

For numbers, the spec additionally says:
* When a comparison prefix in the set gt, lt, ge, le, sa & eb is provided, the implicit precision of the number is ignored, and they are treated as if they have arbitrarily high precision

Previously, we interpreted date search in the same way and so a search like `gt2016`
would return any value above `2016-01-01 00:00:00` in the server's
timezone.  However, we were not consistent with this because, for some reason,
we decided that `lt2016` should return any value below `2017-01-01 00:00:00`.

However, even if we properly implemented this "arbitrarily high precision" clause for date search,
this interpretation leads to confusing and unintuitive results like `gt2016` returning dates in 2016.
Instead, we will now interpret the "search value" of 2016 to be the
range `[2016-01-01 00:00:00, 2017-01-01 00:00:00)` and therefor a search for `gt2016` will
only find target values above 2017-01-01 00:00:00 (inclusive).

### DIFF
--- a/fhir-persistence/src/test/resources/logging.unitTest.properties
+++ b/fhir-persistence/src/test/resources/logging.unitTest.properties
@@ -1,5 +1,5 @@
 # ========================================================================
-#  (C) Copyright IBM Corp. 2000, 2019
+#  (C) Copyright IBM Corp. 2016, 2021
 #
 #  SPDX-License-Identifier: Apache-2.0
 # ========================================================================
@@ -49,5 +49,6 @@ java.util.logging.ConsoleHandler.formatter = java.util.logging.SimpleFormatter
 ############################################################
 
 # Set this to FINE or higher to output the SQL statements and related debug info
+com.ibm.fhir.database.level = INFO
 com.ibm.fhir.persistence.level = INFO
 com.ibm.fhir.search.level = INFO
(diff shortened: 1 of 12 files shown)


prb112:
Should the new  builder util be updated with this change?

fhir-persistence/src/test/resources/logging.unitTest.properties
```diff
@@ -49,5 +49,6 @@ java.util.logging.ConsoleHandler.formatter = java.util.logging.SimpleFormatter
 ############################################################
```
tbieste, line 49: Update copyright date

tbieste (approved):

fhir-persistence/src/test/resources/logging.unitTest.properties
lmsurpre: lol, this one has a beginning copyright of 2000... should we change that to 2016 when our project started?

tbieste: Sounds good to me.

JohnTimm (approved):
LGTM

lmsurpre:
> Should the new builder util be updated with this change?

I reworked the changes so that its now contained in the persistence layer and I updated both the old and new behavior classes

Issue #2503, opened 2021-06-14
Support $reindex operation supports `resourceLogicalId` on Type and Instance


prb112:
**Is your feature request related to a problem? Please describe.**
The $reindex operation supports `resourceLogicalId`. 

|`resourceLogicalId`|string|The ResourceType or the ResourceType/Logical id for targetted reindexing|

```
POST [base]/$reindex would reindex resources of any type
POST [base]/Patient/$reindex would reindex Patient resources
POST [base]/Patient/123/$reindex would reindex just Patient/123
```

It'd be nice to support reindex on the specific Resource and Type.

**Describe the solution you'd like**
Update the Reindex Operation to support the Input Parameter.

**Describe alternatives you've considered**

**Acceptance Criteria**

1.  
   GIVEN [a precondition]
    AND [another precondition]
   WHEN [test step]
    AND [test step]
   THEN [verification step]
    AND [verification step]

**Additional context**

tbieste:
Verified:

- POST [base]/$reindex would reindex resources of any type **[Success]**
- POST [base]/Patient/$reindex would reindex Patient resources **[Success]**
- POST [base]/Patient/123/$reindex would reindex just Patient/123 **[Success]**
- Using both Type/Instance and the`resourceLogicalId` parameter results in the following error: `resourceLogicalId` already specified using call to Operation on Type or Instance` **Success**
- Specifying `resourceLogicalId` with ResourceType of `Resource` or `DomainResource` results in the following error: `ResourceType is not valid` **Success**
- Specifying `[base]/Resource/$reindex` or `[base]/DomainResource/$reindex` results in the following error: `ResourceType is not valid` **Results in no error, but for consistency, probably should**

tbieste:
Verified:

- Specifying [base]/Resource/$reindex or [base]/DomainResource/$reindex results in the following error:`Resource Type 'Resource' is not valid`

Issue #2526, opened 2021-06-17
Deprecate the /Resource and /DomainResource endpoints


lmsurpre:
In our DSTU2 implementation, we allowed a user to search on `[base]/Resource` to invoke whole-system search.

In #261 we made it possible to search via GET on `[base]/` or by POST on `[base]/_search`.
However, we never actually removed the ability to interact with the /Resource and /DomainResource endpoints.

We should remove the /DomainResource and document the /Resource endpoint as deprecated.  Once complete, we should open a separate task for removing the Resource endpoint as well.

Potentially relates to https://github.com/IBM/FHIR/issues/2445

lmsurpre:
I decided to deprecate these endpoints for 4.9.0 which basically just means:
1. logging a warning when we serve requests at these endpoints 
2. adding this to the release notes under the *deprecated* section

michaelwschroeder:
Running with the latest code from the main branch on my local system, I executed the following search requests:
`GET [base]/Resource?_id=test`
`GET [base]/DomainResource?_id=test`
`POST [base]/Resource/_search?_id=test`
`POST [base]/DomainResource/_search?_id=test`

In all cases, I verified that one of the following two WARNING messages was issued as appropriate:
`[WARNING ] Use of abstract resource types like 'Resource' in FHIR URLs is deprecated and will be removed in a future release`
`[WARNING ] Use of abstract resource types like 'DomainResource' in FHIR URLs is deprecated and will be removed in a future release`

Issue verified and closed.

PR #2534, opened 2021-06-21
Issue #2503 - Check resource type on $reindex operation

tbieste:
Issue #2503 - Check resource type on $reindex operation

### DIFF
--- a/fhir-server-test/src/test/java/com/ibm/fhir/server/test/operation/ReindexOperationTest.java
+++ b/fhir-server-test/src/test/java/com/ibm/fhir/server/test/operation/ReindexOperationTest.java
@@ -84,6 +84,31 @@ public class ReindexOperationTest extends FHIRServerTestBase {
     }
 
     @Test(groups = { "reindex" })
+    public void testReindexWithInvalidType() {
+        List<Parameter> parameters = new ArrayList<>();
+        parameters.add(Parameter.builder()
+            .name(string("resourceCount"))
+            .value(of(5))
+            .build());
+
+        Parameters.Builder builder = Parameters.builder();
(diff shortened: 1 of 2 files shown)


lmsurpre:
Its funny, I actually recently opened https://github.com/IBM/FHIR/issues/2526 with a very similar observation.
I'm totally fine with the proposed change but I think we should more wholistically stop allowing requests to the `/Resource` and `/DomainResources`, not just for this one operation.

prb112 (approved):

prb112:
Question: Should this be pushed down one level to AbstractOperation.invoke for Parameters, where it can check if the resourceType is abstract?

```
**@Override
    public final Parameters invoke(
            FHIROperationContext operationContext,
            Class<? extends Resource> resourceType,
            String logicalId, String versionId,
            Parameters parameters,
            FHIRResourceHelpers resourceHelper) throws FHIROperationException {
```

tbieste:
> Question: Should this be pushed down one level to AbstractOperation.invoke for Parameters, where it can check if the resourceType is abstract?
> 
> ```
> **@Override
>     public final Parameters invoke(
>             FHIROperationContext operationContext,
>             Class<? extends Resource> resourceType,
>             String logicalId, String versionId,
>             Parameters parameters,
>             FHIRResourceHelpers resourceHelper) throws FHIROperationException {
> ```

I can push the check down one level in Issue #2195, which adds similar checking in a new custom operation; so that would be a good time to make it more common.

lmsurpre:
> I can push the check down one level in Issue #2195, which adds similar checking in a new custom operation; so that would be a good time to make it more common.

Works for me.

Issue #2535, opened 2021-06-21
Under certain conditions, the root cause of a BundledRequestException is never logged


lmsurpre:
**Describe the bug**
In `FHIRRestHelper.processEntriesByMethod` we catch FHIROperationException and, if the request was a transaction bundle (i.e. failFast=true) we construct a generic FHIRRestBundledRequestException with a message of "Error while processing request bundle." and a set of issues from the original FHIROperationException.

This works well if the original FHIROperationException has high-quality OperationOutcomeIssue entries.
However, if the FHIROperationException does not, it results in swallowing the actual root cause.

**Environment**
4.8.3

**To Reproduce**
1. Submit a transaction bundle that will fail with a FHIROperationException with no issues (not sure how yet)

Note that the exception is logged.

**Expected behavior**
Exceptions should be explicitly handled or wrapped and re-thrown, never lost.

**Additional context**

lmsurpre:
the code is now in place for the next time we see this

PR #2536, opened 2021-06-21
issue #2535 - wrap exceptions instead of losing them

lmsurpre:
In addition to copying their OperationOutcomeIssues, we now include the original
exception in the CausedBy of the FHIRRestBundledRequestExceptions created in 
`FHIRRestHelper.processEntriesByMethod`.

This way, if the FHIROperationException was unexpected and has no
corresponding issues, when it gets mapped to a 500 Server Error we will
automatically log the exception from
FHIRResource.exceptionResponse(FHIROperationException, Status).


michaelwschroeder (approved):
LGTM

PR #2548, opened 2021-06-24
Per Testing with Azure Provider and Fix Skippable Logic

prb112:
- Increase the BlobRange size for Read Speed
- Add Preflight Checks for Azure
- Improve handling of the edge of an $Import read window for Azure Blobs
- Add AzureProviderMain tests for driving tests from local

### DIFF
--- a/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/load/ChunkWriter.java
+++ b/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/load/ChunkWriter.java
@@ -289,7 +289,7 @@ public class ChunkWriter extends AbstractItemWriter {
      */
     public OperationOutcome conditionalFingerprintUpdate(ImportTransientUserData chunkData, boolean skip, Map<String, SaltHash> localCache, FHIRPersistence persistence, FHIRPersistenceContext context, String logicalId, Resource resource) throws FHIRPersistenceException {
         OperationOutcome oo;
-        if (!skip) {
+        if (skip) {
             // Key is scoped to the ResourceType.
             String key = resourceType + "/" + logicalId;
             SaltHash oldBaseLine = localCache.get(key);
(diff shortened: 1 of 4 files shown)


fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/load/ChunkWriter.java
```diff
@@ -289,7 +289,7 @@ public void writeItems(List<java.lang.Object> arg0) throws Exception {
      */
     public OperationOutcome conditionalFingerprintUpdate(ImportTransientUserData chunkData, boolean skip, Map<String, SaltHash> localCache, FHIRPersistence persistence, FHIRPersistenceContext context, String logicalId, Resource resource) throws FHIRPersistenceException {
         OperationOutcome oo;
-        if (!skip) {
+        if (skip) {
```
lmsurpre, line 292: just a minor change :-P

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/AzureProvider.java
```diff
@@ -230,75 +272,137 @@ public void writeResources(String mediaType, List<ReadResultDTO> dtos) throws Ex
...
+                compLine = previous + line;
+            } else {
+                compLine = line;
+            }
+
+            tmpLine = line;
+            line = reader.readLine();
+
+            try (ByteArrayInputStream bais = new ByteArrayInputStream(compLine.getBytes())) {
+                LOG.fine(() -> "Azure Resource [R] " + compLine);
```
prb112, line 362:
```suggestion
                LOG.finer(() -> "Azure Resource [R] " + compLine);
```

lmsurpre (approved):
The new readResource logic is a bit complicated, but I think it needs to be because the problem of reading windows of bytes into full lines is inherently complicated.  LGTM.

PR #2555, opened 2021-06-24
issue #2550 - add default system for `|<code>` searches

lmsurpre:
When we store a token value without a system, we end up using a special
placeholder system instead of saving NULL.

With this PR, we will now use that system when we search for system-less
codes.


michaelwschroeder (approved):
LGTM

prb112 (approved):
LGTM, this is a much more consistent change.

PR #2573, opened 2021-07-06
minor cleanup

lmsurpre:
`set +x` doesn't do what the comment says and I think its not needed


prb112:
it actually trails off on the comment, it's so suppose to say we set -x to debug

prb112 (approved):
LGTM (however, please see my comment)

PR #2584, opened 2021-07-07
issue #2526 - deprecate the /Resource and /DomainResource endpoints

lmsurpre:
Because we use "Resource" as the resource type for whole-system search
and whole-system history interactions, it wasn't as simple as just
checking for abstract types in FHIRRestHelper.validateInteraction().
Instead, I now check that the type in the request path is valid directly
from the JAX-RS endpoints / bundle entry processing.
This has the additional benefit of catching URL path errors up front,
avoiding almost all work associated with this type of bad request.

### DIFF
--- a/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
@@ -174,7 +174,7 @@ public class FHIRRestHelper implements FHIRResourceHelpers {
         log.entering(this.getClass().getName(), "doCreate");
 
         // Validate that interaction is allowed for given resource type
-        validateInteraction(Interaction.CREATE.value(), type);
+        validateInteraction(Interaction.CREATE, type);
 
         FHIRRestOperationResponse ior = new FHIRRestOperationResponse();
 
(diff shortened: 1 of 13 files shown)


fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -3221,8 +3236,8 @@ public int doReindex(FHIROperationContext operationContext, OperationOutcome.Bui
      *            the resource type against which the interaction is to be performed
      * @throws FHIROperationException
      */
-    private void validateInteraction(String interaction, String resourceType) throws FHIROperationException {
-        List<String> interactions = null;
+    private void validateInteraction(Interaction interaction, String resourceType) throws FHIROperationException {
```
lmsurpre, line 3239: I originally made this update when I was thinking to check the interaction and resource type here.  It proved unnecessary, but I ended up leaving it because I prefer passing around the enum and only converting to strings when we need to.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -1946,21 +1948,25 @@ private Entry processEntryForGet(Entry.Request entryRequest, FHIRUrlParser reque
...
+                checkResourceType(pathTokens[0]);
                 resource = doSearch(pathTokens[0], null, null, queryParams, absoluteUri, null);
             }
         } else if (pathTokens.length == 2) {
             // This is a 'read' request.
+            checkResourceType(pathTokens[0]);
             resource = doRead(pathTokens[0], pathTokens[1], true, false, null).getResource();
         } else if (pathTokens.length == 3) {
             if ("_history".equals(pathTokens[2])) {
                 // This is a 'history' request.
```
michaelwschroeder, line 1960: Don't you need to check the resource type here?

lmsurpre: good catch.  i've added it now.

michaelwschroeder (approved):
LGTM

JohnTimm (approved):
LGTM

PR #2585, opened 2021-07-08
Use the "skippableUpdate" optimization during bulk import #2285

prb112:
- Added an Integration Test for re-import which generated an NPE
- Add test-import-skip.ndjson
- Added an Update to include the fhirpersistenceevent object with the
previous resource
- Update copy-server-config.sh


fhir-server-test/src/test/resources/testdata/import-operation/test-import-skip.ndjson
```diff
@@ -0,0 +1,9 @@
+{"resourceType":"Patient","id":"72b0d93c-93d0-43d9-94a8-d5154ce07152","meta":{"versionId":"1","lastUpdated":"2020-04-17T00:09:36.910431Z"},"name":[{"family":"Doe","given":["import-1"]}],"telecom":[{"system":"phone","value":"555-1122","use":"home"}],"birthDate":"1970-01-01"}
```
lmsurpre, line 1: whats the provenance of this sample data.  just a rehash of what we already have out there?

lmsurpre (approved):
LGTM

fhir-server-test/src/test/resources/testdata/import-operation/test-import-skip.ndjson
prb112: exactly, Dag had one that did an update. I originally created the sample data, and wanted a slight difference on this one

Issue #2574, opened 2021-07-07
Client-driven reindex should be restartable


lmsurpre:
**Is your feature request related to a problem? Please describe.**
A new client-driven reindex strategy was added to fhir-bucket under #2502.
This was designed to avoid a slowdown we had seen when doing long reindex jobs on PostgreSQL (#1822 )
To do so, it avoids updating the last indexed tstamp in the database.

In the current implementation, there is a monitor thread which calls the `$retrieve-index` operation to find the next set of resources to reindex and then stores these in memory.
Unfortunately, this means that if the process exits, then all of this state is lost and the client has no good way to "resume" the reindex.

**Describe the solution you'd like**
`$retrieve-index` returns the indexIds (aka the primary key for a given logical resource) in-order.
That means it should be very simple to restart from where we left off.

At minimum, we could include enough detail in the log message so that a user can start from where they left off by altering their arguments to the program.

Alternatively, it might be nice to store the reindex results in the fhir-bucket database.  This way we could better track response times and similar metrics over time.

**Describe alternatives you've considered**
fhir-bucket could write to a file instead of a db.  the benefit of this is that it requires less configuration, but the drawback is that it makes running in a docker container/kubernetes job a little trickier because the file needs to be on a volume (or else it will be lost when the process exits).

**Acceptance Criteria**

1.  
   GIVEN [a precondition]
    AND [another precondition]
   WHEN [test step]
    AND [test step]
   THEN [verification step]
    AND [verification step]

**Additional context**

lmsurpre:
looks like we already log the in-flight requests that failed:
```
```

so we're leaning toward this option:
2. introduce a new parameter to fhir-bucket to resume from a particular indexId.  something like `--reindex-start-id`.  since the actual call starts with the indexId AFTER the passed over the URL, we probably want to treat the id that was passed as a number and subtract 1 from it.

lmsurpre:
I had a reindex job that bombed out.  Using the updated documentation, I was able to restart the reindex from a particular indexId (I chose the lowest one that was logged from the failure).  It appears to have picked up where I left off and so I think this is working as expected.

PR #2592, opened 2021-07-09
Issue #2574 - Improve ability to resume from failed reindex

tbieste:
Issue #2574 - Improve ability to resume from failed reindex

### DIFF
--- a/fhir-bucket/src/main/java/com/ibm/fhir/bucket/reindex/ClientDrivenReindexOperation.java
+++ b/fhir-bucket/src/main/java/com/ibm/fhir/bucket/reindex/ClientDrivenReindexOperation.java
@@ -18,6 +18,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingDeque;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
@@ -46,6 +47,7 @@ public class ClientDrivenReindexOperation extends DriveReindexOperation {
     private static final int MAX_RETRIEVE_COUNT = 1000;
     private static final int OFFER_TIMEOUT_IN_SEC = 30;
     private static final int POLL_TIMEOUT_IN_SEC = 5;
+    private static final int MAX_RESTARTS = 10;
     private static final String RETRIEVE_INDEX_URL = "$retrieve-index";
     private static final String REINDEX_URL = "$reindex";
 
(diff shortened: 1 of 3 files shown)


fhir-bucket/src/main/java/com/ibm/fhir/bucket/reindex/ClientDrivenReindexOperation.java
```diff
@@ -298,30 +370,36 @@ private boolean extractIndexIds(Parameters output) {
...
+                        logger.severe("IllegalStateException caught. FHIR client thread will exit." + x.toString() );
+                    }
+                    if (!ok) {
+                        // stop everything on the first failure
+                        this.active = false;
+                    }
                 }
             }
+        } catch (Throwable t) {
+            logger.severe("Throwable caught. FHIR client thread will exit." + t.toString() );
```
lmsurpre, line 397: does Throwable.toString print the stacktrace?
maybe better to do this?
```suggestion
            logger.log("Throwable caught. FHIR client thread will exit.", t);
```

fhir-bucket/src/main/java/com/ibm/fhir/bucket/reindex/ServerDrivenReindexOperation.java
```diff
@@ -190,27 +190,32 @@ public void waitForStop() {
...
+                    // So probably this will be caught first time monitorLoop calls callOnce and not here.
+                    logger.severe("IllegalStateException caught. FHIR client thread will exit." + x.toString() );
+                }
+                if (!ok) {
+                    // stop everything on the first failure
+                    this.active = false;
+                }
             }
+        } catch (Throwable t) {
+            logger.severe("Throwable caught. FHIR client thread will exit." + t.toString() );
```
lmsurpre, line 214: same as above
```suggestion
            logger.log(Level.SEVERE, "Throwable caught. FHIR client thread will exit.", t);
```

fhir-bucket/src/main/java/com/ibm/fhir/bucket/reindex/ClientDrivenReindexOperation.java
```diff
@@ -298,30 +370,36 @@ private boolean extractIndexIds(Parameters output) {
...
-                    // stop everything on the first failure
-                    this.active = false;
+        try {
+            while (this.running && this.active) {
+                String indexIds = getIndexIdsToReindex();
+                if (!indexIds.isEmpty()) {
+                    boolean ok = false;
+                    try {
+                        ok = callReindexOperation(indexIds);
+                    } catch (DataAccessException x) {
```
JohnTimm, line 380: Should we just catch Throwable here and treat any and all exceptional cases the same? This way we don't need these two specific catches, we need only set active = false once, and we don't need the outer try-catch.

fhir-bucket/src/main/java/com/ibm/fhir/bucket/reindex/ClientDrivenReindexOperation.java
tbieste: Updated.

fhir-bucket/src/main/java/com/ibm/fhir/bucket/reindex/ServerDrivenReindexOperation.java
tbieste: Updated.

lmsurpre (approved):
I checked out the branch and ran it against a large (70+ million resource) db.
When the server went offline this morning, the log output looked better and eventually the tool exited with a message like "Reindexing was not fully completed, restart reindex from index ID '19611009' to resume".  LGTM.

fhir-bucket/src/main/java/com/ibm/fhir/bucket/reindex/ClientDrivenReindexOperation.java
JohnTimm: Created and merged separate PR with suggested changes.

JohnTimm (approved):
LGTM

PR #2594, opened 2021-07-09
Introduce EncodingVisitor and remove redundant resource type checking in AbstractOperation

lmsurpre:
Initially, I thought we could use this EncodingVisitor to encode all string 
fields from OperationOutcome resources that we construct (before returning
them to the client).
However, after further discussion, I think we'll tackle that under #2596 by
encoding the content more selectively.

Additionally, I removed some redundant resource type checking from AbstractOperation.
Now that we ensure that the resource type is valid prior to invoking the
operation, there's no need for having this check in AbstractOperation
and therefor also no need to have the PROPNAME_PATH_PARAMETER in the
OperationContext.

I left the PR open because I still think the EncodingVisitor could be generically useful (and
because some of the other cleanup in here still seems worthwhile to me).

### DIFF
--- a/fhir-model/src/test/java/com/ibm/fhir/model/visitor/test/EncodingVisitorTest.java
+++ b/fhir-model/src/test/java/com/ibm/fhir/model/visitor/test/EncodingVisitorTest.java
@@ -1,0 +1,80 @@
+/*
+ * (C) Copyright IBM Corp. 2021
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.ibm.fhir.model.visitor.test;
+
+import static com.ibm.fhir.model.type.String.string;
+import static org.testng.Assert.assertFalse;
+
+import java.io.StringWriter;
+import java.util.Collections;
+
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+import com.ibm.fhir.model.format.Format;
+import com.ibm.fhir.model.generator.FHIRGenerator;
+import com.ibm.fhir.model.resource.OperationOutcome;
+import com.ibm.fhir.model.resource.OperationOutcome.Issue;
+import com.ibm.fhir.model.test.TestUtil;
+import com.ibm.fhir.model.type.CodeableConcept;
+import com.ibm.fhir.model.type.Extension;
+import com.ibm.fhir.model.type.Narrative;
+import com.ibm.fhir.model.type.Xhtml;
+import com.ibm.fhir.model.type.code.NarrativeStatus;
+import com.ibm.fhir.model.type.code.ResourceType;
+import com.ibm.fhir.model.visitor.EncodingVisitor;
+import com.ibm.fhir.model.visitor.EncodingVisitor.EncodingContext;
+
+public class EncodingVisitorTest {
+    private static OperationOutcome oo;
+
+    @BeforeClass
+    public static void setup() throws Exception {
+        oo = TestUtil.getMinimalResource(ResourceType.OPERATION_OUTCOME);
+        Issue issue = oo.getIssue().get(0).toBuilder()
+                .extension(Extension.builder()
+                    .id("\" onload=\"alert('test1')")
+                    .url("http://example.com/evil")
+                    .value(string("<script>alert('powned')</script>"))
+                    .build())
+                .details(CodeableConcept.builder()
+                    .text(string("<body onload=alert('powned2')>"))
+                    .build())
+                .diagnostics(string("<META HTTP-EQUIV=\"refresh\"\n"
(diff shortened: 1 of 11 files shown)


JohnTimm:
Per our discussion, created this related issue:
https://github.com/IBM/FHIR/issues/2596

JohnTimm (approved):
LGTM - though, as discussed, the single quote escaping is quite annoying especially when operation outcome issue messages are generated _without_ user input.

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -457,7 +443,7 @@ protected FHIROperationException buildUnsupportedResourceTypeException(String re
                             .value(Code.of("resource"))
                             .build())
                         .build())
-                .details(CodeableConcept.builder().text(string(Encode.forHtml(msg))).build())
+                .details(CodeableConcept.builder().text(string(msg)).build())
```
prb112, line 446: I'd suggest a slight change...  on 437 Encode.forHtml(resourceTypeName) this is the part that's returning the reported problem.  Otherwise, I think this is a fine change.

fhir-model/src/test/java/com/ibm/fhir/model/visitor/test/EncodingVisitorTest.java
```diff
@@ -0,0 +1,80 @@
...
+                    .status(NarrativeStatus.EMPTY)
+                    .div(Xhtml.from("Narrative text is tested elsewhere"))
+                    .build())
+                .issue(Collections.singleton(issue))
+                .build();
+    }
+
+    @Test
+    public void testEscape() throws Exception {
+        EncodingVisitor<OperationOutcome> v = new EncodingVisitor<>(EncodingContext.HTML_CONTENT);
```
prb112, line 66: Where is EncodingVisitor.java called in the src/main/java? I didn't find the clear spot. it looks like Encode.forHtml is called directly in some spots.

prb112 (commented):
Left two comments.

fhir-model/src/test/java/com/ibm/fhir/model/visitor/test/EncodingVisitorTest.java
lmsurpre: Only one spot.  In exceptionResponse(OperationOutcome, Status) I replace the Encode.forHtml of the diagnostic field with the EncodingVisitor which will encode all non-narrative string content in the resource:  https://github.com/IBM/FHIR/pull/2594/files#diff-9045cda3af019e8ce706c94d0064fce71e63dcf090b7bbd0c3931fb6a62b3066R310

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -308,23 +307,10 @@ protected Response exceptionResponse(OperationOutcome oo, Status status) {
...
-                issues.add(current);
-            }
-        }
+        EncodingVisitor<OperationOutcome> v = new EncodingVisitor<>(EncodingContext.HTML);
+        oo.accept(v);
         return Response.status(status)
-                .entity(oo.toBuilder()
-                    .issue(issues)
-                    .build())
+                .entity(v.getResult())
```
lmsurpre, line 313: this is where we use the EncodingVisitor to encode the entire OperationOutcome.  but i'm not sure that we actually want to do this because I slightly prefer the more targeted approach that we attempt today (where we only encode content that might contain user-provided input).  John opened https://github.com/IBM/FHIR/issues/2596 for an idea we had to possibly make that a little more explicit / force us to think about whether an error message could contain user-provided content each time we construct an OperationOutcomeIssue.

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
lmsurpre: yeah, I think I like that.  the idea behind removing the Encode.forHtml here is that the EncodingVisitor will take care of all the encoding before we return it to the client, but given my other comment, I think I like the idea of backing that out AND making the change you suggest here (encoding only the part that is user-provided).

fhir-model/src/test/java/com/ibm/fhir/model/visitor/test/EncodingVisitorTest.java
lmsurpre: > it looks like Encode.forHtml is called directly in some spots.
Yeah, I left Encode.forHtml in fhir-provider because I think that is outside of the normal FHIRResource.exceptionResponse.  We could potentially use it there as well, but instead I'm leaning toward going with the approach of encoding only the error message that contain user content.
I *think* the fhir-provider error message could contain such context, but I'm not positive.

prb112 (approved):

Issue #2632, opened 2021-07-22
Dockerfile should ensure directories are writable by root group (gid=0) 


tbieste:
**Describe the bug**
The docker files for ibm-fhir-server and ibm-fhir-schema-tool should ensure any directories that need to be written to are writable by root group (gid=0), to allow the container to be run as any UID that has a GID=0.

**Environment**
IBM FHIR Server 4.8.3

**To Reproduce**
1. Run container as UID that is not root or UID 1001, but does have GID=0.
2. Current error from schema-tool container:
/opt/schematool/run.sh: line 104: /opt/schematool/workarea/persistence.json: Permission denied
3. Current error from fhir-server container:
cp: cannot create regular file '/config/configDropins/overrides/myfile.xml': Permission denied

**Expected behavior**
Should not get permission denied errors.

tbieste:
During QA, tried this out and now get a different permission denied error:

> run.sh - [INFO]: 2021-07-28_15:45:55 - creating the schema
> + /opt/java/openjdk/bin/java -jar /opt/schematool/fhir-persistence-schema-4.9.0-SNAPSHOT-cli.jar ...
> + tee out.log
> tee: out.log: Permission denied
> Jul 28, 2021 3:45:57 PM com.ibm.fhir.schema.app.Main main
> SEVERE: schema tool failed
> java.lang.IllegalStateException: Error while initializing log output file. 
> 	at com.ibm.fhir.database.utils.common.LogFormatter.init(LogFormatter.java:92)
> 	at com.ibm.fhir.schema.app.util.CommonUtil.configureLogger(CommonUtil.java:185)
> 	at com.ibm.fhir.schema.app.util.CommonUtil.configureLogger(CommonUtil.java:157)
> 	at com.ibm.fhir.schema.app.Main.main(Main.java:2193)
> Caused by: java.nio.file.AccessDeniedException: ./fhirschema.log.lck
> 	at java.base/sun.nio.fs.UnixException.translateToIOException(Unknown Source)
> 	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(Unknown Source)
> 	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(Unknown Source)
> 	at java.base/sun.nio.fs.UnixFileSystemProvider.newFileChannel(Unknown Source)
> 	at java.base/java.nio.channels.FileChannel.open(Unknown Source)
> 	at java.base/java.nio.channels.FileChannel.open(Unknown Source)
[...]

PR #2643, opened 2021-07-27
Dockerfile should ensure directories are writable by root group (gid=0) #2632

prb112:
I tested the schematool with the restricted profile.  
I tried the  server and the testing didn't set the right restricted profile.  


tbieste (approved):
LGTM

Issue #2586, opened 2021-07-08
Update fhir-ig-us-core to US-Core STU4


prb112:
Update fhir-ig-us-core to US-Core STU4

prb112:
Per https://github.com/IBM/FHIR/issues/2551#issuecomment-934479841 

Q: What we should do with the technical corrections?
A: ... Hybrid approaches are fine... when we hit the technical corrections.
Let's start up with Major.Minor as individual models. (Could also consider stu3, stu4, r4)

fhir-ig-us-core-stu3 (3.1.0, 3.1.1)
fhir-ig-us-core-stu4 (4.0.0, 4.1.0)

Net: Hybrid approach?

Option 1: Multiple IG versions in a single project / jar (as long as you can exclude and set a default)

Embed the version in the package hierarchy hl7/us/core/stu3/package

prb112:
move the examples to src/test/resources with a version level.

PR #2650, opened 2021-07-29
Update fhir-ig-us-core to US-Core STU4 #2586

prb112:
- Move pre-4.9.0 USCore to STU3 Provider
- Add STU4 package and STU4 Provider
- Format All parts of package with Pretty Print
- Add VersionComparatorTool to show differences

- Errors:
	us-core-pulse-oximetry

### DIFF
--- a/conformance/fhir-ig-us-core/src/test/java/com/ibm/fhir/ig/us/core/test/ConstraintGeneratorTest.java
+++ b/conformance/fhir-ig-us-core/src/test/java/com/ibm/fhir/ig/us/core/test/ConstraintGeneratorTest.java
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright IBM Corp. 2019, 2020
+ * (C) Copyright IBM Corp. 2019, 2021
  *
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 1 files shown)


JohnTimm:
Created #2651 to address the issue with `us-core-pulse-oximetry`

JohnTimm:
There are 3 test failures.

Two of the test failures are related to validation of the US Core examples: `Observation-some-day-smoker.json` and `Observation-some-day-smoker.xml`. Both of these examples are missing the `Observation.effective` element which is required by the generated constraint: `effective.where(is(dateTime)).exists()`.

The other test failure is related to a handwritten constraint that is throwing an exception during validation:
```
Provenance.agent.onBehalfOf and expression: ((%resource.agent.who.resolve() is Practitioner) or (%resource.agent.who.resolve() is Device)) implies exists() at path: Bundle.entry[483].resource
```
An exception is thrown because the `is` type specifier requires an input collection that contains, at most, 1 element per the FHIRPath specification:
http://hl7.org/fhirpath/N1/#is-type-specifier

The constraint could be rewritten as:
```
(%resource.agent.where(who.resolve().is(Practitioner)).exists() or %resource.agent.where(who.resolve().is(Device)).exists()) implies exists()
```
or
```
(%resource.agent.who.where(resolve().is(Practitioner)).exists() or %resource.agent.who.where(resolve().is(Device)).exists()) implies exists()
```
I believe this bug in the US Core specification.

Issue #2654, opened 2021-08-02
Reindex of latest code produced a deadlock in some tables. 


d0roppe:
**Describe the bug**
During a reindex a deadlock was hit

**Environment**
Latest code in main on 7/28/2021

**To Reproduce**

**Expected behavior**

**Additional context**
Here are some log entries from the reindex. 
```
Jul 29 01:55:55 wh-cmsiop-cthon-fhir-internal-7dd96bc9f8-wpmfq fhir-internal WARNING Skipping search parameter with id='us-core-patient-identifier'. Found multiple search parameters, 'http://hl7.org/fhir/SearchParameter/Patient-identifier' and 'http://hl7.org/fhir/us/core/SearchParameter/us-core-patient-identifier', for code 'identifier' on resource type 'Patient'; use search parameter filtering to disambiguate.
Jul 29 01:55:55 wh-cmsiop-cthon-fhir-internal-7dd96bc9f8-wpmfq fhir-internal WARNING Skipping search parameter with id='us-core-patient-gender'. Found multiple search parameters, 'http://hl7.org/fhir/SearchParameter/individual-gender' and 'http://hl7.org/fhir/us/core/SearchParameter/us-core-patient-gender', for code 'gender' on resource type 'Patient'; use search parameter filtering to disambiguate.
Jul 29 01:55:55 wh-cmsiop-cthon-fhir-internal-7dd96bc9f8-wpmfq fhir-internal WARNING Code '_id' is defined for Resource and Patient with differing expressions; using Patient
Jul 29 01:55:55 wh-cmsiop-cthon-fhir-internal-7dd96bc9f8-9wqxl fhir-internal SEVERE INSERT INTO common_canonical_values (canonical_id, url)      SELECT nextval('fhirdata.fhir_ref_sequence'), v.name        FROM     (VALUES (CAST(? AS VARCHAR(1024))) ) AS v(name)  ON CONFLICT DO NOTHING 
org.postgresql.util.PSQLException: ERROR: deadlock detected
  Detail: Process 148774 waits for ShareLock on transaction 33290864; blocked by process 172608.
Process 172608 waits for ShareLock on transaction 33290869; blocked by process 148774.
  Hint: See server log for query details.
  Where: while inserting index tuple (0,13) in relation "common_canonical_values"
    at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2553)
[...]

PR #2655, opened 2021-08-02
Reindex of latest code produced a deadlock in some tables #2654

prb112:

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
@@ -2612,7 +2612,7 @@ public class FHIRPersistenceJDBCImpl implements FHIRPersistence, SchemaNameSuppl
             // It's possible this is a deadlock exception, in which case it could be considered retryable
             if (dax.isTransactionRetryable()) {
                 log.log(Level.WARNING, "Retryable error while performing reindex" + (rir != null ? (" of FHIR Resource '" + rir.getResourceType() + "/" + rir.getLogicalId() + "'") : ""), dax);
-                FHIRPersistenceDataAccessException fpx = new FHIRPersistenceDataAccessException("Data access error while performing a reindex operation.");
+                FHIRPersistenceDataAccessException fpx = new FHIRPersistenceDataAccessException("Data access error while performing a reindex operation.", dax);
                 fpx.setTransactionRetryable(true);
                 throw fpx;
             } else {
--- a/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
@@ -3140,8 +3140,8 @@ public class FHIRRestHelper implements FHIRResourceHelpers {
          * @implNote tried divide and conquer and it caused it to try large pass fail, small pass succeed, and therefore chose a small linear pass.
          */
 
-        // Maximum number to retry
-        final int TX_ATTEMPTS = 5;
+        // Maximum attempts to retry across all windows.
+        final int TX_ATTEMPTS = 10;
 
         int result = 0;
         int max = indexIds.size();
@@ -3150,47 +3150,46 @@ public class FHIRRestHelper implements FHIRResourceHelpers {
         int right = max;
 
         int window = 0;
-        while (left < max) {
+        int attempt = 1;
+        while (left < max && attempt <= TX_ATTEMPTS) {
             window++;
             if (log.isLoggable(Level.FINE)) {
-                log.fine("$reindex window[" + window + "] -> left=[" + left + "] right=[" + right + "] max=[" + max + "]");
+                log.fine("$reindex window [" + window + "/" + attempt + "] -> left=[" + left + "] right=[" + right + "] max=[" + max + "]");
             }
 
             boolean backoff = false;
+
             List<Long> subListIndexIds = indexIds.subList(left, right);
 
-            int attempt = 1;
-            while (attempt <= TX_ATTEMPTS && !backoff) {
-                FHIRTransactionHelper txn = new FHIRTransactionHelper(getTransaction());
-                txn.begin();
-                try {
-                    FHIRPersistenceContext persistenceContext = null;
-                    result += persistence.reindex(persistenceContext, operationOutcomeResult, tstamp, subListIndexIds, null);
-                    attempt = TX_ATTEMPTS + 1; // end the retry loop
-                } catch (FHIRPersistenceDataAccessException x) {
-                    // At this point, the transaction is marked for rollback
-                    if (x.isTransactionRetryable()) {
-                        if (x.getCause() instanceof LockException && ((LockException) x.getCause()).isDeadlock()) {
-                            backoff = true;
-                            long wait = RANDOM.nextInt(5000);
-                            log.info("attempt #" + window + "/" + attempt + " failed, retrying transaction, backing off [wait=" + wait + "ms]");
-                            Thread.sleep(wait);
-                        } else {
-                            log.info("attempt #" + window + "/" + attempt + " failed, retrying transaction, not backing off");
-                        }
+            FHIRTransactionHelper txn = new FHIRTransactionHelper(getTransaction());
+            txn.begin();
+            try {
+                FHIRPersistenceContext persistenceContext = null;
+                result += persistence.reindex(persistenceContext, operationOutcomeResult, tstamp, subListIndexIds, null);
+            } catch (FHIRPersistenceDataAccessException x) {
+                // At this point, the transaction is marked for rollback
+                if (x.isTransactionRetryable() && ++attempt <= TX_ATTEMPTS) {
+                    if (x.getCause() instanceof LockException && ((LockException) x.getCause()).isDeadlock()) {
+                        backoff = true;
+                        long wait = RANDOM.nextInt(5000);
+                        log.info("attempt #" + window + "/" + attempt + " failed, retrying transaction, backing off [wait=" + wait + "ms]");
+                        Thread.sleep(wait);
                     } else {
-                        throw x;
+                        log.info("attempt #" + window + "/" + attempt + " failed, retrying transaction, not backing off");
                     }
-                } finally {
-                    txn.end();
+                } else {
+                    throw x;
                 }
-                attempt++;
+            } finally {
+                txn.end();
             }
 
+            // backoff controls how we increment the window.
             if (backoff) {
                 // we're now going to move over one at a time.
                 right = left + 1;
             } else {
+                // we're sliding over by one more
                 left = right;
                 right += 1;
             }
--- a/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
@@ -3095,7 +3095,7 @@ public class FHIRRestHelper implements FHIRResourceHelpers {
             List<Long> indexIds, String resourceLogicalId) throws Exception {
         int result = 0;
         // Since the try logic is slightly different in the code paths, we want to dispatch to separate methods to simplify the logic.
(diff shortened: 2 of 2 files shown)


fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -3087,8 +3091,124 @@ private void setOperationContextProperties(FHIROperationContext operationContext
...
+     * @param operationOutcomeResult
+     * @param tstamp
+     * @param indexIds
+     * @return
+     * @throws Exception
+     */
+    public int doReindexList(OperationOutcome.Builder operationOutcomeResult, Instant tstamp, List<Long> indexIds) throws Exception {
+        /*
+         * How the backoff works...
+         * indexIds[1,2,3,4,5,6,7,8,9,10]
```
lmsurpre, line 3118: seems like it could have just done them one-at-a-time in a single loop, but I think this should be fine and the description here is very good

prb112: Yes, I considered that.  Robin mentioned a sublist, and this how I ended up doing this approach.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -2612,7 +2612,7 @@ public int reindex(FHIRPersistenceContext context, OperationOutcome.Builder oper
             // It's possible this is a deadlock exception, in which case it could be considered retryable
             if (dax.isTransactionRetryable()) {
                 log.log(Level.WARNING, "Retryable error while performing reindex" + (rir != null ? (" of FHIR Resource '" + rir.getResourceType() + "/" + rir.getLogicalId() + "'") : ""), dax);
-                FHIRPersistenceDataAccessException fpx = new FHIRPersistenceDataAccessException("Data access error while performing a reindex operation.");
```
lmsurpre, line 2615: I think that we were purposefully not including the original exception here because we were worried it could contain sensitive data in the SQL, so we were logging it here and then throwing a more generic error.  I think that was a change that @punktilious made a while back and I'm not certain its needed here, so I'll defer to y'all.

lmsurpre: Yeah, I'm thinking we don't need the causedBy here because its part of the WARNING log statement in the previously line

prb112: I need the original exception.

lmsurpre: Oh, do we walk through the causedBy exceptions somewhere?  I didn't see that.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -3087,8 +3091,124 @@ private void setOperationContextProperties(FHIROperationContext operationContext
...
+                    result += persistence.reindex(persistenceContext, operationOutcomeResult, tstamp, subListIndexIds, null);
+                    attempt = TX_ATTEMPTS + 1; // end the retry loop
+                } catch (FHIRPersistenceDataAccessException x) {
+                    // At this point, the transaction is marked for rollback
+                    if (x.isTransactionRetryable()) {
+                        if (x.getCause() instanceof LockException) {
+                            backoff = true;
+                            long wait = 0;
+                            log.fine(() -> "'DeadLock' executing random retry and backing off on processing");
+                            wait = RANDOM.nextInt(5000);
```
michaelwschroeder, line 3172: Don't see where you actually do the wait.

prb112: thanks, fixed in latest local code, just testing now.

prb112: code is pushed

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
prb112: Yes, in the FHIRRestHelper I check to see if the causedby is LockException and is a deadlock in order to execute a backoff.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
prb112: I switched to one-by-one, it turned out the exponential backoff caused a lot of extra processing when it hit a deadlock. I think it's logical as similar data is inserted at similar times, so backing off using one at a time turned out to move the processing right along.

michaelwschroeder (approved):
LGTM

lmsurpre (approved):
LGTM

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -3087,8 +3091,123 @@ private void setOperationContextProperties(FHIROperationContext operationContext
...
+     * @param operationOutcomeResult
+     * @param tstamp
+     * @param indexIds
+     * @return
+     * @throws Exception
+     */
+    public int doReindexList(OperationOutcome.Builder operationOutcomeResult, Instant tstamp, List<Long> indexIds) throws Exception {
+        // If the indexIds are empty, assume zero.
+        if (indexIds.isEmpty()) {
+            return 0;
```
punktilious, line 3118: Does this imply a failure? I think an empty list is probably worthy of an illegal arg exception if it's a case we don't expect. I can't think of a valid reason to call with an empty list.

prb112: Sure, I can switch to that.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -3087,8 +3091,123 @@ private void setOperationContextProperties(FHIROperationContext operationContext
...
+            boolean backoff = false;
+            List<Long> subListIndexIds = indexIds.subList(left, right);
+
+            int attempt = 1;
+            while (attempt <= TX_ATTEMPTS && !backoff) {
+                FHIRTransactionHelper txn = new FHIRTransactionHelper(getTransaction());
+                txn.begin();
+                try {
+                    FHIRPersistenceContext persistenceContext = null;
+                    result += persistence.reindex(persistenceContext, operationOutcomeResult, tstamp, subListIndexIds, null);
```
punktilious, line 3168: I worry that the caller might interpret a non-zero as success of the whole batch. I know the contract says it's the number successfully processed...worth checking

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
prb112: Changed

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
prb112: Right - it'll only pass up the non-zero value when successfully completed.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -3087,8 +3091,123 @@ private void setOperationContextProperties(FHIROperationContext operationContext
...
+     *
+     * @param operationOutcomeResult
+     * @param tstamp
+     * @param indexIds
+     * @return
+     * @throws Exception
+     */
+    public int doReindexList(OperationOutcome.Builder operationOutcomeResult, Instant tstamp, List<Long> indexIds) throws Exception {
+        // If the indexIds are empty or null, then it's not properly formed.
+        if (indexIds == nulll || indexIds.isEmpty()) {
```
tbieste, line 3117: Typo: nulll

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -3087,8 +3091,123 @@ private void setOperationContextProperties(FHIROperationContext operationContext
...
+        final int TX_ATTEMPTS = 5;
+
+        int result = 0;
+        int max = indexIds.size();
+
+        int left = 0;
+        int right = max;
+
+        int window = 0;
+        while (left < max) {
```
punktilious, line 3153: This could end up trying 5 attempts for each item in the list, which probably risks a client timeout if there's something catastrophic going on. I think you can do the left < max and attempt limit in one loop. This puts an upper bound on the amount of time we're likely to spend here.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
prb112: fixed

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
prb112: OK - we can do that, I'll raise it to 10 attempts then.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -3087,8 +3091,123 @@ private void setOperationContextProperties(FHIROperationContext operationContext
...
+            int attempt = 1;
+            while (attempt <= TX_ATTEMPTS && !backoff) {
+                FHIRTransactionHelper txn = new FHIRTransactionHelper(getTransaction());
+                txn.begin();
+                try {
+                    FHIRPersistenceContext persistenceContext = null;
+                    result += persistence.reindex(persistenceContext, operationOutcomeResult, tstamp, subListIndexIds, null);
+                    attempt = TX_ATTEMPTS + 1; // end the retry loop
+                } catch (FHIRPersistenceDataAccessException x) {
+                    // At this point, the transaction is marked for rollback
```
punktilious, line 3171: I wonder if there's a way to easily abstract this algorithm into a separate strategy class which could allow for an easy unit-test?

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -3087,8 +3091,123 @@ private void setOperationContextProperties(FHIROperationContext operationContext
...
+                    } else {
+                        throw x;
+                    }
+                } finally {
+                    txn.end();
+                }
+                attempt++;
+            }
+
+            if (backoff) {
```
tbieste, line 3190: If I'm understanding this right, if the failed attempts for a subset are isTransactionRetryable(), this will continue to try to process more indexIds, instead of giving up, and return the number that were successful, even though some of them failed.  Is that ok?

prb112: they have all failed, and so we switch to processing them one at a time.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
prb112: refactored so it's one loop up to ten attempts... next commit.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
prb112: agreed - future work item? or resolve now?

tbieste (approved):
LGTM

PR #2661, opened 2021-08-04
Issue 2658

prb112:
- Adds script to cache necessary parquet files
- Add format to the Context used in the push of resources
- Move SLF4J to provided as it causes a loop on parquet usage.


lmsurpre (approved):
LGTM

Issue #2666, opened 2021-08-05
non-fast bulk export writes to many files instead of honoring settings


lmsurpre:
**Describe the bug**
I have an bulkdata config with a ibm-cos storageProvider and COS settings as follows:
```
            "cos" : {
              "partUploadTriggerSizeMB": 10,
              "objectSizeThresholdMB": 200,
              "objectResourceCountThreshold": 200000
            },
```

The fast export seems to be working as desired, but when I add a typeFilter it falls back to the search-based export implementation and the result is that the resources are being split across many more COS objects than they should be.

I turned up tracing and it looks like the currentUploadSize is being computed incorrectly:
```
2021-08-05 16:47:32.530 wh-cmsiop-cthon-fhir-internal-799fc96897-65rg7 fhir-internal FINE isReadyToCheckpoint [readyToWrite=false (37343 >= 10485760), overSizeThreshold=false (37343 >= 209715200), overResourceCountThreshold=false (100 >= 200000), end=false (1 >= 328)]
2021-08-05 16:47:32.674 wh-cmsiop-cthon-fhir-internal-799fc96897-65rg7 fhir-internal FINE isReadyToCheckpoint [readyToWrite=false (73858 >= 10485760), overSizeThreshold=false (111201 >= 209715200), overResourceCountThreshold=false (200 >= 200000), end=false (2 >= 328)]
...
2021-08-05 16:47:44.294 wh-cmsiop-cthon-fhir-internal-799fc96897-65rg7 fhir-internal FINE isReadyToCheckpoint [readyToWrite=false (3796495 >= 10485760), overSizeThreshold=false (199361860 >= 209715200), overResourceCountThreshold=false (10400 >= 200000), end=false (104 >= 328)]
2021-08-05 16:47:45.738 wh-cmsiop-cthon-fhir-internal-799fc96897-65rg7 fhir-internal INFO pushFhirJsonsToCos: '3905969' bytes were successfully appended to COS object - '0vzo_7VIZWnkThMov9MmU8AO7JdwylM9jHTEEGd5JD0/Patient_1.ndjson' uploadId='0100017b-1849-57a1-aa48-8cc67e858702'
```

Specifically, note how the "sizeThreshold" bytes are growing faster than the "readyToWrite" bytes even before we've started an upload and cleared the buffer.

**Environment**
Which version of IBM FHIR Server?

**To Reproduce**
[...]

PR #2668, opened 2021-08-06
issue #2666 - fix for currentUploadSize tracking

lmsurpre:
Previously, we added the entire size of the buffer after each page of
results was read. This leads us to think that we have a lot more data
than we actually do. Now we will add only the new bytes.

This change fixes the over-counting but also started causing us to hit transaction timeouts.
It especially happens if you use an `_elements` filter because the search queries take just as long
but the size of the resources is much smaller and so we need many more resources to reach our
partUploadTriggerSizeMB for the checkpoint.

In the short term, it should be possible to either:
A. increase the transaction timeout (so we can gather more bytes before the timeout occurs); and/or
B. decrease the partUploadTriggerSizeMB (our default is 10 but we can go as low as 5 for S3)

In the longer term, we should possibly consolidate this implementation with the "fast export" one,
or at least use a similar approach. That one uses a maxChunkReadTime to decide when to stop reading.
However, that one also has the luxury of being fast and so we should be able to gather enough bytes
for a multiPartUpload for any moderately sized transaction timeout.

### DIFF
--- a/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/export/system/resource/SystemExportResourceHandler.java
+++ b/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/export/system/resource/SystemExportResourceHandler.java
@@ -43,6 +43,8 @@ public class SystemExportResourceHandler {
             throw new Exception(msg);
         }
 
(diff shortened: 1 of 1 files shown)


JohnTimm (approved):
LGTM

Issue #2669, opened 2021-08-09
SMART interceptor should enforce scope permissions up front


lmsurpre:
**Describe the bug**
Because SMART scopes are only enforced on resources on their way *out* of the FHIR server, it is possible for an application with a valid access token to obtain information about the presence (or absence) of resource types for which they have not been granted read access.

**Environment**
Which version of IBM FHIR Server?

**To Reproduce**
1. load a server with some AllergyIntolerance for a patient
2. configure that server with the fhir-smart jar
3. obtain a token with a patient_id claim for that patient but no scope permitting read access to AllergyIntolerancce
4. perform search
5. obtain a token with a patient_id claim for some other patient with no AllergyIntolerance
6. perform search

Not that a search that returns the AllergyIntolerance comes back with a 403 whereas the other comes back with a 200 (with no results).

**Expected behavior**
A client app without read permission for a given resource type should not be able to infer whether or not any such resources exist.

**Additional context**

lmsurpre:
One of the questions I had while implementing this is whether it should apply only to patient compartment resources (e.g. Condition, Observation) or whether it should also apply to non-compartment resources like Practitioner and Organization.

It wasn't super clear to me whether the existing `afterX` methods checked the scope string for these non-patient-compartment resources or not, but after reviewing https://chat.fhir.org/#narrow/stream/179170-smart/topic/Authorization.20for.20transactions/near/213305185 I felt that it would make the most sense to check the scope string for non-patient-compartment resources in the beforeX methods as well.

Therefore, to invoke read/vread/search/history against ANY resource type's endpoint, it is now definitively required to include the corresponding resourceType (or `*`) in the list of granted resource types.  For example, to read an Organization resource, the user must have granted an application one of `patient/Organization.read`, `patient/*.read`, `user/Organization.read`, or `user/*.read`. (however, use of `user/` permissions is still not implemented for some cases.

kmbarton423:
Walked thru the test of this one with Lee using his Alvearie demo environment.  Confirmed appropriate behaviors with and without limited scope when attempting to access (read/search) non-Patient compartment resources.

PR #2671, opened 2021-08-09
issue #2669 - validate request scopes before read, vread, history, and search

lmsurpre:
In addition to checking the scopes for each resource on the way out (via
the afterX methods), we now check that the requested interaction is
permitted by the passed scopes on the way in (the beforeX methods).
Two reasons:
1. We prevent some unnecessary work for invalid requests
2. We prevent leaking info about the presence/absence of resourceTypes
for which a given application has not been granted access

The request scopes are already checked before create, update, and
delete.

### DIFF
--- a/fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
+++ b/fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
@@ -69,25 +69,33 @@ public class AuthzPolicyEnforcementPersistenceInterceptor implements FHIRPersist
 
     @Override
     public void beforeRead(FHIRPersistenceEvent event) throws FHIRPersistenceInterceptorException {
-        enforceDirectPatientAccess(event);
+        String resourceType = event.getFhirResourceType();
+        DecodedJWT jwt = JWT.decode(getAccessToken());
+        checkScopes(resourceType, Permission.READ, getScopesFromToken(jwt));
+        enforceDirectPatientAccess(resourceType, event.getFhirResourceId(), jwt);
     }
 
     @Override
     public void beforeVread(FHIRPersistenceEvent event) throws FHIRPersistenceInterceptorException {
-        enforceDirectPatientAccess(event);
+        String resourceType = event.getFhirResourceType();
+        DecodedJWT jwt = JWT.decode(getAccessToken());
+        checkScopes(resourceType, Permission.READ, getScopesFromToken(jwt));
+        enforceDirectPatientAccess(resourceType, event.getFhirResourceId(), jwt);
     }
 
     @Override
     public void beforeHistory(FHIRPersistenceEvent event) throws FHIRPersistenceInterceptorException {
-        enforceDirectPatientAccess(event);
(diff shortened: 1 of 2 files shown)


JohnTimm (approved):
LGTM

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
```diff
@@ -355,7 +365,34 @@ public void afterSearch(FHIRPersistenceEvent event) throws FHIRPersistenceInterc
...
     /**
-     * Enforce the authorizations granted by the end user in the form of scope strings
+     * Check whether the permissions required for the current interaction are granted by the approved scopes.
+     *
+     * @param resourceType the resource type used in the endpoint of the request (or "Resource" for whole-system interactions)
+     * @param requiredPermission the permission required for the requested interaction
+     * @param approvedScopes a list of SMART scopes associated with the request
+     * @throws FHIRPersistenceInterceptorException if the interaction is not permitted
+     */
+    private void checkScopes(String resourceType, Permission requiredPermission, List<Scope> approvedScopes) throws FHIRPersistenceInterceptorException {
```
michaelwschroeder, line 375: Seems like this does a subset of what the isAllowed() method is doing - wondering if there's a way to factor out the common code into one method?

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
```diff
@@ -355,7 +365,34 @@ public void afterSearch(FHIRPersistenceEvent event) throws FHIRPersistenceInterc
...
+                }
+            }
+        }
+
+        String msg = requiredPermission.value() + " permission for '" + resourceType +
+                "' is not granted by any of the provided scopes: " + approvedScopes;
+        if (log.isLoggable(Level.FINE)) {
+            log.fine(msg);
+        }
+        throw new FHIRPersistenceInterceptorException(msg)
```
michaelwschroeder, line 390: Can you use the same REQUEST_NOT_PERMITTED message that enforce() uses?

lmsurpre: I could have but I actually liked this one more because it gives the end user / app developer a better idea of why their request didn't go through (while still not leaking anything confidential.
I wouldn't mind making the more generic "Requested interaction is not permitted by any of the passed scopes." a little *less* generic, but for now I just focused on adding this new check.

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
lmsurpre: Yeah, I initially started there but in this case we just have to loop until we confirm the interaction is allowed whereas in the other one we organize the scopes for later use.

michaelwschroeder (approved):
LGTM

Issue #772, opened 2020-03-11
Support for searching via canonical references


lmsurpre:
According to the spec, we should support searching canonical references as either a URI or a Reference.

Presently, we extract the Canonical value and store it in our X_STRING_VALUES table which serves both URI and Reference searches.  However, we *don't* handle the `|<version>` part of it and we certainly don't support advanced reference stuff like chaining or includes/revincludes.

From https://www.hl7.org/fhir/search.html#versions:
```
Elements of type canonical may contain a version specific reference, but this version is different in both meaning and format to version specific references that might be found in a Reference:

   <valueSet value="http://hl7.org/fhir/ValueSet/example|3.0"/>
This version is a reference to the business version of the resource.

For canonical references, servers SHOULD support searching by Canonical URLs, and SHOULD support automatically detecting a |[version] portion as part of the search parameter and interpreting that portion as a search on the business version of the target resource. The modifier :below is used with canonical references, to control whether the version is considered in the search. The search:

GET {base]/Observation?definition:below=http:http://acme.com/some-profile
matches all of these element values:

http://acme.com/some-profile|1.0
http://acme.com/some-profile|1.1
http://acme.com/some-profile|2.0
The search:

GET {base]/Observation?definition:below=http:http://acme.com/some-profile|1
matches the first two element values.
```

And from https://www.hl7.org/fhir/search.html#uri
> Note that for uri parameters that refer to the Canonical URLs of the conformance and knowledge resources (e.g. StructureDefinition, ValueSet, PlanDefinition etc), servers SHOULD support searching by canonical references, and SHOULD support automatically detecting a |[version] portion as part of the search parameter, and interpreting that portion as a search on the version.

lmsurpre:
Need to start with unit tests.
Work we did with "common tokens" could conflict with this.

lmsurpre:
A couple related HL7 JIRA issues (with corresponding discussion on chat.fhir.org):
* https://jira.hl7.org/browse/FHIR-30514
* https://jira.hl7.org/browse/FHIR-30697

csandersdev:
As a related test case, please consider the following...

I have Measure resources that refer to Library resources and I would like to retrieve both resource types in a single call as in 
``/Measure?_id=MEASUREIDHERE&_include=Measure:depends-on:Library``. The Measure resource is defined with canonical URL references as in the snippet below.
```json
{
  "library": [
    "http://ibm.com/fhir/wh-cohort/Library/EXM101"
  ],
}
```
[...]

PR #2683, opened 2021-08-13
Issue #772 - implement canonical search

michaelwschroeder:

### DIFF
--- a/docs/src/pages/Conformance.md
+++ b/docs/src/pages/Conformance.md
@@ -272,8 +272,14 @@ The IBM FHIR Server does not consider the `Quantity.comparator` field as part of
 ### Searching on URI
 URI searches on the IBM FHIR Server are case-sensitive with "exact-match" semantics. The `above` and `below` prefixes can be used to perform path-based matching that is based on the `/` delimiter.
 
+There is one exception to the statement above. The `url` search parameter, which is defined in the base FHIR specification on definitional resource types as a URI search parameter, is actually treated as a canonical search parameter of type REFERENCE, as documented in the [FHIR specification](http://build.fhir.org/references.html#canonical). The following section of this document describes how the IBM FHIR Server processes canonical reference searches.
+
 ### Searching on Reference
-Reference searches on the IBM FHIR Server support search on:
+Reference searches on the IBM FHIR Server support search on elements of type Reference and on elements of type canonical.
+
+**Search on Elements of Type Reference:**
+
+Reference searches may be of the following types:
 
 * a relative reference - `1` where it is reflexsively determined to be a subset of possible targets, such as `Patient/1`, `Group/1`
 * a logical reference - `Patient/1` where it is explicitly set
@@ -294,6 +300,19 @@ Elements of type Reference may contain a versioned reference, such as `Patient/1
 * **Revinclude search**: If a resource has a reference that is versioned, and a `_revinclude` search is performed using the element containing the versioned reference, then the resource containing the versioned reference is returned as an `include` resource only if the version specified in the reference is the referenced resource's current version.
     * This is because the IBM FHIR Server will only return the current version of `match` resources in search results. A reference to a non-current version of the resource is not considered to have met the search criteria, thus the resource containing the reference is not considered a valid `include` resource.
     * Example: A `Condition` resource contains a reference of `Patient/123/_history/2` in its `subject` element, and the current version of the `Patient/123` resource is `2`, and a search of `Patient?_revinclude=Condition:subject` is performed. The `Condition` resource will be returned as an `include` resource since the version of the `Patient` resource specified in the `subject` element is the current version of the `Patient` resource. If the current version of the `Patient/123` resource is `3`, then the `Condition` resource will not be returned as an `include` resource in the search results.
+
+**Search on Elements of Type Canonical:**
+
+Canonical searches search against the canonical URL of a resource. This is for resource types that have a defined element `url` which is the URL that always identifies the resource across all contexts of use. The list of resource types that have a canonical URL and are allowed to be the target of a reference to a canonical URL are documented in the [FHIR specification](http://build.fhir.org/references.html#canonical).
+
+Canonical searches are processed following these rules:
+* The search parameter value is a reference to a definitional FHIR resource (of a type from the list documented in the FHIR specification), where the reference is to the resource’s indexed `url` element value.
+    * The reference can optionally contain a version, delimited by the `|` character (i.e. `http://example.org/fhir/ValueSet/123|1.0.0`). The version will be automatically detected and interpreted as a reference to a resource’s business version - the indexed `version` element value (not to be confused with a resource’s `meta.versionId` element).
+* Searches which do not specify a version will match against all resources whose indexed `url` value matches the search parameter value, regardless if the resource has a `version` value specified.
+* Searches which do specify a version will match only against those resources whose indexed `url` and indexed `version` values match the `url` and `version` components of the search parameter value.
+* Canonical search parameters may be used when performing chained, reverse chained (`_has`), `_include`, or `_revinclude` searches. As described above, versioned canonical references act against the business version of a resource, not it’s versionId. The special rules that apply to chaining and include references specifying versions of the form `/_history/xx` do not apply to versioned canonical references. Versions of the form `/_history/xx` are not supported with canonical references.
+* The `:above` and `:below` modifiers are not supported with canonical search parameters.
+* Canonical references will be resolved via indexed search values. They will not be resolved via registry look-ups.
 
 ### Searching on Special Positional Search
 Positional Search uses [UCUM units](https://unitsofmeasure.org/ucum.html) of distance measure along with common variants:
--- a/docs/src/pages/Conformance.md
+++ b/docs/src/pages/Conformance.md
@@ -2,7 +2,7 @@
 layout: post
 title:  Conformance
 description: Notes on the Conformance of the IBM FHIR Server
-date:   2021-05-26
+date:   2021-08-18
 permalink: /conformance/
 ---
 
@@ -272,7 +272,7 @@ The IBM FHIR Server does not consider the `Quantity.comparator` field as part of
 ### Searching on URI
 URI searches on the IBM FHIR Server are case-sensitive with "exact-match" semantics. The `above` and `below` prefixes can be used to perform path-based matching that is based on the `/` delimiter.
 
-There is one exception to the statement above. The `url` search parameter, which is defined in the base FHIR specification on definitional resource types as a URI search parameter, is actually treated as a canonical search parameter of type REFERENCE, as documented in the [FHIR specification](http://build.fhir.org/references.html#canonical). The following section of this document describes how the IBM FHIR Server processes canonical reference searches.
+There is one exception to the statement above. The `url` search parameter, which is defined in the base FHIR specification on definitional resource types as a URI search parameter, is actually treated as a canonical search parameter of type REFERENCE, as documented in the [FHIR specification](http://hl7.org/fhir/R4/references.html#canonical). The following section of this document describes how the IBM FHIR Server processes canonical reference searches.
 
 ### Searching on Reference
 Reference searches on the IBM FHIR Server support search on elements of type Reference and on elements of type canonical.
@@ -281,7 +281,7 @@ Reference searches on the IBM FHIR Server support search on elements of type Ref
 
 Reference searches may be of the following types:
 
-* a relative reference - `1` where it is reflexsively determined to be a subset of possible targets, such as `Patient/1`, `Group/1`
+* a relative reference - `1` where it is reflectively determined to be a subset of possible targets, such as `Patient/1`, `Group/1`
 * a logical reference - `Patient/1` where it is explicitly set
 * uri searches - where it is explicitly searched using a URI on the server, such as `reference=http://example.org/fhir/Patient/123`
 
@@ -303,13 +303,14 @@ Elements of type Reference may contain a versioned reference, such as `Patient/1
 
 **Search on Elements of Type Canonical:**
 
-Canonical searches search against the canonical URL of a resource. This is for resource types that have a defined element `url` which is the URL that always identifies the resource across all contexts of use. The list of resource types that have a canonical URL and are allowed to be the target of a reference to a canonical URL are documented in the [FHIR specification](http://build.fhir.org/references.html#canonical).
+Canonical searches search against elements of type canonical which contain a reference to a canonical URL. The canonical reference is to a resource's defined element `url` which is the URL that always identifies the resource across all contexts of use. The list of resource types that have a canonical URL and are allowed to be the target of a canonical reference are documented in the [FHIR specification](http://hl7.org/fhir/R4/references.html#canonical). A canonical reference can optionally contain a version, delimited by the `|` character (i.e. `http://example.org/fhir/ValueSet/123|1.0.0`). The version is a reference to a resource’s business version - the indexed `version` element value (not to be confused with a resource’s `meta.versionId` element).
+
+When a resource which contains a canonical element is indexed for search, the reference contained in the canonical element will be parsed into the URL part and the version part (if specified). These values will be stored separately and compared against during canonical searches as described below.
 
 Canonical searches are processed following these rules:
-* The search parameter value is a reference to a definitional FHIR resource (of a type from the list documented in the FHIR specification), where the reference is to the resource’s indexed `url` element value.
-    * The reference can optionally contain a version, delimited by the `|` character (i.e. `http://example.org/fhir/ValueSet/123|1.0.0`). The version will be automatically detected and interpreted as a reference to a resource’s business version - the indexed `version` element value (not to be confused with a resource’s `meta.versionId` element).
-* Searches which do not specify a version will match against all resources whose indexed `url` value matches the search parameter value, regardless if the resource has a `version` value specified.
-* Searches which do specify a version will match only against those resources whose indexed `url` and indexed `version` values match the `url` and `version` components of the search parameter value.
+* The search parameter value is a reference to a canonical URL, which may optionally contain a version delimited by the `|` character. If a version is specified, it will be automatically detected and used in the search.
+* Searches which do not specify a version will match against all resources whose indexed `url` portion of the reference matches the search parameter value, regardless if the reference contains a version. For example, assume one Measure resource contains a canonical reference of `http://example.org/fhir/Library/abc` in its `library` field, and another Measure resource contains a canonical reference of `http://example.org/fhir/Library/abc|1.0.0` in its `library` field. If a search parameter value of `http://example.org/fhir/Library/abc` is used in a canonical search against the `library` field, both of the Measure resources will be a match because the `url` portion of the reference is a match for both resources, regardless of the `version` portion of the reference.
+* Searches which do specify a version will match only against those resources whose indexed `url` portion of the reference and indexed `version` portion of the reference match the `url` and `version` components of the search parameter value. For example, assume the same Measure resources as in the previous example. If a search parameter value of `http://example.org/fhir/Library/abc|2.0.0` is used in a canonical search against the `library` field, neither of the Measure resources will be a match. Although the indexed `url` portion of the reference is a match for both resources, since a version was specified in the search parameter value, the search will also attempt to match the `version` component of the search parameter value with the indexed `version` portion of the references, and neither is a match.
 * Canonical search parameters may be used when performing chained, reverse chained (`_has`), `_include`, or `_revinclude` searches. As described above, versioned canonical references act against the business version of a resource, not it’s versionId. The special rules that apply to chaining and include references specifying versions of the form `/_history/xx` do not apply to versioned canonical references. Versions of the form `/_history/xx` are not supported with canonical references.
 * The `:above` and `:below` modifiers are not supported with canonical search parameters.
 * Canonical references will be resolved via indexed search values. They will not be resolved via registry look-ups.
--- a/docs/src/pages/Conformance.md
+++ b/docs/src/pages/Conformance.md
@@ -315,6 +315,10 @@ Canonical searches are processed following these rules:
 * The `:above` and `:below` modifiers are not supported with canonical search parameters.
 * Canonical references will be resolved via indexed search values. They will not be resolved via registry look-ups.
 
+Notes:
+* For search parameter definitions whose FHIRPath expression selects element types of both Reference and Canonical, the search parameter will be treated as a canonical search parameter, and will follow the rules described above for canonical searches.
+* For search parameter definitions whose FHIRPath expression selects a choice element, and one of the choice element types is Canonical, the search parameter will be treated as a canonical search parameter only if the choice element is explicitly specified to be of type Canonical (for example, if a reference search parameter is defined over an extension's `value` element, it must contain the `as canonical` clause in order to be treated as a canonical search parameter, i.e. `extension.where(url='http://example.org/canonical-value').value as canonical`.
+
 ### Searching on Special Positional Search
 Positional Search uses [UCUM units](https://unitsofmeasure.org/ucum.html) of distance measure along with common variants:
 
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/CanonicalValue.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/CanonicalValue.java
@@ -6,6 +6,8 @@
 
 package com.ibm.fhir.persistence.jdbc.util;
 
+import java.util.Objects;
+
 /**
  * A canonical value, with processing applied to identify the uri, version, and fragment
  */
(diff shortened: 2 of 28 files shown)


michaelwschroeder:
Design doc has been attached to issue #772 - start with that before reviewing to get an overview of the changes.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
```diff
@@ -170,13 +169,58 @@ public boolean visit(java.lang.String elementName, int elementIndex, com.ibm.fhi
...
+            
+            // For REFERENCE search parameters, we need to treat as both string and as composite.
+            if (REFERENCE.equals(this.searchParamType)) {
+                // Parse the canonical value into a uri and a version. Create a normal string
+                // parameter for the input canonical value, and then create as composite parm to correlate
+                // uri and version (even if version is null).
+                CanonicalValue canonicalValue = CanonicalSupport.createCanonicalValueFrom(canonical.getValue());
+                
+                // Just the uri parm
+                StringParmVal p = new StringParmVal();
```
tbieste, line 181: Since this StringParmValue extraction is duplicated in the "else" block, perhaps just move this part outside this entire "if" block.

fhir-persistence/src/test/java/com/ibm/fhir/persistence/search/test/AbstractSearchURITest.java
```diff
@@ -123,4 +123,92 @@ public void testSearchURI_uri_chained_missing() throws Exception {
...
+    @Test
+    public void testSearchURI_canonical_chained() throws Exception {
+        assertSearchReturnsComposition("subject:Basic.canonical", "http://hl7.org/fhir/DSTU2");
+        assertSearchReturnsComposition("subject:Basic.canonical", "urn:uuid:53fefa32-1111-2222-3333-55ee120877b7");
+
+        assertSearchDoesntReturnComposition("subject:Basic.canonical", "http://HL7.org/FHIR/dstu2");
+        assertSearchDoesntReturnComposition("subject:Basic.canonical", "http://hl7.org/fhir/DSTU");
+        assertSearchDoesntReturnComposition("subject:Basic.canonical", "urn:uuid:53FEFA32-1111-2222-3333-55EE120877B7");
+
+        // TODO add test for diacritics and other unusual characters
```
tbieste, line 156: Will you do this TODO, or can this comment be removed?

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -2831,4 +2856,47 @@ public static boolean isCompartmentSearch(FHIRSearchContext searchContext) {
...
+                    } else if (component.startsWith("as(")) {
+                        dataType = ModelSupport.getDataType(component.substring(3, component.indexOf(")")).trim());
+                    } else if (ModelSupport.isResourceType(component)) {
+                        dataType = ModelSupport.getResourceType(component);
+                    } else {
+                        dataType = ModelSupport.getElementType(dataType, component);
+                    }
+                }
+            }
+            if ("Canonical".equals(dataType.getSimpleName())) {
```
tbieste, line 2896: Feels like "Canonical" should be a constant in SearchConstants.java.

fhir-search/src/main/java/com/ibm/fhir/search/SearchConstants.java
```diff
@@ -119,6 +125,12 @@ private SearchConstants() {
 
     // Extracted search parameter suffix for :text modifier
     public static final String TEXT_MODIFIER_SUFFIX = ":text";
+    
+    // Extracted search parameter suffixes for canonical values
+    public static final String CANONICAL_URL_NAME = "internal_canonical_url";
+    public static final String CANONICAL_SUFFIX = ":canonical";
```
tbieste, line 131: Since this is not a modifier, maybe we want to use something other than a colon in the suffix (like an underscore instead).

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
michaelwschroeder: Done.

fhir-search/src/main/java/com/ibm/fhir/search/SearchConstants.java
michaelwschroeder: Suffix was changed to use an underscore. CANONICAL_URL_NAME was removed. For the cases where we now create composite parameters, we use the search parameter name + CANONICAL_SUFFIX for the composite parameter name and we use the SearchUtil.makeCompositeSubCode() method for the component names. This is consistent with how other composite parameters are named.

docs/src/pages/Conformance.md
```diff
@@ -272,8 +272,14 @@ The IBM FHIR Server does not consider the `Quantity.comparator` field as part of
...
+
 ### Searching on Reference
-Reference searches on the IBM FHIR Server support search on:
+Reference searches on the IBM FHIR Server support search on elements of type Reference and on elements of type canonical.
+
+**Search on Elements of Type Reference:**
+
+Reference searches may be of the following types:
 
 * a relative reference - `1` where it is reflexsively determined to be a subset of possible targets, such as `Patient/1`, `Group/1`
```
lmsurpre, line 284: unrelated to the PR but I wonder whether that is supposed to say `reflexively` or `reflectively` :-)

docs/src/pages/Conformance.md
```diff
@@ -272,8 +272,14 @@ The IBM FHIR Server does not consider the `Quantity.comparator` field as part of
 ### Searching on URI
 URI searches on the IBM FHIR Server are case-sensitive with "exact-match" semantics. The `above` and `below` prefixes can be used to perform path-based matching that is based on the `/` delimiter.
 
+There is one exception to the statement above. The `url` search parameter, which is defined in the base FHIR specification on definitional resource types as a URI search parameter, is actually treated as a canonical search parameter of type REFERENCE, as documented in the [FHIR specification](http://build.fhir.org/references.html#canonical). The following section of this document describes how the IBM FHIR Server processes canonical reference searches.
```
lmsurpre, line 275: Unless we're dependent on some updated text, I usually try to scope spec links to R4 vs the latest draft.

docs/src/pages/Conformance.md
```diff
@@ -295,6 +301,19 @@ Elements of type Reference may contain a versioned reference, such as `Patient/1
...
     * Example: A `Condition` resource contains a reference of `Patient/123/_history/2` in its `subject` element, and the current version of the `Patient/123` resource is `2`, and a search of `Patient?_revinclude=Condition:subject` is performed. The `Condition` resource will be returned as an `include` resource since the version of the `Patient` resource specified in the `subject` element is the current version of the `Patient` resource. If the current version of the `Patient/123` resource is `3`, then the `Condition` resource will not be returned as an `include` resource in the search results.
 
+**Search on Elements of Type Canonical:**
+
+Canonical searches search against the canonical URL of a resource. This is for resource types that have a defined element `url` which is the URL that always identifies the resource across all contexts of use. The list of resource types that have a canonical URL and are allowed to be the target of a reference to a canonical URL are documented in the [FHIR specification](http://build.fhir.org/references.html#canonical).
```
lmsurpre, line 306:
```suggestion
Canonical searches search against the canonical URL of a resource. This is for resource types that have a defined element `url` which is the URL that always identifies the resource across all contexts of use. The list of resource types that have a canonical URL and are allowed to be the target of a reference to a canonical URL are documented in the [FHIR specification](http://hl7.org/fhir/R4/references.html#canonical).
```

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/CanonicalValue.java
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.ibm.fhir.persistence.jdbc.util;
+
+/**
+ * A canonical value, with processing applied to identify the uri, version, and fragment
+ */
+public class CanonicalValue {
```
lmsurpre, line 12: wondering if this one should get a equals and hashcode impl so it can be used in more places

docs/src/pages/Conformance.md
```diff
@@ -295,6 +301,19 @@ Elements of type Reference may contain a versioned reference, such as `Patient/1
...
+* Searches which do not specify a version will match against all resources whose indexed `url` value matches the search parameter value, regardless if the resource has a `version` value specified.
+* Searches which do specify a version will match only against those resources whose indexed `url` and indexed `version` values match the `url` and `version` components of the search parameter value.
+* Canonical search parameters may be used when performing chained, reverse chained (`_has`), `_include`, or `_revinclude` searches. As described above, versioned canonical references act against the business version of a resource, not it’s versionId. The special rules that apply to chaining and include references specifying versions of the form `/_history/xx` do not apply to versioned canonical references. Versions of the form `/_history/xx` are not supported with canonical references.
```
lmsurpre, line 313: Is "canonical search parameters" defined somewhere?  I think it means search parameters of type `reference` which target an element of type canonical.  But I'm don't think we know which ones those are when we process a search query, so maybe its just reference search params that have a value which looks like a canonical url?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
```diff
@@ -342,16 +376,15 @@ public boolean visit(String elementName, int elementIndex, Uri uri) {
...
-                p.setName(searchParamCode);
-                p.setUrl(searchParamUrl);
-                p.setVersion(searchParamVersion);
-                p.setValueCode(uri.getValue());
-                result.add(p);
+                try {
+                    result.add(buildReferenceParmVal(uri.getValue(), null));
+                } catch (FHIRSearchException e) {
+                    // Log the error, but skip it because we're not supposed to throw exceptions here
+                    log.log(Level.WARNING, "Error processing reference", e);
```
lmsurpre, line 386: Under which cases would we hit this exception?  Bad input data from clients?  Or only for bad code / bad IGs on the server end?

If it results from some bad/unexpected data from clients, maybe the log level should be reduced.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
```diff
@@ -342,16 +376,15 @@ public boolean visit(String elementName, int elementIndex, Uri uri) {
                 throw invalidComboException(searchParamType, uri);
             }
 
-            // For REFERENCE search parameters, we need to treat Uris as tokens,
+            // For REFERENCE search parameters, we need to treat Uris as references,
```
lmsurpre, line 379: Do you happen to have any example of elements of type `uri` that would get indexed by search parameters of type `reference`?  I know the spec says its allowed, but I was having trouble wrapping my head around it and I think a concrete example might help me.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
lmsurpre: ah, now i see it just matches what we were doing in the Reference visit method.  although now i'm wondering the same thing there...

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -2831,4 +2856,47 @@ public static boolean isCompartmentSearch(FHIRSearchContext searchContext) {
...
+            }
+            
+            // Strip out '.where()' functions, handling nested parentheses
+            subExpression = subExpression
+                    .replaceAll("\\.where(?=\\()(?:(?=.*?\\((?!.*?\\1)(.*\\)(?!.*\\2).*))(?=.*?\\)(?!.*?\\2)(.*)).)+?.*?(?=\\1)[^(]*(?=\\2$)", "");
+            
+            // Split into components
+            for (String component : subExpression.split("\\.")) {
+                component = component.trim();
+                if ((dataType == null || !component.equals(dataType.getSimpleName())) && !component.startsWith("where")) {
```
lmsurpre, line 2884: > !component.startsWith("where")

didn't we just strip out the where functions above

docs/src/pages/Conformance.md
lmsurpre: OK, I think I found my answer in the code.  Specifically SearchUtil.isCanonicalSearchParm is doing some work (understatement) to determine if the search parameter can target elements of type canonical or not.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -2831,4 +2856,47 @@ public static boolean isCompartmentSearch(FHIRSearchContext searchContext) {
         return result;
     }
 
+    /**
+     * Determine if a search parameter expression references an element of type Canonical.
+     * @param resourceType
+     * @param expression
+     * @return
+     */
+    private static boolean isCanonicalSearchParm(Class<?> resourceType, String expression) {
```
lmsurpre, line 2865: given that this is called multiple times from multiple places, I'm a tad worried at the performance cost of this.

I *think* it would be possible to move this into some pre-processing and add an extension onto the SearchParameter definitions which can possibly target canonicals (similar to what we do for computing a token search params implicit system).  Thoughts?

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -1144,7 +1147,7 @@ private static void checkSearchParameterCombinations(Class<?> resourceType, List
      * Common logic from handling a single queryParameterValueString based on its type
      */
     private static List<QueryParameterValue> processQueryParameterValueString(Class<?> resourceType, SearchParameter searchParameter, Modifier modifier,
-        String modifierResourceTypeName, String queryParameterValueString) throws FHIRSearchException, Exception {
+        String modifierResourceTypeName, String queryParameterValueString, boolean isCanonical) throws FHIRSearchException, Exception {
```
lmsurpre, line 1150:
```suggestion
            String modifierResourceTypeName, String queryParameterValueString, boolean isCanonical) throws FHIRSearchException, Exception {
```

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -2082,17 +2100,22 @@ private static QueryParameter parseReverseChainedParameter(Class<?> resourceType
                             }
                         }
 
+                        isCanonical = Type.REFERENCE.equals(type) &&
+                                isCanonicalSearchParm(resourceType, searchParameter.getExpression().getValue());
```
lmsurpre, line 2104: how does it differ from the isCanonical we stored on line 2048?

fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractCanonicalTest.java
```diff
@@ -0,0 +1,767 @@
...
+    private static String uuid = UUID.randomUUID().toString();
+
+    /**
+     * Loads up and saves a bunch of resources with various canonical references to one another
+     */
+    @BeforeClass
+    public void createResources() throws Exception {
+        CarePlan carePlan = TestUtil.readExampleResource("json/ibm/minimal/CarePlan-1.json");
+        Measure measure = TestUtil.readExampleResource("json/ibm/minimal/Measure-1.json");
+        Library library = TestUtil.readExampleResource("json/ibm/minimal/Library-1.json");
```
lmsurpre, line 58: FYI I created TestUtil.getMinimalResource(ResourceType) a while back as a convenience method.  It does the same thing anyway, so doesn't really matter; just wanted to mention it.

If we did use getMinimalResource consistently we could potentially change the implementation to keep the minimal resources around in memory so that we don't need to parse the files for each use.

fhir-persistence/src/test/java/com/ibm/fhir/persistence/search/test/AbstractSearchURITest.java
michaelwschroeder: removed comment

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
michaelwschroeder: I used Canonical.class.getSimpleType() instead.

docs/src/pages/Conformance.md
michaelwschroeder: I think reflectively makes more sense in this context, so changed it to that.

docs/src/pages/Conformance.md
michaelwschroeder: Done.

docs/src/pages/Conformance.md
michaelwschroeder: Done.

docs/src/pages/Conformance.md
michaelwschroeder: I updated this section a bit. Hopefully a little clearer, but you're right - 'canonical search parameter' is determined by looking at the FHIRPath definition

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
michaelwschroeder: It's actually thrown from ReferenceUtil.getServiceBaseUrl() when a base URL is not valid. I believe this would only be from bad user data, but left as warning.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
michaelwschroeder: good catch - removed

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
michaelwschroeder: Done.

fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractCanonicalTest.java
michaelwschroeder: changed to use getMinimalResource

lmsurpre: cool, I decided to just whip up these changes since they were on my mind:  https://github.com/IBM/FHIR/pull/2701
I don't think it will have much impact, but it does mean we can avoid re-reading these same minimal resources over and over and could eventually add up to some savings.

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
lmsurpre: Thinking about this one some more, I'm not sure if the pre-processing idea is a good one or not.  In the case of implicit-systems, it works with or without the extension...having the extension just gives us a performance boost.

Since we allow users to configure their own search params, we probably don't want to rely on some precomputed extensions to get the proper system behavior.  The concern about the complexity/performance still stands though.  Maybe just a place to watch for as we move forward?

michaelwschroeder: I think that sounds reasonable. And this will only get called for Reference search parms, and only once per search parm, so hopefully not a big hit.

docs/src/pages/Conformance.md
lmsurpre: looks good

fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractCanonicalTest.java
```diff
@@ -0,0 +1,769 @@
...
+    private static String uuid = UUID.randomUUID().toString();
+
+    /**
+     * Loads up and saves a bunch of resources with various canonical references to one another
+     */
+    @BeforeClass
+    public void createResources() throws Exception {
+        CarePlan carePlan = TestUtil.getMinimalResource(ResourceType.CARE_PLAN, Format.JSON);
+        Measure measure = TestUtil.getMinimalResource(ResourceType.MEASURE, Format.JSON);
+        Library library = TestUtil.getMinimalResource(ResourceType.LIBRARY, Format.JSON);
```
lmsurpre, line 60: I broke these with https://github.com/IBM/FHIR/pull/2701.  Now it will be:
```suggestion
        CarePlan carePlan = TestUtil.getMinimalResource(CarePlan.class);
        Measure measure = TestUtil.getMinimalResource(Measure.class);
        Library library = TestUtil.getMinimalResource(Library.class);
```

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
lmsurpre: OK, works for me.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/JDBCParameterBuildingVisitor.java
lmsurpre: OK, just something to watch for I guess

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/CanonicalValue.java
michaelwschroeder: done

fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractCanonicalTest.java
michaelwschroeder: done

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/CanonicalValue.java
```diff
@@ -0,0 +1,99 @@
...
+     */
+    public String getFragment() {
+        return fragment;
+    }
+
+    @Override
+    public int hashCode() {
+        final int prime = 31;
+        int result = 1;
+        result = prime * result + ((fragment == null) ? 0 : fragment.hashCode());
```
punktilious, line 57: Objects.hashCode might be more convenient here

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -2655,7 +2655,7 @@ Bundle createSearchBundle(List<Resource> resources, FHIRSearchContext searchCont
             List<QueryParameter> logicalIdReferenceSearchParameters = new ArrayList<>();
             for (QueryParameter queryParameter : searchContext.getSearchParameters()) {
                 if (!queryParameter.isReverseChained()) {
-                    if (queryParameter.isChained()) {
+                    if (queryParameter.isChained() && !queryParameter.isCanonical()) {
```
punktilious, line 2658: What if it is isChained() and isCanonical()?

michaelwschroeder: What we're looking for here with chained parameters is whether or not the reference contained a "logical" version (i.e. `.../_history/xx`), so we can then issue a warning message that we don't act on versioned references . Since that type of version is not valid for canonical references, we don't need to check canonicals. However, I just realized we also don't need to do the check for logical id-only with canonical references, so I'll move the !queryParameter.isCanonical() check up to the line above:
```
if (!queryParameter.isReverseChained() &&  !queryParameter.isCanonical()) {
```

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/util/CanonicalValue.java
michaelwschroeder: changed to use Objects.hash(fragment, uri, version)

lmsurpre (approved):
LGTM

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
punktilious: Can you add that as a comment in the code?

michaelwschroeder: done

punktilious (approved):
LGTM

Issue #2184, opened 2021-03-31
Provide the ability to retrieve a set of COMMON_TOKEN_VALUE_IDs in one request


michaelwschroeder:
**Is your feature request related to a problem? Please describe.**
We have cases when building queries (processing inclusion criteria, processing :in, :not-in, :above, :below modifiers) where we want to attempt to retrieve COMMON_TOKEN_VALUE_IDs for a list of system+code pairs. It could be prohibitive performance-wise to go to the DB one-at-a-time for each cache miss.

**Describe the solution you'd like**
Provide a way to retrieve the COMMON_TOKEN_VALUE_IDs for a list of system+code pairs in a single request.

**Describe alternatives you've considered**
1. Using existing interface, risking going out to the DB for every cache miss
2. Don't use the COMMON_TOKEN_VALUE_IDs

**Acceptance Criteria**
1.  
   GIVEN [a precondition]
    AND [another precondition]
   WHEN [test step]
    AND [test step]
   THEN [verification step]
    AND [verification step]

**Additional context**

lmsurpre:
the new QueryBuilder has support for handling an "in list" of common token values, so it will already handle conversion between `= 'a'` and `in (a,b,c)`, so just a matter of ensuring we get the values before that efficiently (even in face of cache misses)

michaelwschroeder:
This issue has been verified. 

I started with a clean schema and performed a search using an :in modifier:
- verified via log that, as expected, for the values in the specified value set, all were cache misses
- verified that a DB query was executed to retrieve the common token value IDs for the cache misses
- verified that since none were found in the DB, a count query for the search was generated with only a -1 for the COMMON_TOKEN_VALUE_ID value

I then ingested resources which caused some of the values in the value set to be indexed and added to the DB and cache. I reran the search:
- verified via log that, as expected, for the values in the specified value set, all were cache misses except the ones that were indexed for the ingested resources
- verified a DB query was executed to retrieve the common token value IDs for the cache misses
- verified that since none of the misses were in the DB, a count query for the search was generated with only the cache hits specified for the COMMON_TOKEN_VALUE_ID values

I then stopped and restarted the server to clear the cache. I reran the search:
- verified via log that, as expected, for the values in the specified value set, all were cache misses
- verified a DB query was executed to retrieve the common token value IDs for the cache misses, and the ones that were indexed for the ingested resources were returned and added to the cache
[...]

PR #2684, opened 2021-08-13
issue #2184 - support getting all the common token value ids in one go

lmsurpre:
Added:
* getCommonTokenValueIds to JDBCIdentityCache
* resolveCommonTokenValueIds to ICommonTokenValuesCache
* readCommonTokenValueIds to ResourceReferenceDAO

Updated SearchQueryRenderer to use this mechanism in two cases:
* getReferenceFilter for reference params without the `:identifier` modifier (including compartment params)
* getTokenFilter for token params with the IN, ABOVE, BELOW, and NOT-IN modifiers

Also includes cleanup in getReferenceFilter and populateCodesSubSegment (used by getTokenFilter):
* removal of code in getReferenceFilter for handling modifier that we don't support for reference params
* removal of code that uses the token ref view for token filters...I believe that the IN, ABOVE, BELOW, and NOT-IN cases can be handled with COMMON_TOKEN_VALUES now

Needs testing on Db2.

### DIFF
--- a/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/query/expression/StringStatementRenderer.java
+++ b/fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/query/expression/StringStatementRenderer.java
@@ -92,7 +92,7 @@ public class StringStatementRenderer implements StatementRenderer<String> {
 
         if (groupByClause != null) {
             if (this.pretty) {
-                result.append(NEWLINE).append("    ");
+                result.append(NEWLINE).append("   ");
             }
             result.append(SPACE).append(groupByClause.toString());
         }
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/JDBCIdentityCacheImpl.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/JDBCIdentityCacheImpl.java
@@ -7,10 +7,14 @@
 package com.ibm.fhir.persistence.jdbc.dao.impl;
 
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 
 import com.ibm.fhir.persistence.exception.FHIRPersistenceException;
 import com.ibm.fhir.persistence.jdbc.FHIRPersistenceJDBCCache;
@@ -157,6 +161,75 @@ public class JDBCIdentityCacheImpl implements JDBCIdentityCache {
                 cache.getResourceReferenceCache().addTokenValue(new CommonTokenValue(dto.getCodeSystemId(), tokenValue), result);
             }
         }
+        return result;
+    }
+
+//    public List<Long> getCommonTokenValueIdsPreservingOrder(List<CommonTokenValue> tokenValues) {
+//        List<Integer> misses = new ArrayList<>();
+//        List<Long> result = cache.getResourceReferenceCache().resolveCommonTokenValueIds(tokenValues, misses);
+//
+//        if (misses.isEmpty()) {
+//            return result;
+//        }
+//
+//        Map<CommonTokenValue,Integer> valueAndIndex = new LinkedHashMap<>();
+//        for (Integer missIndex : misses) {
+//            CommonTokenValue commonTokenValue = tokenValues.get(missIndex);
+//
+//            if (logger.isLoggable(Level.FINE)) {
+//                logger.fine("Cache miss. Fetching common_token_value_id from database: " + commonTokenValue);
+//            }
+//            valueAndIndex.put(commonTokenValue, missIndex);
+//        }
+//        Set<CommonTokenValueResult> readCommonTokenValueIds = resourceReferenceDAO.readCommonTokenValueIds(valueAndIndex.keySet());
+
+//        if (result == null) {
+//
+//            CommonTokenValueResult dto = resourceReferenceDAO.readCommonTokenValueId(codeSystem, tokenValue);
+//            if (dto != null) {
+//                // Value exists in the database, so we can add this to our cache. Note that we still
+//                // choose to add it the thread-local cache - this avoids any locking. The values will
+//                // be promoted to the shared cache at the end of the transaction. This avoids unnecessary
+//                // contention.
+//                result = dto.getCommonTokenValueId();
+//                if (logger.isLoggable(Level.FINE)) {
+//                    logger.fine("Adding common_token_value_id to cache: '" + codeSystem + "|" + tokenValue + "' = " + result);
+//                }
+//                cache.getResourceReferenceCache().addCodeSystem(codeSystem, dto.getCodeSystemId());
+//                cache.getResourceReferenceCache().addTokenValue(new CommonTokenValue(dto.getCodeSystemId(), tokenValue), result);
+//            }
+//        return result;
+//    }
+
+    @Override
+    public Set<Long> getCommonTokenValueIds(Collection<CommonTokenValue> tokenValues) {
+        Set<CommonTokenValue> misses = new HashSet<>();
+        Set<Long> result = cache.getResourceReferenceCache().resolveCommonTokenValueIds(tokenValues, misses);
+
+        if (misses.isEmpty()) {
+            return result;
+        }
+
+        if (logger.isLoggable(Level.FINE)) {
+            logger.fine("Cache miss. Fetching common_token_value_ids from database: " + misses);
+        }
+
+        Set<CommonTokenValueResult> readCommonTokenValueIds = resourceReferenceDAO.readCommonTokenValueIds(misses);
+        result.addAll(readCommonTokenValueIds.stream()
+            .map(r -> r.getCommonTokenValueId())
+            .collect(Collectors.toSet()));
+
+        for (CommonTokenValueResult dto : readCommonTokenValueIds) {
+            // Value exists in the database, so we can add this to our cache. Note that we still
+            // choose to add it to the thread-local cache - this avoids any locking. The values will
+            // be promoted to the shared cache at the end of the transaction. This avoids unnecessary
+            // contention.
+            if (logger.isLoggable(Level.FINE)) {
+                logger.fine("Adding common_token_value_id to cache: '" + dto.getCodeSystemId() + "|" + dto.getTokenValue() + "' = " + result);
+            }
+            cache.getResourceReferenceCache().addTokenValue(new CommonTokenValue(dto.getCodeSystemId(), dto.getTokenValue()), dto.getCommonTokenValueId());
+        }
+
         return result;
     }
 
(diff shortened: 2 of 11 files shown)


fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/query/expression/StringStatementRenderer.java
```diff
@@ -92,7 +92,7 @@ public String select(boolean distinct, SelectList selectList, FromClause fromCla
 
         if (groupByClause != null) {
             if (this.pretty) {
-                result.append(NEWLINE).append("    ");
```
lmsurpre, line 95: this is a minor formatting fix that is unrelated to the rest of the PR

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/derby/DerbyResourceReferenceDAO.java
```diff
@@ -157,6 +191,29 @@ public void doCanonicalValuesUpsert(String paramList, Collection<String> urls) {
...
+        try (Statement s = getConnection().createStatement()) {
+            s.executeUpdate(upsert.toString());
+        } catch (SQLException x) {
+            logger.log(Level.SEVERE, upsert.toString(), x);
+            throw getTranslator().translate(x);
+        }
+    }
+
+    private void insertToCommonTokenValuesTmp(Collection<CommonTokenValue> tokenValues) {
         final String insert = "INSERT INTO SESSION.common_token_values_tmp(token_value, code_system_id) VALUES (?, ?)";
```
lmsurpre, line 217: I wasn't totally sure if it was safe to use the same temporary table name here for both doCommonTokenValuesUpsert and the new readCommonTokenValueIds.

In the latest CI run I'm seeing errors like the following:
```
[ERROR   ] INSERT INTO common_token_values (token_value, code_system_id)      SELECT src.token_value, src.code_system_id        FROM SESSION.common_token_values_tmp src  LEFT OUTER JOIN common_token_values ctv               ON ctv.token_value = src.token_value              AND ctv.code_system_id = src.code_system_id       WHERE ctv.token_value IS NULL    ORDER BY src.token_value, src.code_system_id
A lock could not be obtained within the time requested
```

So I'm thinking to have this shared helper accept the name for the temporary table as an argument and then ensure that the two different method pass a unique value.

lmsurpre: scratch that, I went with the simpler `WHERE (c.token_value = ? AND c.code_system_id = ?) OR (c.token_value = ? AND c.code_system_id = ?) OR ...` approach instead because it just felt safer to me and I don't want to spend too much time optimizing for derby

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/derby/DerbyResourceReferenceDAO.java
```diff
@@ -49,6 +53,43 @@ public DerbyResourceReferenceDAO(IDatabaseTranslator t, Connection c, String sch
...
 
+    @Override
+    public Set<CommonTokenValueResult> readCommonTokenValueIds(Collection<CommonTokenValue> tokenValues) {
+        if (tokenValues.isEmpty()) {
+            return Collections.emptySet();
+        }
+
+        Set<CommonTokenValueResult> result = new HashSet<>();
+
+        String SQL = ""
```
punktilious, line 64: Should use StringBuilder here

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceReferenceDAO.java
```diff
@@ -152,6 +155,49 @@ public CommonTokenValueResult readCommonTokenValueId(String codeSystem, String t
...
 
+    @Override
+    public Set<CommonTokenValueResult> readCommonTokenValueIds(Collection<CommonTokenValue> tokenValues) {
+        if (tokenValues.isEmpty()) {
+            return Collections.emptySet();
+        }
+
+        Set<CommonTokenValueResult> result = new HashSet<>();
+
+        String SQL = ""
```
punktilious, line 166: StringBuilder

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/domain/SearchQueryRenderer.java
```diff
@@ -522,8 +537,45 @@ protected WhereFragment getTokenFilter(QueryParameter queryParm, String paramAli
...
+                //
+                //
+                // Scenario 2: none of the resources have any of the values from the valueset
+                //
+                //    WHERE NOT EXISTS(
+                //    SELECT 1
+                //    FROM Basic_RESOURCE_TOKEN_REFS AS P2
+                //    WHERE P2.LOGICAL_RESOURCE_ID = LR1.LOGICAL_RESOURCE_ID
+                //      AND P2.PARAMETER_NAME_ID = 20046
+                //      AND ((P2.COMMON_TOKEN_VALUE_ID IN (-1)))))) AS LR
```
punktilious, line 571: Just a thought...if we only have -1 in the set here, it might be worth replacing that with NULL. This _may_ let the database optimize the query more, because it knows that NULL will never match anything and therefore it could short-circuit the join (if it's clever enough). With -1, it still has to go fishing in the index to find that value. May also impact cardinality estimation (hopefully in a positive way).

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dto/CommonTokenValue.java
```diff
@@ -11,18 +11,18 @@
...
-    
+
     private final int codeSystemId;
-    
+
     // tokenValue can be null
     private final String tokenValue;
-    
+
     public CommonTokenValue(int codeSystemId, String tokenValue) {
```
punktilious, line 20: Missing javadoc here (my bad)

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/JDBCIdentityCacheImpl.java
```diff
@@ -160,6 +164,38 @@ public Long getCommonTokenValueId(String codeSystem, String tokenValue) {
...
+        if (misses.isEmpty()) {
+            return result;
+        }
+
+        if (logger.isLoggable(Level.FINE)) {
+            logger.fine("Cache miss. Fetching common_token_value_ids from database: " + misses);
+        }
+
+        Set<CommonTokenValueResult> readCommonTokenValueIds = resourceReferenceDAO.readCommonTokenValueIds(misses);
+        result.addAll(readCommonTokenValueIds.stream()
```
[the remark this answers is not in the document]
lmsurpre: just that it was a tad slower in code with very tight loops.  I don't think we need to avoid ALL uses of streams because of that.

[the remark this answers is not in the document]
lmsurpre: this particular method should happen once per search parameter and typically we don't have a lot of those.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dto/CommonTokenValue.java
lmsurpre: added javadoc

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/domain/SearchQueryRenderer.java
lmsurpre: Its actually a bind variable (I had simplified it for my comment).  I went ahead and changed the behavior so that we bind a null instead of a -1 (at least for the case of token params).
For reference params, we use BaseWhereAdapter.eq which takes primitive values, so in that case I'm still using the `-1`.

We should probably resolve on a consistent approach to these.  Is it better to use literals?  Is it better to use null?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/JDBCIdentityCacheImpl.java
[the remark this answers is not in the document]
punktilious: Yeah, just thought I'm mention it in case you thought it could have an impact. It's a shame about the performance, because the syntactic convenience is nice. Agree we probably don't have to worry unless it shows up in a profile down the road.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/domain/SearchQueryRenderer.java
punktilious: Comparisons with common_token_value_id shouldn't use bind variables - we always want to use literals to make sure we get the best plan. With PostgreSQL it's more interesting, because it definitely peeks more at the bind values when optimizing, but other databases may not. Anyway, our standard approach is to always use literals where it's a database id (where there is little chance the query will be reused).

As for OR vs IN list, the optimizer will usually automatically convert...but I tend to prefer an IN list because it's neater.

If you *only* have -1 - e.g. WHERE foo_id IN (-1) then we should rewrite that to be WHERE foo_id IN (NULL). That gives the optimizer some useful information.

punktilious (approved):
LGTM

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/domain/SearchQueryRenderer.java
lmsurpre: Our current BaseWhereAdapter doesn't seem to support the use of NULL literals (in either eq or in expressions), so unless we change that we need to pick whether to use literals or whether to use NULLs.  For now, I opted to use literals across these two methods now and we can use #2688 if we want to add support for replacing the -1 values with NULLs.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dto/CommonTokenValue.java
```diff
@@ -11,18 +11,23 @@
  * DTO representing a record in COMMON_TOKEN_VALUES
```
tbieste, line 11: Update copyright.

tbieste (approved):
Left one minor comment.  Otherwise, LGTM.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dto/CommonTokenValue.java
lmsurpre: done

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/domain/SearchQueryRenderer.java
punktilious: For completeness, I just wanted to update here and say we should probably use `WHERE foo_id = NULL` for the case where there is a single value. That probably gives us the best chance that the optimizer will see it and short-circuit the join (because nothing ever is equal to NULL). Of course, if we know this we could probably rewrite the query ourselves, but that is complicated, so we are delegating responsibility to the DB layer, in the hope it will do the right thing.

Issue #2690, opened 2021-08-17
Build fails due to missing azure-storage-blob library


MFAshby:
**Describe the bug**
Building with maven as described in [CONTRIBUTING.md](https://github.com/IBM/FHIR/blob/main/CONTRIBUTING.md#setup) fails with the error below. This is because azure-storage-blob library version 12.6.0 has been pulled from maven central (see https://repo.maven.apache.org/maven2/com/azure/azure-storage-blob/ , the link to version 12.6.0 returns 404)

**Environment**
Current `main` branch.

**To Reproduce**
Run the below commands:
```
git clone git@github.com:IBM/FHIR
cd FHIR
mvn clean install -f fhir-parent/pom.xml
```

**Expected behavior**
Successful build.

**Additional context**
Output from build command:
```
[INFO] -----------------< com.ibm.fhir:fhir-bulkdata-webapp >------------------
[INFO] Building fhir-bulkdata-webapp 4.9.0-SNAPSHOT                     [45/53]
[INFO] --------------------------------[ war ]---------------------------------
[WARNING] The POM for com.azure:azure-storage-blob:jar:12.6.0 is missing, no dependency information available
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for IBM FHIR Server 4.9.0-SNAPSHOT:
[INFO] 
[INFO] IBM FHIR Server .................................... SUCCESS [  3.321 s]
[INFO] fhir-core .......................................... SUCCESS [  2.544 s]
[INFO] fhir-database-utils ................................ SUCCESS [  2.082 s]
[INFO] fhir-config ........................................ SUCCESS [  0.530 s]
[INFO] fhir-cache ......................................... SUCCESS [  0.290 s]
[INFO] fhir-model ......................................... SUCCESS [ 13.664 s]
[INFO] fhir-audit ......................................... SUCCESS [  0.645 s]
[INFO] fhir-registry ...................................... SUCCESS [ 15.991 s]
[INFO] fhir-term .......................................... SUCCESS [  0.507 s]
[INFO] fhir-term-graph .................................... SUCCESS [  2.348 s]
[INFO] fhir-provider ...................................... SUCCESS [  1.027 s]
[INFO] fhir-term-remote ................................... SUCCESS [  0.329 s]
[INFO] fhir-profile ....................................... SUCCESS [  0.270 s]
[INFO] fhir-path .......................................... SUCCESS [  5.033 s]
[...]

PR #2691, opened 2021-08-17
Upgrade azure-storage-blob dependency to fix build

MFAshby:
Fixes https://github.com/IBM/FHIR/issues/2690

Also add an extra line to CONTRIBUTING.md. Without first compiling the fhir-examples folder, the remaining build fails.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,7 +74,10 @@ Connect with us through https://chat.fhir.org/ or open an [issue](https://github
 ## Setup
 The IBM FHIR Server is built with Maven and requires Java 8 or higher.  To build the project from the root directory, please execute:
 
-> mvn clean install -f fhir-parent/pom.xml
+```
+mvn clean install -f fhir-examples
+mvn clean install -f fhir-parent
+```
 
 ## Testing
 To ensure a working build, please run the full build from the root of the project before submitting your pull request.
--- a/fhir-parent/pom.xml
+++ b/fhir-parent/pom.xml
@@ -558,7 +558,7 @@
             <dependency>
                 <groupId>com.azure</groupId>
                 <artifactId>azure-storage-blob</artifactId>
-                <version>12.6.0</version>
+                <version>12.13.0</version>
             </dependency>
             <dependency>
                 <groupId>com.azure</groupId>


fhir-parent/pom.xml
```diff
@@ -558,7 +558,7 @@
             <dependency>
                 <groupId>com.azure</groupId>
                 <artifactId>azure-storage-blob</artifactId>
-                <version>12.6.0</version>
+                <version>12.13.0</version>
```
prb112, line 561: Hi Martin, 

Please note Lee merged the fix for the 12.6.0 being removed from OSSRH. 

https://github.com/IBM/FHIR/blob/main/fhir-parent/pom.xml#L561
```
            <dependency>
                <groupId>com.azure</groupId>
                <artifactId>azure-storage-blob</artifactId>
                <version>12.13.0</version>
            </dependency>
```
This is already

prb112: If you Fetch and Merge on the main branch, it'll sync you up with the change. 
image

CONTRIBUTING.md
```diff
@@ -74,7 +74,10 @@ Connect with us through https://chat.fhir.org/ or open an [issue](https://github
 ## Setup
 The IBM FHIR Server is built with Maven and requires Java 8 or higher.  To build the project from the root directory, please execute:
 
-> mvn clean install -f fhir-parent/pom.xml
+```
+mvn clean install -f fhir-examples
```
prb112, line 78: Good point, happy to add

punktilious: I thought fhir-examples was now a dependency available in maven central so this step wouldn't be required?

prb112: It depends ... if in the current release we are using the snapshot.

punktilious: Perhaps also worth pointing people over to the "setting up for development" wiki instead of having instructions twice which may get out of sync.

MFAshby:
Thanks @prb112 and @punktilious  for review, I'll open a new PR with suggested documentation changes since the original issues here is fixed already.

prb112:
Sounds great, happy to have the update to the docs

Issue #2680, opened 2021-08-13
Support for controlling date and time during system testing


MFAshby:
**Is your feature request related to a problem? Please describe.**
I would like to carefully control the current date and time during integration/system testing. Doing so allows me to write tests with predictable outputs that I can easily verify.

**Describe the solution you'd like**
Replace references to `Instant.now()` and `LocalDate.now()` etc with the corresponding signatures that accept a `Clock`.
Provide an API for fixing/changing the clock.

**Describe alternatives you've considered**
[Libraries exist](https://github.com/wolfcw/libfaketime) for intercepting system calls to get the current date and time, however in a Java VM these can have unintended side-effects, and may not work on all JVM implementations.

**Acceptance Criteria**

1.  
   GIVEN an IBM FHIR server configured to accept clock changes
   WHEN I call the API to change the clock
   THEN any FHIR resources or operations which use the current date and time, will get the time from the desired clock
    AND output from those operations is stable and predictable, no matter what time of day the test ran.

**Additional context**
n/a

MFAshby:
I'm finding this slightly difficult due to the lack of dependency injection in a number of places throughout the system. Look for references to `Instant.now()` or `ZonedDateTime.now()` and you'll see that a large number of them are in static methods. This necessitates either passing a `Clock` through a large number of seemingly unrelated signatures, or making those methods non-static and injecting a `Clock` as a class member.

Given the app is based on openliberty, should we be using CDI for proper dependency injection and inversion of control? Specifically I'm looking at DateTimeHandler, ValueSetSupport and classes which depend on it. Although at first glance these appear to be collections of functions with no dependency on application state, they are actually using calls to `.now()` methods which can't be controlled in tests.

punktilious:
@MFAshby discussed this on our team call and agree that DI and IoC patterns such as the one you mention do greatly benefit testing - with clock/time being a really good example - especially for handling tests for things like DST transitions (although we use UTC specifically to avoid this). Not sure when we'll be able to get to this, but wanted to let you know this has sparked discussion/ideas in the team.

MFAshby:
thanks for considering it @punktilious 

What I'll probably do is go ahead and attempt this using a static member for the current date and time factory. If that patch is accepted, it'll close this ticket. I've filed a separate issue regarding the use of dependency injection system.

MFAshby:
[...]

Issue #2687, opened 2021-08-16
Use the dependency injection system


MFAshby:
**Is your feature request related to a problem? Please describe.**
I'd like to implement the extension proposed in https://github.com/IBM/FHIR/issues/2680, but I'm finding it difficult due to many static methods calling `Instant.now()`. This appears to be a wider problem in the FHIR server codebase, it's not easy to change the wiring of dependent services.

**Describe the solution you'd like**
Avoid using static methods except when they are actually pure functions (i.e. no side effects, and they depend only on the passed in parameters). Use the dependency injection system to wire together classes and their dependencies. CDI should be used in this case, as provided by [openliberty server](https://openliberty.io/guides/cdi-intro.html).

This proposal might also allow removal of boilerplate code like `com.ibm.fhir.audit.AuditLogServiceFactory` which appears to be doing manual dependency injection.

**Describe alternatives you've considered**
One possible alternative in this case is to simply have some `static` factory for the current date and time, which can be mutated if necessary.

**Acceptance Criteria**

1.  
   GIVEN [a precondition]
    AND [another precondition]
   WHEN [test step]
    AND [test step]
   THEN [verification step]
    AND [verification step]

**Additional context**

lmsurpre:
Yes, I've actually had this thought for a while, so I'm glad you've opened this enhancement request.  I'm not sure when we will get to it, but if you're interested in working with us on using more of CDI we're definitely open to the idea.

PR #2700, opened 2021-08-17
issue 2680 Allow control of the clock

MFAshby:
Fixes https://github.com/IBM/FHIR/issues/2680

This is useful for system tests in order to produce predictable outputs,
and to test boundary conditions around clock changes.

This should expose a new HTTP endpoint `/.clock`. GET requests will return the system's current date and time, PUT and DELETE requests will allow the operator to set or clear the current date and time, so long as the configuration permits a mutable clock. This switch should typically not be enabled in a production server.

PR is draft as I'm still testing, but I welcome early feedback on the approach. 

Typically I would prefer to use the dependency injection system over static functions, but a larger change is required to introduce dependency injection throughout the application to make this possible, see https://github.com/IBM/FHIR/issues/2687.

### DIFF
--- a/fhir-audit/src/main/java/com/ibm/fhir/audit/cadf/CadfReporterStep.java
+++ b/fhir-audit/src/main/java/com/ibm/fhir/audit/cadf/CadfReporterStep.java
@@ -22,6 +22,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Map;
 
+import com.ibm.fhir.core.clock.ClockFactory;
 import jakarta.json.Json;
 import jakarta.json.JsonArray;
 import jakarta.json.JsonObject;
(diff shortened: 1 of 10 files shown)


lmsurpre:
@MFAshby I see this PR is still in draft status...is this ready for review or are you still working on it?

In addition to flipping this out of Draft mode when ready, it would be great if you could amend the commit so that the sign-off matches the email address in your Author field.  I think you can do that by either changing the author field on the commit or by changing the signoff to your ashbysoft one.

```
Commit sha: d9cb7d1, Author: Martin Ashby, Committer: GitHub; Expected "Martin Ashby martin@ash**soft.com", but got "Martin Ashby MFAshby@users.noreply.github.com".
```
(** redacted by me)

MFAshby:
Hey @lmsurpre , it seems that unfortunately IBM FHIR server isn't going to be suitable for the project I'm on due to lack of STU3 support. It's unfortunate because IBM FHIR seems like one of the better open source java FHIR implementations available right now.

I'll try to get this PR tested and I'll adjust the signature as requested. The code itself should be ready for review, I'll mark it as such.

punktilious:
I like the idea of abstracting the clock but think we need to make this a bigger overhaul of how we handle time in general. There are places where we use currentTimeMillis but should be using nanoTime for time deltas etc.

lmsurpre:
I'm planning to close this one due to lack of activity, Robin's comment, and the lack of DCO signoff.  Feel free to pick it back up if/when you get a chance; thanks for the feedback and the time/effort that went into this.

PR #2701, opened 2021-08-18
add caching to TestUtil.getMinimalResource and standardize on this

lmsurpre:
Our object model is immutable and so its safe to re-use the same
resources across all the tests. This could avoid I/O, parse, and parse
validation time each time its re-used.

I had originally designed TestUtil.getMinimalResource to accept a single
argument of type ResourceType. However, I've always regretted that
because the generic type is more useful if it can be gleaned from an
input argument of type class<T extends Resource> instead of forcing java
to try figuring it out based on the call context.

Finally, we had quite a few search tests that were using the 
`getMinimalResourceType(ResourceType, Format.JSON)` variant.
However, which format the resource we're reading from the examples jar
is in is quite irrelevant to the actual tests, so now these will use the single
argument flavor (which supports this new caching behavior).

### DIFF
--- a/fhir-model/src/test/java/com/ibm/fhir/model/test/TestUtil.java
+++ b/fhir-model/src/test/java/com/ibm/fhir/model/test/TestUtil.java
@@ -18,13 +18,9 @@ import java.io.Reader;
 import java.io.StringWriter;
 import java.nio.charset.Charset;
 import java.util.List;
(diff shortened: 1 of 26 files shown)


JohnTimm (approved):
LGTM

michaelwschroeder (approved):
LGTM

Issue #2621, opened 2021-07-15
Provide overloaded setters that take native Java objects


lmsurpre:
**Is your feature request related to a problem? Please describe.**
fhir-model provides static factory helpers like String.of, Code.of, etc.  However, due to the decision to use the fhir names verbatim, some of the class names conflict with built-in java types (or heavily used types) which means you have to fully quality their usage.

For just this reason, we introduced static factory methods String.string, Boolean.bool, etc.
However, its annoying to explain to new users that they have to statically import those in each class they want to work with the model.

**Describe the solution you'd like**
Start by overriding all the setters in all the fhir-model builders where the argument type conflicts with a common java class name in the jre:
* String
* Integer
* Boolean
* Date
* Instant
* others?

The implementation of these setters would simply call our existing factory methods (at least for any non-null input).

**Describe alternatives you've considered**
Some kind of simplified java facade that is layered on top of the existing model but present a more java-friendly view (e.g. no primitive extensions, but the ability to drill down in the "real" model if/when needed).

**Acceptance Criteria**

1.  
   GIVEN [a precondition]
    AND [another precondition]
   WHEN [test step]
    AND [test step]
   THEN [verification step]
    AND [verification step]

**Additional context**
The Collection<> flavor of the setter cannot be overridden due to type erasure, but I think that is OK since this is really just a convenience mechanism anyway.

lmsurpre:
One unforseen consequence of adding the overloaded methods is that it can break client code that is calling an effected builder setter with null:

![image](https://user-images.githubusercontent.com/5478798/130004865-1dd7d47a-ace2-4830-8e42-c647340ed3dc.png)

Is it a deal-breaker?

lmsurpre:
Another interesting tidbit:  there seems to be a bug in Eclipse where it lists the overloaded method but presents the javadoc of the latter method for *both* instances:

![image](https://user-images.githubusercontent.com/5478798/130009383-e2a4b78f-1c47-402a-8d89-3bb4f1c0b8f3.png)

![image](https://user-images.githubusercontent.com/5478798/130009420-79b34aa3-76b1-4758-8b92-ef484025b8df.png)

Whichever one was declared last is the one the eclipse uses, so I think I'll put the new java-friendly overload setters *before* the existing ones.
This affects String, Boolean, and Integer overrides, but not Date and Instant ones (because those ones have Java equivalents that have different simple names).

prb112:
[...]

PR #2704, opened 2021-08-19
issue #2621 - add overloaded builder setters

lmsurpre:
Draft pull request to see how we like this...
See https://github.com/IBM/FHIR/blob/a7e6ad81c1f723a5541a83088be4993e1ff76508/fhir-model/src/test/java/com/ibm/fhir/model/test/ModelBuilderTest.java#L32-L50 for a quick overview of what this allows.

Note:  We probably need to update fhir-example-generator to have it handle all these new setter methods correctly.


fhir-server-test/src/test/java/com/ibm/fhir/server/test/FHIRPatchTest.java
```diff
@@ -112,7 +113,7 @@ public void testJSONPatchRemoveOperation() throws Exception {
 
         // create a copy of the patient and update it using the model API
         Patient.Builder patientBuilder = patient.toBuilder();
-        patientBuilder.active(null);
```
lmsurpre, line 115: an unfortunate consequence of adding the overloads

fhir-model/src/test/java/com/ibm/fhir/model/test/ModelBuilderTest.java
```diff
@@ -0,0 +1,91 @@
...
+    @Test(expectedExceptions = IllegalStateException.class)
+    public void testVarArgWithNull() {
+        Basic.builder()
+            .code(CodeableConcept.builder()
+                .coding((Coding)null)
+                .build())
+            .build();
+    }
+
+    // This one throws a NullPointerException rather than the expected IllegalStateException...is that an issue?
```
lmsurpre, line 82: is this an issue?

lmsurpre: personally I think its fine to throw the NPE in this case, but maybe we could add it to the javadoc

JohnTimm (approved):
LGTM

PR #2705, opened 2021-08-19
WHFFHIR-207 Write to a protected directory that is not needed, and fails when running SchemaTool/Db2 with a restricted user in openshift

prb112:


tbieste (approved):
LGTM

Issue #2717, opened 2021-08-24
Composition reports a 500 when it's a bad user request.


prb112:
**Describe the bug**
Composition reports a 500 when it's a bad user request.

**Environment**
Which version of IBM FHIR Server?
4.9.0

**To Reproduce**

**Expected behavior**

**Additional context**

prb112:
Passes QA
![image](https://user-images.githubusercontent.com/3016328/131526576-5bd93a0a-78b0-41f0-b1e7-e719c7a975b4.png)

Issue #2715, opened 2021-08-24
Bulk Export Status


dabimahesh:
While doing bulkdata status request  "error":[], should be a part of the response, even if there is no error in this request. Please help us in this. This is required for ONC MU Inferno testing tool.

https://chat.fhir.org/#narrow/stream/212434-ibm/topic/releases/near/250080961

Thanks
Mahesh

prb112:
Thanks, this is an enhancement.

lmsurpre:
Confirmed that, upon successful export, the bulkdata-status endpoint response now contains `"error": []`

Issue #2714, opened 2021-08-24
Bulk Export Status Report


dabimahesh:
Regarding one more scenario for bulk export

A scenario while testing in Inferno

Request URL:
https://***:9443/fhir-server/api/v4/$bulkdata-status?job=HRheZqFY_KT5TXhdXFuY7g

Response Body:
{
"transactionTime": "2021-08-18T10:21:53.82Z",
"request": "https://***:9443/fhir-server/api/v4/Group/17b58c31108-77f6763e-008b-4010-b428-c42ee061aa57/$export",
"requiresAccessToken": false,
"output": [{
"type": "AllergyIntolerance",
"url": "_jVMAj-5zOenYBhcTN5pa5b6DvR2rsx2Cpv4l9wuXOY/AllergyIntolerance_1.ndjson",
"count": 10
},
{
"type": "CarePlan",
"url": "_jVMAj-5zOenYBhcTN5pa5b6DvR2rsx2Cpv4l9wuXOY/CarePlan_1.ndjson",
"count": 2
}
]
}

This test case is a pass, except that it needs few extra things in the response as mentioned below:

 "requiresAccessToken": false. This parameter in the response is false, we need to return this in response as true. We have done all the required changes but are not able to get this. Seems that it is fixed value.
Could you please let us know how to achieve this?

https://chat.fhir.org/#narrow/stream/212434-ibm/topic/releases/near/250499231

prb112:
Marked this as an enhancement.

prb112:
Adding an entry to the storageProvider`fhirServer/bulkdata/storageProviders/" + provider + "/requiresAccessToken`

lmsurpre:
I confirmed that the `"requiresAccessToken"` in the output now takes it value from this new config property and that the default is still `false`.

Issue #2709, opened 2021-08-24
bulkdata preflight checks should verify output provider config


lmsurpre:
**Is your feature request related to a problem? Please describe.**
I submitted a bulk export job against a configure s3 storageProvider.  It accepted the job but then failed with this:
```
StepChunkListener: job[bulkexportfastjob/1/1] --- com.ibm.cloud.objectstorage.services.s3.model.AmazonS3Exception: The specified bucket does not exist. (Service: Amazon S3; Status Code: 404; Error Code: NoSuchBucket; Request ID: 50967bf1-b6f6-4ae3-acf2-78598f2f4d36), S3 Extended Request ID: null
com.ibm.jbatch.container.exception.BatchContainerRuntimeException: com.ibm.cloud.objectstorage.services.s3.model.AmazonS3Exception: The specified bucket does not exist. (Service: Amazon S3; Status Code: 404; Error Code: NoSuchBucket; Request ID: 50967bf1-b6f6-4ae3-acf2-78598f2f4d36), S3 Extended Request ID: null
	at com.ibm.jbatch.container.controller.impl.ChunkStepControllerImpl.readItem(ChunkStepControllerImpl.java:360)
	at com.ibm.jbatch.container.controller.impl.ChunkStepControllerImpl.readAndProcess(ChunkStepControllerImpl.java:245)
	at com.ibm.jbatch.container.controller.impl.ChunkStepControllerImpl.invokeChunk(ChunkStepControllerImpl.java:626)
	at com.ibm.jbatch.container.controller.impl.ChunkStepControllerImpl.invokeCoreStep(ChunkStepControllerImpl.java:795)
	at com.ibm.jbatch.container.controller.impl.BaseStepControllerImpl.execute(BaseStepControllerImpl.java:295)
	at com.ibm.jbatch.container.controller.impl.ExecutionTransitioner.doExecutionLoop(ExecutionTransitioner.java:118)
	at com.ibm.jbatch.container.controller.impl.WorkUnitThreadControllerImpl.executeCoreTransitionLoop(WorkUnitThreadControllerImpl.java:96)
	at com.ibm.jbatch.container.controller.impl.WorkUnitThreadControllerImpl.executeWorkUnit(WorkUnitThreadControllerImpl.java:178)
	at com.ibm.jbatch.container.controller.impl.WorkUnitThreadControllerImpl$AbstractControllerHelper.runExecutionOnThread(WorkUnitThreadControllerImpl.java:503)
[...]

PR #2719, opened 2021-08-24
Minor Fixes

prb112:
- Bulk Export Status #2715
	- Added a default Error
- Bulk Export Status Report #2714
	- Added property for requiresAccessToken
- Composition reports a 500 when it's a bad user request #2717
	- Mutate 500 to 400 in $document
	- Add Test for $document when bad request
- add patch integration tests for Move and Copy to additionally confirm behaviors 
-  For the JSON Patch path for the test option, it is not handled correctly #2720 
     - Skippable updates enabled for patch scenarios 
-  bulkdata preflight checks should verify output provider config #2709 
     - put into an exception handling due to virtualhost and api host handling. 
 
### DIFF
--- a/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
@@ -478,7 +478,8 @@ public class FHIRRestHelper implements FHIRResourceHelpers {
             if (ior.getPrevResource() != null) {
                 performVersionAwareUpdateCheck(ior.getPrevResource(), ifMatchValue);
 
-                if (skippableUpdate && !isDeleted) {
+                // In the case of a patch, we should not be updating meaninglessly.
+                if ((skippableUpdate || patch != null) && !isDeleted) {
                     ResourceFingerprintVisitor fingerprinter = new ResourceFingerprintVisitor();
                     ior.getPrevResource().accept(fingerprinter);
                     SaltHash baseline = fingerprinter.getSaltAndHash();
--- a/operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/client/BulkDataClient.java
+++ b/operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/client/BulkDataClient.java
@@ -354,7 +354,7 @@ public class BulkDataClient {
                 if (OperationConstants.FAILED_BAD_SOURCE.equals(bulkExportJobExecutionResponse.getExitStatus())) {
                     throw export.buildOperationException("A bad source input was used during a call to $import", IssueType.INVALID);
                 } else if (OperationConstants.NO_SUCH_BUCKET.equals(bulkExportJobExecutionResponse.getExitStatus())) {
-                    throw export.buildOperationException("No such bucket exists for the storageProvider", IssueType.EXCEPTION);
+                    throw export.buildOperationException("No such bucket exists for the storageProvider", IssueType.NO_STORE);
                 } else {
                     throw export.buildOperationException("The job has failed", IssueType.EXCEPTION);
                 }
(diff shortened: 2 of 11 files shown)


JohnTimm (approved):
LGTM

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/client/BulkDataClient.java
```diff
@@ -734,6 +739,10 @@ private PollingLocationResponse process(JobExecutionResponse response) {
                 }
             }
             result.setOutput(outputList);
+
+            // Errors need to be added.
+            List<PollingLocationResponse.Output> errors = Collections.emptyList();
+            result.setError(errors);
```
lmsurpre, line 745: do we ever set a non-empty list?  i thought we would put OperationOutcome for an resources that failed to export into this list.

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/client/BulkDataClient.java
```diff
@@ -353,6 +353,8 @@ public PollingLocationResponse status(String job) throws Exception {
                  */
                 if (OperationConstants.FAILED_BAD_SOURCE.equals(bulkExportJobExecutionResponse.getExitStatus())) {
                     throw export.buildOperationException("A bad source input was used during a call to $import", IssueType.INVALID);
+                } else if (OperationConstants.NO_SUCH_BUCKET.equals(bulkExportJobExecutionResponse.getExitStatus())) {
+                    throw export.buildOperationException("No such bucket exists for the storageProvider", IssueType.INVALID);
```
lmsurpre, line 357: will this now result in a 400 error?  it seems like the 500 was actually better here because its actually a server config issue, right?

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -478,7 +478,8 @@ private FHIRRestOperationResponse doPatchOrUpdate(String type, String id, FHIRPa
             if (ior.getPrevResource() != null) {
                 performVersionAwareUpdateCheck(ior.getPrevResource(), ifMatchValue);
 
-                if (skippableUpdate && !isDeleted) {
+                // In the case of a patch, we should not be updating meaninglessly.
+                if ((skippableUpdate || patch != null) && !isDeleted) {
```
lmsurpre, line 482: I wonder if we should eventually make skippableUpdate the default (instead of forcing clients to opt-in)

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/client/BulkDataClient.java
prb112: We don't.  The only time we use it is when we do the Bulk Import.  We have a backlog item  Add operationoutcomes for errors during bulkdata export #1161

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/client/BulkDataClient.java
prb112: Right, let me switch the type.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
prb112: It's probably the preferred way to do it - definitely reduces burden.

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/client/BulkDataClient.java
prb112: pushed the change.

prb112: Changed to IssueType.NO_STORE

lmsurpre (approved):
LGTM

Issue #2713, opened 2021-08-24
BulkData: Delete Job


dabimahesh:
We are getting "An unexpected error has occurred while deleting the job", when a DELETE is initiated. Is it that we are missing any configuration?

prb112:
> We are getting "An unexpected error has occurred while deleting the job", when a DELETE is initiated. Is it that we are missing any configuration?

@dabimahesh Thank you for opening

A few questions: 
1 - What is the topology of your environment? Do you have more than one node? 
2 - can you share your logs? messages.log and trace.log
3 - Can you share the rest call you executed? 

With these questions asked, I can more easily replicate your issue locally.

dabimahesh:
Hello Paul,
     Please find below some points to your questions:

     1 - We have one server on ubuntu and we are using Postgre as DB. We are in the process of testing IBM FHIR server on Inferno testing tool

     2 - We are using SmartOnFHIR and the token used is as below
		{
		  "iss": "https://identityserverforfhir.azurewebsites.net",
		  "nbf": 1629885567,
		  "iat": 1629885567,
		  "exp": 1629889167,
		  "aud": [
			"https://fhirserverurl:9443/fhir-server/api/v4",
			"Inferno"
		  ],
		  "scope": "system/AllergyIntolerance.read system/CarePlan.read system/CareTeam.read system/Condition.read system/Device.read system/DiagnosticReport.read system/DocumentReference.read system/Encounter.read system/Goal.read system/Immunization.read system/Location.read system/Medication.read system/MedicationRequest.read system/Observation.read system/Organization.read system/Patient.read system/Practitioner.read system/Procedure.read system/Provenance.read",
		  "group": [
			"FHIRUsers",
			"FHIRAdmins"
		  ],
		  "client_id": "Inferno-Bulk",
		  "patient": "17b5dc09184-54ce1834-2501-4e85-a3fe-e04223da69cf",
		  "patient_id": "17b5dc09184-54ce1834-2501-4e85-a3fe-e04223da69cf",
		  "upn": "fhiruser",
		  "need_patient_banner": "false",
		  "smart_style_url": "https://identityserverforfhir.azurewebsites.net/SmartOnFHIR/v1.json",
		  "jti": "1BA86FE46C70CFA13522743DFAB4EACA"
		}


     3 - Operation GET (worked fine)                   
		
[...]

PR #2727, opened 2021-08-26
BulkData: Delete Job #2713

prb112:

### DIFF
--- a/operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/client/action/batch/BatchCancelRequestAction.java
+++ b/operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/client/action/batch/BatchCancelRequestAction.java
@@ -19,11 +19,16 @@ import javax.ws.rs.core.Response.Status;
 
 import org.apache.commons.io.IOUtils;
 import org.apache.http.HttpEntity;
+import org.apache.http.auth.AuthScope;
+import org.apache.http.auth.UsernamePasswordCredentials;
 import org.apache.http.client.ClientProtocolException;
+import org.apache.http.client.CredentialsProvider;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpDelete;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpPut;
+import org.apache.http.client.protocol.HttpClientContext;
+import org.apache.http.impl.client.BasicCredentialsProvider;
 import org.apache.http.impl.client.BasicResponseHandler;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.util.EntityUtils;
@@ -60,11 +65,21 @@ public class BatchCancelRequestAction implements BulkDataClientAction {
     private CloseableHttpClient cli = null;
     private String batchUrl = null;
 
+    private HttpClientContext ctx = null;
+
     private Response.Status result = Response.Status.FORBIDDEN;
 
     @Override
     public void prepare(CloseableHttpClient cli, String batchUrl, String batchUser, String batchPass) {
         this.cli = cli;
+
+        // Configure the local context.
+        CredentialsProvider credsProvider = new BasicCredentialsProvider();
+        credsProvider.setCredentials(AuthScope.ANY,
+                new UsernamePasswordCredentials(batchUser, batchPass));
+        ctx = HttpClientContext.create();
+        ctx.setCredentialsProvider(credsProvider);
+
         this.batchUrl = batchUrl;
     }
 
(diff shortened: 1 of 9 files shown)


operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/client/action/batch/BatchCancelRequestAction.java
```diff
@@ -0,0 +1,432 @@
...
+
+        // Stop the Job's execution across all JobExecutions
+        stopJobExecutions(jobExecutions);
+
+        if (supportsDeleteJob()) {
+            // Delete The Job
+            deleteJob(job);
+        } else {
+            // The following is marked as not-found
+            throw export.buildOperationException("The Job is stopped intentionally with an in memory database.", IssueType.NOT_FOUND);
```
lmsurpre, line 122: please help me understand this one

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/client/action/batch/BatchCancelRequestAction.java
```diff
@@ -0,0 +1,432 @@
...
+        return result;
+    }
+
+    @Override
+    public void close() throws FHIROperationException {
+        if (cli != null) {
+            try {
+                cli.close();
+            } catch (IOException ex) {
+                LOG.throwing(CLASSNAME, "close", ex);
```
lmsurpre, line 100: I see this pattern quite a bit in this PR and wanted to confirm:  throwing is logged at FINER level, meaning that we won't normally see the root exception in the logs... is that ok?
if not, then either
A. these should be logged at INFO or WARNING; or 
B. we should include the exception as the cause of the new OperationException we're building

is there a good reason to avoid B?

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/model/JobInstanceResponse.java
```diff
@@ -272,96 +273,123 @@ private Parser() {
...
-
                 return JobInstanceResponse.Parser.parse(in);
+            } catch (Exception e) {
+                throw new FHIROperationException(
+                        "Problem parsing the submission response from the job server", e);
+            }
+        }
 
+        /**
+         * parsers an array of JobInstanceResponses
```
lmsurpre, line 284:
```suggestion
         * parses an array of JobInstanceResponses
```

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/client/action/batch/BatchCancelRequestAction.java
prb112: Right, that's the intent - each is logged at the finer level. In the case of the Close, I opted not to additionally log it out as the Close will then work silently. In many cases, I repack the exception or throw a specific exception in the PR

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/client/action/batch/BatchCancelRequestAction.java
prb112: Chatted with Lee directly. 

The in Memory Persistence Impl that is used with JavaBatch does not support a the 'delete' (purgeJob) and always returns false which results in a 500. This is designed to signal to the user the reason and signal that it is deleted (when in fact it is just stopped).

prb112: It'll only happen in this case when it's a non-production system.

lmsurpre: one idea I had is to change our default to use derby instead of this special in-memory config.  however, since neither one is recommended for production, i think its probably fine as-is.

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/client/action/batch/BatchCancelRequestAction.java
lmsurpre: yeah, I wasn't so concerned with losing this particular exception (which i doubt we'll ever hit).  more just a concern with the pattern...I'll go through and double-check the other instances I saw

lmsurpre (approved):
LGTM

PR #2729, opened 2021-08-27
Include slf4j binding in bulkdata webapp instead of with parquet deps

lmsurpre:
I also removed the unused fhir-server "provided" dependency. This should prevent us from accidentally using classes from fhir-server within the fhir-bulkdata-webapp... the two webapps have isolated classloaders and so its not right to say fhir-server is a provided dependency.

### DIFF
--- a/fhir-bulkdata-webapp/pom.xml
+++ b/fhir-bulkdata-webapp/pom.xml
@@ -1,17 +1,17 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-    
+
     <properties>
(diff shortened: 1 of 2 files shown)


fhir-bulkdata-webapp/pom.xml
```diff
@@ -200,7 +193,6 @@
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-jdk14</artifactId>
-            <scope>provided</scope>
```
prb112, line 203: Would there ever be any terminology operations on this?

lmsurpre: maybe if we get a _typeFilter with some valueset modifiers like :in, :not-in.  potentially with an implicit valueset.

prb112: That's the only case I can see it injecting a conflict.

prb112 (approved):

lmsurpre:
I putzed with this for a while but wasn't able to come up with a fully working configuration via userlib jars without the nop binding.

Issue #2266, opened 2021-04-22
WARNING: An illegal reflective access operation has occurred


rubberduck203:
**Describe the bug**

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.ibm.fhir.model.util.XMLSupport (file:/Users/rubberduck/.gradle/caches/modules-2/files-2.1/com.ibm.fhir/fhir-model/4.7.0/706afb6278d6bf5e56dec035ea17a0f326f41bba/fhir-model-4.7.0.jar) to constructor com.sun.xml.internal.stream.XMLInputFactoryImpl()
```

**Environment**
Which version of IBM FHIR Server?

`4.7.0`

**To Reproduce**

I'm not entirely sure.
Would be happy to help reproduce it given some guidance.

I'm running OpenJDK 11 and it seems that this behavior was introduced in JDK 9.
If you're using an older JDK, you might not see this.

**Expected behavior**

No warnings.

**Additional context**

This StackOverflow Q&A may help someone who understands the problem better than I.
https://stackoverflow.com/questions/50251798/what-is-an-illegal-reflective-access

prb112:
Thanks @rubberduck203 we're aware of the warning.  It's related to the XML library which we is dependent upon uses Reflection to access the xml model in memory.

rubberduck203:
Cool. Thanks @prb112.
I didn't see an issue anywhere, so wanted to make sure you were aware.

JohnTimm:
We are currently using the JVM default implementations of `XMLInputFactory`, `XMLOutputFactory`, and `TransformerFactory`: `com.sun.xml.internal.stream.XMLInputFactoryImpl`, `com.sun.xml.internal.stream.XMLOutputFactoryImpl`, and `com.sun.org.apache.xalan.internal.xsltc.trax.TransformerFactoryImpl` respectively. We found that other, 3rd party implementations cause problems with how we parse / generate XML. I recently retested using Woodstox, Xalan-J, and Saxon without issues. For this particular issue, I think we should upgrade to Java 9 first and then see where things stand. If it turns out that we need to move to a different implementation to avoid the reflective access warning, then we can decide to do so at that time.

rubberduck203:
Just FYI, as of Java 17, this won't even compile anymore.
The 17 LTS is scheduled for release Sept 2021.

prb112:
Confirmed with JDK 11 it no longer is reported.

[...]

PR #2738, opened 2021-09-02
issue #2266 - avoid reflection while instantiating XML factories

lmsurpre:
The javadocs say that these system properties are the first things consulted.
I think we shied away from relying on system properties for proper functionality in the past,
but with Java 17 around the corner it would be nice to remove the reflection here before it breaks.

If we drop java 8 support, we could possibly just use these methods instead:
* https://docs.oracle.com/en/java/javase/11/docs/api/java.xml/javax/xml/stream/XMLInputFactory.html#newDefaultFactory()
* https://docs.oracle.com/en/java/javase/11/docs/api/java.xml/javax/xml/stream/XMLOutputFactory.html#newDefaultFactory()
* https://docs.oracle.com/en/java/javase/11/docs/api/java.xml/javax/xml/transform/TransformerFactory.html#newDefaultInstance()

### DIFF
--- a/fhir-model/src/main/java/com/ibm/fhir/model/util/XMLSupport.java
+++ b/fhir-model/src/main/java/com/ibm/fhir/model/util/XMLSupport.java
@@ -27,6 +27,14 @@ public final class XMLSupport {
     public static final String FHIR_NS_URI = "http://hl7.org/fhir";
     public static final String XHTML_NS_URI = "http://www.w3.org/1999/xhtml";
 
+    private static final String PROP_XML_INPUT_FACTORY = "javax.xml.stream.XMLInputFactory";
+    private static final String PROP_XML_OUTPUT_FACTORY = "javax.xml.stream.XMLOutputFactory";
+    private static final String PROP_TRANSFORMER_FACTORY = "javax.xml.stream.XMLTransformerFactory";
+
+    private static final String XML_INPUT_FACTORY_IMPL = "com.sun.xml.internal.stream.XMLInputFactoryImpl";
+    private static final String XML_OUTPUT_FACTORY_IMPL = "com.sun.xml.internal.stream.XMLOutputFactoryImpl";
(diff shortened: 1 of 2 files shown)


fhir-model/src/main/java/com/ibm/fhir/model/util/XMLSupport.java
```diff
@@ -197,15 +209,32 @@ private static XMLInputFactory createXMLInputFactory() {
 
     private static XMLOutputFactory createXMLOutputFactory() {
         try {
-            return (XMLOutputFactory) Class.forName(XML_OUTPUT_FACTORY_IMPL).getConstructor().newInstance();
+            boolean isSet = System.getProperty(PROP_XML_OUTPUT_FACTORY) != null;
+            if (!isSet) {
+                System.setProperty(PROP_XML_OUTPUT_FACTORY, XML_OUTPUT_FACTORY_IMPL);
+            }
+            XMLOutputFactory factory = XMLOutputFactory.newFactory();
```
punktilious, line 216: Just a thought here (although perhaps this is overly conservative)...if `!isSet`, then do a little loop and check the result of getProperty matches the  `XML_OUTPUT_FACTORY_IMPL` value after the newFactory call. As discussed, the race condition here is pretty theoretical, so this may be overkill.

prb112 (approved):

fhir-model/src/main/java/com/ibm/fhir/model/util/XMLSupport.java
lmsurpre: looping seems overkill to me, but it wouldn't hurt to check the value after we instantiate the factory.  the question is what would we do if it doesn't match...throw?

lmsurpre: I'm thinking maybe just a WARNING message for that case, but the more I think about it, the more I feel it shouldn't be needed.  I think we should merge as-is, but do let me know if you want the added check.

punktilious: I'm OK as-is...just throwing out a possible option for completeness.

Issue #2504, opened 2021-06-14
Enforce OperationDefinition affectsState


prb112:
**Is your feature request related to a problem? Please describe.**
Enforce OperationDefinition affectsState per the Specification 

```
OperationDefinition.affectsState
--
Element Id | OperationDefinition.affectsState
Definition | Whether the operation affects state. Side effects such as producing audit trail entries do not count as 'affecting  state'.
Cardinality | 0..1
Type | boolean
Summary | true
Comments | What http methods can be used for the  operation depends on the .affectsState value and whether the input  parameters are primitive or complex:   Servers SHALL support POST method for all operations.   Servers SHALL support GET method if all the parameters for the  operation are primitive or there are no parameters and the operation has  affectsState = false.
```

https://www.hl7.org/fhir/operationdefinition-definitions.html#OperationDefinition.affectsState

**Describe the solution you'd like**
The FHIROperation framework should read the affectsState per the Specification, and enforce the use of get/post

**Describe alternatives you've considered**

**Acceptance Criteria**

1.  
   GIVEN [a precondition]
    AND [another precondition]
   WHEN [test step]
    AND [test step]
   THEN [verification step]
    AND [verification step]

**Additional context**
Per Comment https://github.com/IBM/FHIR/pull/2489#discussion_r650881272

PR #2744, opened 2021-09-07
Issue #2504 - Check affectsState

tbieste:
Issue #2504 - Check affectsState

### DIFF
--- a/fhir-server/src/main/java/com/ibm/fhir/server/operation/spi/AbstractOperation.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/operation/spi/AbstractOperation.java
@@ -14,6 +14,7 @@ import com.ibm.fhir.exception.FHIROperationException;
 import com.ibm.fhir.model.resource.OperationDefinition;
 import com.ibm.fhir.model.resource.OperationOutcome;
 import com.ibm.fhir.model.resource.Parameters;
+import com.ibm.fhir.model.resource.Parameters.Parameter;
 import com.ibm.fhir.model.resource.Resource;
 import com.ibm.fhir.model.type.code.IssueType;
 import com.ibm.fhir.model.type.code.OperationParameterUse;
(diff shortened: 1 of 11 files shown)


fhir-server/src/main/java/com/ibm/fhir/server/operation/spi/AbstractOperation.java
```diff
@@ -157,6 +168,45 @@ protected void validateOperationContext(FHIROperationContext operationContext, C
...
 
+    /**
+     * Determines if the operation disallows the GET method.
+     * This is determined by the affectsState value of the OperatorDefinition and whether the
+     * OperatorDefinition contains any non-primitive parameters.
+     * @param operationDefinition the operation definition
+     * @param parameters the parameters
+     * @return true or false
+     */
+    private boolean isGetMethodAllowed(OperationDefinition operationDefinition, Parameters parameters) {
```
prb112, line 179: The OperationDefinitions probably don't need to be reprocessed for each request.  Maybe in this case it should be cached?  Since this is at the tail end of the expression, I have some doubts about the need.  

A cache feels ideal for the below... once warmed up, the cacheis sued.

fhir-server/src/main/java/com/ibm/fhir/server/operation/spi/AbstractOperation.java
```diff
@@ -157,6 +168,45 @@ protected void validateOperationContext(FHIROperationContext operationContext, C
...
+        return true;
+    }
+
+    /**
+     * Determines if any methods (except GET and POST) are allowed for the operation.
+     * This can be overridden by an operation to allow additional methods.
+     * @return true or false
+     */
+    protected boolean isAdditionalMethodAllowed(String method) {
+        return false;
```
prb112, line 207: Does this mean it blocks bulkdata?

prb112: Got it... found below... it's overriden in extended classes

prb112 (approved):
Looks good....

fhir-server/src/main/java/com/ibm/fhir/server/operation/spi/AbstractOperation.java
tbieste: Since the passed in parameters on the request can be different on each request, they would still need to be checked each time, so I'm not sure what we'd be able to cache.

prb112: you could minimally cache the OperationDefinition rather than process it each time. a point lookup after the processing of the OperationDefinition is then possible afterwards.

PR #2745, opened 2021-09-08
Deprecate our old static jdbc caches for future removal

lmsurpre:
These were largely replaced by JDBCIdentityCache and the JDBCIdentityCacheImpl.

A couple of the older caches are still used by live code, so we should only
deprecate them if we know that they aren't really needed.

For example, the ParameterNamesCache is used (transitively) from
JDBCIdentityCacheImpl.getParameterNameId.


michaelwschroeder (approved):
LGTM

prb112 (approved):

PR #2749, opened 2021-09-08
FHIRPathSpecTest update and related changes

lmsurpre:
1. Update to the latest from
https://github.com/FHIR/fhir-test-cases/blob/master/r4/fhirpath/tests-fhir-r4.xml
2. Comment out the test cases we consider to be invalid (and add
comments to explain why)
3. Minor fixes to the FHIRPathSpecTest harness
4. Minor fixes to our FHIRPath engine
   * add guard in FHIRPathAbstractTemporalValue.isComparableTo
   * FHIRPathEvaluator.exists fix for applying predicate against all
context nodes (not just the first one)
   * Update allTrue() to throw when called with non-boolean inputs
   * rename private validateEqualityOperands helper

Update results:
```
Total tests run: 723, Passes: 544, Failures: 0, Skips: 179
```


fhir-path/src/main/java/com/ibm/fhir/path/evaluator/FHIRPathEvaluator.java
```diff
@@ -873,7 +883,9 @@ private IllegalArgumentException unexpectedNumberOfArguments(int arity, String f
                 return afterEvaluation(ctx, SINGLETON_FALSE);
             }
 
-            if (!validateEqualityOperands(left, right)) {
+            // for quantity operands: Attempting to operate on quantities with invalid units will result in empty ({ }).
+            // for temporal operands: If one input has a value for the precision and the other does not, the comparison stops and the result is empty ({ })
+            if (!isEqualityOperationValid(left, right)) {
```
JohnTimm, line 888: we are really checking the operands here. I think perhaps:
```java
if (!equalityOperandsValid(left, right))
```

JohnTimm (approved):
LGTM

PR #2757, opened 2021-09-14
Mark janusgraph-cql and janusgraph-es as provided

lmsurpre:
When we bumped to JanusGraph 0.6.0, our webapp went from 82MB to 126MB.
Most of this is from the `janusgraph-cql` module pulling in all of
Cassandra.
Since this is an experimental feature, I think it makes sense to require
user action to enable cassandra and elasticsearch options for graphs.
In our experience, the berkeleyje + lucene option performs better and is
easier to manage anyway.

Note: we don't actually have any compile-time dependencies on these
modules at this time, but I opted to mark the as provided (instead of
omitting them entirely) to make it as easy as possible to bring them back in.
This is similar to what we did with our optional Spark dependencies in
fhir-bulkdata-webapp.


prb112 (approved):
LGTM, there should be a docs update too on the Terminology guide?

ldettman (approved):
LGTM

Issue #2758, opened 2021-09-15
Using fhir-persistence-schema CLI tool to drop schema gives an error


d0roppe:
**Describe the bug**
using the fhir-persistance-schema CLI command to drop fhir schemas hits a bug
com.ibm.fhir.database.utils.api.DataAccessException: org.postgresql.util.PSQLException: ERROR: cannot drop table fhirdata.logical_resources because other objects depend on it
  Detail: constraint fk_logical_resource_tags_lr on table fhirdata.logical_resource_tags depends on table fhirdata.logical_resources
constraint fk_logical_resource_security_lr on table fhirdata.logical_resource_security depends on table fhirdata.logical_resources
  Hint: Use DROP ... CASCADE to drop the dependent objects too.
	at com.ibm.fhir.database.utils.postgres.PostgresTranslator.translate(PostgresTranslator.java:104)
	at com.ibm.fhir.database.utils.common.JdbcTarget.runStatement(JdbcTarget.java:48)
	at com.ibm.fhir.database.utils.common.CommonDatabaseAdapter.runStatement(CommonDatabaseAdapter.java:243)
	at com.ibm.fhir.database.utils.common.CommonDatabaseAdapter.dropTable(CommonDatabaseAdapter.java:282)
	at com.ibm.fhir.database.utils.model.Table.drop(Table.java:176)
	at com.ibm.fhir.database.utils.model.PhysicalDataModel.drop(PhysicalDataModel.java:185)
	at com.ibm.fhir.schema.app.Main.dropSchema(Main.java:502)
	at com.ibm.fhir.schema.app.Main.process(Main.java:2117)
	at com.ibm.fhir.schema.app.Main.main(Main.java:2218)
Caused by: org.postgresql.util.PSQLException: ERROR: cannot drop table fhirdata.logical_resources because other objects depend on it
  Detail: constraint fk_logical_resource_tags_lr on table fhirdata.logical_resource_tags depends on table fhirdata.logical_resources
**Environment**
Latest main and postgreSQL

**To Reproduce**

**Expected behavior**

**Additional context**
The issue highlights a testing gap in our CI/CD.  We should add a drop schema step that runs against each db...probably by extending the current migration tests with an extra step at the end.

lmsurpre:
Also relates to #2136

prb112:
test commands 
```
[...]

PR #2775, opened 2021-09-17
fhir-persistence-schema CLI drop schema gives an error #2758

prb112:
- Add missing tags to whole system tables to enforce a FK drop prior to
table drop
- Add tags to the OAuthSchemaGenerator and JavaBatchGenerator (drop
wasn't working without the tags)
- Tags are unique for the Generator

No schema change this is in the runtime of the CLI only.


fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/JavaBatchSchemaGenerator.java
```diff
@@ -142,23 +142,26 @@ VALUE VARCHAR(255),
...
                 .addVarcharColumn(VALUE, 4096, NULL) // VALUE VARCHAR(255) (updated to 4096)
                 .addBigIntColumn(FK_JOBEXECID, NOT_NULL) // FK_JOBEXECID BIGINT
                 .addForeignKeyConstraintAltTarget("JBPRMETERFKJBXECID", schemaName, JOBEXECUTION_TABLE, JOBEXECID, FK_JOBEXECID)
                 .addIndex("JP_FKJOBEXECID_IX", FK_JOBEXECID)
                 .addPrivileges(generateGroupPrivilege())
                 .build(model);
-        model.addTable(parameter);
-        model.addObject(parameter);
+
+        table.addTag(FhirSchemaGenerator.SCHEMA_GROUP_TAG, FhirSchemaGenerator.FHIRDATA_GROUP);
```
lmsurpre, line 154: should we use a different group id for the jbatch schema objects?

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/OAuthSchemaGenerator.java
```diff
@@ -102,6 +102,7 @@ protected void addCacheTable(PhysicalDataModel model) {
                 .addUniqueIndex(OAUTH20CACHE + "_" + EXPIRES, EXPIRES) // ASC is the default
                 .addPrivileges(generateGroupPrivilege())
                 .build(model);
+        cache.addTag(FhirSchemaGenerator.SCHEMA_GROUP_TAG, FhirSchemaGenerator.FHIRDATA_GROUP);
```
lmsurpre, line 105: should we use a different group id for the OAuth schema objects?

prb112: I think that implied much more work were we split out the Physical Data model.  If not true, is that OK to change... @punktilious ?

lmsurpre: yeah, i just had a quick look in Main and we have logic like this:
```
if (dropFhirSchema || dropOauthSchema || dropJavaBatchSchema) {
            logger.info("Dropping FK constraints in the data schema: " + this.schema.getSchemaName());
            dropForeignKeyConstraints(pdm, FhirSchemaGenerator.SCHEMA_GROUP_TAG, FhirSchemaGenerator.FHIRDATA_GROUP);
        }

...

if (dropFhirSchema || dropOauthSchema || dropJavaBatchSchema) {
                        // Just drop the objects associated with the FHIRDATA schema group
                        pdm.drop(adapter, FhirSchemaGenerator.SCHEMA_GROUP_TAG, FhirSchemaGenerator.FHIRDATA_GROUP);
                    }
```

I guess I don't really understand what the "group" is supposed to signify here.

prb112: I tried it out. It'll fail miserably. it has to be part of separating the physical model.

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/JavaBatchSchemaGenerator.java
prb112: see the following comment about PhysicalDataModel

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/OAuthSchemaGenerator.java
prb112: it's part of the PhysicalDataModel

lmsurpre: Do we know anywhere else that the SCHEMA_GROUP_TAG is used?

prb112: got it to work.... changed the tag names

lmsurpre: yes.  this looks exactly how i was expecting it.  marking LGTM but lets wait for robin's review to be sure we're not missing something!

lmsurpre (approved):
LGTM

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirSchemaGenerator.java
```diff
@@ -717,6 +717,9 @@ public Table addLogicalResourceTags(PhysicalDataModel pdm) {
                 })
                 .build(pdm);
 
+        tbl.addTag(SCHEMA_GROUP_TAG, FHIRDATA_GROUP);
+        this.procedureDependencies.add(tbl);
+
         pdm.addTable(tbl);
         pdm.addObject(tbl);
```
punktilious, line 724: Do you think it's worth adding a check to make sure there's at least one tag on the object? A better change (although a lot more work) would be to make tagging part of the builder and then we could make it mandatory.

punktilious (commented):
Should we extend the schema migration test to include a drop phase...then we can check that all the objects have been dropped.

fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirSchemaGenerator.java
prb112: I've added the checks in place to tests, and modified where the tags are missing. 
The build will fail without it.

PR #2790, opened 2021-09-21
introduce disabled configDropin for a derby-backed job repository

lmsurpre:
In the future, we may want to use this as the default instead of the default in-memory store because its better aligned with our default db config.
The biggest advantage is that the derby-backed store supports delete (purge) whereas the in-memory store does not.


prb112:
makes no functional changes (offers a new bulkdata testing dropin)

Issue #2454, opened 2021-06-02
Potential race condition when create overlaps with erase (hard-delete)


punktilious:
**Describe the bug**
The persistence layer contains logic to handle race conditions when two threads attempt to create a new resource at the same time. The current implementation handles the locking/create/update sequence correctly with the proviso that a logical_resource could never be deleted.

This is no longer true now that the IBM FHIR Server supports `$erase` hard-delete. Although the chance of create/erase conflicting is low, for completeness the locking/create/update logic should be enhanced to handle the possibility of a row being deleted between the (failed) create and update steps. This needs to be implemented as a loop.

**Environment**
4.8.2

**To Reproduce**
Run concurrent create/update/erase operations

**Expected behavior**
Data consistency at all times.

lmsurpre:
Probably should start with the test.  Maybe a fork of our ConcurrentUpdateTest.

prb112:
I found no concurrency issues - opened PR with the test cases which mixes the workload.
PR https://github.com/IBM/FHIR/pull/2792

PR #2792, opened 2021-09-21
Potential race condition when create overlaps with erase (hard-delete) #2454

prb112:
Introduced an Interaction Factory with a public final class to support Concurrent Tests which mix a number of various test cases so we can test the locking.  Good news is that it worked properly and the concurrency didn't show any deadlock situations with the test (it's a very aggressive test and can be tuned up or down appropriately).

### DIFF
--- a/fhir-server-test/src/test/java/com/ibm/fhir/server/test/concurrent/InteractionFactory.java
+++ b/fhir-server-test/src/test/java/com/ibm/fhir/server/test/concurrent/InteractionFactory.java
@@ -198,7 +198,7 @@ public final class InteractionFactory {
     }
 
     /**
-     * This inner class invokes the FHIRClient update API for the Patient instance it encapsulates,
+     * This inner class invokes the InteractionType FHIRClient API for the resource instance it encapsulates,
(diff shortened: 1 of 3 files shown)


fhir-server-test/src/test/java/com/ibm/fhir/server/test/concurrent/InteractionFactory.java
```diff
@@ -0,0 +1,226 @@
...
+                }
+            } while (status == Response.Status.CONFLICT.getStatusCode());
+
+            assertTrue(status == Response.Status.OK.getStatusCode() || status == Response.Status.NOT_FOUND.getStatusCode(), "Status=" + status);
+            return null;
+        }
+    }
+
+    /**
+     * This inner class invokes the FHIRClient update API for the Patient instance it encapsulates,
```
lmsurpre, line 201:
```suggestion
     * This inner class invokes the InteractionType FHIRClient API for the resource instance it encapsulates,
```

lmsurpre (approved):
LGTM

PR #2793, opened 2021-09-22
Rev kafka-client version to stay on latest working release

prb112:
Re https://kafka.apache.org/cve-list https://nvd.nist.gov/vuln/detail/CVE-2021-38153
- We are not impacted as we are using kafka-clients

Re 3.0.0 we cannot upgrade as it requires us to drop Java8 support. 
- per https://kafka.apache.org/downloads


lmsurpre (approved):
LGTM

Issue #2803, opened 2021-09-23
'migration' workflow fails to upload server logs


lmsurpre:
I am trying to debug a failed migration pipeline run, but it is hard to do that when we are missing the server logs in the artifacts section:  https://github.com/IBM/FHIR/actions/runs/1266952122

In our other workflows (e.g. `integration.yaml`) we have a `post-integration-test-docker.sh` script that runs after the integration tests and gathers to docker console output to a directory named `integration-test-results`.  Then it uploads all that as part of the `Upload logs` step.

The migration.yaml workflow seems to be doing something similar because it has an `Upload logs` step that uploads the content from `fhir/build/migration/integration-test-results`.  However, I don't think thats working because when I download the artifacts I see only these two files:
* current-integration-tests.log
* prev-integration-tests.log

-----

Note: the run linked above also lists some test failures like the following:
```
migration (db2, last)
expected:<201> but was:<500>
```
However, when I view the current-integration-tests.log from the attached artifact, it says that all of the tests were successful.  I'm hoping that the extra log output could possibly help untangle this.

lmsurpre:
I noticed the following comment in build/migration/bin/7_current-integration-test.sh`:
> # Add || docker container logs "$(docker container ls --format {{.Names}} | grep -i ${migration} | grep -i fhir)"

Not sure if thats just for debug or some kind of TODO related to this.

lmsurpre:
In our other pipelines, we gather the logs in two different spots:
* on success:  post-integration-test-docker.sh (at the end of the test step)
* on failure:  the Gather error logs step

I don't see a post-integration-test equivalent on the migration workflow, so I think we need to either:
A. add a step for that; or
B. *always* run the "Gather error logs" step

To be honest, I'm not totally sure why we don't do B in our other workflows...maybe just because this step was added later and we didn't want to mess with the existing (working) post-integration-test script/

prb112:
> I noticed the following comment in build/migration/bin/7_current-integration-test.sh`:
> 
> ```
> # Add || docker container logs "$(docker container ls --format {{.Names}} | grep -i ${migration} | grep -i fhir)"
> ```
> 
> Not sure if thats just for debug or some kind of TODO related to this.

We can talk about it. You are looking at it a bit out of step, I'll explain when we chat.

prb112:
I checked the latest PRs and can confirm they are present.

PR #2804, opened 2021-09-24
Issue #2803 - always gather the migration test logs

lmsurpre:
Signed-off-by: Lee Surprenant <lmsurpre@us.ibm.com>

PR #2805, opened 2021-09-24
issue #2803 - checkout to base dir instead of subdir

lmsurpre:
Accidentally did this on top of liberty-dev-mode, but pushed anyway because I want to see if it works or not.
The original design was to checkout the "current" code in the fhir subdir so that it doesn't conflict with the "prev" code in the "prev" subdir.  But that causes some trouble with "common" utilities.  This PR checks out "current" code in the default (base) dir and hopefully that still works (and fixes 2803).

My backup idea was to set WORKSPACE to the fhir subdir in this particular case.

### DIFF
--- a/build/migration/postgres/4_current-migrate.sh
+++ b/build/migration/postgres/4_current-migrate.sh
@@ -16,13 +16,13 @@ set -o pipefail
 pushd $(pwd) > /dev/null
 
 # Change to the migration/bin directory
-cd "${WORKSPACE}/fhir/build/migration/postgres"
+cd "${WORKSPACE}/build/migration/postgres"
 
 # In order not to hit this after packaging everything up,w e want to run this before we start up the db.
 # waiting for server to start....2021-07-16 20:42:03.136 UTC [9] FATAL:  data directory "/db/data" has invalid permissions
 # 2021-07-16 20:42:03.136 UTC [9] DETAIL:  Permissions should be u=rwx (0700) or u=rwx,g=rx (0750).
-sudo chown -R 70:70 ${WORKSPACE}/fhir/build/migration/postgres/workarea/volumes/dist/db
-sudo chmod -R 0750 ${WORKSPACE}/fhir/build/migration/postgres/workarea/volumes/dist/db
+sudo chown -R 70:70 ${WORKSPACE}/build/migration/postgres/workarea/volumes/dist/db
+sudo chmod -R 0750 ${WORKSPACE}/build/migration/postgres/workarea/volumes/dist/db
 
 # Startup db
 docker-compose up --remove-orphans -d db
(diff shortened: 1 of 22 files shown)


build/migration/postgres/4_current-migrate.sh
```diff
@@ -50,33 +50,33 @@ done
...
     --db-type derby --prop db.database=${DB_LOC}/profile --prop db.create=Y \
     --prop resourceTypes=Patient,Group,Practitioner,PractitionerRole,Person,RelatedPerson,Organization,Location,Observation,MedicationAdministration,StructureDefinition,ElementDefinition,CodeSystem,ValueSet,Resource \
     --update-schema
-java -jar ${WORKSPACE}/fhir/fhir-persistence-schema/target/fhir-persistence-schema-*-cli.jar \
+java -jar ${WORKSPACE}/fhir-persistence-schema/target/fhir-persistence-schema-*-cli.jar \
     --db-type derby --prop db.database=${DB_LOC}/reference --prop db.create=Y \
     --prop resourceTypes=Patient,Group,Practitioner,PractitionerRole,Device,Organization,Location,Medication,Observation,MedicationAdministration,StructureDefinition,ElementDefinition,CodeSystem,ValueSet,Resource \
     --update-schema
-java -jar ${WORKSPACE}/fhir/fhir-persistence-schema/target/fhir-persistence-schema-*-cli.jar \
+java -jar ${WORKSPACE}/fhir-persistence-schema/target/fhir-persistence-schema-*-cli.jar \
```
prb112, line 79: I opted not to do take this approach.  It's not ideal and makes the git history polluted.  I personally prefer the two separate directories.

lmsurpre: I wouldn't recommend anyone do this for working with the repo, but for just checking out the code I think it'll work just fine.

lmsurpre:
with these changes, the output now has the docker-console.txt, fhir-server-test report, and server-logs:
```
integration-test-results-db2-last
  current-integration-tests.log
  docker-console.txt
  fhir-server-test
  prev-integration-tests.log
  runtime.txt
  server-logs
```

build/migration/postgres/4_current-migrate.sh
prb112: This is an anti-pattern.  While it works, I suggest altering the approach to use the sub directory.

lmsurpre: Would it help if we add a comment explaining what we're doing and why it is ok in this case (but not in general)?

Something like this:
```
# Note: here we are cloning a previous version of the repo into a subdirectory of itself.
# This is not advised, but in this case it is OK because it is used strictly for pulling down the code.
# Avoid this pattern for repositories that you plan to continue working with over time.
```

lmsurpre:
closed in favor of #2804

Issue #2806, opened 2021-09-24
Split out the Term Graph Loaders


prb112:
**Is your feature request related to a problem? Please describe.**
Split out the Term Graph Loaders

**Describe the solution you'd like**
Split out the Term Graph Loaders

**Describe alternatives you've considered**

**Acceptance Criteria**

1.  
   GIVEN [a precondition]
    AND [another precondition]
   WHEN [test step]
    AND [test step]
   THEN [verification step]
    AND [verification step]

**Additional context**

prb112:
PR https://github.com/IBM/FHIR/pull/2813

lmsurpre:
The project has been split out and everything compiles.  #2798 will add additional checking to ensure the feature works as expected.

PR #2813, opened 2021-09-27
Split out the Term Graph Loaders #2806

prb112:
- refactor the pom module
- split loader and graph


term/fhir-term-loader/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>com.ibm.fhir</groupId>
+        <artifactId>fhir-term-parent</artifactId>
+        <version>4.10.0-SNAPSHOT</version>
+        <relativePath>../</relativePath>
+    </parent>
+    <artifactId>fhir-term-loader</artifactId>
```
lmsurpre, line 11: do we expect this project to evolve beyond loading just the fhir-term-graph?
if not, we might want to use `fhir-term-graph-loader` instead.  I'm good either way thought.

lmsurpre (approved):
Very glad to have the loader separated from the termProvider.  I'm not sure we needed to move all term-related projects into a subdirectory, but I'm fine with it.  LGTM.

term/fhir-term-loader/pom.xml
prb112: Changed to be more specific

lmsurpre:
I missed it / forgot about it in my first review, but we should also update the README to reflect this new module (and the updated paths to the term modules)

Issue #2814, opened 2021-09-27
AdoptOpenJDK Images are deprecated and we need to convert to Eclipse builds


prb112:
**Describe the bug**
AdoptOpenJDK Images are deprecated and we need to convert to Eclipse builds

**Environment**
Which version of IBM FHIR Server? 4.9.1 and below

IBM Semeru https://github.com/docker-library/official-images/pull/10666
eclipse-temurin:<version>-alpine
https://hub.docker.com/_/eclipse-temurin/?tab=description

**Expected behavior**
n/a

**Additional context**
n/a

prb112:
We have to use UBI.  
Pick up the Adoptium one if Semeru is not available.

prb112:
PR https://github.com/IBM/FHIR/pull/2815

lmsurpre:
This one caused our 4.9.2 release automation to hiccup and so our schematool image wasn't published at first.  Paul manually addressed the missing image on dockerhub.  Moving forward, we just need to monitor this docker repo as its location may change.

punktilious:
Tested on RHEL8:
```
ibm-fhir-schematool$ docker run  --add-host=host.docker.internal:127.0.0.1 --env ENV_TOOL_INPUT=$(cat persistence.json | base64 -w 0) localhost/ibm-fhir-schematool:4.10.0
...
2021-10-07 07:51:10.568 00000001    INFO   com.ibm.fhir.schema.app.Main Processing took:   7.063 s
2021-10-07 07:51:10.572 00000001    INFO   com.ibm.fhir.schema.app.Main SCHEMA CHANGE: OK
+ set +x
Skipping refresh-tenants as it's not needed on 'postgresql'
```

Verified.

PR #2815, opened 2021-09-27
AdoptOpenJDK Images are deprecated and we need to convert to Eclipse …

prb112:
…builds #2814


lmsurpre (approved):
We'll need to watch this as I'm not sure ibmsemeruruntime/open-11-jdk will be the long-term home for these images. LGTM.

AdamBrousseau:
Correct, unofficial ubi images were added as a stop gap until they can be added to RedHat's container registry. I'm not yet sure if we plan to maintain them once RH ones are there.

prb112:
Thanks for the feedback @AdamBrousseau we'll definitely be watching

Issue #2802, opened 2021-09-23
Global caches in ProfileSupport should be tenant-specific 


lmsurpre:
**Describe the bug**
Global caches in ProfileSupport are not currently tenant-specific which means that tenant-specific profiles acquired through the ServerRegistryResourceProvider may be visible globally. Instead, these caches should be tenant-specific.

**Environment**
Which version of IBM FHIR Server?

**To Reproduce**
1. Enable ServerRegistryResourceProvider on a server with two tenants (A and B)
2. As a user of tenant A, create a StructureDefinition that defines an extension
3. As a user of tenant A, validate a resource that contains this extension (loading this StructureDefinition and its generated constraints in our caches)
4. As a user of tenant B, validate a resource that contains this extension

Note that the extension is validated with the constraints from tenant A even though tenant B may have a different version of the extension definition.

**Expected behavior**
Users on tenant B should not have visibility into any conformance resources posted to tenant A.

**Additional context**
Very similar to #2256.

I hope there's not more of these hiding, but we should consider advising users to avoid multiTenant deployments with serverRegistryResourceProvider enabled (unless they control who is allowed to create conformance resources).

prb112:
QA: Passed

I ran with two different datastores (Postgres and Db2). One for default and one for tenant1. 
I created a StructureDefinition, I showed that the cache'd rule was not applicable for the default tenant and available for the tenant1 and failed. 

Negative tests showed the only way to screw this up is to misconfigure the data store so they share the tenants share the same exact schema.

PR #2816, opened 2021-09-27
issue #2802 - use CacheManager for tenant-specific caches

lmsurpre:
I followed the pattern from CodeSystemSupport.

### DIFF
--- a/fhir-profile/src/main/java/com/ibm/fhir/profile/ExtensionBuilder.java
+++ b/fhir-profile/src/main/java/com/ibm/fhir/profile/ExtensionBuilder.java
@@ -1,0 +1,32 @@
+/*
+ * (C) Copyright IBM Corp. 2021
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.ibm.fhir.profile;
+
+import static com.ibm.fhir.model.type.String.string;
+
+import com.ibm.fhir.model.resource.StructureDefinition.Context;
+import com.ibm.fhir.model.type.Extension;
+
+/**
+ * Specializes ProfileBuilder for making it more convenient to build extension definitions
+ */
+public class ExtensionBuilder extends ProfileBuilder {
+
+    /**
+     * Construct a ProfileBuilder with a fixed Extension.url and a type constraint on Extension.value[x]
+     *
+     * @param url
+     * @param version
+     * @param context
+     * @param type
+     */
+    public ExtensionBuilder(String url, String version, Context context, String type) {
+        super(Extension.class, url, version, context);
+        fixed("Extension.url", string("http://ibm.com/fhir/StructureDefinition/favorite-team"));
+        type("Extension.value[x]", ProfileBuilder.type(type));
+    }
+
+}
--- a/fhir-profile/src/main/java/com/ibm/fhir/profile/ProfileSupport.java
+++ b/fhir-profile/src/main/java/com/ibm/fhir/profile/ProfileSupport.java
@@ -6,6 +6,8 @@
 
 package com.ibm.fhir.profile;
 
+import static com.ibm.fhir.cache.CacheKey.key;
+
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
@@ -15,11 +17,13 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Predicate;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
+import com.ibm.fhir.cache.CacheKey;
+import com.ibm.fhir.cache.CacheManager;
+import com.ibm.fhir.cache.CacheManager.Configuration;
 import com.ibm.fhir.model.annotation.Constraint;
 import com.ibm.fhir.model.constraint.spi.ConstraintProvider;
 import com.ibm.fhir.model.resource.Resource;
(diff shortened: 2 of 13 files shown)


fhir-profile/src/main/java/com/ibm/fhir/profile/ProfileSupport.java
```diff
@@ -203,11 +211,8 @@ public static Binding getBinding(String path) {
     }
 
     public static Map<String, Binding> getBindingMap(String url) {
-        Map<String, Binding> bindingMap = BINDING_CACHE.get(url);
-        if (bindingMap == null) {
-            bindingMap = BINDING_CACHE.computeIfAbsent(url, ProfileSupport::computeBindingMap);
-        }
```
lmsurpre, line 209: @JohnTimm I think the null checks here aren't really needed because `computeIfAbsent` should already handle that case, right?  Anything I'm missing?

fhir-profile/src/main/java/com/ibm/fhir/profile/ProfileSupport.java
```diff
@@ -291,11 +294,8 @@ public static ElementDefinition getElementDefinition(String path) {
     }
 
     public static Map<String, ElementDefinition> getElementDefinitionMap(String url) {
-        Map<String, ElementDefinition> elementDefinitionMap = ELEMENT_DEFINITION_CACHE.get(url);
-        if (elementDefinitionMap == null) {
-            elementDefinitionMap = ELEMENT_DEFINITION_CACHE.computeIfAbsent(url, ProfileSupport::computeElementDefinitionMap);
-        }
-        return elementDefinitionMap;
+        Map<String, Map<String, ElementDefinition>> elementDefinitionMapCache = CacheManager.getCacheAsMap(ELEMENT_DEF_CACHE_NAME, ELEMENT_DEF_CACHE_CONFIG);
+        return elementDefinitionMapCache.computeIfAbsent(url, ProfileSupport::computeElementDefinitionMap);
```
prb112, line 298: It'd be nice to get a test case for this using `default`.  I think we enable server registry by default. if not, we do so on tenant1.

lmsurpre: Sure.  The hardest part is creating a simple extension definition (surprisingly difficult).
I just tested in locally with an extension from https://alvearie.io/alvearie-fhir-ig/StructureDefinition-process-name.html ...maybe I'll just suck that one down locally (unless we have a similar one laying around somewhere already).

prb112: just given we have a 'security' label on this, I think we should.

prb112 (approved):
LGTM, approving (hoping to have integration test)

fhir-profile/src/main/java/com/ibm/fhir/profile/ProfileSupport.java
lmsurpre: ok, I pushed a new test class named TenantIsolationTest

fhir-profile/src/main/java/com/ibm/fhir/profile/ExtensionBuilder.java
```diff
@@ -0,0 +1,32 @@
+/*
+ * (C) Copyright IBM Corp. 2021
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.ibm.fhir.profile;
```
lmsurpre, line 6: I put this one next to ProfileBuilder under `src/main`, but I'm wondering if we should move them both to `src/test`...any thoughts?

prb112: since it's hard coding the url, I think src/test is the right place

fhir-server-test/src/test/java/com/ibm/fhir/server/test/TenantIsolationTest.java
```diff
@@ -0,0 +1,116 @@
...
+                    .code(Code.of("Brewers"))
+                    .build())
+                .concept(Concept.builder()
+                    .code(Code.of("Bucks"))
+                    .build())
+                .concept(Concept.builder()
+                    .code(Code.of("Badgers"))
+                    .build())
+                .build();
+        createResourceAndReturnTheLogicalId("CodeSystem", codeSystem);
```
prb112, line 64: Well... I can tell where you stand :)

fhir-server-test/src/test/java/com/ibm/fhir/server/test/TenantIsolationTest.java
```diff
@@ -0,0 +1,116 @@
...
+        FHIRResponse response = client.invoke("CodeSystem", "$lookup", params);
+        Resource resource = response.getResource(Resource.class);
+        assertTrue(resource instanceof Parameters);
+        System.out.println(resource);
+
+        response = client.invoke("CodeSystem", "$lookup", params,
+            FHIRRequestHeader.header("X-FHIR-TENANT-ID", "tenant1"),
+            FHIRRequestHeader.header("X-FHIR-DSID", "reference"));
+        OperationOutcome oo = response.getResource(OperationOutcome.class);
+        assertEquals(oo.getIssue().get(0).getSeverity(), IssueSeverity.FATAL, oo.getIssue().get(0).getDetails().toString());
```
prb112, line 89: should there be a positive case? e.g. one that succeeds to prove the structuredefinition/extension above was correct.

prb112: I was also thinking we should push two structuredefinitions with the same name and two distinct behaviors.

lmsurpre: this is what lines 80 through 83 are doing

prb112: Right, so the same named structure def would be the only stretch goal I could for see

lmsurpre: though I should probably remove that sysout...
I can throw a `assertEquals(response.getStatus(), 200);` in there while at it for good measure

lmsurpre: ok, I restructured the test completely.  please have another look; i'll plan to merge if everything passes

lmsurpre: There was a challenge in that the ServerRegistryResourceProvider caches results for a minute.
Fortunately, that cache is kept on a per-datastore basis and so we can avoid the trouble by using different tenant1 datastores for the two ("no definition" vs "conflicting definition") cases.

prb112: nice

Issue #2605, opened 2021-07-12
Reject FHIR resources with the null char \u0000 or similar control chars


lmsurpre:
**Is your feature request related to a problem? Please describe.**
During a database reindex, 

It turns out that one of the resources contained a string that included `\u0000` but PostgreSQL does not allow the null char in its strings.

The original resource content was stored successfully because we gzip the resource contents before saving it to the DB.
However, now the element that contains this character is matching on a new search parameter and so that is what causes the problem.

The spec says this:
> Strings SHOULD not contain Unicode character points below 32, except for u0009 (horizontal tab), u0010 (carriage return) and u0013 (line feed).

**Describe the solution you'd like**
Strictly enforce this guidance as part of constructing model types that extend String or Uri.  Instead of allowing it in resource content but not in extracted fields, just reject resources with it up front.

This is the preferred approach because it avoids data quality issues up-front.  For example, the NULL char can be used for certain exploits as described at https://owasp.org/www-community/attacks/Embedding_Null_Code

As part of https://github.com/IBM/FHIR/issues/2424 we are no longer running parse validation when we read resources from the database, so it should be safe to add without introducing an "escape hatch" that disables the added validation.
A consequence of this is that certain actions like Patch or Reindex could fail (potentially in a confusing way) until the previously-ingested data is fixed to avoid these problem characters.

**Describe alternatives you've considered**
The main alternative is to allow these characters in both resource content and extracted/indexed fields.
In this case, we should still probably add a WARNING to the output so that users are still informed that they are not following this SHOULD clause in the spec.

To work around the specific issues with the \u0000 char in postgresql, we'd want to replace this character with something like https://www.fileformat.info/info/unicode/char/fffd/index.htm.

**Acceptance Criteria**

1.  
   GIVEN a resource contains the \u0000 char
   WHEN its posted to the server
   THEN it is rejected with a clear error message

1.  
   GIVEN a resource has already been ingested with the \u0000 char
   WHEN the resource is reindexed
   THEN it fails with a clear error message

**Additional context**
https://github.com/IBM/FHIR/issues/2424 contains some related discussion, but I don't think that we ever really addressed the root issue.

lmsurpre:
team decision:  default should be to reject resources that include invalid Unicode character points below 32.  validation to be done during object construction.  we should also introduce a new config parameter on ModelConfig and expose that from fhir-server-config (default to strict and keep out of default/example configs)

prb112:
Test cases: 
[...]

PR #2820, opened 2021-10-01
Reject FHIR resources with the control chars #2605

prb112:

### DIFF
--- a/fhir-model/src/main/java/com/ibm/fhir/model/util/ValidationSupport.java
+++ b/fhir-model/src/main/java/com/ibm/fhir/model/util/ValidationSupport.java
@@ -70,6 +70,7 @@ public final class ValidationSupport {
         }
     };
     private static final Set<Character> WHITESPACE = new HashSet<>(Arrays.asList(' ', '\t', '\r', '\n'));
+    private static final Set<Character> UNSUPPORTED_UNICODE = buildUnsupportedUnicodeCharacterSet();
     private static final char [] BASE64_CHARS = {
         'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P',
         'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z', 'a', 'b', 'c', 'd', 'e', 'f',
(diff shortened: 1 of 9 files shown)


fhir-model/src/main/java/com/ibm/fhir/model/util/ValidationSupport.java
```diff
@@ -80,6 +81,22 @@ public Validator initialValue() {
...
+     */
+    private static Set<Character> buildUnsupportedUnicodeCharacterSet() {
+        Set<Character> chars = new HashSet<>();
+        for (int i = 0; i < 32; i++) {
+            if (i != 9 && i != 10 && i != 13) {
+                chars.add(Character.valueOf((char) i));
+            }
+        }
+        return chars;
+    }
```
prb112, line 98: Generates the Forbidden Character Map (Unicode Character Points)

fhir-model/src/main/java/com/ibm/fhir/model/util/ValidationSupport.java
```diff
@@ -106,17 +123,29 @@ public static void checkString(String s) {
...
             throw new IllegalStateException(String.format("Trimmed String value length: %d is less than minimum required length: %d", count, MIN_STRING_LENGTH));
         }
     }
 
+    /**
+     * Helper method to check if there is unsupported unicode.
+     * @param ch
+     */
+    private static void checkUnsupportedUnicode(String s, Character ch) {
+        if (UNSUPPORTED_UNICODE.contains(ch)) {
```
prb112, line 144: The key check

punktilious: Is there a micro-benchmark for this? I'm concerned about performing a Set lookup for every inbound character. I think John had some benchmarking stuff for the model parsing, perhaps that would be a good reference.

punktilious (changes requested):
I think we need to compare performance.

fhir-model/src/main/java/com/ibm/fhir/model/util/ValidationSupport.java
lmsurpre: FHIRParserBenchmark (under `fhir-benchmark`) could probably be used for this.  I think it defaults to using different resources each time, but there's options to feed it a specific set of resources (either in main or maybe I just have that locally).

prb112: I'm currently running the FHIRParserBenchmark. 
I'll report back and change if needed.

prb112:
Implemented, waiting on ci/cd

PR #2822, opened 2021-10-01
Handle null values in fhir-server-config and throw for exceptional case

lmsurpre:
Previously, we'd throw an IllegalStateException if we encountered a
null value. This came up because the helm `toJson` function produces all
possible fields for an object, even when some are null.

Additionally, if we have an invalid config (such as described above), in
FHIRRestHelper.validateResource we were swallowing the underlying
exception and instread constructing a generic OperationOutcomeIssue with
message "Error retrieving profile configuration". This meant we were missing
the root cause in our logs and also meant that this exception was being handled
just like any other validation error and so, even though it was a server
config issue, this was not clear to the client.

### DIFF
--- a/fhir-config/src/main/java/com/ibm/fhir/config/PropertyGroup.java
+++ b/fhir-config/src/main/java/com/ibm/fhir/config/PropertyGroup.java
@@ -237,12 +237,16 @@ public class PropertyGroup {
     /**
      * Returns the properties contained in the PropertyGroup in the form of a list of
      * PropertyEntry instances. If no properties exist, then an empty list will be returned.
+     * Properties with a value of null will be omitted from the list.
      * @throws Exception
      */
     public List<PropertyEntry> getProperties() throws Exception {
         List<PropertyEntry> results = new ArrayList<>();
         for (Map.Entry<String, JsonValue> entry : jsonObj.entrySet()) {
-            results.add(new PropertyEntry(entry.getKey(), convertJsonValue(entry.getValue())));
+            Object jsonValue = convertJsonValue(entry.getValue());
+            if (jsonValue != null) {
+                results.add(new PropertyEntry(entry.getKey(), convertJsonValue(entry.getValue())));
+            }
         }
         return results;
     }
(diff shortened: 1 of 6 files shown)


fhir-config/src/main/java/com/ibm/fhir/config/PropertyGroup.java
```diff
@@ -292,6 +292,8 @@ public static Object convertJsonValue(JsonValue jsonValue) throws Exception {
         case FALSE:
             result = Boolean.FALSE;
             break;
+        case NULL:
```
punktilious, line 295: This method can return null now. There are loads of places in the code where we don't check for null when calling PropertyGroup#getValue(), so we'd just end up with a NPE if things were misconfigured.

prb112 (approved):

fhir-config/src/main/java/com/ibm/fhir/config/PropertyGroup.java
lmsurpre: I ended up doing the above only for getProperties (and not for convertJsonArray).  IMHO that should be sufficient, but let me know if you think we should do something for convertJsonArray as well...

punktilious: sounds good

fhir-config/src/main/java/com/ibm/fhir/config/PropertyGroup.java
```diff
@@ -237,12 +237,16 @@ else if (jsonValue instanceof JsonString) {
...
+     * Properties with a value of null will be omitted from the list.
      * @throws Exception
      */
     public List<PropertyEntry> getProperties() throws Exception {
         List<PropertyEntry> results = new ArrayList<>();
         for (Map.Entry<String, JsonValue> entry : jsonObj.entrySet()) {
-            results.add(new PropertyEntry(entry.getKey(), convertJsonValue(entry.getValue())));
+            Object jsonValue = convertJsonValue(entry.getValue());
+            if (jsonValue != null) {
+                results.add(new PropertyEntry(entry.getKey(), convertJsonValue(entry.getValue())));
```
michaelwschroeder, line 248:
```suggestion
                results.add(new PropertyEntry(entry.getKey(), jsonValue));
```

michaelwschroeder (approved):
LGTM

PR #2828, opened 2021-10-04
inline version vars

lmsurpre:
It seems like depend-a-bot is only opening PRs for dependencies with their versions inlined in fhir-parent/pom.xml
so I drafted these changes to help dependabot  help us...but first we should probably confirm the assumption


lmsurpre:
I posed my question at https://github.community/t/does-dependabot-resolve-maven-variables/204707 but I havn't gotten any response...I think we should just pull the trigger and see how it goes.

prb112 (approved):

PR #2830, opened 2021-10-05
issue-1834 set PostgreSQL fillfactor for updatable tables

punktilious:

### DIFF
--- a/fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirSchemaGenerator.java
+++ b/fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirSchemaGenerator.java
@@ -1196,10 +1196,6 @@ public class FhirSchemaGenerator {
     public Table addResourceTokenRefs(PhysicalDataModel pdm) {
 
         final String tableName = RESOURCE_TOKEN_REFS;
-        
-        // Custom tuning for this table
-        List<With> withs = new ArrayList<>(addWiths());
-        withs.add(With.with(FhirSchemaConstants.PG_FILLFACTOR_PROP, Integer.toString(FhirSchemaConstants.PG_FILLFACTOR_VALUE)));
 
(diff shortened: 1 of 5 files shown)


fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/control/FhirSchemaGenerator.java
```diff
@@ -789,7 +810,7 @@ public void addResourceChangeLog(PhysicalDataModel pdm) {
                 .setTablespace(fhirTablespace)
                 .addPrivileges(resourceTablePrivileges)
                 .enableAccessControl(this.sessionVariable)
-                .addWiths(addWiths()) // New Column for V0017
+                .addWiths(customWiths) // Does not require fillfactor tuning
```
prb112, line 813: makes sense

prb112 (approved):
LGTM

Q: Is there any odd behavior we should let downstream users know about post? e.g. a rebalancing after the fillfactor change, or is it naturally going to happen over time

punktilious:
> LGTM
> 
> Q: Is there any odd behavior we should let downstream users know about post? e.g. a rebalancing after the fillfactor change, or is it naturally going to happen over time

It'll happen naturally over time. The fillfactor will only affect newly allocated blocks or if a full vacuum is performed (which is a heavyweight operation and should be guided by a DBA). I think it's sufficient to mention in the release notes.

prb112:
> > LGTM
> > Q: Is there any odd behavior we should let downstream users know about post? e.g. a rebalancing after the fillfactor change, or is it naturally going to happen over time
> 
> It'll happen naturally over time. The fillfactor will only affect newly allocated blocks or if a full vacuum is performed (which is a heavyweight operation and should be guided by a DBA). I think it's sufficient to mention in the release notes.

Great - That's what I was hoping.

PR #2833, opened 2021-10-05
Remove jwt lib dependency

lmsurpre:


fhir-parent/pom.xml
```diff
@@ -415,7 +415,7 @@
             <dependency>
                 <groupId>com.auth0</groupId>
                 <artifactId>java-jwt</artifactId>
-                <version>3.13.0</version>
```
prb112, line 418: Is this now just for test?

lmsurpre: yes

Issue #2845, opened 2021-10-11
Refactor safeSleep to use single helper


punktilious:
```
private void safeSleep(long millis) {
        try {
            Thread.sleep(millis);
        } catch (InterruptedException x) {
            // NOP
        }
    }
```

is used frequently in the code. Should be a static method in a helper.

prb112:
PR https://github.com/IBM/FHIR/pull/2876

Issue #2751, opened 2021-09-09
Protect from concurrent update schema commands


lmsurpre:
**Is your feature request related to a problem? Please describe.**
In https://github.com/Alvearie/alvearie-helm/issues/27 we were hoping to move from helm hooks (once and only once execution) to init containers (runs for each pod) so that we can add a PostgreSQL subchart.

I have two main concerns with that:
1. multiple pods start at the same time and i'm not super confident that we can safely execute the schema tool in parallel in all cases (and its hard to test for)
2. it makes pod startup slower

**Describe the solution you'd like**
1. create a new table in the FHIR_ADMIN schema for tracking schema upgrades
2. when a schema action begins (anything other than "create-schema" I think), write a single row to this table to represent its "lease" 
3. include an expiration timestamp in the row and update the schema tool to move that timestamp forward while its running (it should never expire while upgrade is still running)
4. before executing a schema-update (and other actions?) ensure that there is no lease row;  or if there is one, ensure that the lease is expired.  otherwise, exit with a successful status.

**Describe alternatives you've considered**
At first I was thinking we could use a simple table lock on the FHIR_ADMIN.VERSION_HISTORY table, but because we use multiple transactions for some schema updates, we don't think this is quite sufficient.

**Acceptance Criteria**

1.  
   GIVEN [a precondition]
    AND [another precondition]
   WHEN [test step]
    AND [test step]
   THEN [verification step]
    AND [verification step]

**Additional context**
The described solution addresses my first concern, which is the most important one.
However, it would also be great if we could cover my second concern with is the slow pod startup.
Therefor, it would be great if we could add the overall schema version to this table.  Then we could do logic like this:
1. once the schema-update is successful, mark this "lease" row as "completed"
2. when doing an upgrade, if there is a completed lease row for *this* version of the schema (or higher), skip the update

d0roppe:
Moving back to in progress as the new tables used for this is not removed with the drop-schema part of the fhir-persistence-schema-cli tool

d0roppe:
Verified that issuing multiple change schemas at the same time, they will not interfere with each other, 1 will do the work while the other will time out and you can try again.

PR #2876, opened 2021-10-20
Refactor safeSleep to use single helper #2845

prb112:
- Unifies the usage of safeSleep into a helper method


prb112:
> Oh funny! I had introduced a helper for this too as part of #2751. Will need to consolidate after this gets merged.

Merged :)

Issue #2923, opened 2021-11-01
Move kafka-clients from 2.8.1 to 3.0.0


prb112:
Move kafka-clients from 2.8.1 to 3.0.0

Failing PR https://github.com/IBM/FHIR/pull/2895

lmsurpre:
I ran `mvn dependency:tree -f fhir-parent` and confirmed all instances were now at 3.0.0.

Intersetingly, I found `com.amazonaws:aws-java-sdk-kafka:jar:1.11.596:provided` in with the aws stuff, but that is marked as provided and so should be ok.

PR #2927, opened 2021-11-01
Move kafka-clients from 2.8.1 to 3.0.0 #2923

prb112:
- Need to make a slight code change and dependency change. 
- Added slf4j-api to fhir-audit
- KafkaService - Change timeout to use Duration.of 


lmsurpre (approved):
LGTM

PR #2928, opened 2021-11-01
Demo and Fix for Member Match

prb112:
- This is a small demo used for integration tests.  (not sure if we want an example folder instead of demo)
- @kmbarton423 found an issue with name and system use.  I added a check for system and fixed how names were processed.


lmsurpre:
> not sure if we want an example folder instead of demo

my intent for the demo folder was to have runnable demos that use some config of the fhir server to show some capability.  this sort of fits, but fits much better if you could wrap it in a docker-compose env.
if we choose to go this route, I could imagine moving the current contents of `/demo` down one level.  if we do that we need to give consumers (especially the benchmark project) a heads up.

"examples" kind of conflicts with our fhir-examples project name/meaning (which I had originally been calling fhir-samples).

prb112:
I can move the project into operation and not add it to the module.

lmsurpre:
alternatively, maybe we could just provide some example command for adding the member-match operation to userlib (maybe via mvn dependency:copy) and enabling it in the config under `demo/fhir/config/` ?

lmsurpre:
> I can move the project into operation and not add it to the module.

yeah, maybe that would be easiest/best.  then just clearly label it as an example

operation/fhir-operation-member-match/src/main/java/com/ibm/fhir/operation/davinci/hrex/provider/strategy/DefaultMemberMatchStrategy.java
```diff
@@ -250,13 +250,25 @@ public boolean visit(String elementName, int elementIndex, Identifier identifier
...
-            if (val != null && val.getValue() != null) {
-                this.value = val.getValue();
+            if (identifier.getType() != null
+                    && identifier.getType().getCoding() != null) {
+                // If there is  more than one Coding it has to be MB.
+                for (Coding coding : identifier.getType().getCoding()) {
+                    if (coding.getSystem() != null
+                            &&  "http://terminology.hl7.org/CodeSystem/v2-0203"
+                                    .equals(coding.getSystem().getValue())
+                            && "MB".equals(coding.getCode().getValue())) {
```
lmsurpre, line 260: technically i think coding.getCode() could be null, but I think its probably rare we'd hit that

lmsurpre (approved):
LGTM

operation/fhir-operation-member-match/src/main/java/com/ibm/fhir/operation/davinci/hrex/provider/strategy/DefaultMemberMatchStrategy.java
prb112: Yeah - in negative testing

operation/fhir-operation-member-match/src/main/java/com/ibm/fhir/operation/davinci/hrex/provider/strategy/DefaultMemberMatchStrategy.java
```diff
@@ -250,13 +250,25 @@ public boolean visit(String elementName, int elementIndex, Identifier identifier
...
-            Uri sys = identifier.getSystem();
-            com.ibm.fhir.model.type.String val = identifier.getValue();
-            if (sys != null && sys.getValue() != null) {
-                this.system = sys.getValue();
-            }
-            if (val != null && val.getValue() != null) {
-                this.value = val.getValue();
+            if (identifier.getType() != null
+                    && identifier.getType().getCoding() != null) {
+                // If there is  more than one Coding it has to be MB.
```
lmsurpre, line 255: already got merged, but i thought of this so decided to submit it anyway (no need to action):
```suggestion
                // Look for an identifier with the 'MB' code from http://terminology.hl7.org/CodeSystem/v2-0203.
```

PR #2929, opened 2021-11-01
Issue 2843 - store both versioned and unversioned canonicals in ParametersMap

lmsurpre:
To make this simpler, I also updated the way we apply the filter rules.

Note: this changeset introduces the fhir-ig-us-core as a test dependency for
fhir-search. This was strictly out of convenience...the alternative
would be to create a new test-only PackagedRegistryResourceProvider that
provides multiple versions of the same search parameter(s).

### DIFF
--- a/fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersMap.java
+++ b/fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersMap.java
@@ -15,17 +15,19 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 
 import com.ibm.fhir.model.resource.SearchParameter;
 
 /**
- * A multi-key map that indexes a set of search parameters by SearchParameter.code and SearchParameter.url
+ * A multi-key map that indexes a set of search parameters by SearchParameter.code and
+ * SearchParameter.url / canonical (url + "|" + version)
  */
 public class ParametersMap {
     private static final Logger log = Logger.getLogger(ParametersMap.class.getName());
 
     private final Map<String, Set<SearchParameter>> codeMap;
-    private final Map<String, SearchParameter> urlMap;
+    private final Map<String, SearchParameter> canonicalMap;
 
     /**
      * Construct a ParametersMap from a Bundle of SearchParameter
@@ -33,7 +35,7 @@ public class ParametersMap {
     public ParametersMap() {
         // LinkedHashMaps to preserve insertion order
         codeMap = new LinkedHashMap<>();
-        urlMap = new LinkedHashMap<>();
+        canonicalMap = new LinkedHashMap<>();
     }
 
     public void insert(String code, SearchParameter parameter) {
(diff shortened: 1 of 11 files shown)


fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersMap.java
```diff
@@ -15,54 +15,78 @@
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
```
prb112, line 18: Copyright year

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersMap.java
```diff
@@ -15,54 +15,78 @@
...
+                if (previous.getExpression() == null || previous.getExpression().equals(parameter.getExpression())) {
+                    if (!code.equals(previous.getCode().getValue())) {
+                        log.info("SearchParameter '" + canonical + "' already exists with the same expression; "
+                                + "adding additional code '" + code + "'");
+                    }
+                } else {
+                    log.warning("SearchParameter '" + canonical + "' already exists with a different expression;\n"
+                            + "replacing [id=" + previous.getId() + ", expression=" + previous.getExpression().getValue()
+                            + "] with [id=" + parameter.getId() + ", expression=" + parameter.getExpression().getValue() + "]");
+                }
```
prb112, line 86: Rather than duplicate this segment, maybe this deserves a helper method to limit the additional complexity in updating / maintaining ?

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersMap.java
```diff
@@ -77,12 +101,24 @@ public void insertAll(ParametersMap map) {
...
     }
 
     public Collection<SearchParameter> values() {
-        return urlMap.values();
+        // use List to preserve order
+        return Collections.unmodifiableList(canonicalMap.entrySet().stream()
+                .filter(e -> !e.getKey().contains("|"))
+                .map(e -> e.getValue())
+                .collect(Collectors.toList()));
     }
```
prb112, line 122: I think this is a good change.  How expensive is this?

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersMap.java
```diff
@@ -97,7 +133,21 @@ public int size() {
...
+    /**
+     * @deprecated use {@link #canonicalEntries()} instead
+     */
+    @Deprecated
     public Set<Entry<String, SearchParameter>> urlEntries() {
-        return Collections.unmodifiableSet(urlMap.entrySet());
+        return Collections.unmodifiableSet(canonicalMap.entrySet().stream()
+                .filter(e -> !e.getKey().contains("|"))
+                .collect(Collectors.toSet()));
+    }
```
prb112, line 144: Maybe just remove these deprecated methods? No other team should be depending on this behavior.  While observable, it shouldn't be a problem.

fhir-search/src/test/java/com/ibm/fhir/search/parameters/ParametersSearchUtilTest.java
```diff
@@ -146,11 +148,39 @@ public void testGetSearchParameters6Default() throws Exception {
...
+
+        FHIRRequestContext.set(new FHIRRequestContext("tenant5"));
+
+        result = SearchUtil.getApplicableSearchParameters("Device");
+        assertNotNull(result);
+        printSearchParameters("testVersionedSearchParameterFilter/Device", result);
+        for (SearchParameter sp : result) {
+            System.out.println(sp.getUrl().getValue() + "|" + sp.getVersion().getValue());
+            if ("http://hl7.org/fhir/us/core/SearchParameter/us-core-device-type".equals(sp.getUrl().getValue())) {
+                assertNotEquals("3.1.1", sp.getVersion());
```
prb112, line 182: small comment on this test and the for loop above. in rare circumstances the for loop may complete without actually hitting an assert condition.

fhir-search/src/test/java/com/ibm/fhir/search/parameters/ParametersUtilTest.java
```diff
@@ -40,10 +40,10 @@
 import com.ibm.fhir.search.test.BaseSearchTest;
 
 /**
- *
  * Tests ParametersUtil
  */
 public class ParametersUtilTest extends BaseSearchTest {
+    public static final boolean DEBUG = true;
```
prb112, line 46: turn it off?

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersMap.java
lmsurpre: fixed

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersMap.java
```diff
@@ -15,54 +15,78 @@
...
-                        + "] with [id=" + parameter.getId() + ", expression=" + parameter.getExpression().getValue() + "]");
+                log.info("SearchParameter '" + url + "' already exists with a different expression;\n"
+                        + "replacing [id=" + previous.getId() + ", version=" + previous.getVersion().getValue() + ", expression=" + previous.getExpression().getValue()
+                        + "] with [id=" + parameter.getId() + ", version=" + parameter.getVersion().getValue() + ", expression=" + parameter.getExpression().getValue() + "]");
+            }
+        }
+        canonicalMap.put(url, parameter);
+
+        if (version != null) {
+            String canonical = url + (version == null ? "" : "|" + version);
```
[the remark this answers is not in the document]
lmsurpre: changed as suggested

fhir-search/src/test/java/com/ibm/fhir/search/parameters/ParametersUtilTest.java
lmsurpre: done (and also in ParametersSearchUtilTest)

fhir-search/src/test/java/com/ibm/fhir/search/parameters/ParametersSearchUtilTest.java
lmsurpre: good catch.  I updated it to ensure that the expected search parameter is found (or else it fails)

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersMap.java
lmsurpre: the two uses here are very similar but there is a subtle difference which is that the first stanza will print the version (if there is one) whereas this second block is only entered when canonical includes the version and so I removed that from the `[ ]` content.

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersMap.java
lmsurpre: good question.  currently its only used from `SearchUtil.getUserDefinedSearchParameters(resourceType)` which is just for the extension parameters.  typically we don't have many of those, but it would be more costly if someone has a lot of them.

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersMap.java
lmsurpre: i was trying to preserve backwards compatibility in the java where possible, but i'm ok removing them now if we don't want to wait

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersMap.java
prb112: Right no worries - more of a consistency check point.   LGTM

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersMap.java
prb112: Great

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersMap.java
prb112: I  vote for removing

prb112 (approved):
LGTM (left one comment unresolved for the deprecated code)

fhir-search/src/main/java/com/ibm/fhir/search/parameters/ParametersMap.java
lmsurpre: done

PR #2939, opened 2021-11-03
Fix for Threading problem in tenant configurations in tests

prb112:


fhir-search/src/test/java/com/ibm/fhir/search/test/ExamplesDriverTest.java
```diff
@@ -27,6 +35,7 @@ public void processExamples() throws Exception {
         String index = System.getProperty(this.getClass().getName()
             + ".index", Index.MINIMAL_JSON.name());
         driver.processIndex(Index.valueOf(index));
+        FHIRRequestContext.remove();
      }
```
prb112, line 40: Change in this branch (built on lee-main which had additional commits)

PR #2943, opened 2021-11-05
issue 2751 - handle race condition for create control table

punktilious:


prb112 (approved):
LGTM - makes sense... basically catch and control the violation so it maps to an expected return code.

Issue #2471, opened 2021-06-04
Audit record for PUTs should not always use "Update"


d0roppe:
**Is your feature request related to a problem? Please describe.**
With Audit records on, anything with the PUT command is always treated as an update even when you get a 201 return code, and that is for created. 

**Describe the solution you'd like**
Every PUT that gets a return code of 201 should be treated as a create in the audit record.

**Describe alternatives you've considered**

**Acceptance Criteria**

1.  
   GIVEN [a precondition]
    AND [another precondition]
   WHEN [test step]
    AND [test step]
   THEN [verification step]
    AND [verification step]

**Additional context**
If you are creating a resource that is not version 1 of the resource, should that be also treated as a create if you got a 201.  Same for a delete, should that be treated as an update or a delete if it is a soft delete?

lmsurpre:
We do distinguish between creates and updates (for PUTs) for our whole-system history implementation.  Please consult that.

d0roppe:
Now every Put is recorded as a create in the Audit record.  So moving this back to in progress.

d0roppe:
Verified that the Put with a 200 return code is a Update, and a put with a 201 return code is a Create.  Verified the same behavior for bulk data import. Closing issue.

PR #2949, opened 2021-11-05
Audit record for PUTs should not always use "Update" #2471

prb112:


lmsurpre (approved):
LGTM

PR #2950, opened 2021-11-06
Stop swallowing exceptions in the $everything operation

lmsurpre:

### DIFF
--- a/operation/fhir-operation-everything/src/main/java/com/ibm/fhir/operation/everything/EverythingOperation.java
+++ b/operation/fhir-operation-everything/src/main/java/com/ibm/fhir/operation/everything/EverythingOperation.java
@@ -55,7 +55,6 @@ import com.ibm.fhir.server.spi.operation.FHIRResourceHelpers;
  * This class implements the <a href="https://www.hl7.org/fhir/operation-patient-everything.html">$everything</a> operation
  * which is used to return all the information related to one or more patients described in the resource or context on
  * which this operation is invoked.
- *
  */
 public class EverythingOperation extends AbstractOperation {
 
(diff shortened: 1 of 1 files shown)


operation/fhir-operation-everything/src/main/java/com/ibm/fhir/operation/everything/EverythingOperation.java
```diff
@@ -250,7 +258,8 @@ protected Parameters doInvoke(FHIROperationContext operationContext, Class<? ext
         try {
             outputParameters = FHIROperationUtil.getOutputParameters(bundleBuilder.build());
         } catch (Exception e) {
-            FHIROperationException exceptionWithIssue = buildExceptionWithIssue("An unexpected error occurred while creating the operation output parameters for the resulting Bundle.", IssueType.EXCEPTION);
+            FHIROperationException exceptionWithIssue = buildExceptionWithIssue("An unexpected error occurred while "
+                    + "creating the operation output parameters for the resulting Bundle.", IssueType.EXCEPTION, e);
```
prb112, line 262: I believe the PR review we asked that Luis not send back the exceptions.  This is a bit of a design change.  
I think this deserves a bit of javadoc explaining the choice so we don't go revisit this in the future and flip the other way.

lmsurpre: Are you thinking javadoc at each location where we catch + wrap?   Or just mention it once at the class level?

prb112: I was just thinking at the class level.

prb112 (approved):
LGTM, (adding the Javadoc is just a nice to have)

operation/fhir-operation-everything/src/main/java/com/ibm/fhir/operation/everything/EverythingOperation.java
lmsurpre: OK, how about I add something to our style guide at https://github.com/IBM/FHIR/wiki/Style-Guide about it instead?
We already cover exception handling there, so I could add a blurb about "wrap, rethrow, log and throw new, or handle; don't swallow unless you handle it"?

PR #2957, opened 2021-11-08
MemberMatchFactory should throw custom exception when Strategy is not found

prb112:
- MemberMatchFactory should throw custom exception when Strategy is not found

### DIFF
--- a/operation/fhir-operation-member-match/src/main/java/com/ibm/fhir/operation/davinci/hrex/provider/MemberMatchFactory.java
+++ b/operation/fhir-operation-member-match/src/main/java/com/ibm/fhir/operation/davinci/hrex/provider/MemberMatchFactory.java
@@ -10,9 +10,12 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.ServiceLoader;
 
+import com.ibm.fhir.exception.FHIROperationException;
+import com.ibm.fhir.model.type.code.IssueType;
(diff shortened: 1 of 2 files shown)


operation/fhir-operation-member-match/src/main/java/com/ibm/fhir/operation/davinci/hrex/provider/MemberMatchFactory.java
```diff
@@ -48,12 +51,13 @@ public static MemberMatchFactory factory() {
...
      * @param config
      * @return the MemberMatchStrategy or the DefaultMemberMatchStrategy
+     * @throws FHIROperationException
      */
-    public MemberMatchStrategy getStrategy(ConfigurationAdapter config) {
+    public MemberMatchStrategy getStrategy(ConfigurationAdapter config) throws FHIROperationException {
         String key = config.getStrategyKey();
         MemberMatchStrategy strategy = keyToMemberMatchStrategy.get(key);
         if (strategy == null) {
-            strategy = keyToMemberMatchStrategy.get(DEFAULT_STRATEGY);
```
lmsurpre, line 56: the default strategy is still the default though, right?  so this would only apply if someone has explicitly configured a strategy that is not in the config?

prb112: Correct

lmsurpre (approved):
LGTM

Issue #2958, opened 2021-11-08
Bulk Data Import skipping a resource marked deleted


d0roppe:
**Describe the bug**
Importing a resource that has been deleted does not ingest the resource again, but skips the import. 

Regression at this line
https://github.com/IBM/FHIR/commit/3d877c59542#diff-ad9adffb91185e4b3058adfc485a5a90f173fbd23b223afd39dbaba9fb7911f4R222
f41d8045be7 (Robin Arnold      2021-10-03 12:40:31 +0100 321)         resource = FHIRPersistenceUtil.copyAndSetResourceMetaFields(resource, logicalId, newVersionNumber, lastUpdated);

**Environment**
local/Main

**Expected behavior**
Should not skip a resource that is marked deleted.

d0roppe:
Verified that the bulk data import now works as expected.

PR #2959, opened 2021-11-08
Two Fixes - MemberMatch and BulkData Regression

prb112:
1 - Regression on $import - Bulk Data Import skipping a resource marked
deleted #2958 (https://github.com/IBM/FHIR/commit/3d877c59542#diff-ad9adffb91185e4b3058adfc485a5a90f173fbd23b223afd39dbaba9fb7911f4R222 introduced the regression as `delete status` is not checked)
2 - Fix for edge case when no identifier exists in MemberMatch

### DIFF
--- a/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/load/ChunkWriter.java
+++ b/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/load/ChunkWriter.java
@@ -220,7 +220,7 @@ public class ChunkWriter extends AbstractItemWriter {
 
                                 FHIRPersistenceEvent event = new FHIRPersistenceEvent(fhirResource, props);
 
-                                // Set up the persistence context to include NOT include the deleted resources when we read
+                                // Set up the persistence context to include the deleted resources when we read
                                 FHIRPersistenceContext persistenceContext = FHIRPersistenceContextFactory.createPersistenceContext(event, true);
                                 long startTime = System.currentTimeMillis();
                                 operationOutcome = conditionalFingerprintUpdate(chunkData, skip, localCache, fhirPersistence, persistenceContext, id, fhirResource);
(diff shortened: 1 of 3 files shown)


fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/load/ChunkWriter.java
```diff
@@ -219,7 +220,7 @@ public void writeItems(List<java.lang.Object> arg0) throws Exception {
 
                                 FHIRPersistenceEvent event = new FHIRPersistenceEvent(fhirResource, props);
 
-                                // Set up the persistence context to include deleted resources when we read
+                                // Set up the persistence context to include NOT include the deleted resources when we read
```
lmsurpre, line 223:
```suggestion
                                // Set up the persistence context to NOT include the deleted resources when we read
```

operation/fhir-operation-member-match/src/main/java/com/ibm/fhir/operation/davinci/hrex/provider/strategy/DefaultMemberMatchStrategy.java
```diff
@@ -186,6 +186,17 @@ public MemberMatchResult executeMemberMatch() throws FHIROperationException {
 
             Patient patient = patientBundle.getEntry().get(0).getResource().as(Patient.class);
             patient.accept(getPatientIdentifier);
+
+            /*
+             * QA: Since there is filtering ton the codesystem and value, the MB has to exist.
```
lmsurpre, line 191:
```suggestion
             * QA: Since there is filtering on the codesystem and value, the MB has to exist.
```

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/load/ChunkWriter.java
lmsurpre: the comment says we're gonna stop including deleted resources, but line 224 still retrieves the deleted instances.

I think we actually need to retrieve the deleted instance so that we can set the appropriate versionId.

prb112: Yes, we do. Thanks for catching this

lmsurpre (approved):
LGTM

Issue #2965, opened 2021-11-09
Location URI is wrong when resource id begins with resource type name


lmsurpre:
**Describe the bug**

**Environment**
main

**To Reproduce**
1. issue a PUT to `{{base}}/Patient/{{id}}` with an id like "Patient1"
2. note the location header in the response

The resource is created successfully, but the location header in the response has an extra `/Patient`:
`https://localhost:9443/fhir-server/api/v4/Patient/Patient/Patient1/_history/1`

**Expected behavior**
The location header should point to the proper absolute url of the created resource

**Additional context**
When we try to construct the baseUrl from the request, we mistake the `/Patient1` for the `/Patient` segment of the path.

d0roppe:
Verified this no longer happens.  Verified with 3 different resource types.

PR #2966, opened 2021-11-09
issue #2965 - modify getRequestBaseUri logic

lmsurpre:
Previously, we looked for the last instance of `/[resourceType]` in the path and assumed that was the first thing after the baseUrl. However, if a resourceId happened to match the resource type, this proved erroneous.

The updated logic looks for the first instance of `/[resourceType]/` in the path instead.
If that doesn't exist, then we fall back to the old approach of using the last index of `/[resourceType]` (and because we now know there was no `/[resourceType]/` segment before this, this approach should now be safe).

This updated logic could break if someone insisted on using a baseUrl that includes a path segment that matches `/[resourceType]/` (e.g. https://example.com/my/Patient/api/ ) but I think that is an acceptable risk because that would be very dumb to do. We could even add that guidance to the docs if we feel its needed.


fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -419,10 +422,11 @@ protected String getRequestBaseUri(String type) throws Exception {
 
         // Strip off any path elements after the base
         if (type != null && !type.isEmpty()) {
-            int resourceNamePathLocation = baseUri.lastIndexOf("/" + type);
```
lmsurpre, line 422: this is the key change.  see the commit description for considerations

prb112: what happens if this is hosted under `/Patient` or some other resource type. 
The intent of lastIndexOf was to be resilient to that approach. 

I think this is a trading off one approach for another which has similar issues. 

What happens if the the Logical ID is `Patient` in the code to the right? Wouldn't that exhibit a similar issue? 

Or if the server was hosted at https://server.me/PatientData/api/v4/Patient

The best way to harden this would be to pass in the JAXRS URL Path Segments and the JAXRS Query parameters and use these at this level to establish a resourceNamePathLocation as that will be clearly delineated among the  various options.   

This can be quickly packed into the FHIRRequestContext in the Jaxrs Filter

lmsurpre: > What happens if the the Logical ID is Patient in the code to the right? Wouldn't that exhibit a similar issue?

the proposed update here addresses exactly this issue (and even adds a test case for it)

> Or if the server was hosted at https://server.me/PatientData/api/v4/Patient

the propose update makes us flaky if someone changes the base path in a way such as this.  the difference is this base bath is controlled by the operator and not end users, and so i think its a much better option than what we have today


> I think this is a trading off one approach for another which has similar issues.

for sure.  i think its much better with the proposed change but its still just heuristic.  we can discuss if its worth doing something more robust tomorrow

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -419,12 +422,17 @@ protected String getRequestBaseUri(String type) throws Exception {
...
+            int resourceNamePathLocation = baseUri.indexOf("/" + type + "/");
```
michaelwschroeder, line 425: I think you want to use lastIndexOf here, so if you did happen to have the case where a path segment matched a particular resource type exactly, but the resource type was specified at the end, you'd still parse correctly (e.g. `https://example.com/my/Patient/api/Patient/123`),

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -419,12 +422,17 @@ protected String getRequestBaseUri(String type) throws Exception {
...
             if (resourceNamePathLocation != -1) {
```
michaelwschroeder, line 426: In this case, where you've found `/Patient/`, maybe you want to do an `endsWith("/" + type)` check and bump out to that index if found. Would correctly parse this path: `https://example.com/my/Patient/api/Patient`. Guess you'd want to check that it didn't directly follow a `/Patient` to make sure it wasn't a logical ID (i.e. `https://example.com/my/Patient/api/Patient/Patient`)

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
lmsurpre: the problem with lastIndexOf here is if a user created a resource with id "Patient" that logic would break.
for example:  `https://example.com/fhir/Patient/Patient/_history`

i'd rather err on the side of "operators should do the right thing" rather than "users should do the right thing".

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
lmsurpre: sorry, i'm not sure I understand the suggestion

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
michaelwschroeder: I'm ok with that. Should probably document it if there's somewhere that it makes sense to do so.

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
michaelwschroeder: This is a moot point if we're assuming no path segment matching resource type exactly and check in line 425 doesn't change to use `lastIndexOf`

michaelwschroeder (approved):
LGTM

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
prb112: I'd ask that you please add documentation stating the limitation or add a test case showing that we don't support it.

lmsurpre: OK, I added a test and the following documentation to the sections on `X-FHIR-FORWARDED-URL` and `fhirServer/core/externalBaseUrl`:

> Note that the base URL must not include a path segment that matches any FHIR resource type name (case-sensitive). For example, "https://example.com" or "https://example.com/my/patient/api" are fine, but "https://example.com/my/Patient/api" is not.

prb112 (approved):

Issue #2050, opened 2021-03-10
Support conditional create-on-update via If-None-Match


lmsurpre:
**Is your feature request related to a problem? Please describe.**
In https://github.com/IBM/FHIR/issues/160 we added support for conditional read via `If-None-Match`.

However, this header is also useful on the update/create-on-update side.  Specifically, it can be used to avoid creating the same resource multiple times.

**Describe the solution you'd like**
Clients can pass the `If-None-Match` header to us in their PUT request.
If `If-None-Match` is set to `*` and a resource exists at the passed location, then no update performed.

**Describe alternatives you've considered**
The spec defines a conditional create via a custom `If-None-Exists` header that is based on FHIR search.  We support that today, but it would be extremely hard to get the locking right if we want to guarantee no duplicates (see #2051).

**Acceptance Criteria**
1.  
  GIVEN a PUT to [base]/Observation/123
    WHEN the `If-None-Match` header is set to *
    AND a resource (any version) already exists at [base]/Observation/123
    THEN the response status is HTTP 304 (Not Modified)
    AND the response body is set according to the `Prefer: return` preference (representation=the existing resource, minimal=empty, OperationOutcome=a short description of why the resource wasn't updated)

2.  
  GIVEN a PUT to [base]/Observation/123
    WHEN the `If-None-Match` header is set to `W/"1"`
    AND a resource with version "1" already exists at [base]/Observation/123
    THEN the response status is HTTP 500 (Server Error)
    AND the response body is set according to the `Prefer: return` preference (minimal=empty, OperationOutcome=a description of the error explaining that If-None-Match with a specific version is not supported)

3.  
  GIVEN a PUT to [base]/Observation/123
    WHEN the `If-None-Match` header is set to *
    AND no resource exists at [base]/Observation/123
    THEN the response status is HTTP 201 (Created)
    AND the response body is set according to the `Prefer: return` preference (representation=the newly created resource, minimal=empty, OperationOutcome=any warnings or informational messages from validating the new resource version)

4.  
  GIVEN a PUT to [base]/Observation/123
    WHEN the `If-None-Match` header is not passed
    AND a resource exists (any version) at [base]/Observation/123
    THEN the response status is HTTP 200 (OK)
    AND the response body is set according to the `Prefer: return` preference (representation=the newly updated resource, minimal=empty, OperationOutcome=any warnings or informational messages from validating the new resource version)

5.
  GIVEN a PUT to [base]/Observation/123
[...]

PR #2968, opened 2021-11-10
issue #2050 - perform extra read in case of IfNoneMatch conflict

lmsurpre:
Question: should undeleting a deleted resource be handled like an
updateCreate?

Without this, we could run into cases where the locationURI could list
the deleted version of a resource.
Basically, handling the undelete like create-on-update means that we can
avoid doing the extra read in the normal IfNoneMatch case.


lmsurpre:
closing in favor of #2971

PR #2974, opened 2021-11-11
remove outdated property

lmsurpre:
I guess `additionalparam` went away with javadoc plugin 3.0.0.
In Eclipse its just a warning, but in IDEA its an error (by default).


sroorda (approved):
There are more references that need to be removed
fhir-parent/pom.xml - line 1121, 1172, 1429
fhir-examples/pom.xml - line 191, 419

prb112 (approved):

PR #2960, opened 2021-11-09
issue #2946 - add conditional reference value and entry index (again)

lmsurpre:
Signed-off-by: Lee Surprenant <lmsurpre@us.ibm.com>

PR #2978, opened 2021-11-12
issue 2050 ifNoneMatch returns 412 unless ifNoneMatchReturnsNotModified

punktilious:

### DIFF
--- a/fhir-server/src/main/java/com/ibm/fhir/server/rest/FHIRRestInteractionVisitorPersist.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/rest/FHIRRestInteractionVisitorPersist.java
@@ -277,8 +277,7 @@ public class FHIRRestInteractionVisitorPersist extends FHIRRestInteractionVisito
             }
         } catch (FHIRPersistenceResourceNotFoundException e) {
             if (failFast) {
-                String msg = "Error while processing request bundle.";
-                throw new FHIRRestBundledRequestException(msg, e).withIssue(e.getIssues());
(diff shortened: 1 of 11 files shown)


fhir-server/src/main/java/com/ibm/fhir/server/rest/FHIRRestInteractionVisitorPersist.java
```diff
@@ -300,6 +302,21 @@ private void doInteraction(int entryIndex, String requestDescription, long initi
                         .build())
                     .build();
             setEntryComplete(entryIndex, entry, requestDescription, initialTime);
+        } catch (FHIRPersistenceIfNoneMatchException e) {
+            if (failFast) {
+                String msg = "Error while processing request bundle.";
+                throw new FHIRRestBundledRequestException(msg, e).withIssue(e.getIssues());
+            }
```
lmsurpre, line 309: I updated some very similar code in the past to include the entryIndex in both the error message and the issue...maybe we should do the same in each catch block here?

lmsurpre: https://github.com/IBM/FHIR/pull/2960/files#diff-35b7e53e0ed72c5da33f937ee4634c89bb018a496a8c2db4e220a75117c59507R396

punktilious: Yep, I'll do that

lmsurpre (approved):
LGTM

PR #2980, opened 2021-11-12
issue 2979 fix schema version tracking regression

punktilious:

### DIFF
--- a/fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
+++ b/fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
@@ -459,37 +459,38 @@ public class Main {
                 buildFhirDataSchemaModel(pdm);
                 boolean isNewDb = updateSchema(pdm);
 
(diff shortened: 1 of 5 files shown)


fhir-persistence-schema/src/main/java/com/ibm/fhir/schema/app/Main.java
```diff
@@ -523,13 +524,15 @@ protected void updateOauthSchema() {
...
-                if (grantTo != null) {
-                    grantPrivilegesForOAuth();
+                if (this.exitStatus == EXIT_OK) {
+                    // Apply privileges if asked
+                    if (grantTo != null) {
+                        grantPrivilegesForOAuth();
+                    }
+    
+                    // Mark the schema as up-to-date
+                    svm.updateSchemaVersion();
```
prb112, line 534: I'm sorry I didn't see this before.  

What happens when someone runs an older version than the current one, would this ever end up with an out of sync version? It seems to insert without checking the version number.

lmsurpre (approved):
LGTM

PR #3008, opened 2021-11-18
upgrade to jakarta.ws.rs 3.0.0

lmsurpre:
This involves updating the javax.ws.rs imports to the jakarta.ws.rs package.
I also removed some fully-qualified class names from the middle of a handful of classes.

Finally, I grouped a number of test dependencies in the fhir-parent/pom.xml and
moved them to the end of the dependencies list.


lmsurpre:
Looks like we can't actually make this change until either:
A. CXF updates their implementation to use the updated package name (specifically `META-INF/services/javax.ws.rs.ext.RuntimeDelegate` from cxf-rt-frontend-jaxrs); or
B. we move to RESTEasy and they support this same change (5.0.0 doesn't support it but the next major release will)

lmsurpre:
from https://lists.jboss.org/archives/list/resteasy@lists.jboss.org/thread/ZCXCCH76MNZ7LS2LPX3HHIVSUUTMVNVJ/
> I don't think that development of a "native" 3.0 implementation has started
yet (other than Jersey), but what we are doing in Open Liberty betas is
taking the latest RESTEasy modules and "transforming" them so that they
will use the new namespace. We're using a tool called Eclipse Transformer
to do that (more info here: https://github.com/eclipse/transformer ).

PR #3027, opened 2021-11-22
Update README.md to remove projects that are not in main

prb112:


lmsurpre (approved):
good catch

PR #3040, opened 2021-11-24
Remove undocumented feature for overriding the http method via header and related cleanup

lmsurpre:
1. We had an undocumented / untested feature where you can override the
HTTP method via a header. This changeset removes that. CXF has a
slightly different header for doing exactly the same thing. Not sure
about RESTEasy.

2. We had a whole lot of unnecessary override methods in
FHIRHttpServletRequestWrapper. Since the parent class defaults to
calling the delegate already, its cleaner to leave those out of this
subclass.

3. Committing minor edits to the JAX-RS Search resource...it doesn't
actually consume APPLICATION_FHIR media types in any of the methods and
there's no need to override the class-level path with the same value

### DIFF
--- a/fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRHttpServletRequestWrapper.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRHttpServletRequestWrapper.java
@@ -1,33 +1,27 @@
 /*
- * (C) Copyright IBM Corp. 2016, 2020
+ * (C) Copyright IBM Corp. 2016, 2021
  *
  * SPDX-License-Identifier: Apache-2.0
  */
 
 package com.ibm.fhir.server.filter.rest;
 
-import java.io.BufferedReader;
-import java.io.IOException;
+import static javax.ws.rs.core.HttpHeaders.ACCEPT;
+import static javax.ws.rs.core.HttpHeaders.ACCEPT_CHARSET;
+
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
-import java.security.Principal;
 import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 import java.util.Vector;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-import javax.servlet.RequestDispatcher;
-import javax.servlet.ServletInputStream;
-import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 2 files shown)


fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRHttpServletRequestWrapper.java
```diff
@@ -763,247 +487,19 @@ public String getQueryString() {
...
-    /**
-     * @param arg0
-     * @throws UnsupportedEncodingException
-     * @see javax.servlet.ServletRequest#setCharacterEncoding(java.lang.String)
-     */
-    @Override
-    public void setCharacterEncoding(String arg0) throws UnsupportedEncodingException {
-        delegate.setCharacterEncoding(arg0);
+    @SuppressWarnings("unused")
+    private String displayHeaderValues(Enumeration<String> headers) {
```
lmsurpre, line 493: this was there before, i just added the javadoc for it.  i think we could probably axe this one as well.

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRHttpServletRequestWrapper.java
```diff
@@ -6,27 +6,22 @@
 
 package com.ibm.fhir.server.filter.rest;
 
-import java.io.BufferedReader;
-import java.io.IOException;
+import static javax.ws.rs.core.HttpHeaders.ACCEPT;
```
prb112, line 9: copyright year

lmsurpre: fixed

prb112 (approved):
LGTM - good cleanup

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRHttpServletRequestWrapper.java
```diff
@@ -227,7 +216,7 @@ private void formParameters(HttpServletRequest req) {
      */
     private String updateAcceptHeader(String s) {
         if (s!= null && !s.contains(CHARSET)) {
-            String originalHeaderValue = delegate.getHeader(ACCEPT);
+            String originalHeaderValue = delegate.getHeader(HttpHeaders.ACCEPT);
```
lmsurpre, line 219:
```suggestion
            String originalHeaderValue = delegate.getHeader(ACCEPT);
```

lmsurpre: forgot to clean this up after adding static import

PR #3041, opened 2021-11-24
bump spark-sql version to 3.2.0

lmsurpre:
its only a provided dependency, but there's a few vulnerabilities in the transitive dependencies for 3.1.0.

dependabot opened this for it in the past https://github.com/IBM/FHIR/pull/2985 ...we closed but I forget why

### DIFF
--- a/fhir-parent/pom.xml
+++ b/fhir-parent/pom.xml
@@ -122,7 +122,7 @@
             <dependency>
                 <groupId>org.slf4j</groupId>
                 <artifactId>slf4j-api</artifactId>
-                <version>1.7.30</version>
+                <version>1.7.32</version>
             </dependency>
             <dependency>
                 <groupId>org.slf4j</groupId>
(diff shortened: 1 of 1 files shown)


prb112:
yeah - it doesn't work properly with our environment.  If you've tested this. I'll approve, I'd rather not accept it until it's tested

lmsurpre:
I ran the following local smoketest:
1. marked the spark and stocator dependencies as runtime (instead of `provided`)
2. deployed to a local test server
3. configured the server with an ibm-cos storageProvider that has `"enableParquet": true`1
4. loaded a db with 1690 practitioner resources, 
5. ran an export with `_outputFormat=application/fhir+parquet` and `_type=Practitioner`

I noted this, but I *think* we had the same warning with 3.1.0 and I don’t think there’s much we could do about it:
```
Illegal reflective access by org.apache.spark.unsafe.Platform (file:/Users/lmsurpre/.m2/repository/org/apache/spark/spark-unsafe_2.12/3.2.0/spark-unsafe_2.12-3.2.0.jar) to constructor java.nio.DirectByteBuffer(long,int)
```

I also noted some odd logging behavior, but I'm not sure if thats a regression either...still investigating.

prb112 (approved):
LGTM - discussed with Lee offline

lmsurpre:
I confirmed that the odd logging behavior is not a regression (we get the same odd behavior with the current versions).
I opened https://github.com/IBM/FHIR/issues/3070 for that one and I think we're good to merge this in the meantime.

PR #3045, opened 2021-11-25
issue 3044 - use correct TransformerFactory impl class

### LINKED ISSUE #3044: issue-2266 implementation causes regression of 1226 FHIRGeneratorException when serializing resource <div> element as XML
**Describe the bug**
Issue #1226 was implemented to ensure that fhir-model always uses the correct TransformerFactory class.

Issue #2266 changed the implementation to address an illegal reflective action warning which would become an error in a future version of Java. However, this causes a regression in behavior in one case where the user is using fhir-validation-dist, not the whole server.

**Environment**
Which version of IBM FHIR Server: 4.10.0 and 4.10.1.

**To Reproduce**
1. See #1226 

**Expected behavior**
The test described by #1226 should pass.

**Additional context**

punktilious:

Fixes #3044.

### DIFF
--- a/fhir-model/src/main/java/com/ibm/fhir/model/util/XMLSupport.java
+++ b/fhir-model/src/main/java/com/ibm/fhir/model/util/XMLSupport.java
@@ -26,12 +26,16 @@ import javax.xml.transform.TransformerFactory;
 public final class XMLSupport {
     public static final String FHIR_NS_URI = "http://hl7.org/fhir";
     public static final String XHTML_NS_URI = "http://www.w3.org/1999/xhtml";
-
+    
+    // System properties used to control XML factory behavior
     private static final String PROP_XML_INPUT_FACTORY = "javax.xml.stream.XMLInputFactory";
+    private static final String PROP_XML_INPUT_FACTORY_CONFIG = "com.ibm.fhir.xml.input.factory";
     private static final String PROP_XML_OUTPUT_FACTORY = "javax.xml.stream.XMLOutputFactory";
-    private static final String PROP_TRANSFORMER_FACTORY = "javax.xml.stream.XMLTransformerFactory";
+    private static final String PROP_XML_OUTPUT_FACTORY_CONFIG = "com.ibm.fhir.xml.output.factory";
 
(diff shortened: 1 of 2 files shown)


lmsurpre (approved):
I had tried leaving this configurable per the design of these XML factories, but I agree it makes more sense just to instantiate a known-good impl since we know there are issues with other implementations.  LGTM.

fhir-model/src/main/java/com/ibm/fhir/model/util/XMLSupport.java
```diff
@@ -226,15 +226,8 @@ private static XMLOutputFactory createXMLOutputFactory() {
 
     private static TransformerFactory createTransformerFactory() {
         try {
-            boolean isSet = System.getProperty(PROP_TRANSFORMER_FACTORY) != null;
-            if (!isSet) {
-                System.setProperty(PROP_TRANSFORMER_FACTORY, TRANSFORMER_FACTORY_IMPL);
```
lmsurpre, line 231: we can probably remove the PROP_TRANSFORMER_FACTORY now that its not used here

fhir-persistence-schema/README.md
```diff
@@ -493,3 +493,282 @@ If there is data in the DOMAINRESOURCE and RESOURCE table groups, which is unexp
 
 
 FHIR® is the registered trademark of HL7 and is used with the permission of HL7.
+
+
+# Physical Data Model
```
lmsurpre, line 498: looks like some great doc!  did you mean to include it in this PR?

lmsurpre (approved):
LGTM

fhir-model/src/main/java/com/ibm/fhir/model/util/XMLSupport.java
punktilious: Ah yes...copied it back when I reverted the isSet stuff. I'll take it out

punktilious: Wait...no...it's not there - already gone

Issue #3067, opened 2021-12-02
Sonatype lift reports three vulnerabilities on fhir-term-graph-loader


prb112:
**Describe the bug**
Sonatype lift reports three vulnerabilities on fhir-term-graph-loader

**Environment**
Which version of IBM FHIR Server? 4.10.2

**Details**
pkg:maven/org.apache.thrift/libthrift@0.9.2
pkg:maven/org.apache.ant/ant@1.7.0
pkg:maven/org.hibernate/hibernate-validator@4.3.0.Final

Link to Report
https://sbom.lift.sonatype.com/report/T1-0ff0976f7f21c391f20f-ebcc64500db6f-1638456711-641a223fd2bb4f16b87b0f4d3f84c7b3

PR #3068, opened 2021-12-02
Sonatype lift reports three vulnerabilities on fhir-term-graph-loader #3067

prb112:
Excluded Three Dependencies which are not used


lmsurpre (approved):
LGTM

PR #3079, opened 2021-12-06
Fix for Bad text

prb112:


lmsurpre (approved):
From the diff I don't see how this would change things, but it looks fine to merge.

Issue #3084, opened 2021-12-06
Error when trying to evaluate cql when us-core ig jar also loaded


ranum:
**Describe the bug**
Trying to create an extended fhir server by placing jar files into the userlib directory using the ibm-fhir-server services approach.  Wanted to add both **fhir-operation-cpg-4.10.1-shaded.jar** and **fhir-operation-cqf-4.10.1-shaded.jar** as well as **fhir-ig-us-core-4.10.1.jar**.  cql evaluation works without the us-core jar but breaks after it is added.  

Example cql, something like

<base url>/$cql?expression = [Patient] p where p.gender = 'male' & subject= Patient/...

**Environment**
Which version of IBM FHIR Server?
4.10.1

**To Reproduce**
These are the steps used when deploying via health patterns helm chart.  You may be able to reproduce by simply adding the three jars to your dev environment server.

Within the Alvearie health patterns, use the pom.xml below to build a new container for the fhir server that includes all three jars mentioned above

```
<project xmlns="http://maven.apache.org/POM/4.0.0"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
	<modelVersion>4.0.0</modelVersion>
	<groupId>com.ibm.fhir</groupId>
	<artifactId>fhir-ibmcloud</artifactId>
	<version>99-SNAPSHOT</version>
	<packaging>pom</packaging>

	<properties>
		<fhir.version>4.10.1</fhir.version>
	</properties>

	<dependencies>
		<dependency>
			<groupId>${project.groupId}</groupId>
			<artifactId>fhir-operation-cqf</artifactId>
			<version>${fhir.version}</version>
			<classifier>shaded</classifier>
		</dependency>
		<dependency>
			<groupId>${project.groupId}</groupId>
			<artifactId>fhir-operation-cpg</artifactId>
			<version>${fhir.version}</version>
			<classifier>shaded</classifier>
		</dependency>
		<dependency>
			<groupId>${project.groupId}</groupId>
			<artifactId>fhir-ig-us-core</artifactId>
			<version>${fhir.version}</version>
			<exclusions>
				<exclusion>
					<groupId>*</groupId>
					<artifactId>*</artifactId>
				</exclusion>
			</exclusions>
		</dependency>
	</dependencies>

	<build>
		<plugins>
			<plugin>
[...]

Issue #3064, opened 2021-12-01
$submit-data operation fails 


kmbarton423:
Running IBM FHIR Server 4.10.0 with CPG and CQF operations.
Executed $submit-data operation for an existing MeasureReport:  
   POST {{local1}}{{context1}}/Measure/$submit-data  
      (see attached requestBody.txt)

Status 200 received yet operation fails:
```{
  "resourceType": "OperationOutcome",
  "id": "ac-11-0-3-cde64786-f73c-408f-a87b-3b1a174101be",
  "issue": [
    {
      "severity": "fatal",
      "code": "exception",
      "details": {
        "text": "FHIROperationException: Operation failed"
      }
    }
  ]
}
```

Logs indicate "PSQLException: ERROR: Concurrent update - mismatch of version in JSON"
(see attached out.log)

The core issue is related to MeasureSubmitDataOperation.java on line 80. 
The switch between PUT/POST depends on a VERSION. 
On the second submission of the same data element, the version is passed into the persistence layer, and the persistence layer sees a violation of the stored version (v1) violates the incoming version (v1). 

Options, remove the version in that code, or use ifNoneMatch and handle the response.

**Additional context**

- [requestBody.txt](https://github.com/IBM/FHIR/files/7637140/requestBody.txt)
- [out.log](https://github.com/IBM/FHIR/files/7637154/out.log)

kmbarton423:
Receive the following after the fix:

Status 409 Conflict

`{
	"resourceType": "OperationOutcome",
	"id": "ac-11-0-3-e8b4cd0e-8ba2-494c-af1d-b34a53d711c0",
	"issue": [
		{
			"severity": "fatal",
			"code": "conflict",
			"details": {
				"text": "Encountered version id mismatch while inserting Resource"
			},
			"expression": [
				"<empty>",
				"Bundle.entry[1]"
			]
		}
	]
}`

Issue #3061, opened 2021-12-01
null pointer exception exercising operations $collect-data and $evaluate-measure


kmbarton423:
Running IBM FHIR Server 4.10.0 with CPG and CQF operations.
When attempting to confirm required and optional parameters, hit null pointer exception:

```
   "resourceType": "OperationOutcome",
  "issue": 
      "severity": "fatal",
      "code": "exception",
      "details": {
        "text": "NullPointerException: &lt;null message&gt;"
```

Null pointer exception received for each of the following cases:
- /Measure/$collect-data?periodStart=2014-01-31&periodEnd=2014-12-31&measure=Measure/measure-EXM104-8.2.000
- /Measure/$evaluate-measure?periodStart=2014-01-31&periodEnd=2014-12-31&measure=Measure/measure-EXM104-8.2.000
- /Measure/measure-EXM104-8.2.000/$evaluate-measure?periodStart=2014-01-31&periodEnd=2014-12-31

csandersdev:
Looks like I'm not handling the "Measure not found" use case very well. You need to load the Measure resource you are trying to use into your server prior to invoking these methods. The resources that I use to test with are in [fhir-server-test/src/test/resources/testdata/Bundle-EXM74-10.2.000.json](https://github.com/IBM/FHIR/blob/main/fhir-server-test/src/test/resources/testdata/Bundle-EXM74-10.2.000.json), [cql/operation/fhir-operation-cpq/src/test/resources](https://github.com/IBM/FHIR/tree/main/cql/operation/fhir-operation-cpg/src/test/resources), and [cql/operation/fhir-operation-cqf/src/test/resources](https://github.com/IBM/FHIR/tree/main/cql/operation/fhir-operation-cqf/src/test/resources)

[12/1/21, 13:26:14:360 EST] 00000046 com.ibm.fhir.server.resources.FHIRResource                   E An unexpected exception occurred while processing the request^M
java.lang.NullPointerException^M
    at com.ibm.fhir.operation.cqf.AbstractMeasureOperation.doMeasureEvaluation(AbstractMeasureOperation.java:90)^M
    at com.ibm.fhir.operation.cqf.EvaluateMeasureOperation.doInvoke(EvaluateMeasureOperation.java:91)^M
    at com.ibm.fhir.operation.cqf.MeasureCollectDataOperation.doInvoke(MeasureCollectDataOperation.java:51)^M
[...]

Issue #2618, opened 2021-07-15
Add support for prefetchData and remote content in CPG cql operations


csandersdev:
**Is your feature request related to a problem? Please describe.**
The $cql and Library/$evaluate operations implemented in #2577 do not currently support the prefetchData or remote content, data, and endpoint optional input parameters. This ticket is for adding that support.

**Describe the solution you'd like**
Support the input parameters defined in the operation definitions.

**Describe alternatives you've considered**
The remote terminology endpoint is already partially supported through the use of the term service configuration in server properties. 

**Acceptance Criteria**
Operations behave in the manner described in the operation definitions.

csandersdev:
The initial implementation of this issue ignored these unsupported operation parameters. In #3093, logic was added to warn the user when unsupported operation parameters were used in a request, so that behavior is more clear to the end user. The core issue documented here for adding support is as yet unaddressed. We should leave it open.

PR #3093, opened 2021-12-07
Bug fixes for CQL-related functions

csandersdev:
This PR addresses a number of issues related to the CQL / clinical reasoning module. 

#3061 - Added more robust reference resolution and better error handling
#2618 - Added error handling for the unsupported features. Keeping that issue open for future implementation of the feature.
#3064 - Clearing the meta.versionId on posted resources now. After discussing with the team, it is a known limitation that users will not be able to post the same resource twice as input to this operation. It will fail during the bundle processing.
#3084 - Switched the parameter resolution from SearchParameter.name to SearchParameter.code to deal with the fact that us-core parameter names do not match the code value like they do in the core parameter definitions.


cql/operation/fhir-operation-cpg/src/test/java/com/ibm/fhir/operation/cpg/CqlOperationTest.java
```diff
@@ -54,6 +60,75 @@
...
+                .clazz(reason)
+                .period(Period.builder()
+                    .start(DateTime.now())
+                    .end(DateTime.now())
+                    .build())
+                .build();
+
+        String expression = "[Encounter] e where e.status = 'finished'";
+
+        Parameter pSubject = Parameter.builder().name(fhirstring(CqlOperation.PARAM_IN_SUBJECT)).value(fhirstring("Patient/" + patient.getId())).build();
```
lmsurpre, line 84: you shouldn't need to use `fhirstring` any more, you can pass your java string directly

lmsurpre (approved):

cql/fhir-cql/src/main/java/com/ibm/fhir/cql/engine/searchparam/SearchParameterResolver.java
```diff
@@ -41,7 +41,7 @@ public SearchParameter getSearchParameterDefinition(String resourceType, String
         Map<String, SearchParameter> params = SearchUtil.getSearchParameters(resourceType);
 
         for (SearchParameter param : params.values()) {
-            if (name != null && param.getName().getValue().equals(name)) {
+            if (name != null && param.getCode().getValue().equals(name)) {
```
prb112, line 44: This is a good find. I went ahead and checked the call hierarchy for getCode and the above are the only situations where it is used.

prb112 (approved):

Issue #2453, opened 2021-06-02
CXF LoggingFeature is deprecated, and it'd be nice to move the fhir-client to the latest approach


prb112:
**Is your feature request related to a problem? Please describe.**
CXF LoggingFeature is deprecated, and it'd be nice to move the fhir-client to the latest approach

https://stackoverflow.com/questions/54989929/apache-cxf-loggingininterceptor-is-deprecated-what-to-use-instead

**Describe the solution you'd like**
Upgrade the LoggingFeature to the latest

**Describe alternatives you've considered**
Do Nothing or Move to RestEasy

*Please do NOT implement until a team review*

**Acceptance Criteria**

1.  
   GIVEN [a precondition]
    AND [another precondition]
   WHEN [test step]
    AND [test step]
   THEN [verification step]
    AND [verification step]

**Additional context**

lmsurpre:
I think we should consider moving the whole thing over to RestEasy...

lmsurpre:
We should also look at removing the `provided` dependency on CXF from `fhir-term-remote`.  We think its just using JAX-RS / isn't actually tied to CXF or the fhir-client.

To use the non-deprecated approach for CXF: probably a couple points

To switch over to RestEasy wholesale, upwards of 5 points.

PR #3104, opened 2021-12-09
CXF LoggingFeature is deprecated, and it'd be nice to move the fhir-client to the latest approach #2453

prb112:
- Turns out the LoggingFeature is moved and slightly different
- Updated to make it work, and not-use the deprecated feature

### DIFF
--- a/fhir-client/pom.xml
+++ b/fhir-client/pom.xml
@@ -80,10 +80,6 @@
                 </exclusion>
             </exclusions>
         </dependency>
-<!--         <dependency> -->
-<!--             <groupId>org.slf4j</groupId> -->
-<!--             <artifactId>slf4j-nop</artifactId> -->
-<!--         </dependency> -->
         <dependency>
             <groupId>org.testng</groupId>
             <artifactId>testng</artifactId>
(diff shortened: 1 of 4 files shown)


fhir-client/pom.xml
```diff
@@ -65,6 +66,24 @@
...
+                <exclusion>
+                    <groupId>org.jboss.spec.javax.rmi</groupId>
+                    <artifactId>jboss-rmi-api_1.0_spec</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+<!--         <dependency> -->
+<!--             <groupId>org.slf4j</groupId> -->
+<!--             <artifactId>slf4j-nop</artifactId> -->
+<!--         </dependency> -->
```
lmsurpre, line 86: can we remove this?  if not, please add a comment to indicate why its commented out / when to uncomment

lmsurpre (approved):

fhir-client/pom.xml
prb112: Whoops... yes, we can remove it.

Issue #3106, opened 2021-12-09
Large Bundles with lots of new CodeSystems and Ids hits JDBC Parameter Limit


prb112:
**Describe the bug**

```
Caused by: java.io.IOException: Tried to send an out-of-range integer as a 2-byte value: 54842
```

**Environment**
Which version of IBM FHIR Server? main (and any with common_token_values)

**To Reproduce**
1. Create an empty db 
2. Load fhir-server-test/src/test/resources/testdata/Bundle-ValueSets.json in a Transaction
3. See the exception 
```
Size 27421
[AUDIT   ] J2CA0056I: The Connection Manager received a fatal connection error from the Resource Adapter for resource jdbc/fhir_default_default. The exception is: org.postgresql.util.PSQLException: An I/O error occurred while sending to the backend.:java.io.IOException: Tried to send an out-of-range integer as a 2-byte value: 54842
[...]

PR #3108, opened 2021-12-10
Large Bundles with lots of new CodeSystems and Ids hits JDBC Parameter Limit #3106

prb112:
Large Bundles with lots of new CodeSystems and Ids hits JDBC Parameter Limit #3106

- Added Optional Max Query Parameter for Postgres, Db2 
- Add Test Updates for CQL
- Formatted the Bundle-valuesets.json
- Added Cleanup for the Test Data
- Added SkipException for Profiles that are not loaded (Bulk Data too) 
- Added a Cache for Metadata requests across TestNG Threads


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceReferenceDAO.java
```diff
@@ -756,78 +753,98 @@ public void upsertCommonTokenValues(List<ResourceTokenValueRec> values) {
...
+            // Make sure we don't overflow.
+            if (high > sortedTokenValues.size()) {
+                high = sortedTokenValues.size();
             }
-            inList.append("(?,?)");
-        }
 
-        final String paramListStr = paramList.toString();
-        doCommonTokenValuesUpsert(paramListStr, sortedTokenValues);
+            List<CommonTokenValue> sortedTokenValuesSub = sortedTokenValues.subList(low, high - 1);
```
prb112, line 777: This is the new bit...

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceReferenceDAO.java
```diff
@@ -756,78 +753,98 @@ public void upsertCommonTokenValues(List<ResourceTokenValueRec> values) {
...
+                    if (id != null) {
+                        xr.setCommonTokenValueId(id);
+
+                        // update the thread-local cache with this id. The values aren't committed to the shared cache
+                        // until the transaction commits
+                        cache.addTokenValue(key, id);
+                    }
                 }
             }
+            low = high;
```
prb112, line 847: This is the bit that moves the list forward

fhir-server-test/src/test/java/com/ibm/fhir/server/test/cpg/ServerCqlOperationTest.java
```diff
@@ -33,57 +33,58 @@
...
         setUp(testProperties);
-        
+
         JsonObject jsonObject = TestUtil.readJsonObject("testdata/Patient_SallyFields.json");
         Entity<JsonObject> entity = Entity.entity(jsonObject, FHIRMediaType.APPLICATION_FHIR_JSON);
         Response response = getWebTarget().path("/" + TEST_PATIENT_ID).request().put( entity );
         assertResponse( response, Response.Status.Family.SUCCESSFUL );
+
+        // Add to a registry so the AfterClass method deletes it.
+        this.addToResourceRegistry("Patient", "sally-fields");
```
prb112, line 43: one change to remove resource

fhir-server-test/src/test/java/com/ibm/fhir/server/test/cpg/ServerLibraryEvaluateOperationTest.java
```diff
@@ -36,83 +37,109 @@
...
         assertEquals( outcome.getIssue().get(0).getCode(), IssueType.INVALID);
     }
-    
+
     @Test
     public void testLibraryEvaluateWithDebug() {
         Response response = getWebTarget().path("/Library/$evaluate").queryParam("library", "http://ibm.com/health/Library/EXM74|10.2.000").queryParam("subject", TEST_PATIENT_ID).queryParam("debug", "true").request().get();
         assertResponse( response, 200 );
-        
+
```
prb112, line 142: I formatted the contents in the JSON that follows.

cql/fhir-cql/src/main/java/com/ibm/fhir/cql/engine/searchparam/ReferenceParameter.java
```diff
@@ -68,4 +69,13 @@ public ReferenceParameter setValue(String value) {
...
     }
+
+    @Override
+    public String toString() {
+        String rtm = null;
+        if (resourceTypeModifier != null && resourceTypeModifier.getValue() != null) {
+            rtm = resourceTypeModifier.getValue();
+        }
+        return "ReferenceParameter [resourceTypeModifier=" + rtm + ", chainedProperty=" + chainedProperty + ", value=" + value + "]";
+    }
```
prb112, line 80: There was some funky multiline logging where ReferenceParameter caused JSON content to output as Pretty... instead I used the primitive here.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/api/IDatabaseTranslator.java
```diff
@@ -214,4 +215,12 @@
...
      */
     default boolean isIndexUseSchemaPrefix() { return true; }
+
+    /**
+     * @return the total number of parameters supported by the data store during a query.
+     *  An empty optional indicates no maximum is set.
+     */
+    default Optional<Integer> maximumQueryParameters() {
+        return Optional.empty();
+    }
```
prb112, line 225: I chose to add this as we hit issues, we can implement for other providers.  My guess is derby/db2 have a very high parameter count.

fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/postgres/PostgresTranslator.java
```diff
@@ -267,4 +268,10 @@ public String pagination(int offset, int rowsPerPage) {
         return result.toString();
     }
 
+    @Override
+    public Optional<Integer> maximumQueryParameters() {
+        // https://github.com/pgjdbc/pgjdbc/blob/13a644b4945996dff5274e45741313380d4a47e6/pgjdbc/src/main/java/org/postgresql/core/PGStream.java#L273
+        // it's Short.MAX_VALUE
+        return Optional.of(Integer.valueOf(32767));
```
prb112, line 275: Per the JDBC driver it's Short.MAX_VALUE

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceReferenceDAO.java
```diff
@@ -756,78 +753,104 @@ public void upsertCommonTokenValues(List<ResourceTokenValueRec> values) {
...
-            while (rs.next()) {
-                // SELECT code_system_id, token_value...note codeSystem not required
-                CommonTokenValue key = new CommonTokenValue(null, rs.getInt(1), rs.getString(2));
-                idMap.put(key, rs.getLong(3));
+            // The sublist as we have a maxQuery
+            List<CommonTokenValue> sortedTokenValuesSub;
+            if (!maxQuery.isEmpty()) {
+                sortedTokenValuesSub = sortedTokenValues.subList(low, high - 1);
+            } else {
+                sortedTokenValuesSub = sortedTokenValues;
```
prb112, line 782: Up to here is new...

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceReferenceDAO.java
```diff
@@ -756,78 +753,104 @@ public void upsertCommonTokenValues(List<ResourceTokenValueRec> values) {
...
+                    if (id != null) {
+                        xr.setCommonTokenValueId(id);
+
+                        // update the thread-local cache with this id. The values aren't committed to the shared cache
+                        // until the transaction commits
+                        cache.addTokenValue(key, id);
+                    }
                 }
             }
+            low = high + 1;
```
prb112, line 853: moves the list forward

prb112:
@csandersdev if you can review the CQL test changes and ReferenceParameter changes. 

@punktilious @lmsurpre if you can review the rest... I added some test optimizations and cache for the CapabilitiesStatement.

csandersdev (approved):
The CQL changes look fine to me.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceReferenceDAO.java
punktilious: What is the semantic impact of only using the subset of values? If we're hitting the limit here, I think we should be using an alternative strategy (perhaps a temporary table).

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceReferenceDAO.java
punktilious: with subList the high is exclusive, so I think you're skipping an entry here. Should be `low = high` I believe.

punktilious (changes requested):
See comment about subList where high is exclusive.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceReferenceDAO.java
prb112: It means that insert is at least 1 operation and potentially more.  `1..*` executions 

Semantically the `merge into` is probably the best option rather than a pseudo `upsert`. 

I'd say the likelihood of that many unique code systems values and ids is fairly low once a system is somewhat loaded with data, and the cache is active and warm.  Then the potential for multiple inserts is low.  

Yes, a temporary table is an option.  It's probably the better option in the longer run.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceReferenceDAO.java
prb112: I think you are right... let me double check.  I know I added low/high logging just to verify this.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceReferenceDAO.java
punktilious: I think we can come back to this when we revisit the cache stuff which will require us to move everything to transaction data and push everything to the database during commit. In some cases it might be easier just to use batch operations.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceReferenceDAO.java
```diff
@@ -756,78 +753,104 @@ public void upsertCommonTokenValues(List<ResourceTokenValueRec> values) {
...
-        }
+        // Process segments up to maximum query parameter size.
+        for (int low = 0; low < sortedTokenValues.size();/*at end  of loop*/) {
+            int high = sortedTokenValues.size();
 
-        final String paramListStr = paramList.toString();
-        doCommonTokenValuesUpsert(paramListStr, sortedTokenValues);
+            if (!maxQuery.isEmpty() && high > maxQuery.get()/2) {
+                // 2 as there a pair inserted.
+                high = low + maxQuery.get()/2;
```
punktilious, line 769: What if maxQuery were set to 1?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceReferenceDAO.java
```diff
@@ -756,78 +753,104 @@ public void upsertCommonTokenValues(List<ResourceTokenValueRec> values) {
...
-        select.append("             AND ctv.code_system_id = v.code_system_id ");
-
-        // Grab the ids
-        Map<CommonTokenValue, Long> idMap = new HashMap<>();
-        try (PreparedStatement ps = connection.prepareStatement(select.toString())) {
-            int a = 1;
-            for (CommonTokenValue tv: sortedTokenValues) {
-                ps.setString(a++, tv.getTokenValue());
-                ps.setInt(a++, tv.getCodeSystemId());
+            // Make sure we don't overflow.
```
punktilious, line 772: this if block can be inside the above if block, I think. It only applies of the high has been recalculated.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceReferenceDAO.java
```diff
@@ -756,78 +753,104 @@ public void upsertCommonTokenValues(List<ResourceTokenValueRec> values) {
...
 
-            ResultSet rs = ps.executeQuery();
-            while (rs.next()) {
-                // SELECT code_system_id, token_value...note codeSystem not required
-                CommonTokenValue key = new CommonTokenValue(null, rs.getInt(1), rs.getString(2));
-                idMap.put(key, rs.getLong(3));
+            // The sublist as we have a maxQuery
+            List<CommonTokenValue> sortedTokenValuesSub;
+            if (!maxQuery.isEmpty()) {
+                sortedTokenValuesSub = sortedTokenValues.subList(low, high - 1);
```
punktilious: subList upper value is exclusive. foo.subList(0, foo.size()) is the whole list. The above is chopping the last element.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceReferenceDAO.java
prb112: changed the implementation to use a single loop

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/ResourceReferenceDAO.java
```diff
@@ -756,76 +754,96 @@ public void upsertCommonTokenValues(List<ResourceTokenValueRec> values) {
...
-            // also build the inList for the select statement later
-            if (inList.length() > 0) {
-                inList.append(",");
-            }
-            inList.append("(?,?)");
+        // The maximum number of query parameters that are available for a particular persistence layer.
+        // There are two '?' parameters declared for each CommonTokenValue.
+        Optional<Integer> maxQuery = translator.maximumQueryParameters();
+        int maxSub;
+        if (maxQuery.isPresent() && (max / 2) > maxQuery.get()) {
```
punktilious, line 770: shouldn't that be max * 2, because max is the number of CommonTokenValue elements in the list, and each requires two bind variables.

prb112, line 770:
```suggestion
        if (maxQuery.isPresent() && (max * 2) > maxQuery.get()) {
```

prb112: Yes, you are correct

punktilious (approved):
LGTM

Issue #3107, opened 2021-12-10
Misleading completed bundle request message


prb112:
**Describe the bug**

The message shows the running time across the bundle, not the Bundle.Entry

```
[INFO    ] Completed bundle request took:[2.415 secs]: entryIndex:[39] correlationId:[5ad0421b-f0bd-49f9-9e23-ded3f62f997f] method:[PUT] uri:[ValueSet/2.16.840.1.113883.3.117.1.7.1.219] status:[201]
[INFO    ] Completed bundle request took:[2.435 secs]: entryIndex:[40] correlationId:[5ad0421b-f0bd-49f9-9e23-ded3f62f997f] method:[PUT] uri:[ValueSet/2.16.840.1.113883.3.117.1.7.1.222] status:[201]
[INFO    ] Completed bundle request took:[2.454 secs]: entryIndex:[41] correlationId:[5ad0421b-f0bd-49f9-9e23-ded3f62f997f] method:[PUT] uri:[ValueSet/2.16.840.1.113883.3.117.1.7.1.223] status:[201]
[INFO    ] Completed bundle request took:[2.475 secs]: entryIndex:[42] correlationId:[5ad0421b-f0bd-49f9-9e23-ded3f62f997f] method:[PUT] uri:[ValueSet/2.16.840.1.113883.3.117.1.7.1.230] status:[201]
[INFO    ] Completed bundle request took:[2.497 secs]: entryIndex:[43] correlationId:[5ad0421b-f0bd-49f9-9e23-ded3f62f997f] method:[PUT] uri:[ValueSet/2.16.840.1.113883.3.117.1.7.1.232] status:[201]
[INFO    ] Completed bundle request took:[2.53 secs]: entryIndex:[44] correlationId:[5ad0421b-f0bd-49f9-9e23-ded3f62f997f] method:[PUT] uri:[ValueSet/2.16.840.1.113883.3.117.1.7.1.247] status:[201]
[INFO    ] Completed bundle request took:[2.867 secs]: entryIndex:[45] correlationId:[5ad0421b-f0bd-49f9-9e23-ded3f62f997f] method:[PUT] uri:[ValueSet/2.16.840.1.113883.3.117.1.7.1.255] status:[201]
[...]

PR #3037, opened 2021-11-24
fhir-bucket local dir support and db bootstrapping

punktilious:
#1640 don't require a schema just to run reindex
#2997 add bootstrap option for schema
#2935 load from local directory without need for database

PR #3109, opened 2021-12-10
Misleading completed bundle request message #3107 

prb112:
Update Logging to be as accurate as possible in Bundle.Entry logging,… some cases not covered are Bundle Validation related as it would mutate the signature

Instead of only reporting the linearly increasing time, it reports the time spent in most bundle.entry processing.


punktilious (changes requested):
Blocking this for now because this has already been addressed in #3037

prb112:
> Blocking this for now because this has already been addressed in #3037

How are you accounting for the elapsedTime for the entry? accumulatedTime is fine, however it seems a bit more is required.

prb112:
Fixed in separate PR [INFO    ] Completed request[0.041924812 secs]: tenantId:[default] dsId:[default] user:[fhiruser] method:[GET] uri:[https://localhost:9443/fhir-server/api/v4/Patient?family=e0ffca67-9c80-4d57-bce8-72076a7bca84] status:[200]

Issue #3111, opened 2021-12-10
Can't start the docker container when Docker Resource/Memory is less than 3.25 GB


bruneves:
I am not exactly sure what the minimum memory requirement is for running the FHIR Server container image (tag 4.10.2), however by testing with diff configurations I was able to start the container as soon as the Docker Resource/Memory is set to 3.25 GB. Just bringing this to your attention in the case the server was supposed to run with less than that.

## Error Message

```
JVMJ9GC020E -Xms too large for heap
JVMJ9VM015W Initialization error for library j9gc29(2): Failed to initialize
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

Reference https://www.eclipse.org/openj9/docs/xxusecontainersupport/

prb112:
@lmsurpre suggestion is to update docs or add a jvm.options file for minimal environments.

prb112:
Raised a proposal for development specific examples/docs

lmsurpre:
Thanks Paul.  Personally, I'd still like to use a smaller fixed initial heap (or remove it and go with the jvm-default if we can prove that resizing the heap has minimal impact).  However, I'm OK with the documentation-focused PR and have added my approval.

lmsurpre:
ended up with a doc-only change...no further QA needed

PR #3117, opened 2021-12-15
Can't start the docker container when Docker Resource/Memory is less than 3.25 GB #3111

prb112:
- Updated the fhir-install/Dockerfile to point to a Docker container specific jvm.options file
- Update the jvm.otpions in disabled to use XX MaxMemory Resources
- Tested with 2G max assigned to container (we could document this as tested, I only assigned 2G max to Docker / Racher Desktop)

### DIFF
--- a/fhir-install/Dockerfile
+++ b/fhir-install/Dockerfile
@@ -46,6 +46,7 @@ ENV FHIR_CONFIG_HOME=/opt/ol/wlp/usr/servers/defaultServer \
 COPY target/LICENSE /licenses/
 
 COPY --chown=1001:0 --from=base /opt/ol/wlp/usr/servers/defaultServer/server.xml /opt/ol/wlp/usr/servers/defaultServer/
+COPY --chown=1001:0 --from=base /opt/ol/wlp/usr/servers/defaultServer/configDropins/disabled/jvm.options /opt/ol/wlp/usr/servers/defaultServer/jvm.options
 COPY --chown=1001:0 --from=base /opt/ol/wlp/usr/servers/defaultServer/configDropins /opt/ol/wlp/usr/servers/defaultServer/configDropins
 
 RUN features.sh
--- a/fhir-install/Dockerfile
+++ b/fhir-install/Dockerfile
@@ -47,7 +47,9 @@ COPY target/LICENSE /licenses/
 
 COPY --chown=1001:0 --from=base /opt/ol/wlp/usr/servers/defaultServer/server.xml /opt/ol/wlp/usr/servers/defaultServer/
 COPY --chown=1001:0 --from=base /opt/ol/wlp/usr/servers/defaultServer/configDropins /opt/ol/wlp/usr/servers/defaultServer/configDropins
-COPY --chown=1001:0 --from=base /opt/ol/wlp/usr/servers/defaultServer/configDropins/disabled/jvm.options /opt/ol/wlp/usr/servers/defaultServer/jvm.options
+
+# Movies the Container Specific Settings inplace
+COPY --chown=1001:0 --from=base /opt/ol/wlp/usr/servers/defaultServer/configDropins/disabled/jvm.options /opt/ol/wlp/usr/servers/defaultServer/configDropins/defaults/jvm.options
 
 RUN features.sh
 
--- a/fhir-install/Dockerfile
+++ b/fhir-install/Dockerfile
@@ -48,7 +48,7 @@ COPY target/LICENSE /licenses/
 COPY --chown=1001:0 --from=base /opt/ol/wlp/usr/servers/defaultServer/server.xml /opt/ol/wlp/usr/servers/defaultServer/
 COPY --chown=1001:0 --from=base /opt/ol/wlp/usr/servers/defaultServer/configDropins /opt/ol/wlp/usr/servers/defaultServer/configDropins
 
-# Movies the Container Specific Settings inplace
+# Moves the Container Specific Settings in place
 COPY --chown=1001:0 --from=base /opt/ol/wlp/usr/servers/defaultServer/configDropins/disabled/jvm.options /opt/ol/wlp/usr/servers/defaultServer/configDropins/defaults/jvm.options
 
 RUN features.sh
--- a/fhir-install/Dockerfile
+++ b/fhir-install/Dockerfile
@@ -46,8 +46,8 @@ ENV FHIR_CONFIG_HOME=/opt/ol/wlp/usr/servers/defaultServer \
 COPY target/LICENSE /licenses/
 
 COPY --chown=1001:0 --from=base /opt/ol/wlp/usr/servers/defaultServer/server.xml /opt/ol/wlp/usr/servers/defaultServer/
-COPY --chown=1001:0 --from=base /opt/ol/wlp/usr/servers/defaultServer/configDropins/disabled/jvm.options /opt/ol/wlp/usr/servers/defaultServer/jvm.options
 COPY --chown=1001:0 --from=base /opt/ol/wlp/usr/servers/defaultServer/configDropins /opt/ol/wlp/usr/servers/defaultServer/configDropins
+COPY --chown=1001:0 --from=base /opt/ol/wlp/usr/servers/defaultServer/configDropins/disabled/jvm.options /opt/ol/wlp/usr/servers/defaultServer/jvm.options
 
 RUN features.sh
 
--- a/fhir-server-webapp/src/main/liberty/config/configDropins/disabled/jvm.options
+++ b/fhir-server-webapp/src/main/liberty/config/configDropins/disabled/jvm.options
@@ -1,3 +1,21 @@
+# ----------------------------------------------------------------------------
+# (C) Copyright IBM Corp. 2016, 2021
+# 
+# SPDX-License-Identifier: Apache-2.0
+# ----------------------------------------------------------------------------
+
+# This file contains options that are passed directly to the JVM on startup
+# See https://openliberty.io/docs/ref/config/serverConfiguration.html for more info
+
+# Use only TLSv1.2
+-Dhttps.protocols=TLSv1.2
+-Djdk.tls.client.protocols=TLSv1.2
+-Dcom.ibm.jsse2.overrideDefaultProtocol=TLSv1.2
+-Dcom.ibm.jsse2.renegotiate=DISABLED
+
+# Prevents Apache Xerces from overactively using the service loader to find a class
+-Dorg.apache.xml.dtm.DTMManager=org.apache.xml.dtm.ref.DTMManagerDefault
+
 # Container Configuration - https://www.eclipse.org/openj9/docs/xxinitialrampercentage/
 # We need to leave room to operate on the VM - curl.
 -XX:InitialRAMPercentage=0.25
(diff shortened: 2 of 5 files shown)


fhir-server-webapp/src/main/liberty/config/configDropins/disabled/jvm.options
```diff
@@ -1,5 +1,7 @@
-# Set a fixed max heap size
--Xmx3072M
+# Container Configuration - https://www.eclipse.org/openj9/docs/xxinitialrampercentage/
+# We need to leave room to operate on the VM - curl.
+-XX:InitialRAMPercentage=0.25
+-XX:MaxRAMPercentage=0.90
```
prb112, line 4: leaves room to work

fhir-server-webapp/src/main/liberty/config/jvm.options
```diff
@@ -1,12 +1,14 @@
...
-# (C) Copyright IBM Corp. 2016, 2020
+# (C) Copyright IBM Corp. 2016, 2021
 # 
 # SPDX-License-Identifier: Apache-2.0
 # ----------------------------------------------------------------------------
 
 # This file contains options that are passed directly to the JVM on startup
 # See https://openliberty.io/docs/ref/config/serverConfiguration.html for more info
 
+# Be sure to update the jvm.options in the disabled folder.
```
prb112, line 10:
```suggestion
```

fhir-server-webapp/src/main/liberty/config/jvm.options
```diff
@@ -1,12 +1,14 @@
 # ----------------------------------------------------------------------------
-# (C) Copyright IBM Corp. 2016, 2020
+# (C) Copyright IBM Corp. 2016, 2021
```
prb112, line 2:
```suggestion
# (C) Copyright IBM Corp. 2016, 2020
```

fhir-server-webapp/src/main/liberty/config/jvm.options
```diff
@@ -7,6 +7,7 @@
 # This file contains options that are passed directly to the JVM on startup
 # See https://openliberty.io/docs/ref/config/serverConfiguration.html for more info
 
+
```
prb112, line 10:
```suggestion

```

fhir-server-webapp/src/main/liberty/config/jvm.options
```diff
@@ -6,7 +6,6 @@
 
 # This file contains options that are passed directly to the JVM on startup
 # See https://openliberty.io/docs/ref/config/serverConfiguration.html for more info
-
 # Use only TLSv1.2
```
prb112, line 9:
```suggestion

# Use only TLSv1.2
```

fhir-server-webapp/src/main/liberty/config/configDropins/disabled/windows-jvm.options
```diff
@@ -0,0 +1,27 @@
...
+# See https://openliberty.io/docs/ref/config/serverConfiguration.html for more info
+
+# Use only TLSv1.2
+-Dhttps.protocols=TLSv1.2
+-Djdk.tls.client.protocols=TLSv1.2
+-Dcom.ibm.jsse2.overrideDefaultProtocol=TLSv1.2
+-Dcom.ibm.jsse2.renegotiate=DISABLED
+
+# Prevents Apache Xerces from overactively using the service loader to find a class
+-Dorg.apache.xml.dtm.DTMManager=org.apache.xml.dtm.ref.DTMManagerDefault
```
prb112, line 18:
```suggestion
```

fhir-server-webapp/src/main/liberty/config/configDropins/disabled/windows-jvm.options
```diff
@@ -0,0 +1,27 @@
...
+
+# Set a fixed max heap size
+-Xmx3072M
+
+# Turn on verbose classloading
+#-verbose:class
+
+# Turn on verbose garbage collection
+#-verbose:gc
```
prb112, line 27:
```suggestion
```

fhir-server-webapp/src/main/liberty/config/configDropins/disabled/jvm.options
```diff
@@ -1,8 +1,9 @@
...
+# ----------------------------------------------------------------------------
 
-# Turn on verbose classloading
-#-verbose:class
-
-# Turn on verbose garbage collection
-#-verbose:gc
+# https://www.eclipse.org/openj9/docs/xxinitialrampercentage/
+-XX:InitialRAMPercentage=50.00
+-XX:MaxRAMPercentage=90.00
```
prb112, line 9:
```
<gc-end id="443" type="scavenge" contextid="439" durationms="57.408" usertimems="223.188" systemtimems="0.172" stalltimems="4.439" timestamp="2021-12-20T18:00:43.360" activeThreads="4">

  <mem-info id="444" free="469525144" total="858914816" percent="54">

    <mem type="nursery" free="269655128" total="468451328" percent="57">

      <mem type="allocate" free="269655128" total="349700096" percent="77" />

      <mem type="survivor" free="0" total="118751232" percent="0" />

    </mem>

    <mem type="tenure" free="199870016" total="390463488" percent="51" macro-fragmented="99381968">

      <mem type="soa" free="180346432" total="370939904" percent="48" />

      <mem type="loa" free="19523584" total="19523584" percent="100" />

    </mem>

    <pending-finalizers system="0" default="1887" reference="1537" classloader="0" />

    <remembered-set count="18969" />

  </mem-info>

</gc-end>

<cycle-end id="445" type="scavenge" contextid="439" timestamp="2021-12-20T18:00:43.361" />

<allocation-satisfied id="446" threadId="000000000070AB00" bytesRequested="48" />

<af-end id="447" timestamp="2021-12-20T18:00:43.361" threadId="000000000070B508" success="true" from="nursery"/>

<exclusive-end id="448" timestamp="2021-12-20T18:00:43.361" durationms="58.933" />
```
I analyzed the settings, and it is able to grow from the default 469M to the maximum supported in a 2GB container.

fhir-install/Dockerfile
```diff
@@ -48,6 +48,9 @@ COPY target/LICENSE /licenses/
 COPY --chown=1001:0 --from=base /opt/ol/wlp/usr/servers/defaultServer/server.xml /opt/ol/wlp/usr/servers/defaultServer/
 COPY --chown=1001:0 --from=base /opt/ol/wlp/usr/servers/defaultServer/configDropins /opt/ol/wlp/usr/servers/defaultServer/configDropins
 
+# Movies the Container Specific Settings inplace
```
lmsurpre, line 51:
```suggestion
# Moves the Container Specific Settings in place
```

fhir-server-webapp/src/main/liberty/config/configDropins/disabled/jvm.options
lmsurpre: given that the openj9 defaults for max heap are never more than 75%, I'm a tad nervous that 90% is too aggressive

prb112: What else is running in the container? I can only think of curl, curl is really designed to be memory efficient. For 2G allocation, that's 200M. 4G is 400M.

prb112: I didn't account for the tty, so the shell is one thing additional.

prb112, line 9:
```suggestion
-XX:MaxRAMPercentage=80.00
```
I suggest higher than the default. We're not running parallel processes, even the heapdump is lightweight.

fhir-server-webapp/src/main/liberty/config/configDropins/disabled/jvm.options
```diff
@@ -1,8 +1,9 @@
...
+# SPDX-License-Identifier: Apache-2.0
+# ----------------------------------------------------------------------------
 
-# Turn on verbose classloading
-#-verbose:class
-
-# Turn on verbose garbage collection
-#-verbose:gc
+# https://www.eclipse.org/openj9/docs/xxinitialrampercentage/
+-XX:InitialRAMPercentage=50.00
```
lmsurpre, line 8: I think this could lead to a larger-than-needed initial heap size in cases where the memory limit is set fairly high.
For example, currently the alvearie/ibm-fhir-server helm chart defaults to a request of 1 Gi and a limit of 4Gi.
With this setting, I think that request will be way off which could lead to trouble.

I think I slightly prefer a fixed min heap size since we've observed idle memory usage in the 512 - 768 Mi range.

If we do go with the initialRAMPercent, I think we need to:
* document it at https://hub.docker.com/r/ibmcom/ibm-fhir-server
* change the helm chart so that the memory limit is roughly twice the memory request (probably dropping our limit from 4Gi to 2 Gi)

lmsurpre: One nice thing about using RAMPercentage in our defaults is that the `-Xms` and `-Xmx` options naturally override those.  So we certainly could have a default initialRAMPercent and then change the default to a fixed min heap size in the helm chart (if we wanted).

prb112: Yes, that is possible.

fhir-server-webapp/src/main/liberty/config/configDropins/disabled/jvm.options
lmsurpre: I think its more the 'off-heap' storage that i'd be worried about.  I'm not sure whether the zip stream stuff uses much memory or not.  also not sure how big the stacks can be.  90% may be totally fine...was just communicating a gut feel

lmsurpre: totally good with higher-than-default

lmsurpre: maybe even 85% ?

prb112 (commented):

fhir-server-webapp/src/main/liberty/config/configDropins/disabled/jvm.options
punktilious: What about all the class/JIT and stack space? 80% memory utilization is a common threshold for alerting, so I agree that 90% is probably too aggressive.

prb112, line 9:
```suggestion
```

fhir-server-webapp/src/main/liberty/config/configDropins/disabled/jvm.options
prb112: opted for the default.

lmsurpre: > opted for the default

I think that comment was intended for the max heap, right?  50% is definitely not the default on initial heap.
Personally, I think I'd still prefer a fixed initial heap size.  The default is 8 Mi.  Previously we went with 768 Mi.  I think it makes sense to go somewhere between there so that we can:
1. start in more constrained environments 
2. start in environments with lots of RAM available without oversizing the heap
3. avoid unnecessary resizing (although its still not clear to me how costly that actually is)

prb112: I propose an alterative. Let's close the current PR/Issue as won't fix. and update the documentation.

PR #3121, opened 2021-12-16
Update per vscode issue

prb112:


lmsurpre:
Thanks for jumping on this one Paul.  I still can't quite explain what I was seeing in vscode, but I'm now able to see the request and respond bodies from the vscode "DEBUG CONSOLE" with either System.err or System.out output.  Lets leave it as System.err for now (unless there is another reason to change it).

Issue #2972, opened 2021-11-11
Add an option to restrict Whole System History api to return only certain resource types


mpayne2:
**Is your feature request related to a problem? Please describe.**
When I'm only interested in the history of one resource type, the Whole System History api returns all resource types and requires me to page through all the resources to find them.  This would cause a lot of paging if the FHIR server contains millions of records.

**Describe the solution you'd like**
Allow an option on the Whole System History api to pass in the resourceType and only return entries of that resource type.

Preferably just support the type-level history interaction defined at https://www.hl7.org/fhir/http.html#history

**Describe alternatives you've considered**
Page through all the resources on the client and filter out everything but that resourceType.

**Acceptance Criteria**

1.  
   GIVEN a Whole System History api call
   WHEN the api has a resourceType option specified
   THEN only entries with that resourceType are returned in the bundle

**Additional context**
This is for a service looking for Consents being stored in the FHIR server and need to be registered in Consent Manager.  We are looking at using the Whole System History api to detect create, updates, and deletes of consents in order to get the consents in Consent Manager in sync with the FHIR server

d0roppe:
Verified that it only includes the resource type that you restrict with the _type= parameter. Closing issue

PR #3128, opened 2021-12-17
Issue #2972 type filter on system history plus new resource history interaction

punktilious:
Breaking change. The whole-system-history search endpoints have been changed to address spec-compliance. Pagination has changed. See Conformance.md for a detailed description.

### DIFF
--- a/docs/src/pages/Conformance.md
+++ b/docs/src/pages/Conformance.md
@@ -16,7 +16,6 @@ The IBM FHIR Server supports only version 4.0.1 of the specification.
 
 ## FHIR HTTP API
 The HL7 FHIR specification is more than just a data format. It defines an [HTTP API](https://www.hl7.org/fhir/R4/http.html) for creating, reading, updating, deleting, and searching over FHIR resources. The IBM FHIR Server implements the full API for every resource defined in the specification, with the following exceptions:
-* history is not supported at the resource-type level (only the resource instance level and the whole-system level)
 * whole-system history is not conformant to the HL7 FHIR specification; we chose not to include the resource in the history response bundle
 * there are parts of the FHIR search specification which are not fully implemented as documented in the following section
 
@@ -124,26 +123,52 @@ The whole system history interaction can be used to obtain a list of changes (cr
     curl -k -u '<username>:<password>' 'https://<host>:<port>/fhir-server/api/v4/_history'
 ```
 
-The response is a history bundle as described by the [FHIR specification](https://www.hl7.org/fhir/http.html#history) with one exception. Whereas the specification requires each entry in the history bundle to contain the full contents of the resource (at least for entries with a `entry.request.method` of PUT or POST), the IBM FHIR Server whole-system history response bundle contains only references to the resources. Clients may choose which resources they fetch. They may also fetch the resources in parallel, which may greatly improve throughput.
+The response is a history bundle as described by the [FHIR specification](https://www.hl7.org/fhir/http.html#history) with one exception. Whereas the specification requires each entry in the history bundle to contain the full contents of the resource (at least for entries with a `entry.request.method` of PUT or POST), clients can specify the HTTP header value `Prefer: return=none` in which case the whole-system history response bundle contains only references to the resources. This allows clients to decide which resources they actually need to read.
 
-To return all changes that have occurred since a known point in time, use the `_since` query parameter:
+There are three traversal paths through the data which are requested using the `_sort` request parameter:
+
+1. `_sort=-_lastUpdated` [Default] Decreasing time - most recent changes first
+2. `_sort=_lastUpdated` Increasing time - oldest changes first
+3. `_sort=none` Increasing id order - oldest changes first
+
+Option 1 is useful for users interested in finding the most recent changes that have been applied to the server. Option 2 is useful for system-to-system synchronization because it represents the change history of the server. Option 3 is also useful for system-to-system synchronization cases, and benefits guaranteeing uniqueness in the resources being returned when following the `next` links.
+
+To return all changes that have occurred since a known point in time, use both `_sort=_lastUpdated` and `_since` query parameters:
 
 ```
-    curl -k -u '<username>:<password>' 'https://<host>:<port>/fhir-server/api/v4/_history?_since=2021-02-21T00:00:00Z'
+    curl -k -u '<username>:<password>' 'https://<host>:<port>/fhir-server/api/v4/_history?_since=2021-02-21T00:00:00Z&_sort=_lastUpdated'
 ```
 
-By default, the returned bundle will contain up to 100 resource references. The number of resources can be increased (up to 1000) using the `_count` parameter. Specifying `_count` values larger than 1000 will return no more than 1000 resource references:
+By default, the returned bundle will contain up to 100 resources. The number of resources can be increased (up to 1000) using the `_count` parameter. Specifying `_count` values larger than 1000 will return no more than 1000 resources:
 ```
     curl -k -u '<username>:<password>' 'https://<host>:<port>/fhir-server/api/v4/_history?_count=1000'
 ```
 
-Client applications can use the `_since` parameter to scan sequentially through all the changes recorded by the IBM FHIR Server. Each result bundle contains a `next` link which can be used to fetch the next set of data. To simplify client implementations, the IBM FHIR Server also includes a custom attribute `_afterHistoryId`. This value can be used by a client to checkpoint where they are in the list of changes, and ask for only changes that come after the given id. This is convenient, because the `_afterHistoryId` is based on a unique identifier. The client does not need to worry about duplicates which otherwise occur when using `_since` to sequentially fetch history:
+Client applications can use the `_since` parameter to scan sequentially through all the changes recorded by the IBM FHIR Server. Each result bundle contains a `next` link which can be used to fetch the next set of data. 
+
+As mentioned above, to simplify client implementations in system-to-system synchronization scenarios, the client may specify `_sort=none` as follows:
 
 ```
-    curl -k -u '<username>:<password>' 'https://<host>:<port>/fhir-server/api/v4/_history?_count=1000&_afterHistoryId=3004'
+    curl -k -u '<username>:<password>' 'https://<host>:<port>/fhir-server/api/v4/_history?_count=100&_sort=none'
+
+
+In this case, the IBM FHIR Server uses a custom paging attribute `_afterHistoryId`. This value can be used by a client to checkpoint where they are in the list of changes, and ask for only changes that come after the given id. The simplest way to do this is to follow the `next` link returned in the response Bundle. If the next link is not present, the end has been reached.
+
+Clients cannot use both `_afterHistoryId` and `_since` together nor should they switch to using the other without understanding how concurrency, transaction boundaries and cluster clock synchronization issues can affect how ids and timestamps are assigned during ingestion.
+
+The query parameter `_type` can be used to limit which resource types are returned from a `_history` request. For example, the following request will return only Patient and Observation resources:
+
+```
+    curl -k -u '<username>:<password>' 'https://<host>:<port>/fhir-server/api/v4/_history?_count=1000&_afterHistoryId=3004&_type=Patient,Observation'
 ```
 
-Clients cannot use both `_afterHistoryId` and `_since` together nor should they switch to using the other without understanding how concurrency, transaction boundaries and cluster clock synchronization issues can affect how ids and timestamps are assigned during ingestion. If `_since` is specified, the `next` link in the returned bundle will include the `_since` parameter using the greatest `lastModified` from the returned resources as its value. If `_afterHistoryId` is specified, the `next` link in the returned bundle will include the `_afterHistoryId` parameter using the greatest internal id from the returned resources as its value. If you do not specify any filter parameter, the IBM FHIR Server uses `_afterHistoryId=0` as the initial start point and so will use `_afterHistoryId` for the subsequent `next` links.
+If only a single resource type is required, the `[type]/_history` endpoint can be used:
+
+```
+    curl -k -u '<username>:<password>' 'https://<host>:<port>/fhir-server/api/v4/Patient/_history?_count=1000&_afterHistoryId=3004'
+```
+
+This form has identical behavior to the whole-system-history endpoint with a single `_type=[type]` value.
 
 In a highly concurrent system, several resources could share the same timestamp. Also, the internal id used to identify individual resource changes may not correlate perfectly with the `lastModified` time. Take the following, for example:
 
@@ -160,7 +185,9 @@ Note how the change time for `patient-3` and `patient-4` is the same, although t
 
 Clients must also exercise caution when reading recently ingested resources. When processing large bundles in parallel, an id may be assigned by the database but ACID isolation means that the record will not be visible to a reader until the transaction is committed. This could be up to 120s or longer if a larger transaction-timeout property has been defined. If a smaller bundle starts after the larger bundle and its transaction is committed first, its change ids and timestamps will be visible to readers before the resources from the other bundle, which will have earlier change ids and timestamps. If clients do not take this into account, they may miss some resources. This behavior is a common concern in databases and not specific to the IBM FHIR Server.
 
-To guarantee no data is skipped, clients should not process resources with a `lastModified` timestamp which is after `{current_time} - {transaction_timeout} - {max_cluster_clock_drift}`. By waiting for this time window to close, the client can be sure the data being returned is complete and in order, and new data will not appear. The default value for transaction timeout is 120s but this is configurable. A value of 2 seconds is a reasonable default to consider for `max_cluster_clock_drift` in lieu of specific information. Implementers should check with server administrators on the appropriate values to use.
+To guarantee no data is skipped, clients should not process resources with a `lastModified` timestamp which is after `{current_time} - {transaction_timeout} - {max_cluster_clock_drift}`. By waiting for this time window to close, the client can be sure the data being returned is complete and in order, and new data will not appear. The default value for transaction timeout is 120s but this is configurable. A value of 2 seconds is a reasonable default to consider for `max_cluster_clock_drift` in lieu of specific information. Implementers should check with server administrators on the appropriate values to use. If the server has been configured with the environment variable `FHIR_TRANSACTION_MANAGER_TIMEOUT` to control the transaction timeout, clients may specify the optional query parameter `_excludeTransactionTimeoutWindow=true` to perform this filtering within the server. If the environment variable `FHIR_TRANSACTION_MANAGER_TIMEOUT` is not configured, a default transaction timeout of 120s is used, although this may not be the actual timeout if the server is configured in a non-standard way.
+
+Note that using `_excludeTransactionTimeoutWindow=true` intentionally introduces a delay into when freshly ingested resources become visible from the `_history` endpoints. This is done to ensure that clients can safely iterate along the resource history timeline using the `next` links in the Bundle response without the risk of missing data.
 
 ## Search
 The IBM FHIR Server supports all search parameter types defined in the specification:
--- a/docs/src/pages/Conformance.md
+++ b/docs/src/pages/Conformance.md
@@ -123,56 +123,67 @@ The whole system history interaction can be used to obtain a list of changes (cr
     curl -k -u '<username>:<password>' 'https://<host>:<port>/fhir-server/api/v4/_history'
 ```
 
-The response is a history bundle as described by the [FHIR specification](https://www.hl7.org/fhir/http.html#history) with one exception. Whereas the specification requires each entry in the history bundle to contain the full contents of the resource (at least for entries with a `entry.request.method` of PUT or POST), clients can specify the HTTP header value `Prefer: return=none` in which case the whole-system history response bundle contains only references to the resources. This allows clients to decide which resources they actually need to read.
+The response is a history bundle as described by the [FHIR specification](https://www.hl7.org/fhir/http.html#history) with one exception. Whereas the specification requires each entry in the history bundle to contain the full contents of the resource (at least for entries with a `entry.request.method` of PUT or POST), clients can specify the HTTP header value `Prefer: return=minimal` in which case the whole-system history response bundle contains only references to the resources. This allows clients to decide which resources they actually need to read (and can do so in parallel to further improve throughput).
 
-There are three traversal paths through the data which are requested using the `_sort` request parameter:
+There are three possible traversal paths through the data which are requested using the `_sort` request parameter:
 
 1. `_sort=-_lastUpdated` [Default] Decreasing time - most recent changes first
 2. `_sort=_lastUpdated` Increasing time - oldest changes first
 3. `_sort=none` Increasing id order - oldest changes first
 
-Option 1 is useful for users interested in finding the most recent changes that have been applied to the server. Option 2 is useful for system-to-system synchronization because it represents the change history of the server. Option 3 is also useful for system-to-system synchronization cases, and benefits guaranteeing uniqueness in the resources being returned when following the `next` links.
+Option 1 is useful for users interested in finding the most recent changes that have been applied to the server. Option 2 is useful for system-to-system synchronization because it represents the change history of the server. Option 3 is also useful for system-to-system synchronization cases, but guarantees uniqueness across pages when following the `next` links (versus option 2 which may repeat entries at the time window boundary when multiple resources share the same modification timestamp).
 
-To return all changes that have occurred since a known point in time, use both `_sort=_lastUpdated` and `_since` query parameters:
+To return all changes that have occurred since a known point in time, use both `_since` and `_sort=_lastUpdated` query parameters:
 
 ```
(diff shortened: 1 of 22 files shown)


docs/src/pages/Conformance.md
```diff
@@ -16,7 +16,6 @@ The IBM FHIR Server supports only version 4.0.1 of the specification.
 
 ## FHIR HTTP API
 The HL7 FHIR specification is more than just a data format. It defines an [HTTP API](https://www.hl7.org/fhir/R4/http.html) for creating, reading, updating, deleting, and searching over FHIR resources. The IBM FHIR Server implements the full API for every resource defined in the specification, with the following exceptions:
-* history is not supported at the resource-type level (only the resource instance level and the whole-system level)
```
lmsurpre, line 19: 🙌 yay!

docs/src/pages/Conformance.md
```diff
@@ -124,26 +123,52 @@ The whole system history interaction can be used to obtain a list of changes (cr
...
 
-To return all changes that have occurred since a known point in time, use the `_since` query parameter:
+There are three traversal paths through the data which are requested using the `_sort` request parameter:
+
+1. `_sort=-_lastUpdated` [Default] Decreasing time - most recent changes first
+2. `_sort=_lastUpdated` Increasing time - oldest changes first
+3. `_sort=none` Increasing id order - oldest changes first
+
+Option 1 is useful for users interested in finding the most recent changes that have been applied to the server. Option 2 is useful for system-to-system synchronization because it represents the change history of the server. Option 3 is also useful for system-to-system synchronization cases, and benefits guaranteeing uniqueness in the resources being returned when following the `next` links.
```
lmsurpre, line 134: minor wording suggestion 
```suggestion
Option 1 is useful for users interested in finding the most recent changes that have been applied to the server. Option 2 is useful for system-to-system synchronization because it represents the change history of the server. Option 3 is also useful for system-to-system synchronization cases guarantees uniqueness across pages when following the `next` links (as opposed to option 2 which may repeat entries at the time window boundary).
```

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/FetchResourceChangesDAO.java
```diff
@@ -33,27 +35,40 @@
...
-     * @param resourceTypeName
      * @param fromTstamp
      * @param afterResourceId
+     * @param resourceTypeIds
+     * @param excludeTransactionTimeoutWindow
+     * @param historySortOrder
      */
-    public FetchResourceChangesDAO(IDatabaseTranslator tx, String schemaName, int resourceCount, String resourceTypeName, Instant fromTstamp, Long afterResourceId) {
+    public FetchResourceChangesDAO(IDatabaseTranslator tx, String schemaName, int resourceCount, Instant fromTstamp, Long afterResourceId, List<Integer> resourceTypeIds,
+        boolean excludeTransactionTimeoutWindow, HistorySortOrder historySortOrder) {
```
lmsurpre, line 63:
```suggestion
            boolean excludeTransactionTimeoutWindow, HistorySortOrder historySortOrder) {
```

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/FetchResourceChangesDAO.java
```diff
@@ -64,6 +79,21 @@ public FetchResourceChangesDAO(IDatabaseTranslator tx, String schemaName, int re
      */
     public List<ResourceChangeLogRecord> run(Connection c) throws FHIRPersistenceException {
         List<ResourceChangeLogRecord> result = new ArrayList<>();
+        
+        Instant before = null;
```
lmsurpre, line 83: I feel like we also talked about possibly taking `_before` as query parameter (similar to `_since` but for paging through results in the opposite direction).  I don't know if its needed or not, but wanted to see if you'd thought about that one.

fhir-persistence/src/main/java/com/ibm/fhir/persistence/HistorySortOrder.java
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * The ordering specified in the client request for whole system history
+ * interactions. This is not defined in the R4 specification but has been
+ * proposed for R5. The IBM FHIR Server implements it for R4 because it is 
+ * useful to support system-system synchronization scenarios
+ * @formatter:off
+ * _sort=_lastUpdated            - ORDER BY last_updated ASC
+ * _sort=-_lastUpdated (default) - ORDER BY last_updated DESC
+ * _sort=none                    - ORDER BY resource_id ASC
+ * @formatter:on
```
lmsurpre, line 19: the formatter tags will keep eclipse from squashing these lines, but javadoc / html will still lose the linebreaks unless you do something like this:
```suggestion
 * <pre>
 * _sort=_lastUpdated            - ORDER BY last_updated ASC
 * _sort=-_lastUpdated (default) - ORDER BY last_updated DESC
 * _sort=none                    - ORDER BY resource_id ASC
 * </pre>
```

fhir-persistence/src/main/java/com/ibm/fhir/persistence/context/FHIRSystemHistoryContext.java
```diff
@@ -36,4 +40,34 @@
...
+     */
+    HistorySortOrder getHistorySortOrder();
+    
+    /**
+     * Get the return preference
+     * @formatter:off
+     *   Prefer: return=minimal          response bundle summary without Resources
+     *   Prefer: return=representation   response bundle includes Resources
+     *   Prefer: return=OperationOutcome 400 Bad Request
+     * @formatter:on
```
lmsurpre, line 68:
```suggestion
     * <pre>
     *   Prefer: return=minimal          response bundle summary without Resources
     *   Prefer: return=representation   response bundle includes Resources
     *   Prefer: return=OperationOutcome 400 Bad Request
     * </pre>
```

fhir-persistence/src/main/java/com/ibm/fhir/persistence/context/impl/FHIRSystemHistoryContextImpl.java
```diff
@@ -51,6 +81,14 @@ public void setAfterHistoryId(long id) {
         this.afterHistoryId = id;
     }
 
+    /**
+     * Add the resource type to include in the response
+     * @param resourceType
+     */
+    public void addResourceType(String resourceType) {
+        this.resourceTypes.add(resourceType);
+    }
```
lmsurpre, line 90: does casing matter?

fhir-server-spi/src/main/java/com/ibm/fhir/server/spi/operation/FHIRResourceHelpers.java
```diff
@@ -381,7 +381,25 @@ default Resource doVRead(String type, String id, String versionId) throws Except
...
+     * Implement the system level history operation to obtain a list of changes to resources
+     * with an optional resourceType which supports for example [base]/Patient/_history
+     * requests to return the complete history of changes filtered to a specific resource type
+     *
+     * @param queryParameters
+     *            a Map containing the query parameters from the request URL
+     * @param requestUri
+     *            the request URI
+     * @param resourceType
+     *            optional resourceType to filter the history
```
lmsurpre, line 398: is this a comma-separated list if you're filtering on more than one?  if so, I think we should mention that here

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
```diff
@@ -256,12 +255,14 @@ private HTTPHandlingPreference computeHandlingPref(ServletRequest request) throw
         return handlingPref;
     }
 
-    private HTTPReturnPreference computeReturnPref(ServletRequest request, HTTPHandlingPreference handlingPref) throws FHIRException {
+    private void computeReturnPref(FHIRRequestContext context, ServletRequest request, HTTPHandlingPreference handlingPref) throws FHIRException {
         HTTPReturnPreference returnPref = defaultHttpReturnPref;
```
lmsurpre, line 259: Now that we have the `setReturnPreferenceDefault` flag and different defaults based on the interaction, I wonder if it makes sense to keep this null in cases where its not explicitly passed (forcing everywhere using it to use their own "isDefault"->set returnPref logic)...

lmsurpre: If not, then it might be worth leaving a quick comment here to explain that this `defaultHttpReturnPref` is for the spec-defined uses but that interactions may have their own defaults

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -2949,18 +2951,32 @@ public void validateInteraction(Interaction interaction, String resourceType) th
...
+        
+        if (requestContext.isReturnPreferenceDefault()) {
+            // For _history, override the default preference to be REPRESENTATION so resources
+            // are returned in the response bundle per the R4 spec.
+            requestContext.setReturnPreference(HTTPReturnPreference.REPRESENTATION);
+        }
         FHIRSystemHistoryContext historyContext =
                 FHIRPersistenceUtil.parseSystemHistoryParameters(queryParameters, HTTPHandlingPreference.LENIENT.equals(requestContext.getHandlingPreference()));
 
+        // If HTTPReturnRepresentation is REPRESENTATION, we fetch the resources and include them
```
lmsurpre, line 2975:
```suggestion
        // If HTTPReturnPreference is REPRESENTATION, we fetch the resources and include them
```

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -3028,11 +3072,15 @@ public Bundle doHistory(MultivaluedMap<String, String> queryParameters, String r
...
+            responseBuilder.location(Url.of(changeRecord.getResourceTypeName() + "/" + changeRecord.getLogicalId() + "/_history/" + changeRecord.getVersionId()));
 
+            // Per the R4 spec, the fullUrl should not contain _history/:vid
             Entry.Builder entryBuilder = Entry.builder();
-            entryBuilder.fullUrl(Url.of(changeRecord.getResourceTypeName() + "/" + changeRecord.getLogicalId() + "/_history/" + changeRecord.getVersionId()));
+            entryBuilder.id(Long.toString(changeRecord.getChangeId()));
+            entryBuilder.fullUrl(Url.of(changeRecord.getResourceTypeName() + "/" + changeRecord.getLogicalId()));
             entryBuilder.request(requestBuilder.build());
             entryBuilder.response(responseBuilder.build());
+            entryBuilder.resource(resource); // remember resource may be null if returnPreference=minimal
```
lmsurpre, line 3083: this says `returnPreference=minimal` which matched my proposal in the issue, but in the docs we say `returnPreference=none`.
based on the logic above the two should work the same (since it only checks `returnPreference=representation`), but I think we should settle on one or the other to avoid any confusion.

docs/src/pages/Conformance.md
```diff
@@ -124,26 +123,52 @@ The whole system history interaction can be used to obtain a list of changes (cr
...
 ```
 
-The response is a history bundle as described by the [FHIR specification](https://www.hl7.org/fhir/http.html#history) with one exception. Whereas the specification requires each entry in the history bundle to contain the full contents of the resource (at least for entries with a `entry.request.method` of PUT or POST), the IBM FHIR Server whole-system history response bundle contains only references to the resources. Clients may choose which resources they fetch. They may also fetch the resources in parallel, which may greatly improve throughput.
+The response is a history bundle as described by the [FHIR specification](https://www.hl7.org/fhir/http.html#history) with one exception. Whereas the specification requires each entry in the history bundle to contain the full contents of the resource (at least for entries with a `entry.request.method` of PUT or POST), clients can specify the HTTP header value `Prefer: return=none` in which case the whole-system history response bundle contains only references to the resources. This allows clients to decide which resources they actually need to read.
```
lmsurpre, line 126: should this say `return=minimal` instead of `return=none` or were you trying to introduce a new value?

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
lmsurpre: ok, i left a suggestion on the doc to say return=minimal ...assuming we make that update then this is good-to-go

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -3051,19 +3099,38 @@ public Bundle doHistory(MultivaluedMap<String, String> queryParameters, String r
...
+                nextRequest.append("&_sort=" + HistorySortOrder.ASC_LAST_UPDATED.toString());
                 nextRequest.append("&_since=").append(lastChangeTime.atZone(UTC).format(DateTime.PARSER_FORMATTER));
-            } else {
-                // in all other cases we fetch ordered by resource_id, and so the client
-                // should get the next page using the _afterHistoryId marker, allowing
-                // easy paging through the stream of changes without duplicates
+                break;
+            case DESC_LAST_UPDATED:
+                nextRequest.append("&_sort=" + HistorySortOrder.DESC_LAST_UPDATED.toString());
+                nextRequest.append("&_since=").append(lastChangeTime.atZone(UTC).format(DateTime.PARSER_FORMATTER));
```
lmsurpre, line 3126: if we introduce a `_before` parameter, we could use that here instead of changing the meaning of `_since`

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/FetchResourceChangesDAO.java
```diff
@@ -84,31 +114,53 @@ public FetchResourceChangesDAO(IDatabaseTranslator tx, String schemaName, int re
         }
 
         if (fromTstamp != null) {
-            query.append(" AND c.change_tstamp >= ? ");
+            if (historySortOrder == HistorySortOrder.ASC_LAST_UPDATED) {
+                query.append(" AND c.change_tstamp >= ? ");
+            } else {
+                query.append(" AND c.change_tstamp <= ? ");
```
lmsurpre, line 120: does this mean we basically flip the meaning of `_since` if the sort order is DESC_LAST_UPDATED?

fhir-server-spi/src/main/java/com/ibm/fhir/server/spi/operation/FHIRResourceHelpers.java
lmsurpre: after reading the impl, it looks like param is only used for the resourceType path parameter and that the optional resourceTypes query parameter is passed separately (in the query params which are then baked into the context via FHIRPersistenceUtil.parseSystemHistoryParameters).

docs/src/pages/Conformance.md
punktilious: good

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/FetchResourceChangesDAO.java
punktilious: Yeah, less necessary now because you can switch the sort order, so _since behaves that way. If we add _before, then we'd need to make the semantics the same for _since and _before regardless of sort order, which may not be a bad thing because I was struggling about using _since for both directions. Gonna chew on this.

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
punktilious: Yeah, I'll add the comment.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
punktilious: I knew I'd make that mistake somewhere :-)

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/FetchResourceChangesDAO.java
punktilious: at the moment, yes.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
punktilious: My struggle exactly (from the earlier conversation). I think this makes sense. It'll need good documentation, especially when interacting with the transaction timeout window.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/FetchResourceChangesDAO.java
punktilious: ok

fhir-persistence/src/main/java/com/ibm/fhir/persistence/context/FHIRSystemHistoryContext.java
punktilious: ok

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
punktilious: fixed

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/FetchResourceChangesDAO.java
```diff
@@ -64,6 +79,21 @@ public FetchResourceChangesDAO(IDatabaseTranslator tx, String schemaName, int re
...
+            // compute the start time of the current transaction timeout window. Note that
+            // if the FHIR_TRANSACTION_MANAGER_TIMEOUT variable is not set, we can't determine
+            // the timeout value, so we can't use the feature
+            long txTimeoutSecs = SystemConfigHelper.getDurationFromEnv("FHIR_TRANSACTION_MANAGER_TIMEOUT", "120s");
+            if (txTimeoutSecs < 0) {
+                logger.warning("Ignoring excludeTransactionTimeoutWindow because FHIR_TRANSACTION_MANAGER_TIMEOUT is invalid");
+            } else {
+                // add a couple of seconds to cover any clock drift in a cluster environment.
+                final int clockDriftBuffer = 2;
+                before = Instant.now().minus(txTimeoutSecs + clockDriftBuffer, ChronoUnit.SECONDS);
```
prb112, line 94: smart bit of craft her

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/FetchResourceChangesDAO.java
punktilious: Per our conversation I've now included support for both `_before` and `_since` - both are supported for the lastUpdated ASC and DESC ordering.

punktilious: Updated the Conformance.md doc accordingly.

fhir-persistence/src/main/java/com/ibm/fhir/persistence/HistorySortOrder.java
punktilious: ok

fhir-persistence/src/main/java/com/ibm/fhir/persistence/context/impl/FHIRSystemHistoryContextImpl.java
punktilious: Resource types are checked via ModelSupport, so yes if I undertood the question correctly

docs/src/pages/Conformance.md
punktilious: fixed

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
punktilious: Changed the semantics so that both _since and _before can be used regardless of sort direction.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/FetchResourceChangesDAO.java
punktilious: changed the semantics. See Conformance.md

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/impl/FetchResourceChangesDAO.java
punktilious: thanks :-)

fhir-persistence/src/main/java/com/ibm/fhir/persistence/context/impl/FHIRSystemHistoryContextImpl.java
lmsurpre: ok, I was debating whether 
A. the javadoc should reflect that; or
B. that should be enforced here;

but I guess we pretty consistently use the TitleCase/CamelCase for resource names throughout the project, so its probably fine

fhir-server/src/test/java/com/ibm/fhir/server/test/MockPersistenceImpl.java
```diff
@@ -180,8 +180,9 @@ public ResourcePayload fetchResourcePayloads(Class<? extends Resource> resourceT
     }
 
     @Override
-    public List<ResourceChangeLogRecord> changes(int resourceCount, java.time.Instant fromLastModified, Long afterResourceId, List<String> resourceTypeNames,
-            boolean excludeTransactionTimeoutWindow, HistorySortOrder historySortOrder) throws FHIRPersistenceException {
+    public List<ResourceChangeLogRecord> changes(int resourceCount, java.time.Instant sinceLastModified, java.time.Instant beforeLastModified,
+        Long afterResourceId, List<String> resourceTypeNames, boolean excludeTransactionTimeoutWindow, HistorySortOrder historySortOrder)
+        throws FHIRPersistenceException {
```
lmsurpre, line 185:
```suggestion
            Long afterResourceId, List<String> resourceTypeNames, boolean excludeTransactionTimeoutWindow, HistorySortOrder historySortOrder)
            throws FHIRPersistenceException {
```

docs/src/pages/Conformance.md
```diff
@@ -124,30 +123,67 @@ The whole system history interaction can be used to obtain a list of changes (cr
...
 ```
 
-By default, the returned bundle will contain up to 100 resource references. The number of resources can be increased (up to 1000) using the `_count` parameter. Specifying `_count` values larger than 1000 will return no more than 1000 resource references:
+To return changes within a known window of time, specify both the `_since` and `_before` query parameters and choose either ascending or descending sort. 
+
+When sorting with ascending time, the `_since` parameter in the `next` link is calculated to fetch the next page of resources and the `_before` parameter (if given) remains constant. When sorting with descending time, the `_before` parameter in the `next` link is calculated to fetch the next page of resources and the `_since` parameter (if given) remains constant.
+
+When sorting using ascending or descending time, the `next` link also includes a value for `_lastChangeId`. This value is used to exclude this resource change record from the next result page. This is required because multiple resources may share the same `_lastUpdated` time, so filters for `_since` and `_before` must always be inclusive which can lead to duplicates across result page boundaries.
```
lmsurpre, line 152: `_lastChangeId` felt a bit ambiguous to me because I felt it could be interpretted as "the last change id that the client would like to receive in their request" whereas it actually is "the last change id the client received in the previous request"

Robin and I brainstormed  alternative names:

- _lastChangeIdReveived
- _pagingRef
- _pagingMarker
- _pageMarker
- _pageMark
- _changeIdMarker
- _changeIdMark
- _idMark
- _mark

We finally settled on `_changeIdMarker`

fhir-server/src/test/java/com/ibm/fhir/server/test/MockPersistenceImpl.java
punktilious: fixed

docs/src/pages/Conformance.md
punktilious: done

lmsurpre (approved):
LGTM

Issue #3135, opened 2021-12-20
fhir schema lease not always given up after completed schema update


lmsurpre:
**Describe the bug**
Seen once during CI (https://github.com/IBM/FHIR/runs/4581232255?check_suite_focus=true):
```
2021-12-20 10:57:33.398 00000001    INFO   com.ibm.fhir.schema.app.Main Refreshing tenant partitions
2021-12-20 10:57:33.427 00000001    INFO m.fhir.schema.app.LeaseManager Canceling lease for schema 'FHIRDATA'
2021-12-20 10:57:33.429 00000001    INFO m.fhir.schema.app.LeaseManager Lease canceled for schema 'FHIRDATA'
2021-12-20 10:57:33.431 00000001    INFO m.fhir.schema.app.LeaseManager Requesting update lease for schema 'FHIR_OAUTH' [attempt 1]
2021-12-20 10:57:33.558 00000001    INFO   com.ibm.fhir.schema.app.Main Collecting model update tasks
2021-12-20 10:57:33.559 00000001    INFO   com.ibm.fhir.schema.app.Main Starting model updates
2021-12-20 10:57:33.636 0000000f    INFO hir.task.core.impl.TaskManager Task complete callback for taskId: TABLE:FHIR_OAUTH.OAUTH20CACHE:1
2021-12-20 10:57:33.685 0000000f    INFO hir.task.core.impl.TaskManager Task complete callback for taskId: TABLE:FHIR_OAUTH.OAUTH20CLIENTCONFIG:1
2021-12-20 10:57:33.718 0000000f    INFO hir.task.core.impl.TaskManager Task complete callback for taskId: TABLE:FHIR_OAUTH.OAUTH20CONSENTCACHE:1
2021-12-20 10:57:33.723 00000001    INFO m.fhir.schema.app.LeaseManager Canceling lease for schema 'FHIR_OAUTH'
2021-12-20 10:57:33.724 00000001    INFO m.fhir.schema.app.LeaseManager Lease canceled for schema 'FHIR_OAUTH'
2021-12-20 10:57:33.725 00000001    INFO m.fhir.schema.app.LeaseManager Requesting update lease for schema 'FHIR_JBATCH' [attempt 1]
2021-12-20 10:57:33.780 00000001    INFO   com.ibm.fhir.schema.app.Main Collecting model update tasks
2021-12-20 10:57:33.780 00000001    INFO   com.ibm.fhir.schema.app.Main Starting model updates
[...]

PR #3140, opened 2021-12-22
issue #3135 stop lease maintenance thread before cancel lease

punktilious:


lmsurpre (approved):
LGTM

PR #3168, opened 2022-01-10
workflow to build and push updated docker images

lmsurpre:
currently set to run once-a-day (same schedule as our liberty base images)... might want to adjust the schedule (maybe weekly?) once we see how its working


.github/workflows/rebuild.yml
```diff
@@ -0,0 +1,49 @@
...
+        OSSRH_USERNAME: ${{ secrets.OSSRH_USERNAME }}
+        OSSRH_PASSWORD: ${{ secrets.OSSRH_PASSWORD }}
+        OLD_JAVA_HOME: /opt/hostedtoolcache/AdoptOpenJDK/1.0.0-ga-11-jdk-hotspot-linux-x64-normal-latest/x64/
+      run: |
+        . build/release/global.sh
+        bash build/release/00_prep.sh
+        bash build/release/10_build.sh
+        bash build/release/20_test.sh
+        bash build/release/30_release.sh
+        bash build/release/40_drop.sh
```
prb112, line 49: you'll need to exclude the MavenCentral stuff.

lmsurpre: it seemed like those were in the later phases (which I removed from this workflow), maybe there's some setup work i missed in one of these...any pointers?

lmsurpre: ah, i see it now, the first step of 30_release.sh:

```
bash ${WORKSPACE}/build/release/bin/30_release/0_sonatype.sh

# This should always be last
bash ${WORKSPACE}/build/release/bin/30_release/1_dockerhub.sh
```

maybe i can just call `30_release/1_dockerhub.sh` directly from the workflow instead of clalling 30_release.sh?

prb112: I think that'd be preferrable.

lmsurpre: done

build/release/bin/20_test/1_code_coverage.sh
```diff
@@ -3,23 +3,39 @@
 set -eu -o pipefail
 
 ###############################################################################
-# (C) Copyright IBM Corp. 2020, 2021
+# (C) Copyright IBM Corp. 2020, 2022
```
prb112, line 6: maybe this one could be skipped too?

prb112 (approved):
LGTM - as a phase1 I think this is good.

tbieste (approved):

Issue #3032, opened 2021-11-22
Warn and skip update-schema when deployed version > new version


lmsurpre:
**Is your feature request related to a problem? Please describe.**
In https://github.com/IBM/FHIR/pull/2880/files#diff-b4577e1635a5c980b9e41f30be0c6abcbf3fe5e6915b118fd319963bdac70a78R440 we added a check so that we skip `--update-schema` when the deployed version is already up-to-date.

However, this got implemented in SchemaVersionsManager as an `==` check:  https://github.com/IBM/FHIR/pull/2880/files#diff-125e38fd75a71288d1716b5c3c920f85cdca913358ce3ec5b236973890ccddfcR101

**Describe the solution you'd like**
If the currently-deployed schema version is greater than the one being "updated to", then log a warning and skip the work.
Most importantly, do NOT update (i.e. downgrade) the stored procs and do NOT update the whole-schema version.

**Describe alternatives you've considered**

**Acceptance Criteria**

1.  
   GIVEN [a precondition]
    AND [another precondition]
   WHEN [test step]
    AND [test step]
   THEN [verification step]
    AND [verification step]

**Additional context**

d0roppe:
With this issue we may also want to address the drop-schema (incorrect version used issue) that an error is issued if you try to drop a older schema with a  newer tool, or vice versa.  Maybe even tell them what version of the tool needs to be used.

lmsurpre:
A related scenario I thought of:
1. User specifies a particular set of resourceTypes to deploy in the schema (this feature still isn't documented but its there... #819)
2. User wants to change the set of resoureTypes in the schema, but at the same schema version

Currently, there is no way to force it to do its thing when the schema version matches the expected schema version. For this use case, and maybe also for the case described in "Additional context" above, I think it would be nice to add a `--force` option which would bypass the version check and attempt to apply the schema update.

punktilious:
> 

For the drop schema, I'm thinking we want to make it dictionary-driven so that it doesn't matter which version you have.

punktilious:
The new `--force` option will only attempt to apply the schema update if the schema version actually matches the database version. This is important so that we don't inadvertently apply the wrong procedures. But this does allow us to bypass the version check when adding more resource types, as requested by @lmsurpre

d0roppe:
Any issues with multi Tenant on different versions?

lmsurpre:
to QA:
1. create a schema at level V26 (e.g. via R4B)
[...]

PR #3200, opened 2022-01-17
issue #3032 only update when current schema is behind code version

punktilious:

Also added `--force` option which will cause the schema upgrade steps to be performed (partially bypassing the whole-schema-version check) but only if the code schema version is equal to the database schema version. Note that if the database schema version is behind the code schema version, you don't need `--force` because the changes will be applied anyway.

### DIFF
--- a/fhir-bucket/src/main/java/com/ibm/fhir/bucket/app/Main.java
+++ b/fhir-bucket/src/main/java/com/ibm/fhir/bucket/app/Main.java
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright IBM Corp. 2020, 2021
+ * (C) Copyright IBM Corp. 2020, 2022
  *
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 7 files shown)


fhir-database-utils/src/main/java/com/ibm/fhir/database/utils/schema/SchemaVersionsManager.java
```diff
@@ -90,15 +88,28 @@ public void updateSchemaVersionId(int versionId) {
...
+        // The schema version should never regress, so make sure we don't change
+        // anything if this code is older than what's currently in the database
+        if (latestCodeVersion > getVersionForSchema()) {
+            updateSchemaVersionId(this.latestCodeVersion);
+        }
+    }
+
+    /**
+     * Returns true if the current schema version recorded in WHOLE_SCHEMA_VERSION is older
+     * the last version in FhirSchemaVersion and therefore needs to be updated
```
lmsurpre, line 100: super minor, only update if making other edits
```suggestion
     * than the last version in FhirSchemaVersion and therefore needs to be updated
```

fhir-bucket/src/main/java/com/ibm/fhir/bucket/app/Main.java
```diff
@@ -753,13 +753,13 @@ public void bootstrapDb() {
...
-            if (svm.isLatestSchema()) {
-                logger.info("Already at latest version; skipping update for: '" + schemaName + "'");
-            } else {
+            if (svm.isSchemaOld()) {
                 buildSchema();
                 
                 // Update the whole schema version
                 svm.updateSchemaVersion();
+            } else {
+                logger.info("Already at latest version; skipping update for: '" + schemaName + "'");
```
lmsurpre, line 762: is this still true if `--force` was passed?

lmsurpre: never mind, I don't think `--force` was added for the fhir-bucket schema

lmsurpre (approved):
LGTM

prb112 (approved):
LGTM

Issue #3212, opened 2022-01-19
Bundle Change impacted the member-match behavior.


prb112:
**Describe the bug**

Bundle Change impacted the member-match behavior.

**Environment**
Which version of IBM FHIR Server? main

**To Reproduce**

**Expected behavior**

**Additional context**
The original change that uncovered this one was for #3058

kmbarton423:
Created multiple resourceTypes with the same resource.id (Patient, Coverage, etc).  
Confirmed $member-match works as expected when processing these resources.

PR #3213, opened 2022-01-19
Bundle Change impacted the member-match behavior. #3212

prb112:
- During the Bundle behavior refactoring, the Bundle.Entry.id is no longer populated. The member-match needs this id (which was a proxy for the Resource.id). 
- The code change redirects to Resource.id. 
- The shell script change makes sure we test Member-Match


lmsurpre (approved):
LGTM

PR #3225, opened 2022-01-20
Update to clean up the dependency trees

prb112:
- Remove unused surface area. 

- Address 
fhir-persistence-blob:
kotlin-stdlib-1.3.72.jar (pkg:maven/org.jetbrains.kotlin/kotlin-stdlib@1.3.72, cpe:2.3:a:jetbrains:kotlin:1.3.72:*:*:*:*:*:*:*) : CVE-2020-29582
kotlin-stdlib-common-1.3.70.jar (pkg:maven/org.jetbrains.kotlin/kotlin-stdlib-common@1.3.70, cpe:2.3:a:jetbrains:kotlin:1.3.70:*:*:*:*:*:*:*) : CVE-2020-29582
- also an issue in bulkdata.

### DIFF
--- a/fhir-parent/pom.xml
+++ b/fhir-parent/pom.xml
@@ -641,6 +641,7 @@
                 <artifactId>azure-core</artifactId>
                 <version>1.24.0</version>
             </dependency>
+            <!-- Used in OKHTTP as part of azure-core -->
             <dependency>
                 <groupId>org.jetbrains.kotlin</groupId>
                 <artifactId>kotlin-stdlib</artifactId>
--- a/fhir-parent/pom.xml
+++ b/fhir-parent/pom.xml
@@ -642,6 +642,16 @@
                 <version>1.24.0</version>
             </dependency>
             <dependency>
+                <groupId>org.jetbrains.kotlin</groupId>
+                <artifactId>kotlin-stdlib</artifactId>
+                <version>1.6.10</version>
+            </dependency>
+            <dependency>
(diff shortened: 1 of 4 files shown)


fhir-parent/pom.xml
```diff
@@ -641,6 +641,16 @@
...
+            <dependency>
+                <groupId>org.jetbrains.kotlin</groupId>
+                <artifactId>kotlin-stdlib</artifactId>
+                <version>1.6.10</version>
+            </dependency>
+            <dependency>
+                <groupId>org.jetbrains.kotlin</groupId>
+                <artifactId>kotlin-stdlib-common</artifactId>
+                <version>1.6.10</version>
+            </dependency>
```
lmsurpre, line 653: maybe a comment indicating these are used in azure blob sdk from fhir-persistence-blob?

fhir-parent/pom.xml
```diff
@@ -641,6 +641,16 @@
                 <artifactId>azure-core</artifactId>
                 <version>1.24.0</version>
             </dependency>
+            <dependency>
```
prb112, line 644:
```suggestion
            <!-- Used in OKHTTP as part of azure-core -->
            <dependency>
```

lmsurpre (approved):
Assuming it still works after the exclusions, then LGTM.

prb112:
> Assuming it still works after the exclusions, then LGTM.

It should since it's part of the original dep.

Issue #3245, opened 2022-01-25
Encode the datetime values in the bundle.Link.URL in the resource


sagarsarvankar:
**Describe the bug**
When there are many records for a resource, it gives "next" url in the bundle. We have a case as below
"link": [
{
"relation": "self",
"url": "https://fhirserver.justtest.in:9443/fhir-server/api/v4/Observation?_count=10&category=vital-signs&date=gt1940-10-04T00:00:00+00:00&patient=17e77411359-8837d878-ea1e-45b5-bb89-882fdfb2be4b&_page=1"
},
{
"relation": "next",
"url": "https://fhirserver.justtest.in:9443/fhir-server/api/v4/Observation?_count=10&category=vital-signs&date=gt1940-10-04T00:00:00+00:00&patient=17e77411359-8837d878-ea1e-45b5-bb89-882fdfb2be4b&_page=2"
}
Here in the URL, there is "+" sign in the date parameter, and hence the GET on this URL to FHIR server fails saying below
{
"resourceType": "OperationOutcome",
"id": "ac-1f-5-a3-1d80ad82-06d4-41c1-9f6a-2cd2c2ecfc7a",
"issue": [
{
"severity": "fatal",
"code": "invalid",
"details": {
"text": "Invalid Date Time Format found please use 'yyyy-mm-ddThh:mm:ss[Z|(+|-)hh:mm].'"
},}]}

**Environment**
4.10.1

**To Reproduce**
1. search with a timestamp that include timezone offset
2. check the self url

**Expected behavior**
The returned URL after encoding should look like https://fhirserver.justtest.in:9443/fhir-server/api/v4/Observation?_count=10&category=vital-signs&date=gt1940-10-04T00:00:00%2B00:00&patient=17e77411359-8837d878-ea1e-45b5-bb89-882fdfb2be4b&_page=2
where + is replaced with %2B

More details are located at https://www.hl7.org/fhir/http.html#paging it says "The links are opaque to the client, have no dictated structure, and only the server understands them. The client must use the server supplied links in order to traverse the pages."

Also, there was a discussion in the chat and is available at https://chat.fhir.org/#narrow/stream/212434-ibm/topic/Bundle.20Link.20URL

tbieste:
I verified that each '+' in the link URLs are encoded as %2B.

PR #3246, opened 2022-01-25
Encode the datetime values in the bundle.Link.URL in the resource #3245

prb112:

### DIFF
--- a/fhir-search/src/main/java/com/ibm/fhir/search/uri/UriBuilder.java
+++ b/fhir-search/src/main/java/com/ibm/fhir/search/uri/UriBuilder.java
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright IBM Corp. 2019, 2021
+ * (C) Copyright IBM Corp. 2019, 2022
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -100,7 +100,10 @@ public class UriBuilder {
         URI selfUri = new URI(requestUri.getScheme(), requestUri.getAuthority(), requestUri.getPath(),
                 queryString.toString(), null);
 
-        return selfUri.toString();
+        // ... originally the approach was to modify the Type.Date parameter value, however that's unreliable
+        // as a '+' is a control character which the URI code then tries to escape the escaped value.
+        String urlString = selfUri.toString();
+        return urlString.replace("+", "%2B");
     }
 
     /**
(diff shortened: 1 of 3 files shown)


fhir-search/src/main/java/com/ibm/fhir/search/uri/UriBuilder.java
```diff
@@ -100,7 +100,10 @@ public String toSearchSelfUri() throws URISyntaxException {
         URI selfUri = new URI(requestUri.getScheme(), requestUri.getAuthority(), requestUri.getPath(),
                 queryString.toString(), null);
 
-        return selfUri.toString();
+        // ... originally the approach was to modify the Type.Date parameter value, however that's unreliable
+        // as a '+' is a control character which the URI code then tries to escape the escaped value.
+        String urlString = selfUri.toString();
+        return urlString.replace("+", "%2B");
```
prb112, line 106: Originally I tried manipulating the Type.Date on 294's stream. This turns out that java.net.URI has an encoding for '%' which causes it to get encoded so you end up with `%252B` which is not ideal. There are alternatives such as switching the framework to use Guava and URLEscapers, however that is a heavy lift.  This case it converts the + to be a plus and the framework on read doesn't convert to SPACE

lmsurpre (approved):
LGTM

tbieste (commented):
I looked at the ways the self/next/prev links are built in FHIRRestHelper.java, and I'm not sure if the prev link would see this change.  Perhaps a ServerTest (FVT) level test to ensure it's correct.

prb112:
> I looked at the ways the self/next/prev links are built in FHIRRestHelper.java, and I'm not sure if the prev link would see this change. Perhaps a ServerTest (FVT) level test to ensure it's correct.

FHIRRestHelper uses the selfUrl from SearchUtil to build the links, and then uses the selfUrl with the fix to build the prev and next links. I'll add an integration test.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SearchAllTest.java
```diff
@@ -576,6 +578,24 @@ public void testSearchAllUsingOneTag() throws Exception {
         }
     }
 
+    @Test(groups = { "server-search-all" }, dependsOnMethods = { "testSearchAllUsingOneTag" })
+    public void testSearchAllUsingOneTagForPrevNext() throws Exception {
```
prb112, line 582: @tbieste here is the FVT

tbieste (approved):

Issue #3250, opened 2022-01-26
Implement the $versions operation


tbieste:
**Is your feature request related to a problem? Please describe.**
If the IBM FHIR Server is going to support both R4 and R4B, it would be useful to implement the FHIR $versions operation that allows the client to discover which versions are supported.  This tells the users which FHIR Version Mime Type Parameter values are supported by the server.

**Describe the solution you'd like**
Implement this operation:  https://www.hl7.org/fhir/capabilitystatement-operation-versions.html

**Describe alternatives you've considered**
Only mention which FHIR versions are supported in documentation, and do not provide a programmable way to discover it.

**Acceptance Criteria**

1.  
   GIVEN IBM FHIR server supports R4 and R4B
   WHEN $versions operation is called
   THEN return the versions supported, R4 [4.0] and R4B [4.3]

**Additional context**
It's a very simple API operation, so it would be convenient to add at the same time as the additional R4B support.

PR #3273, opened 2022-02-02
Issue #3250 - Implement the $versions operation

tbieste:
Issue #3250 - Implement the $versions operation


operation/fhir-operation-versions/src/main/java/com/ibm/fhir/operation/versions/VersionsOperation.java
```diff
@@ -0,0 +1,62 @@
...
+            throws FHIROperationException {
+        try {
+
+            Parameters.Builder builder = Parameters.builder();
+            builder.parameter(Parameter.builder()
+                .name(string(PARAM_VERSION))
+                .value(Code.of("4.0")).build());
+            builder.parameter(Parameter.builder()
+                .name(string(PARAM_DEFAULT))
+                .value(Code.of("4.0")).build());
```
lmsurpre, line 50: shouldn't need to use the static string method any more
```suggestion
            builder.parameter(Parameter.builder()
                .name(PARAM_VERSION)
                .value(Code.of("4.0"))
                .build());
            builder.parameter(Parameter.builder()
                .name(PARAM_DEFAULT)
                .value(Code.of("4.0"))
                .build());
```

docs/src/pages/Conformance.md
```diff
@@ -474,6 +474,7 @@ System operations are invoked at `[base]/$[operation]`
 | [$export](https://hl7.org/fhir/uv/bulkdata/STU1/OperationDefinition-export.html) | Export data from the server | exports to an S3-compatible data store; see the [user guide](https://ibm.github.io/FHIR/guides/FHIRServerUsersGuide#4101-bulk-data-export) for config info |
 | [$import](https://github.com/smart-on-fhir/bulk-import/blob/main/import.md) | Import FHIR Resources from a source| see the [user guide](https://ibm.github.io/FHIR/guides/FHIRServerUsersGuide#4101-bulk-data-export) for config info. This implementation is based on the proposed operation.|
 | [$healthcheck](https://github.com/IBM/FHIR/blob/main/operation/fhir-operation-healthcheck/src/main/resources/healthcheck.json) | Check the health of the server | Checks for a valid connection to the database |
+| [$versions](https://hl7.org/fhir/capabilitystatement-operation-versions.html) | Returns the list of supported FHIR versions | Returning output in non-FHIR formats is not supported |
```
lmsurpre, line 477: > Returning output in non-FHIR formats is not supported
there actually is a way to coerce our operation framework to do this, but its not real pretty.  paul introduced it for the bulkdata-status operation

tbieste: Did you want that implemented right now in this issue?  Or can that wait?

lmsurpre: not sure what the $versions response is supposed to be...was just commenting on the statement in the doc

lmsurpre: ok, i looked it up:
> For the convenience of clients, servers that implement this operation SHALL also support the relevant XML or JSON format

so it does look like its required

tbieste: Correct, that's why for now I added the statement in the Conformance doc to say we don't support returning the output -- at least not yet.  If that needs to implemented in this PR, I'll need help to be pointed to what you mentioned that Paul did.  Otherwise, I can open a follow-on issue to add that support later, and we can prioritize it appropriately.

tbieste: I think I found Paul's example.  I'm going to give it a try.

operation/fhir-operation-versions/src/main/java/com/ibm/fhir/operation/versions/VersionsOperation.java
```diff
@@ -0,0 +1,189 @@
...
+    private String getNonFhirOutputFormat(FHIROperationContext operationContext) {
+        String headerString = operationContext.getHeaderString(ACCEPT_HEADER);
+        if (headerString != null) {
+            for (String headerStringElement : headerString.split(",")) {
+                if (MediaType.APPLICATION_JSON.equals(headerStringElement)) {
+                    return MediaType.APPLICATION_JSON;
+                }
+                if (MediaType.APPLICATION_XML.equals(headerStringElement)) {
+                    return MediaType.APPLICATION_XML;
+                }
```
lmsurpre, line 175: To be robust to content-type and other MIME-type parameters, we should probably split the headerStringElement strings by ";", take just the first component, then trim().

operation/fhir-operation-versions/src/main/java/com/ibm/fhir/operation/versions/VersionsOperation.java
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    /**
+     * Gets the non-FHIR format for returning output. If the first Accept header
+     * is either 'application/json' or 'application/xml', the output is returned in
+     * that non-FHIR format.
+     * @param operationContext the operation context
+     * @return 'application/json', 'application/xml', or null
+     */
+    private String getNonFhirOutputFormat(FHIROperationContext operationContext) {
```
lmsurpre, line 166: We should probably revisit this private helper when we merge with the r4b-rebased branch...we might be able to pass the MediaType through the OperationContext instead of having to parse the header here.

lmsurpre (approved):
LGTM

PR #3286, opened 2022-02-05
Refactor ResourceTypeHelper and update callers

lmsurpre:
Replace getNewOrBreakingResourceTypeNames() with a more
generic `isCompatible(resType, fhirVersion1, fhirVersion2)` method.

To implement it, I introduced new members to the ResourceTypeName enums
so that we can compute which resource types are associated with which
FHIR versions.

The compatibility knowledge is still encapsulated in
ResourceTypeHelper, but now we more cleanly distinguish between
R4B_ONLY_RESOURCE_TYPES and BACKWARD_BREAKING_R4B_ENUMS (previously they
were both together in the R4B_ONLY_RESOURCE_TYPES list). And now we have
a way to get the inverse (valid 4.3 resources that maybe not be valid in
4.0), although this isn't currently used by the server.

### DIFF
--- a/fhir-core/src/main/java/com/ibm/fhir/core/util/ResourceTypeHelper.java
+++ b/fhir-core/src/main/java/com/ibm/fhir/core/util/ResourceTypeHelper.java
@@ -19,90 +19,208 @@ import com.ibm.fhir.core.ResourceTypeName;
  * Helper methods for working with FHIR Resource Type Strings
  */
 public class ResourceTypeHelper {
-    private static final Set<ResourceTypeName> REMOVED_RESOURCE_TYPES = collectRemovedResourceTypes();
-    private static final Set<ResourceTypeName> R4B_ONLY_RESOURCE_TYPES = collectR4bOnlyResourceTypes();
-    private static final Set<ResourceTypeName> ABSTRACT_TYPES = Collections.unmodifiableSet(new HashSet<>(
+    private static final Set<ResourceTypeName> R4_ENUMS = collectResourceTypesFor(FHIRVersionParam.VERSION_40);
+    private static final Set<ResourceTypeName> R4B_ENUMS = collectResourceTypesFor(FHIRVersionParam.VERSION_43);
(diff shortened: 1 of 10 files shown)


fhir-server-spi/src/main/java/com/ibm/fhir/server/spi/operation/FHIRResourceHelpers.java
```diff
@@ -76,7 +76,8 @@ public static Interaction from(String value) {
      * @param interaction
      *            the interaction to be performed
      * @param resourceType
-     *            the resource type against which the interaction is to be performed
+     *            the resource type against which the interaction is to be performed; use "Resource"
+     *            for whole-system interactions
```
lmsurpre, line 80: this is the only intentional change in this file...the rest is just whitespace cleanup

fhir-core/src/main/java/com/ibm/fhir/core/util/ResourceTypeHelper.java
```diff
@@ -19,90 +19,208 @@
...
                 .collect(Collectors.toSet()));
 
     /**
      * @param fhirVersion The value of the MIME-type parameter 'fhirVersion' for the current interaction
-     *          (e.g. "4.3" not "4.3.0")
-     * @return a set of resource type names that corresponds to the requested fhirVersion
+     * @return the set of resource type names from FHIR R4B that are backwards-compatible with the requested fhirVersion
      */
-    public static Set<String> getResourceTypesFor(FHIRVersionParam fhirVersion) {
+    public static Set<String> getR4bResourceTypesFor(FHIRVersionParam fhirVersion) {
```
lmsurpre, line 90: on the fence about whether to keep this flavor for convenience or force everything into `getCompatibleResourceTypes(sourceFhirVersion, targetFhirVersion)`.  thoughts?

punktilious (approved):
LGTM

Issue #3289, opened 2022-02-07
Remove duplication of ResourceType strings and enums


lmsurpre:
**Is your feature request related to a problem? Please describe.**
Because of the way we generate code subtypes from code elements with a required binding, fhir-model in 4.x has two code subtype classes that are basically identical:
* ResourceType
* FHIRResourceType

Because working with resource types is so common in the codebase, and having an enum of all resource types is handy, we have a lot of places that use the underlying Value enums from these two classes (ResourceType.Value and FHIRResourceType.Value).

For https://github.com/IBM/FHIR/pull/3268 I wanted to introduce a new ResourceType enum that could be used from fhir-config (and preferably without the complexity overhead of the existing ones) and so I introduce a new ResourceTypeName enum with almost the same content.

**Describe the solution you'd like**
1. Core, ubiquitous enum constants for each resource type in the spec.
2. A single code subtype that wraps this enum for use in the model.

**Describe alternatives you've considered**
Consolidate ResourceType and FHIRResourceType but keep the resulting class in fhir-model and do not have a separate enum for ResourceType in fhir-core.

**Acceptance Criteria**

1.  
   GIVEN [a precondition]
    AND [another precondition]
   WHEN [test step]
    AND [test step]
   THEN [verification step]
    AND [verification step]

**Additional context**

lmsurpre:
I confirm that the ResourceType and FHIRResourceType code subtypes have been replaced with a unified Code subtype *ResourceTypeCode* and this code subtype uses the *ResourceType* enum from fhir-core for its values.

PR #3290, opened 2022-02-07
issue #3289 - rationalize ResourceType and ResourceTypeCode

lmsurpre:
1. Update the CodeGenerator to combine the ResourceType and
FHIRResourceType code subtypes into a single new code subtype named
`ResourceTypeCode`
2. Rename the ResourceTypeName enum in fhir-core to ResourceType, move
the `from(String)` method there, and update ResourceTypeCode to use this
enum instead of duplicating the values in a nested class
3. Update all uses of ResourceType.Value and FHIRResourceType.Value to
use the new ResourceType enum instead

The result is
1. a simple enum we can use for ResourceType constants throughout the
codebase (without the added overhead of optional id and extension fields); and
2. a single code subtype that builds on this enum for when its used
within the fhir model


cql/fhir-cql/src/main/java/com/ibm/fhir/cql/engine/model/FHIRModelResolver.java
```diff
@@ -102,7 +103,13 @@ public String getPackageName() {
 
         // add all code subtypes
         for (Class<?> codeSubtype : ModelSupport.getCodeSubtypes()) {
+            if (codeSubtype == ResourceTypeCode.class) {
+                // special case handling for ResourceTypeCode
+                typeMap.put("ResourceType", codeSubtype);
+                typeMap.put("FHIRResourceType", codeSubtype);
+            }
```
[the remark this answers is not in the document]
lmsurpre: > Is that your thinking here?

precisely

> Are there other places where clients of the model would expect those old type names?

I hope not.  but thats a big reason I wanted to do this change in our 5.0 release...I feel a bit better about Java api-breaking changes like this to fhir-model if they are in a major release bump

> My first thought when I encountered this was that the mapping from our internal representation to the external FHIR representation is logic that might be more properly placed in ModelSupport.

Not exactly sure what you mean by that, but happy to discuss.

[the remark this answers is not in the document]
csandersdev: Basically, I was wondering if ModelSupport should have something like `isResourceTypeCode(Class<?>)` or something like that. A utility that could be reused if other code outside of the FHIRModelResolver cared about this change. If the CQL engine is the only player, then it doesn't matter. I am ok with these changes (I don't have formal approval rights).

conformance/fhir-ig-davinci-pdex/src/test/java/com/ibm/fhir/ig/davinci/pdex/tool/SearchParameterAugmenter.java
```diff
@@ -16,6 +16,7 @@
 import java.util.stream.Collectors;
```
tbieste, line 16: Update copyright year. (Same comment on many of the files.)

tbieste (approved):

conformance/fhir-ig-davinci-pdex/src/test/java/com/ibm/fhir/ig/davinci/pdex/tool/SearchParameterAugmenter.java
lmsurpre: yeah, i purposefully didn't bother with that since most of the edits were auto-refactoring via eclipse...wanted to make sure we wanted these changes first.  happy to update the copyrights if we like this direction

cql/fhir-cql/src/main/java/com/ibm/fhir/cql/engine/model/FHIRModelResolver.java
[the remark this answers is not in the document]
lmsurpre: OK.  This is the only spot in our codebase that would benefit from such a method.  And I'm hoping most folks aren't dependent on the class names of our code subtypes because they aren't really even real artifacts from FHIR (we derive them from the binding names on ElementDefinitions shipped with the spec). I think its mostly just luck/coincidence that we happened to use the same field name as HAPI for these Classnames in the first place.

conformance/fhir-ig-davinci-pdex/src/test/java/com/ibm/fhir/ig/davinci/pdex/tool/SearchParameterAugmenter.java
lmsurpre: ok, I went through and tried to fix these.  i included an updated codegen-support/header.txt (used by CodeGenerator for generating all model classes) but didn't regenerate the whole model so as not to pollute the PR.

tbieste (approved):
LGTM

fhir-model/src/main/java/com/ibm/fhir/model/resource/CompartmentDefinition.java
```diff
@@ -1083,7 +1083,7 @@ protected Builder from(CompartmentDefinition compartmentDefinition) {
             valueSet = "http://hl7.org/fhir/ValueSet/resource-types|4.3.0-CIBUILD"
         )
         @Required
-        private final ResourceType code;
+        private final ResourceTypeCode code;
```
prb112, line 1086: The headers.txt needing to be updated.

prb112 (changes requested):
CodeGenerator uses an out of date copyright

prb112 (approved):

fhir-model/src/main/java/com/ibm/fhir/model/resource/CompartmentDefinition.java
lmsurpre: headers.txt was updated in previous commit.  now i pushed an update that sets the 2022 copyright on all remaining generated classes in fhir-model

PR #3296, opened 2022-02-07
Generators not following Swagger and APIC spec

mpayne2:
Signed-off-by: mpayne2 <mpayne2@us.ibm.com>

### DIFF
--- a/fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
+++ b/fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
@@ -567,7 +567,11 @@ public class FHIROpenApiGenerator {
         info.add("version", "4.0.0");
         swagger.add("info", info);
 
-        swagger.add("basePath", CONTEXT_ROOT);
+        JsonArrayBuilder servers = factory.createArrayBuilder();
+        JsonObjectBuilder server = factory.createObjectBuilder();
+        server.add("url", CONTEXT_ROOT);
+        servers.add(server);
+        swagger.add("servers", servers);
 
         JsonObjectBuilder paths = factory.createObjectBuilder();
         JsonObjectBuilder definitions = factory.createObjectBuilder();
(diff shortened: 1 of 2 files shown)


fhir-swagger-generator/src/main/java/com/ibm/fhir/openapi/generator/FHIROpenApiGenerator.java
```diff
@@ -1465,10 +1477,6 @@ private static void generateWholeSystemHistoryPathItem(JsonObjectBuilder path, S
         }
         get.add("operationId", operationId);
 
-        JsonArrayBuilder produces = factory.createArrayBuilder();
-        produces.add(FHIRMediaType.APPLICATION_FHIR_JSON);
-        get.add("produces", produces);
```
prb112, line 1470: I'd expect this to exist going forward. Where is it set if it's not compliant?

mpayne2: Top level produces array has been removed.  It is specified in the responses content.  For example:
                "responses": {
                    "200": {
                        "description": "Search Condition operation successful",
                        "content": {
                            "application/fhir+json": {
                                "schema": {
                                    "$ref": "#/components/schemas/Bundle"
                                }
                            }
                        }
                    }
                }

prb112: Right, not what I would expect, since you can specify Content-Type as one (say application/fhir+json), and Accept as another (say application/fhir+xml).  Just a bit surprising.

mpayne2: Accept type would be specified in the requestBody.  Similar pattern:

```
                "requestBody": {
                    "content": {
                        "application/x-www-form-urlencoded": {
                            "schema": {
                                "type": "object",
                                "properties": {
                                }
                            }
                        }
                    }
                },

```

fhir-openapi/src/main/webapp/META-INF/openapi.json
```diff
@@ -914,9 +914,6 @@
                 ],
                 "summary": "Get the whole system history for Account resources",
                 "operationId": "wholeSystemHistoryAccount",
-                "produces": [
-                    "application/fhir+json"
-                ],
                 "parameters": [
                     {
                         "$ref": "#/components/parameters/_sortParam"
```
lmsurpre, line 919: seems unrelated to your PR, but is this even valid JSON?  don't we need a closing curly-brace here?

lmsurpre (approved):
LGTM

prb112 (approved):

Issue #3298, opened 2022-02-07
$member-match should throw FHIROperationException instead of rewrapping the exception


prb112:
**Describe the bug**
$member-match should throw FHIROperationException instead of rewrapping the exception. 

https://github.com/IBM/FHIR/blob/124c2b6686efb5e40e3efd4196590177300f33db/operation/fhir-operation-member-match/src/main/java/com/ibm/fhir/operation/davinci/hrex/provider/strategy/DefaultMemberMatchStrategy.java#L207

**Environment**
Which version of IBM FHIR Server? main

**To Reproduce**
1. Create Two Patients
2. Throw Exception directly

**Expected behavior**

**Additional context**

prb112:
QA: Logging shown in debugging session with IHI team

PR #3299, opened 2022-02-07
$member-match should throw FHIROperationException instead of rewrapping the exception #3298

prb112:
- Facilitate throwing the actual exception

### DIFF
--- a/operation/fhir-operation-member-match/src/main/java/com/ibm/fhir/operation/davinci/hrex/provider/strategy/DefaultMemberMatchStrategy.java
+++ b/operation/fhir-operation-member-match/src/main/java/com/ibm/fhir/operation/davinci/hrex/provider/strategy/DefaultMemberMatchStrategy.java
@@ -202,6 +202,8 @@ public class DefaultMemberMatchStrategy extends AbstractMemberMatch {
                         .responseType(ResponseType.NO_MATCH)
                         .build();
             }
+        } catch (FHIROperationException foe){
+            throw foe;
         } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


operation/fhir-operation-member-match/src/main/java/com/ibm/fhir/operation/davinci/hrex/provider/strategy/DefaultMemberMatchStrategy.java
```diff
@@ -202,6 +202,8 @@ public MemberMatchResult executeMemberMatch() throws FHIROperationException {
                         .responseType(ResponseType.NO_MATCH)
                         .build();
             }
+        } catch (FHIROperationException foe){
+            throw foe;
         } catch (Exception e) {
             LOG.throwing(getClass().getSimpleName(), "executeMemberMatch", e);
             throw FHIROperationUtil.buildExceptionWithIssue("Error executing the MemberMatch", IssueType.EXCEPTION);
```
lmsurpre, line 209: you sure we don't want to include the caused by in this case too?  worried about leaking too much info to the client in the response?

lmsurpre (approved):

operation/fhir-operation-member-match/src/main/java/com/ibm/fhir/operation/davinci/hrex/provider/strategy/DefaultMemberMatchStrategy.java
prb112: Yeah, that's why.  I'm intentionally keeping it light.

Issue #3159, opened 2022-01-06
Dockerfile doesn't update yum on build


prb112:
**Describe the bug**

Dockerfile doesn't update yum on build. In release 4.7.0, we removed the yum update. Need to understand the reasons and if this is intended. 

https://github.com/IBM/FHIR/commit/6b5e844b39abe0d17c0a0cd86d99a67989341d06#diff-bb21e41aa1608e829c45be2d17d3831a0e77f0c602fc8e0213753f2ee7298ded

```USER root
RUN yum update -y && \
    yum clean all && \
    rm -rf /var/cache/yum

USER 1001
```

**Environment**
Which version of IBM FHIR Server?

**To Reproduce**

**Expected behavior**

**Additional context**

prb112:
PR #3302

lmsurpre:
the dockerfile is updated

Issue #1412, opened 2020-08-13
Establish process/procedure for rebuilding old images


lmsurpre:
Currently, we build and push our Docker image to DockerHub as part of our release process.

This works, but ends up leaving vulnerable images because the underlying layers get patched but we never re-build our images on top of them.

What we *should* be doing is rebuilding the images at some interval so that our image is always up-to-date at the lower levels (OS, Java, Liberty, etc), probably via either [autobuilds](https://docs.docker.com/docker-hub/builds/) or [GitHub Actions scheduled events](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events)

As a forcing function, Docker recently changed their terms of service and will start deleting stale images that haven't been rebuilt in the last 6 months.

prb112:
Team to setup a side discussion on how to approach

prb112:
Proposal
- Rebuild on a Schedule 
- All Tags

Go from there.

lmsurpre:
now that it only rebuilds the latest (4.11.0), we finally have an automated image rebuild!

first success: https://github.com/IBM/FHIR/actions/runs/2200236782

d0roppe:
This seems to be building correctly now. Closing issue.

PR #3302, opened 2022-02-08
Dockerfile doesn't update yum on build #3159

prb112:


lmsurpre (approved):
Agree we should do this, at least until we have the "image rebuild" (https://github.com/IBM/FHIR/issues/1412) working properly and have seen a history of OS-level patches in the upstream liberty image.

PR #3307, opened 2022-02-08
Do not return OperationOutcome resources or duplicates from the batch 

sfc42:


prb112 (approved):

operation/fhir-operation-everything/src/main/java/com/ibm/fhir/operation/everything/EverythingOperation.java
```diff
@@ -828,12 +829,14 @@ private void readsOfAdditionalAssociatedResources(String compartmentMemberType,
...
             for (Entry entry: responseBundle.getEntry()) {
-                String resourceId = ModelSupport.getTypeName(entry.getResource().getClass()) + "/" + entry.getResource().getId();
-                resourceIds.add(resourceId);
-                // remove this next line if fullUrls start getting set on batch reads
-                Bundle.Entry.Builder entryBuilder = entry.toBuilder()
-                        .fullUrl(Uri.of(url + resourceId));
-                allEntries.add(entryBuilder.build());
+                String type = ModelSupport.getTypeName(entry.getResource().getClass());
+                // OperationOutcomes come back if the resource isn't found, so take them out
+                if (!type.equals("OperationOutcome")) {
```
lmsurpre, line 834: A good catch.  I think an alternative would be to check the `Bundle.entry.search.mode` but I think just excluding the OperationOutcome by type should work just fine.

Issue #3303, opened 2022-02-08
Race condition in delete interaction


punktilious:
**Describe the bug**
If the resource to be deleted is already deleted, an extra read is performed. This read could potentially return a resource which has just been undeleted (by another thread).
```
                // Read the resource so it will be available to the beforeDelete interceptor methods.
                try {
                    resourceToDelete = doRead(type, id, !THROW_EXC_ON_NULL, !INCLUDE_DELETED, null, null, !CHECK_INTERACTION_ALLOWED).getResource();
                    if (resourceToDelete != null) {
                        responseBundle = Bundle.builder().type(BundleType.SEARCHSET)
                                .id(UUID.randomUUID().toString())
                                .entry(Entry.builder().id(id).resource(resourceToDelete).build())
                                .total(UnsignedInt.of(1))
                                .build();
                    } else {
                        warnings.add(buildOperationOutcomeIssue(IssueSeverity.WARNING, IssueType.NOT_FOUND, "Cannot find "
                                + type + " with id '" + id + "'."));
                    }
                } catch (FHIRPersistenceResourceDeletedException e) {
                    // Absorb this exception.
                    ior.setResource(doRead(type, id, !THROW_EXC_ON_NULL, INCLUDE_DELETED, null, null, !CHECK_INTERACTION_ALLOWED).getResource());
                    warnings.add(buildOperationOutcomeIssue(IssueSeverity.WARNING, IssueType.DELETED, "Resource of type '"
                        + type + "' with id '" + id + "' is already deleted."));
                }
```

See the second doRead call in the catch block.

The doRead method returns a SingleResourceResult which now includes a ResourceResult object holding the deletion status of the resource. The first call can therefore be made with INCLUDE_DELETED and the deletion check logic can be performed without the need for an exception handler and second read.

**Environment**
Which version of IBM FHIR Server? Latest main.

**To Reproduce**
1. Run multiple deletes/undeletes with high concurrency.

**Expected behavior**
No race condition.

**Additional context**

lmsurpre:
non-concurrent regression testing to be performed as part of https://github.com/IBM/FHIR/issues/3293#issuecomment-1046986301

for testing the concurrent delete, maybe open a new issue to extend our ConcurrentUpdate test in fhir-server-test?

Issue #3293, opened 2022-02-07
Avoid storing the resource contents in the case of a soft delete


lmsurpre:
**Is your feature request related to a problem? Please describe.**
In versioned FHIR servers (like ours), the DELETE action is a "soft delete"... its actually more like an update where we set the delete flag to Y.

Despite the resource content being logically deleted, for historical reasons we always write a new version of the entire resource payload in the db on this deleted version.

**Describe the solution you'd like**
Stop writing the resource payload on the deleted version of this record.

**Describe alternatives you've considered**

**Acceptance Criteria**

1.  
   GIVEN [a precondition]
    AND [another precondition]
   WHEN [test step]
    AND [test step]
   THEN [verification step]
    AND [verification step]

**Additional context**
In a previous iteration of the DB, I think we always needed non-null data for the x_Resources tables.
We may have a few places that still assume you can read it back out (maybe audit?), but we think we can manage those.

lmsurpre:
I confirmed that the rows in the x_resources table that correspond to deleted versions now have a null data column:
![image](https://user-images.githubusercontent.com/5478798/156180748-f146f51c-5c3b-457f-ad85-75c577d53aab.png)

lmsurpre:
I deployed this one to our cluster on IBM Cloud and ran the touchstone conformance tests against it.  100% successful.

kmbarton423:
confirmed conditional delete is working as expected
did write up an enhancement issue #3419

kmbarton423:
confirmed if-none-match and if-none-match within a bundle functionality

kmbarton423:
confirmed $erase functionality
found issue 3439 during testing but it appears to be unrelated to these changes for soft delete

lmsurpre:
issue #3439 has been addressed and this feature seems to be working

Issue #3145, opened 2021-12-24
Refactor payload persistence to its own visitor in Bundle processing


punktilious:
**Describe the bug**
Currently the payload persistence step is invoked within the FHIRRestInteractionVisitorReferenceMapping visitor, which is the last time the resource is modified and therefore the first opportunity to store the payload.

This, is a little obscure. It could be much clearer (with probably very little performance impact) to loop through the Bundle entries again to perform the offload in a dedicated visitor. This would only need to be done if offload were enabled - which requires a FHIRPersistence method to support this check.

**Environment**
Which version of IBM FHIR Server? main

lmsurpre:
QA for this will be included with the QA for 3293:  https://github.com/IBM/FHIR/issues/3293#issuecomment-1046986301

PR #3314, opened 2022-02-09
issue #3145 move offload processing to its own interaction visitor

punktilious:
issue #3145 move offload processing to its own interaction visitor for bundle processing

Plus:
#3293. We no longer store a copy of the resource payload in the soft-delete record. This simplifies things for offload, but required a number of API changes which relied on the resource being returned even after it had been deleted. The API now uses the SingleResourceResult class as a carrier which includes the deletion status. Callers can use values from the SingleResourceResult to obtain the meta information about the deletion marker without requiring the resource itself (which no longer exists anyway).

#3303. The doDelete method in FHIRRestHelper now uses a single call to doRead to determine the current state of the resource. This avoids the race condition described in the issue.


fhir-server/src/main/java/com/ibm/fhir/server/rest/FHIRRestInteractionVisitorOffload.java
```diff
@@ -0,0 +1,246 @@
...
+    /**
+     * Unified exception handling for each of the operation calls
+     * @param entryIndex
+     * @param v
+     * @param failFast
+     * @param requestDescription
+     * @param accumulatedTime
+     * @throws Exception
+     */
+    private FHIRRestOperationResponse doOperation(int entryIndex, String requestDescription, long accumulatedTime, Callable<FHIRRestOperationResponse> v) throws Exception {
```
lmsurpre, line 187: maybe `doInteraction` instead?

punktilious: Would need to also change this in all the other visitors, plus that means renaming FHIRRestOperationResponse to FHIRRestInteractionResponse, which becomes a bigger refactor.

fhir-server/src/main/java/com/ibm/fhir/server/rest/FHIRRestInteractionVisitorOffload.java
```diff
@@ -0,0 +1,246 @@
...
+     * @param helpers
+     */
+    public FHIRRestInteractionVisitorOffload(boolean transaction, FHIRResourceHelpers helpers, Map<String, String> localRefMap, Entry[] responseBundleEntries) {
+        super(helpers, localRefMap, responseBundleEntries);
+        this.transaction = transaction;
+    }
+
+    @Override
+    public FHIRRestOperationResponse doSearch(int entryIndex, String requestDescription, FHIRUrlParser requestURL, long accumulatedTime, String type, String compartment, String compartmentId,
+        MultivaluedMap<String, String> queryParameters, String requestUri, Resource contextResource, boolean checkInteractionAllowed) throws Exception {
```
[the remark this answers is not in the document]
punktilious: ok

fhir-server/src/main/java/com/ibm/fhir/server/rest/FHIRRestInteractionVisitorOffload.java
```diff
@@ -0,0 +1,246 @@
...
+    public FHIRRestOperationResponse doInvoke(String method, int entryIndex, Entry validationResponseEntry, String requestDescription, FHIRUrlParser requestURL, long accumulatedTime, FHIROperationContext operationContext, String resourceTypeName, String logicalId,
+            String versionId, Resource resource, MultivaluedMap<String, String> queryParameters) throws Exception {
+        // NOP
+        return null;
+    }
+
+    @Override
+    public FHIRRestOperationResponse doDelete(int entryIndex, String requestDescription, FHIRUrlParser requestURL, long accumulatedTime, String type, String id, String searchQueryString) throws Exception {
+        // NOP
+        return null;
```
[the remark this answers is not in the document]
punktilious: When I was implementing the offload I was working through avoiding storing the deletion marker payload because now we have better handling for return null resources in the persistence layer read methods, but I think there are more use-cases to cover. I think for now I need to store the payload and then we can deal with skipping the store as a later optimization.

fhir-server/src/main/java/com/ibm/fhir/server/rest/FHIRRestInteractionVisitorOffload.java
lmsurpre: oh, is it inhereted?  i thought it was just a private helper.  i'm fine either way...just prefer to save "operation" for the customer operation framework and use "interaction" when dealing with the FHIR REST API.

fhir-persistence/src/main/java/com/ibm/fhir/persistence/FHIRPersistence.java
```diff
@@ -78,14 +78,22 @@
...
-     * @param resource
+     * @param context the FHIRPersistenceContext instance associated with the current request
+     * @param resourceType the resource type of the Resource instance to be deleted
+     * @param logicalId the logical id of the Resource instance to be deleted
+     * @param versionId the current version of the Resource instance to be deleted
+     * @param lastUpdated the modification timestamp to use for the deletion
      * @throws FHIRPersistenceException
      */
-    default <T extends Resource> void delete(FHIRPersistenceContext context, T resource) throws FHIRPersistenceException {
+    default <T extends Resource> void delete(FHIRPersistenceContext context, Class<T> resourceType, String logicalId, int versionId,
```
lmsurpre, line 95: by returning void, we're sort of locked in to the approach of handling all error cases via exceptions.
but its probably fine?

fhir-server-test/src/test/java/com/ibm/fhir/server/test/DeleteTest.java
```diff
@@ -7,6 +7,7 @@
 package com.ibm.fhir.server.test;
```
tbieste, line 7: Update copyright date.

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -8,6 +8,8 @@
```
tbieste, line 8: Update copyright date.

fhir-server/src/main/java/com/ibm/fhir/server/resources/Delete.java
```diff
@@ -138,6 +138,8 @@ private Response buildResponse(FHIRRestOperationResponse ior) {
```
tbieste, line 138: Update copyright date.

fhir-server/src/main/java/com/ibm/fhir/server/rest/FHIRRestInteractionVisitorOffload.java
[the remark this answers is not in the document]
lmsurpre: payload is no longer stored for the delete case, so this is resolved

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -1067,6 +1063,7 @@ private FHIRPersistenceNotSupportedException buildNotSupportedException(String m
     }
 
     /**
+     * Read the current version of the resource from the database.
```
lmsurpre, line 1066: This is an implementation of the public interface...I don't think we should override the javadoc.
If we want to add additional info to the javadoc which is specific to this implementation, we can use `@implNote`.

lmsurpre: I don't think `@throws FHIRPersistenceResourceDeletedException` overrides the interface's javadoc, but if does then we should probably move that to the javadoc on the interface level.

fhir-persistence/src/main/java/com/ibm/fhir/persistence/FHIRPersistence.java
punktilious: Yeah, I'd actually like to refactor our overall error handling approach so that data issues are handled as return values, leaving exceptions to be used for unexpected behavior or system failures etc. But that's a much bigger piece of work. In this case, the pre-checks are performed by the FHIRRestHelper, so when we hit the persistence delete method we are expecting things to work. If something goes wrong here, it's unexpected, so I think an exception is OK.

fhir-server/src/main/java/com/ibm/fhir/server/rest/FHIRRestInteractionVisitorOffload.java
```diff
@@ -0,0 +1,246 @@
...
+       return helpers.storePayload(resource, logicalId, newVersionNumber, resourcePayloadKey);
+    }
+
+    /**
+     * Unified exception handling for each of the operation calls
+     * @param entryIndex
+     * @param v
+     * @param failFast
+     * @param requestDescription
+     * @param accumulatedTime
```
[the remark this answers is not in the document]
punktilious: ok

fhir-server-spi/src/main/java/com/ibm/fhir/server/spi/operation/FHIRRestOperationResponse.java
```diff
@@ -25,6 +25,9 @@
     private OperationOutcome operationOutcome;
```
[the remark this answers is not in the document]
punktilious: thanks

fhir-server-test/src/test/java/com/ibm/fhir/server/test/DeleteTest.java
punktilious: thanks

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
punktilious: thanks

fhir-server/src/main/java/com/ibm/fhir/server/resources/Delete.java
punktilious: thanks

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -457,7 +457,8 @@ private void doCachePrefill() throws FHIRPersistenceException {
      * @throws IOException
      * @throws FHIRGeneratorException
      */
-    private com.ibm.fhir.persistence.jdbc.dto.Resource createResourceDTO(String logicalId, int newVersionNumber,
+    private com.ibm.fhir.persistence.jdbc.dto.Resource createResourceDTO(Class<? extends Resource> resourceType,
```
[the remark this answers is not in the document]
punktilious: ok

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
punktilious: agreed. Removed the comment block

lmsurpre (approved):
LGTM

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
```diff
@@ -231,6 +237,17 @@ protected ResponseBuilder addHeaders(ResponseBuilder rb, Resource resource) {
                 .lastModified(Date.from(resource.getMeta().getLastUpdated().getValue().toInstant()));
     }
 
+    /**
+     * Adds the Etag header to the specified response object.
+     * @param rb
+     * @param resource
```
tbieste, line 243: Extra "resource" param does not exist in method.

tbieste (approved):
One minor comment.  Otherwise, LGTM.

fhir-server/src/main/java/com/ibm/fhir/server/resources/FHIRResource.java
lmsurpre: Since this is the only outstanding comment, I'm going to:
1. add the `ci-skip` label to the PR
2. make a single commit to remove this one line
3. press merge on the PR

```suggestion
```

PR #3320, opened 2022-02-09
Exclude LGPL licensed dependencies

prb112:


lmsurpre (commented):
does it break any of the cql or term-graph stuff?

lmsurpre:
At least the CQL stuff has some tests...so hopefully everything green means its good-to-merge.  Will approve but lets wait for @csandersdev 's approval as well.

lmsurpre (approved):

csandersdev:
The dependency being excluded from the CQL modules is a transitive dependency from some JAXB code (org.jvnet.jaxb2_commons). I'm not really clear on how the javaparser LGPL code is used in that module, so I'm going to have to take some guesses. There are a few areas where I think a JAXB issue might manifest - parsing ELM attachments of libraries, reading a model info file for CQL translation, and then writing the translated ELM back out. All of those things are exercised by the unit tests / server tests which are still passing after this change. I think this is ok for now.

Issue #3232, opened 2022-01-21
Whole system search with _type parameter and global params should be much faster


lmsurpre:
**Describe the bug**
a whole-system search like this:
```
runQueryTest(Resource.class, "_type", "Basic,EvidenceVariable,ServiceRequest")
```
results in queries like these.

count:
```
      SELECT COUNT(*) AS CNT 
        FROM LOGICAL_RESOURCES AS LR0
       WHERE LR0.IS_DELETED = 'N'
         AND LR0.RESOURCE_TYPE_ID IN (8,52,126)
         AND EXISTS (
      SELECT 1 
        FROM LOGICAL_RESOURCES AS LR1
       WHERE LR1.LOGICAL_RESOURCE_ID = LR0.LOGICAL_RESOURCE_ID)
```
fetch ids:
```
      SELECT LR0.RESOURCE_TYPE_ID, LR0.LOGICAL_RESOURCE_ID 
        FROM LOGICAL_RESOURCES AS LR0
       WHERE LR0.IS_DELETED = 'N'
         AND LR0.RESOURCE_TYPE_ID IN (8,52,126)
         AND EXISTS (
      SELECT 1 
        FROM LOGICAL_RESOURCES AS LR1
       WHERE LR1.LOGICAL_RESOURCE_ID = LR0.LOGICAL_RESOURCE_ID)
    ORDER BY LR0.LOGICAL_RESOURCE_ID
 FETCH FIRST 10 ROWS ONLY
```
fetch resources:
```
      SELECT RESOURCE_ID, LOGICAL_RESOURCE_ID, VERSION_ID, LAST_UPDATED, IS_DELETED, DATA, LOGICAL_ID, RESOURCE_PAYLOAD_KEY, RESOURCE_TYPE_ID 
        FROM (
      SELECT R.RESOURCE_ID, R.LOGICAL_RESOURCE_ID, R.VERSION_ID, R.LAST_UPDATED, R.IS_DELETED, R.DATA, LR.LOGICAL_ID, R.RESOURCE_PAYLOAD_KEY, LR.RESOURCE_TYPE_ID AS RESOURCE_TYPE_ID 
        FROM (
      SELECT LR.LOGICAL_RESOURCE_ID, LR.LOGICAL_ID, LR.CURRENT_RESOURCE_ID, 8 AS RESOURCE_TYPE_ID 
        FROM Basic_LOGICAL_RESOURCES AS LR
       WHERE LR.IS_DELETED = 'N'
[...]

PR #3329, opened 2022-02-10
issue #3232- add the extra IN clause more selectively

lmsurpre:
Previously, we were adding this extra IN clause for the _type parameter
in all cases.
In the case of a whole-system search *count* query, this was slowing
things down a bit. We still add the IN clause to the query where it
seemed to help, but now we avoid that for the count query.

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/domain/SearchQuery.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/domain/SearchQuery.java
@@ -29,7 +29,7 @@ public abstract class SearchQuery {
     private final List<SearchParam> searchParams = new ArrayList<>();
 
     // A list of non-parameter related extensions (e.g. location processing)
-    private final List<SearchExtension> extensions = new ArrayList<>();
+    protected final List<SearchExtension> extensions = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/domain/SearchQuery.java
```diff
@@ -76,11 +76,8 @@ public void add(SearchExtension ext) {
         T query = getRoot(visitor);
         visitExtensions(query, visitor);
 
-        // Add the parameters subquery and add the extensions there too
+        // Add the parameters subquery
         T parameterBase = visitor.getParameterBaseQuery(query);
-        for (SearchExtension ext: this.extensions) {
```
punktilious, line 81: But what about the extensions used by other query types?

lmsurpre: we still visit the extensions and add them to the outer query.
we had added these three lines (81-83) just to help postgresql choose a more efficient plan in the case of the WholeSystemFilterQuery.  since it caused unanticipated regression on other queries (e.g. the count), i've moved this to the specific subtype where we were seeing the original issue.

Issue #3216, opened 2022-01-20
FhirPath resolve function


abedultamimi:
entry.first().resource.eventCoding.code='medication-dispense' implies entry.first().resource.focus.all( $this.resolve() is MedicationDispense)

 i'm trying this constraint but without any luck, any idea how resolve works in ibm fhir. i also tried:
entry.first().resource.eventCoding.code='medication-dispense' implies entry.first().resource.focus.all( resolve() is MedicationDispense)

abedultamimi:
Hi @lmsurpre ,

Thanks for taking time to review this, my intention is to have this validated resolving other bundle entry.

Assuem that i have a bundle of 2 entries like the below:
1. MedicationRequest
2. Patient ( fullurl is http://xyz.com/Patient/123)

And im having a MedicationRequest.subject as "Patient/123"

Xyz.com is an arbitrary value not equal to fhir server base url.
My question is can resolve work against bundle entries or it needs to be server resources.

Thanks Again

lmsurpre:
Hi @abedultamimi sorry for the delay.  I think the answer is that our current ResolveFunction implementation does not support the "Resolving references in Bundles" algorithm outlined at https://www.hl7.org/fhir/bundle.html#references

I'm debating between:
A. turning this issue into a feature request for that feature; or
B. labeling this one as a `question`, closing it, and opening a separate feature request for that feature

Let me know if you have a preference.

lmsurpre:
Closing due to lack of response.  We believe this is covered sufficiently in the updated ResolveFunction implementation to be shipped with 4.11.

Issue #2512, opened 2021-06-15
Bundle transaction processing should replace all local references, not just ones that start with "urn:"


lmsurpre:
**Is your feature request related to a problem? Please describe.**
From https://www.hl7.org/fhir/http.html#trules
> A transaction may include references from one resource to another in the bundle, including circular references where resources refer to each other. When the server assigns a new id to any resource in the bundle which has a POST method as part of the processing rules above, it SHALL also update any references to that resource in the same bundle as they are processed (see about Ids in a bundle). References to resources that are not part of the bundle are left untouched. Version-specific references should remain as version-specific references after the references have been updated. Note that when building a transaction, a client can use arbitrarily chosen version references since they will all be re-assigned anyway. Servers SHALL replace all matching links in the bundle, whether they are found in the resource ids, resource references, elements of type uri, url, oid, uuid, and <a href=“” & <img src=“” in the narrative. Elements of type canonical are not replaced.

Today, the IBM FHIR Server only updates literal references when they start with the "urn:" prefix.

**Describe the solution you'd like**
Update all literal references in the bundle when they match a fullUrl in the bundle that is associated with a resource being POSTed.

To determine whether a literal reference needs updating or not, we should follow the guidance at
https://www.hl7.org/fhir/bundle.html#references and I believe that we already have code for this.

**Describe alternatives you've considered**

**Acceptance Criteria**

1.  
   GIVEN a transaction bundle with resources A and B
    AND resource A is a POST
    AND resource B is a POST or PUT with a literal *relative* reference to the fullUrl for resource A
   WHEN the bundle is posted to the server
   THEN the literal reference in resource B is updated to the assigned resource id of resource A

2.  
   GIVEN a transaction bundle with resources A and B
    AND resource A is a POST
    AND resource B is a POST or PUT with a literal *absolute* reference to the fullUrl for resource A
   WHEN the bundle is posted to the server
   THEN the literal reference in resource B is updated to the assigned resource id of resource A

**Additional context**

punktilious:
This should be taken care of in #1869. Need to double-check the implementation.

lmsurpre:
#1869 was closed, but I don't think we got this one in there.  Sound right, @punktilious ?

michaelwschroeder:
The scope of this issue is to replace all local references (not just those starting with the `urn:` prefix) in elements of type `Reference` only. Issue #2993 has been opened to address replacing local references in the following element types, per the FHIR spec:
- resource ids
[...]

PR #3335, opened 2022-02-11
issues #3216 and #2512 - support for alternative reference schemes and bundle ref resolution from FHIRPath ResolveFunction

lmsurpre:
This PR adds support for alternative schemes, like `resource:`, in our
Reference type checking / reference value validation.
Previously, we required reference values to use the `http:`, `https:`, or `urn:` schemes,
but that prevented us from supporting alternative schemes like `resource:` which is
used in the SMART health cards spec.

I thought that I would need to do additional logic for supporting
reference rewriting for this scheme but, based on the tests I added in
FHIRRestHelperTest and ReferenceMappingVisitorTest, apparently not
...we just needed to let it past our parser.

I also included updates to our FHIRPath ResolveFunction implementation so that we can more easily
resolve local bundle references from FHIRPath.


fhir-path/src/main/java/com/ibm/fhir/path/function/ResolveFunction.java
```diff
@@ -94,16 +112,49 @@ public int getMaxArity() {
...
+                            // we know the root of the tree is a Bundle and the current node is of type Reference,
+                            // so walk up the tree until we get to highest Bundle.entry.fullUrl in the tree
+                            // and save the value of its fullUrl
+                            String fullUrl = null;
+                            FHIRPathTree tree = evaluationContext.getTree();
+                            FHIRPathNode nodeUnderTest = tree.getParent(node);
+                            while (nodeUnderTest != root) {
+                                if (nodeUnderTest.isResourceNode()) {
+                                    FHIRPathNode sibling = tree.getSibling(nodeUnderTest, "fullUrl");
+                                    fullUrl = sibling.getValue().asStringValue().string();
```
tbieste, line 125: Looks like null checks are done a lot, except here.  Just making sure this won't ever be null (or check for it, if not).

lmsurpre: honestly not sure if its possible for it to be null, but won't hurt to add some guards...will update

lmsurpre: done

fhir-path/src/test/java/com/ibm/fhir/path/test/ResolveInternalFragmentReferenceTest.java
```diff
@@ -1,67 +0,0 @@
...
-                        .build())
-                    .build())
-                .build();
-
-        EvaluationContext evaluationContext = new EvaluationContext(bundle);
-        FHIRPathResourceNode resourceNode = evaluationContext.getTree().getNode("Bundle.entry[0].resource.contained[0]").asResourceNode();
-
-        Collection<FHIRPathNode> result = FHIRPathEvaluator.evaluator().evaluate(evaluationContext, "providedBy.exists() and providedBy.resolve().partOf.resolve().is(Organization)", resourceNode);
-
-        assertEquals(result, SINGLETON_TRUE);
```
prb112, line 65: What are the thoughts behind the test? too brittle? covered else where?

prb112: Found it

fhir-path/src/test/java/com/ibm/fhir/path/test/ResolveFunctionTest.java
```diff
@@ -1,40 +1,249 @@
...
             result = evaluator.evaluate(evaluationContext, "Observation.subject.where(resolve() is Device)");
-            System.out.println("result: " + result);
-            System.out.println("");
+            assertNotNull(result);
+            assertEquals(result.size(), 0);
+        }
+    }
+
+    @Test
+    public void testResolveInternalFragmentReference() throws Exception {
```
prb112, line 64: from another class...

fhir-path/src/main/java/com/ibm/fhir/path/function/ResolveFunction.java
```diff
@@ -94,16 +112,51 @@ public int getMaxArity() {
...
                     } else {
-                        Matcher matcher = REFERENCE_PATTERN.matcher(referenceReference);
-                        if (matcher.matches()) {
-                            resourceType = matcher.group(RESOURCE_TYPE_GROUP);
-                            if (referenceType != null && !resourceType.equals(referenceType)) {
-                                throw new IllegalArgumentException("Resource type found in reference URL does not match reference type");
+                        if (isBundleContext) {
+                            // we know the root of the tree is a Bundle and the current node is of type Reference,
+                            // so walk up the tree until we get to highest Bundle.entry.fullUrl in the tree
+                            // and save the value of its fullUrl
```
prb112, line 118: While the comment really does help... .

Could you layout a small example? or link to a test example? 

If I have to maintain it, it'd help to laser focus on the 'aaah-ha' that's what it does.

fhir-path/src/main/java/com/ibm/fhir/path/function/ResolveFunction.java
```diff
@@ -94,16 +112,51 @@ public int getMaxArity() {
...
+                                    }
+                                }
+                                nodeUnderTest = tree.getParent(nodeUnderTest);
                             }
-                            String baseUrl = matcher.group(BASE_URL_GROUP);
-                            if ((baseUrl == null ||  matchesServiceBaseUrl(baseUrl)) && evaluationContext.resolveRelativeReferences()) {
-                                // relative reference
-                                resource = resolveRelativeReference(evaluationContext, node, resourceType, matcher.group(LOGICAL_ID_GROUP), matcher.group(VERSION_ID_GROUP));
+
+                            String bundleReference = FHIRUtil.buildBundleReference(reference, fullUrl);
```
prb112, line 132: Note for others reviewing... when fuller is null, it'll return the referenceUriString = ref.getReference().getValue();

when this condition is true (ref.getReference() != null && ref.getReference().getValue() != null)

prb112 (approved):
Left comments for clarification. 

LGTM

fhir-path/src/main/java/com/ibm/fhir/path/function/ResolveFunction.java
```diff
@@ -72,6 +78,17 @@ public int getMaxArity() {
...
+        boolean isBundleContext = root != null && root.type() == FHIRPathType.FHIR_BUNDLE;
+        Map<String, Resource> bundleResources = new HashMap<>();
+        if (isBundleContext) {
+            Bundle bundle = root.asResourceNode().resource().as(Bundle.class);
+            for (Entry e : bundle.getEntry()) {
+                if (e.getResource() != null && e.getFullUrl() != null && e.getFullUrl().hasValue()) {
+                    bundleResources.put(e.getFullUrl().getValue(), e.getResource());
+                }
+            }
+        }
```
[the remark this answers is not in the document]
lmsurpre: my assumption is that we set the `rootResource` to the outermost bundle (the one we're evaluating against).
i tried to account for the "bundle-in-a-bundle" case in the algorithm below by continuing to loop until we've reached the top of the tree.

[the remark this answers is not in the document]
JohnTimm: We have some helper functions (in FHIRPathUtil) that get the right resource depending on how it might be nested. These are typically stashed in the EvaluationContext.

fhir-path/src/main/java/com/ibm/fhir/path/function/ResolveFunction.java
lmsurpre: I added some examples to this comment (and also a test case for the Bundle-in-Bundle case)

fhir-path/src/main/java/com/ibm/fhir/path/function/ResolveFunction.java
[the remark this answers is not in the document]
lmsurpre: I added the test.  I can think of a couple "interesting edge cases" but they seem so edge to me, that I don't care so much about what our behavior is there.

1. what to do if we have a Bundle thats contained within some other resource that is *not* a bundle (and that outer resource is the root/context).  With the current impl, we will not do local bundle resolution in that case, which seems fine to me.

2. what to do if we have a relative literal reference (e.g. `Patient/1`) in a Bundle.entry where that Bundle is inside another Bundle but the outer Bundle has no fullUrl for this entry.  in this scenario, I think the current algorithm will use the fullUrl from the inner bundle's entry to convert the relative URL into an absolute one.  I don't know if that behavior is desired or not, but again it seems like such an edge case that the impl seems ok to me

fhir-path/src/test/java/com/ibm/fhir/path/test/ResolveFunctionTest.java
```diff
@@ -1,40 +1,293 @@
...
+import com.ibm.fhir.model.type.code.BundleType;
+import com.ibm.fhir.model.type.code.HTTPVerb;
 import com.ibm.fhir.path.FHIRPathNode;
+import com.ibm.fhir.path.FHIRPathResourceNode;
+import com.ibm.fhir.path.FHIRPathType;
 import com.ibm.fhir.path.evaluator.FHIRPathEvaluator;
 import com.ibm.fhir.path.evaluator.FHIRPathEvaluator.EvaluationContext;
+import com.ibm.fhir.path.util.FHIRPathUtil;
 
 public class ResolveFunctionTest {
```
punktilious, line 40: javadoc

fhir-path/src/test/java/com/ibm/fhir/path/test/ResolveFunctionTest.java
```diff
@@ -1,40 +1,293 @@
...
-    public static void main(String[] args) throws Exception {
+    @Test
+    public static void testResolve() throws Exception {
```
punktilious, line 42: A comment block would be useful to describe what this is testing

fhir-path/src/test/java/com/ibm/fhir/path/test/ResolveFunctionTest.java
```diff
@@ -1,40 +1,293 @@
...
+        FHIRPathResourceNode resourceNode = evaluationContext.getTree()
+                .getNode("Bundle.entry[0].resource.contained[0]").asResourceNode();
+
+        Collection<FHIRPathNode> result = FHIRPathEvaluator.evaluator().evaluate(evaluationContext,
+                "providedBy.exists() and providedBy.resolve().partOf.resolve().is(Organization)", resourceNode);
+
+        assertEquals(result, SINGLETON_TRUE);
+    }
+
+    @Test
```
punktilious, line 101: comment block

fhir-path/src/test/java/com/ibm/fhir/path/test/ResolveFunctionTest.java
lmsurpre: added

fhir-path/src/test/java/com/ibm/fhir/path/test/ResolveFunctionTest.java
lmsurpre: done

punktilious (approved):
LGTM

fhir-model/src/test/java/com/ibm/fhir/model/util/test/ReferenceMappingVisitorTest.java
```diff
@@ -193,11 +193,40 @@ public void testUpdateReferencesWithRelativeFullUrlAndRelativeReferenceMatch() t
         assertEquals(result.getGeneralPractitioner().get(0).getReference().getValue(), "Practitioner/1");
```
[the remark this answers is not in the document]
lmsurpre: thanks Troy.  I'll
1. throw a ci-skip label on here to prevent the tests from running again
2. ammend my last commit to fix this
3. merge the PR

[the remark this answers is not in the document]
lmsurpre: done

Issue #3039, opened 2021-11-24
Datasource hits an I/IO exception with bulkdata datasource


prb112:
**Describe the bug**

Datasource hits an I/IO exception with bulkdata datasource

```
   Nov 24 09:25:44 test-ibm-fhir-server-b5765889-gqcqg ibm-fhir-server INFO Received request: tenantId:[default] dsId:[default] user:[fhiruser] method:[POST] uri:[https://uffda.cluster1-blue-250babbbe4c3000e15508cd07c1d282b-0000.us-east.containers.appdomain.cloud/fhir-server/api/v4/$import]
    Nov 24 09:25:44 test-ibm-fhir-server-b5765889-gqcqg ibm-fhir-server AUDIT J2CA0056I: The Connection Manager received a fatal connection error from the Resource Adapter for resource jdbc/fhirbatchDB. The exception is: org.postgresql.util.PSQLException: An I/O error occurred while sending to the backend.:java.net.SocketException: Connection reset
    Nov 24 09:25:44 test-ibm-fhir-server-b5765889-gqcqg ibm-fhir-server ERROR WTRN0074E: Exception caught from before_completion synchronization operation: javax.persistence.PersistenceException: Exception [EclipseLink-4002] (Eclipse Persistence Services - 2.7.9.v20210604-2c549e2208): org.eclipse.persistence.exceptions.DatabaseException
    Internal Exception: org.postgresql.util.PSQLException: An I/O error occurred while sending to the backend.
    Error Code: 0
    Call: INSERT INTO FHIR_JBATCH.JOBINSTANCE (AMCNAME, BATCHSTATUS, CREATETIME, EXITSTATUS, INSTANCESTATE, JOBNAME, JOBXMLNAME, JOBXML, UPDATETIME, NUMEXECS, RESTARTON, SUBMITTER) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
        bind => [12 parameters bound]
    Query: InsertObjectQuery(For JobInstanceEntity:  instanceId = 0, batchStatus = STARTING, instanceState = SUBMITTED)
        at org.eclipse.persistence.internal.jpa.EntityManagerSetupImpl$1.handleException(EntityManagerSetupImpl.java:788)
        at org.eclipse.persistence.transaction.AbstractSynchronizationListener.handleException(AbstractSynchronizationListener.java:277)
[...]

PR #3348, opened 2022-02-15
 Datasource hits an I/IO exception with bulkdata datasource #3039

prb112:
- add validationTimeout=30s to the bulkdata datasources


lmsurpre (approved):
validationTimeout=30s matches what we have on our webapp datasources as well...i assume its ok although I'm wondering if we should confirm with @punktilious

Issue #3180, opened 2022-01-12
$import allows adding Resources of multiple types in the same ndjson which could include unsupported resources.


prb112:
**Describe the bug**
$import allows adding Resources of multiple types in the same ndjson which could include unsupported resources.

**Environment**
Which version of IBM FHIR Server? 4.10.2

**To Reproduce**
1. Create an NDJSON with a Condition resource
2. Create a fhir-server-config.json which doesn't have Condition
3. Condition should not be loaded but is.... 

**Expected behavior**

I expect a filtering of the unsupported resources.

**Additional context**

d0roppe:
Verified that $import did not import the resource if it was not known.

PR #3351, opened 2022-02-16
#3180 - $import allows adding Resources of multiple types in the same ndjson which could include unsupported resources.

prb112:
- Adds a configuration feature to check to see if there are mixed resource types in the ndjson (false by default, and optionally able to be set to true).  It might need to be a multivalued enum - allow, disallow, Provenance,Attachment ... for now I think this is sufficient. 
- Add checks in the Import/Load
- Cleans up the ChunkReader so it uses a isLoggable


fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/load/ChunkWriter.java
```diff
@@ -182,19 +183,20 @@ public void writeItems(List<java.lang.Object> arg0) throws Exception {
                     List<Resource> fhirResourceList = (List<Resource>) objResJsonList;
 
                     for (Resource fhirResource : fhirResourceList) {
-                        if (this.resourceType != fhirResource.getClass().getSimpleName()) {
-                            failValidationIds.add(fhirResource.getClass().getSimpleName() + "/" + fhirResource.getId());
+                        String assertedResourceType = fhirResource.getClass().getSimpleName();
+                        if (this.resourceType.equals(assertedResourceType)) {
```
lmsurpre, line 187: is this missing a `!`?
```suggestion
                        if (!this.resourceType.equals(assertedResourceType)) {
```

lmsurpre (approved):
LGTM

Issue #3083, opened 2021-12-06
BulkData Export checks that the Resource Types exported are valid based on the fhir-model and not configurations


prb112:
**Describe the bug**

BulkData Export checks that the Resource Types exported are valid based on the fhir-model. 
In specific configurations users may choose to deploy only a subset of ResourceTypes

Bulk Data should verify the resources are supported in the configuration or in the custom compartment.

```
            if (ExportType.PATIENT.equals(exportType) || ExportType.GROUP.equals(exportType)) {
                if (types != null && !types.isEmpty()) {
                    export.checkExportPatientResourceTypes(types);
                } else {
                    types = export.addDefaultsForPatientCompartment();
                }
            } else if ((ExportType.SYSTEM.equals(exportType) )
                            && (types == null || types.isEmpty())) {
                types = export.defaultResourceTypes();
            }
```

```
public class BulkDataExportUtil {
    private static Set<String> RESOURCE_TYPES = ModelSupport.getResourceTypes(false).stream()
                                                    .map(Class::getSimpleName)
                                                    .collect(Collectors.toSet());
```

**Environment**
Which version of IBM FHIR Server? main

**To Reproduce**

**Expected behavior**

**Additional context**

lmsurpre:
Same sort of issue as #2963 ...would be nice to have a utility we can use which abstracts the config and just gives us a list of which resource types are enabled.

lmsurpre:
when #3355 gets merged, this is expected to work in the following cases:
1. `/fhirServer/resources` config where `open=false` and one or more resource types are listed
2. no `/fhirServer/resources` config (i.e. everything supported)

It is not expected to work in the case of `open=true` with one or more resource types listed.
The fix for that is in the `r4b-rebased` branch.

@prb112 any thoughts on whether to keep this one open (probably mark it blocked by that other one and unassign you) vs opening a separate issue to track that?

prb112:
How about we target the r4b-rebased branch for this one? Thoughts?
I could also do a custom thing where we check open=false.  

I'd like to close it out as complete, rather than leave it incomplete

lmsurpre:
To me, its looking more and more likely like we'll want to cut a release prior to R4B being ready, so I have a slight preference for coding up the logic to handle this case in the current PR (if its not too much trouble)

prb112:
QA: Configure a open: false for resource then run a request for an invalid Resource

d0roppe:
[...]

PR #3355, opened 2022-02-16
BulkData Export checks that the Resource Types exported are valid based on the fhir-model and not configurations #3083

prb112:
- Harden the Processing of _type with bulkdata
- Considers the supported designations in the Configuration
- There is a nuance captured with Lee's r4b branch where it improves the use of the FHIRConfigHelper


operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataExportUtil.java
```diff
@@ -103,7 +103,7 @@ public void checkExportPatientResourceTypes(List<String> resourceTypes) throws F
     public Set<String> getSupportedResourceTypes() {
         try {
             List<String> rts = FHIRConfigHelper.getSupportedResourceTypes();
```
lmsurpre, line 105: as paul alludes to in the PR description `FHIRConfigHelper.getSupportedResourceTypes()` is much improved in the `r4b` branch.  in the target branch of this PR, `main`, it just reads the resources config without considering whether the server is open (supports all types) or closed (supports only the listed ones).

I'm not sure if the logic below handles all the cases properly, but is the idea to wait on the updated handling from `FHIRConfigHelper.getSupportedResourceTypes()` for that?

operation/fhir-operation-bulkdata/src/test/java/com/ibm/fhir/operation/bulkdata/util/BulkDataImportUtilTest.java
```diff
@@ -155,6 +158,15 @@ public void testCheckAllowedTotalSizeForTenantOrSystemFail() throws FHIROperatio
         util.checkAllowedTotalSizeForTenantOrSystem(100);
     }
 
+    @Test//(expectedExceptions = { FHIROperationException.class })
```
lmsurpre, line 161: why is expectedExceptions commented out?
can it be removed?

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataExportUtil.java
lmsurpre: if so, I think thats fine with me but maybe we should leave a note somewhere..maybe in #3083 ?

operation/fhir-operation-bulkdata/src/test/java/com/ibm/fhir/operation/bulkdata/util/BulkDataImportUtilTest.java
prb112: Fixed, it should have been expected

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/util/BulkDataExportUtil.java
prb112: That's right, since the adapter is there, I think that's the long term answer. Rather than recreate in main

lmsurpre (approved):
LGTM

Issue #3318, opened 2022-02-09
Prefer header processing uses ';' instead of ',' and should support multiple Prefer headers


kmbarton423:
Working with the current pre-release of FHIR on main branch
Per the conformance documentation:  https://ibm.github.io/FHIR/Conformance

_The _sort, _count, _summary, _elements, and _total parameters may each only be specified once in a search. In lenient mode, only the first occurrence of each of these parameters is used; additional occurrences are ignored._

Received an error when trying to use multiple occurrences of the _sort, _count, _summary, _elements, and _total when setting Prefer handling=lenient via the header.  See attached screenshot.

UPDATE:  the issue only occurs when either multiple Prefer headers are set or when a single Prefer header contains multiple components.  currently the code tries to split the components of the Prefer header by `;` but the relevant rfc states it should be a `,` instead.

**To Reproduce**
Default or set 'true' fhir server config parameter 'fhirServer/core/allowClientHandlingPref'

Set the following headers for all requests (in this order):
* Prefer: return=representation
* Prefer: handling=lenient 

Attempt the following:
  GET [base]/Measure?_sort=_lastUpdated&_sort=status
  GET [base]/Measure?_count=1&_count=1000
  GET [base]/Measure?_summary=text&_summary=count
  GET [base]/Measure?_elements=status&_elements=name
  GET [base]/Measure?_total=none&_total=accurate

**Expected behavior**
Expected only the first occurrence of the parameters to be used successfully as stated above.
![2022-02-09 13_41_10-Insomnia - Localhost – search Measure sort](https://user-images.githubusercontent.com/69682345/153268408-ea8eb923-defd-4cce-872e-edae2f3b4f16.png)

**QA**
Test "Prefer" header with "handling=" and "return=" tokens in various permutations.

lmsurpre:
we're pretty sure it used to work as documented.
investigate to make sure what we understand whats happening, confirming:
1. handling=lenient is working as intented
2. code seems to be logical / make sense

if both those are working as implemented, it should be a doc-only update

prb112:
She's specifically trying some of the more restrictive search types. 
We verified she had that correct.

kmbarton423:
Confirmed using the single header with ";" works as expected.  Should ',' separator work as well?

tbieste:
Based on https://datatracker.ietf.org/doc/html/rfc7240:

[...]

PR #3356, opened 2022-02-16
Issue #3318 - Case insensitive check of HTTP header name

tbieste:
Issue #3318 - Case insensitive check of HTTP header name

### DIFF
--- a/fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
@@ -302,16 +302,23 @@ public class FHIRRestServletFilter extends HttpFilter {
      */
     private String getHeaderTokenValue(FHIRRequestContext context, String headerName, String tokenName) {
         String preferValue = null;
-        for (String headerValue : context.getHttpHeaders().getOrDefault(headerName, Collections.emptyList())) {
-            for (String headerValueElement : headerValue.split(",")) {
-                headerValueElement = headerValueElement.split(";")[0];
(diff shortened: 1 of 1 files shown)


fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
```diff
@@ -302,16 +302,23 @@ private void computeReturnPref(FHIRRequestContext context) throws FHIRException
...
-        for (String headerValue : context.getHttpHeaders().getOrDefault(headerName, Collections.emptyList())) {
-            for (String headerValueElement : headerValue.split(",")) {
-                headerValueElement = headerValueElement.split(";")[0];
-                String[] tokenAndValue = headerValueElement.split("=", 2);
-                if (tokenName.equals(tokenAndValue[0].trim())) {
-                    preferValue = "";
-                    if (tokenAndValue.length > 1) {
-                        preferValue = tokenAndValue[1].trim();
+        for (String curHeaderName : context.getHttpHeaders().keySet()) {
+            if (curHeaderName.equalsIgnoreCase(headerName)) {
```
lmsurpre, line 306: might be nice to have a test that exercises the case-insensitive flavor, but looks good to me

lmsurpre (approved):

fhir-server/src/main/java/com/ibm/fhir/server/filter/rest/FHIRRestServletFilter.java
tbieste: Yeah, I added integration tests that use various cases (lower/uppercase/etc.), though the case always showed up the same when it reached this code.  That was the best I could come up with.

Issue #3000, opened 2021-11-16
ImportPartitionAnalyzer not reporting infly rate on threshold


prb112:
**Describe the bug**
ImportPartitionAnalyzer not reporting infly rate on threshold

**Environment**
Which version of IBM FHIR Server? main

**To Reproduce**
1. Start up IBM FHIR Server with FileProvider pointing to the output folder
2. Execute curl 
```
curl --location --request POST 'https://localhost:9443/fhir-server/api/v4/$import' \
--header 'X-FHIR-TENANT-ID: default' \
--header 'Content-Type: application/fhir+json' \
--header 'X-FHIR-BULKDATA-PROVIDER: default' \
--header 'X-FHIR-BULKDATA-PROVIDER-OUTCOME: default' \
--header 'Authorization: Basic XXXX' \
--data-raw '{
    "resourceType": "Parameters",
    "id": "30321130-5032-49fb-be54-9b8b82b2445a",
    "parameter": [
        {
            "name": "inputSource",
            "valueUri": "https://my-server/source-fhir-server"
        },
        {
            "name": "inputFormat",
            "valueString": "application/fhir+ndjson"
        },
        {
            "name": "input",
            "part": [
                {
                    "name": "type",
                    "valueString": "AllergyIntolerance"
                },
                {
                    "name": "url",
                    "valueUrl": "r4_AllergyIntolerance.ndjson"
                }
            ]
        },
        {
            "name": "storageDetail",
            "valueString": "file"
        }
    ]
}'
```
3. Check the console and you should see in fly rate popup at 2000 resources.  However we use a border of 21, so you have to execute a LOT to hit that magic match up. 

Increments independent of the actual resources.
```
                // Add info to the object thats already in the list              importedResourceTypeInFlySummary.setNumOfProcessedResources(importedResourceTypeInFlySummary.getNumOfProcessedResources()
                        + adapter.getImportNumberOfFhirResourcesPerRead(null));
```

**Expected behavior**
Should report out on an interval. 

**Additional context**

d0roppe:
moving the issue back to in progress as the message does not calculate the rate of ingestion, but states it know it.

prb112:
QA: In fly rate should show with a non-zero number

d0roppe:
in-fly message now has the importRate
[...]

PR #3358, opened 2022-02-16
ImportPartitionAnalyzer not reporting infly rate on threshold #3000

prb112:
ImportPartitionAnalyzer - modify the infly rate reporting message and changed the message to report out Resources/second and report which  workitem is triggering the message.
Improve accuracy of the infly rate calculation

Sample Message: 

```
[INFO ] Import in-fly rate:
        [286/286/r4_AllergyIntolerance.ndjson/AllergyIntolerance]
        reportingThreshold=[2000] resources imported in 12610 milliseconds,
        ImportRate: [0.064] Resources/milliseconds
            Instance/Execution/File/ResourceType threshold milliseconds and
            rate.
```


lmsurpre (approved):
Not sure I totally get it, but the changes look ok to me

prb112:
Since it was so quick it was actually rounding to zero.

Issue #3016, opened 2021-11-19
Bulk data import with missing "id"


d0roppe:
**Describe the bug**
If a Resource "id" is not part of the bulk data import the resource is created with a generated ID, but the audit log entry does not have the Resource ID of the created resource. 
If the import file has Resource "meta" with "versionId"  the passed in version ID is added as the "version_id" in the audit record, not the "versionId" of 1 for the created resource.

Also on a import of the same resource over and over with a Resource "id" specified, it creates it the Resource the first time, and skips the subsequent imports. but the Audit record says it is "version_id": 2, but the database never is updated past version 1 of the Resource

**Environment**
IBM FHIR Server (main)

d0roppe:
with the latest change the Patient created has the patient ID in the audit record, and the Patient that already exists still has a audit entry, but it states the correct version of the patient, and that the import was skipped. `"description": "FHIR BulkData Update Resource Skipped request",`
Closing issue

PR #3359, opened 2022-02-16
Bulk Data Import with missing 'id' on Audit #3016

prb112:
- Switched the conditionalUpdate to skip and pass in the oldResource as
- they exactly match rather than the Meta updated version with the unused new version
- update the Create to ignore the Version when passed in
- Enable the Version on Update by using the Create with Version.
- Refactor the BulkAuditLogger to eliminate the dual resource usage

### DIFF
--- a/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/audit/BulkAuditLogger.java
+++ b/fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/audit/BulkAuditLogger.java
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright IBM Corp. 2021
+ * (C) Copyright IBM Corp. 2021, 2022
  *
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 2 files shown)


fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/audit/BulkAuditLogger.java
```diff
@@ -60,15 +60,15 @@ public boolean shouldLog() {
         return svc.isEnabled();
```
tbieste, line 60: Update copyright date.

tbieste (approved):

lmsurpre:
Only tangentially related to this PR but I'm wondering whether https://github.com/IBM/FHIR/pull/3314 will have an impact on the audit log message for our delete interactions

prb112:
> Only tangentially related to this PR but I'm wondering whether #3314 will have an impact on the audit log message for our delete interactions

It probably should

lmsurpre (commented):
Will change our audit log payload?
Probably fine either way, but wondering if we need to update our docs at all...like `attachments/content/data/resource_type` under https://ibm.github.io/FHIR/guides/FHIRServerUsersGuide#4111-cadf-audit-log-entry

prb112:
that's a good question, seems like an issue to confirm.

Issue #3038, opened 2021-11-24
Bulk Data S3Provider with Prefix Matrix's import status does not have correct count when the job is completed


d0roppe:
**Describe the bug**
When multiple files are imported using prefix matching on COS, the files are not summarized together.
So the bulk data status call does not have the correct count when the job is completed. 
**Environment**
IBM FHIR Server - main

**To Reproduce**

**Expected behavior**

**Additional context**

d0roppe:
Ran a test with the $import file from S3Provider with matching segments, now gives correct count after the job is complete and viewing the $bulkdata-status

PR #3361, opened 2022-02-16
Bulk Data S3Provider with Prefix Matrix's import status does not have  correct count when the job is completed #3038

prb112:
- Add an intermediate step to organize counts and then sort and then convert to ExitStatus


lmsurpre (approved):
I could probably use an explainer on this one.
I get that we're sorting the ExitStatusVals now but I'm not totally sure that I understand how that addresses the issue.

prb112:
> I could probably use an explainer on this one. I get that we're sorting the ExitStatusVals now but I'm not totally sure that I understand how that addresses the issue.

Yep... 

It's an issue when we use the prefix search on S3. 
We generate multiple partitions for the match 

for instance.... 

folder/Patient_
matches

folder/Patient_1
folder/Patient_2
folder/Patient_3

The reason we sort is to organize the ExitStatuses for the Partitions/Matrixed elements so we can uniformly add them together. 

Dag found it, and the only way I could organize was sorting.

Issue #3034, opened 2021-11-23
bulk data delete followed by status caused no output


d0roppe:
**Describe the bug**
Issuing a Delete bulk data status on a job.  then issuing a get bulk data status for the same job return code of 200, but no data in the response body just {}

**Environment**
IBM FHIR Server - main

**To Reproduce**
Issueing a delete bulk data status of a job getting back a 202 code, then changing the call to a delete for the same job, it returned a 200 return code but the response body was empty except {}

**Expected behavior**
If the status of a bulk data job returns a 200 return code it should have info in the response body.

prb112:
when a job is stopped, we don't return 202 when we restart it

Create an import

Stop the job curl --location --request PUT 'https://localhost:9443/ibm/api/batch/jobexecutions/406?action=stop'

curl --location --request GET 'https://localhost:9443/fhir-server/api/v4/$bulkdata-status?job=406' \
--header 'X-FHIR-TENANT-ID: default' \
--header 'Content-Type: application/fhir+json' 
{}

d0roppe:
Not able to verify this issue.  with a long running $export, and then issueing a Delete of that job, and then issuing a $bulkdata-status of that job, the status gets a 202, but it also seems to restart the job.   And then when the job is done, I get a 500 error on the $bulkdata-status.
If I issue delete job of a long running $export, and then issue the delete again.  It actually stops and deletes the job. and then the $bulkdata-status call gets a 404 not found as expected when the job is done.

prb112:
We can debug - points to a different issue.

PR #3362, opened 2022-02-16
bulk data delete followed by status caused no output #3034

prb112:
- missing null return, otherwise it hits an empty pollingresponse.

Unfortunately, this depends on calls to the batch framework underneath to test, and is not ideal for Unit or Integration Test.


tbieste (approved):
LGTM

lmsurpre (approved):

PR #3363, opened 2022-02-17
Patch ModelResolver to return Resource.id as FHIR.string

csandersdev:
We found while testing the EXM104 measure from the fhir401 connectathon repo that when the CQL tries to reference the Resource.id field (such as in `[Encounter] e return Last(Split(e.id,'/'))`) that we would get an error saying could not resolve function 
FHIRHelpers.ToString(java.lang.String). The core problem is that the IBM FHIR model implements Resource.id as java.lang.String and the HAPI model which is used as the reference for the CQL implementation defines Resource.id as a FHIR.String. The CQL translator automatically inserts a FHIRHelpers.ToString into the ELM whenever Resource.id is used and the only way around that would be to create a complete new modelinfo and I don't really want to do that, so I patched our ModelResolver to do the type conversion to FHIR String whenever an `id` property is referenced. 

Signed-off-by: Corey Sanders <corey.thecolonel@gmail.com>


lmsurpre (approved):
LGTM

prb112 (approved):

Issue #3003, opened 2021-11-17
Failed to execute the s3 access, check the s3 configuration is a bit misleading


prb112:
**Describe the bug**

```
[ERROR   ] Failed to execute the s3 access, check the s3 configuration
Failed to execute the s3 access, check the s3 configuration
```

```
[11/17/21, 15:34:12:028 UTC] 00000025 FHIRResource  E   Failed to execute the s3 access, check the s3 configuration
                                 com.ibm.fhir.exception.FHIROperationException: Failed to execute the s3 access, check the s3 configuration  [probeId=c0-a8-56-25-89f85884-d500-4728-abcc-ab39772696a4]
	at com.ibm.fhir.operation.bulkdata.util.BulkDataExportUtil.buildOperationException(BulkDataExportUtil.java:143)
	at com.ibm.fhir.operation.bulkdata.config.preflight.impl.S3Preflight.preflight(S3Preflight.java:75)
	at com.ibm.fhir.operation.bulkdata.ImportOperation.doInvoke(ImportOperation.java:78)
	at com.ibm.fhir.server.spi.operation.AbstractOperation.invoke(AbstractOperation.java:57)
	at com.ibm.fhir.server.util.FHIRRestHelper.doInvoke(FHIRRestHelper.java:1428)
	at com.ibm.fhir.server.resources.Operation.invoke(Operation.java:126)
	at com.ibm.fhir.server.resources.Operation$Proxy$_$$_WeldClientProxy.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at com.ibm.ws.jaxrs20.cdi.component.JaxRsFactoryImplicitBeanCDICustomizer.serviceInvoke(JaxRsFactoryImplicitBeanCDICustomizer.java:348)
	at com.ibm.ws.jaxrs20.server.LibertyJaxRsServerFactoryBean.performInvocation(LibertyJaxRsServerFactoryBean.java:641)
[...]

PR #3382, opened 2022-02-22
bulkdata preflight checks should verify output provider config #2709

prb112:
- Move the Client Generation to the Server (AmazonS3) 
- Added cos to the fhir-server/pom.xml 
- Enhanced preflight with a write check 
- Example warning that occurs with the operation now (storageProvider id/bad
bucketName)

```
{
    "id": "c0-a8-56-25-977e7573-20c9-4a0b-ba92-b07e640ddc1c",
    "issue": [
        {
            "severity": "error",
            "code": "exception",
            "diagnostics": "The bucket does not exist
&#39;minio/fhirbulkdataxx&#39;"
        }
    ]
}
```

And 

https://github.com/IBM/FHIR/issues/3003

- Relaxed the S3Preflight to support http connection with warning

```
        MSG: BulkData Request is using a provider that connects to an
        insecure url: http://bucket.bucket:9000/
```

- Improved the Response Handling to capture the ex and package it back up the stack

### DIFF
--- a/fhir-server/pom.xml
+++ b/fhir-server/pom.xml
@@ -169,6 +169,10 @@
             </exclusions>
         </dependency>
         <dependency>
+            <groupId>com.ibm.cos</groupId>
+            <artifactId>ibm-cos-java-sdk</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.testng</groupId>
             <artifactId>testng</artifactId>
             <scope>test</scope>
--- a/operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/impl/AbstractSystemConfigurationImpl.java
+++ b/operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/impl/AbstractSystemConfigurationImpl.java
@@ -72,7 +72,7 @@ public abstract class AbstractSystemConfigurationImpl implements ConfigurationAd
 
     private static final byte[] NDJSON_LINESEPERATOR = "\r\n".getBytes();
 
-    public static final int IMPORT_NUMOFFHIRRESOURCES_PERREAD = 20;
+    public static final int IMPORT_NUMOFFHIRRESOURCES_PERREAD = 50;
     public static final int IMPORT_INFLY_RATE_NUMOFFHIRRESOURCES = 2000;
 
     // The following are set on startup:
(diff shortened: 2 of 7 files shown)


fhir-server/pom.xml
```diff
@@ -168,6 +168,10 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>com.ibm.cos</groupId>
+            <artifactId>ibm-cos-java-sdk</artifactId>
```
lmsurpre, line 173: I think we talked about potentially going straight to the aws client (removing the ibm-cos dependency), but agree it makes more sense to stick with the ibm-cos to match what we currently have in bulkdata-webapp.
@punktilious heads up in case it conflicts with any changes you made on the s3 side.

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/impl/AbstractSystemConfigurationImpl.java
```diff
@@ -72,7 +72,7 @@
 
     private static final byte[] NDJSON_LINESEPERATOR = "\r\n".getBytes();
 
-    public static final int IMPORT_NUMOFFHIRRESOURCES_PERREAD = 20;
+    public static final int IMPORT_NUMOFFHIRRESOURCES_PERREAD = 50;
```
lmsurpre, line 75: is this default config value documented anywhere?
not actually sure it needs to be, just trying to make sure we keep docs up-to-date

prb112: It's always been hard coded. https://github.com/IBM/FHIR/blob/d5b8962ca2110afea41b4e4b78247be10e53c84d/operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/config/impl/AbstractSystemConfigurationImpl.java#L344 

20 had just felt low when I ran my tests (50 seems to be a decent chunk)

lmsurpre: ok, sounds good

lmsurpre (approved):
LGTM

fhir-server/pom.xml
lmsurpre: I also considered whether its worth trying to make this a shared dependency between fhir-server-webapp and bulkdata-webapp, but I don't think its worth the pain.  Better to just package it with each war.

prb112:
image
Adds the highlighted for 95M in size

PR #3383, opened 2022-02-23
Issue 3136 azure blob offload support

punktilious:


fhir-persistence-blob-app/README.md
```diff
@@ -0,0 +1,225 @@
+# Payload Offload for Azure Blob
+
+The IBM FHIR Server can be configured to store the resource payload records in an Azure Blob container. Each record is stored as a JSON string with UTF-8 encoding. The IBM FHIR Server relies on the blob service to compress/encrypt the data at rest. The blob service must apply the necessary security controls required when storing PHI, and connection to the service must use an encrypted transport.
+
+## Configurating the IBM FHIR Server to use Payload Offloading with Azure Blob
```
lmsurpre, line 5:
```suggestion
## Configuring the IBM FHIR Server to use Payload Offloading with Azure Blob
```

fhir-persistence-blob-app/README.md
```diff
@@ -0,0 +1,225 @@
...
+In the main `default/fhir-server-config.json`, configure the `fhirServer/persistence/factoryClassname` as shown below and add a `fhirServer/persistence/payload` block containing the connection information for each datasource you have defined under `fhirServer/persistence/datasources` (typically there is just one datasource called `default`). 
+```
+{
+    "__comment": "FHIR Server configuration",
+    "fhirServer": {
+        ...
+        "persistence": {
+            "factoryClassname": "com.ibm.fhir.persistence.blob.FHIRPersistenceJDBCBlobFactory",
+            "datasources": {
+                ...
```
lmsurpre, line 50: might be worth including the default dsid here just to drive it home:
```suggestion
                "default": {
                ...
```

lmsurpre: applies for section 4 as well

fhir-persistence-blob-app/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+When new resources are ingested, the class `com.ibm.fhir.persistence.blob.BlobStorePayload` logs a `FINE` message which can be seen in the Liberty Profile `trace.log` file:
+
+```
+Payload storage path: 97/my-blob-5/7/b0a2fb75-1ef8-45c8-b5d7-658350fb45cc
+```
+Where:
+| Value | Meaning |
+| ----- | ------- |
+|        97 | `resource_type_id` primary key from RDBMS `resource_types` table |
```
lmsurpre, line 188: still on the fence about whether we'd be better off just using the resource type name

fhir-persistence-blob-app/README.md
```diff
@@ -0,0 +1,225 @@
...
+| my-blob-5 | resource logical-id |
```
lmsurpre, line 189: just double-checking:  all the allowed chars for FHIR Resource.id are valid in an azure blob path, right?
```
[A-Za-z0-9\-\.]{1,64}
```

fhir-persistence-blob-app/README.md
```diff
@@ -0,0 +1,225 @@
...
+                        "containerName": "default-default"
+                    }
+                }
+            }
+        }
+    }
+}
+```
+
+Container names allowed by Azure Blob are more restrictive than IBM FHIR Server tenant and datasource names (for example `_` is not allowed). For this reason, the container name for each tenant and datasource must be specified in the `fhirServer/persistence/payload/connectionProperties/containerName` property.
```
lmsurpre, line 67: I found this one in their docs:
> The container name must be lowercase. For more information about naming containers, see [Naming and Referencing Containers, Blobs, and Metadata](https://docs.microsoft.com/en-us/rest/api/storageservices/Naming-and-Referencing-Containers--Blobs--and-Metadata).

Would probably be worth noting that here (along with the `_` one).

fhir-persistence-blob-app/pom.xml
```diff
@@ -0,0 +1,181 @@
...
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>fhir-examples</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>fhir-persistence-schema</artifactId>
+            <version>${project.version}</version>
+            <scope>provided</scope>
```
lmsurpre, line 29: what are we using from this?  found it curious that its marked 'provided'

fhir-persistence-blob-app/README.md
```diff
@@ -0,0 +1,225 @@
+# Payload Offload for Azure Blob
```
lmsurpre, line 1: great README!

fhir-persistence-blob-app/src/main/java/com/ibm/fhir/persistence/blob/app/CreateContainer.java
```diff
@@ -0,0 +1,67 @@
...
+        this.dsId = dsId;
+        this.dryRun = dryRun;
+    }
+
+    /**
+     * Create the container for the configured tenant and datasource pair
+     */
+    public void run() throws FHIRException {
+        // Set up the request context for the configured tenant and datastore
+        FHIRRequestContext.set(new FHIRRequestContext(tenantId, dsId));
```
lmsurpre, line 43: was surprised to see FHIRRequestContext here since I thought this utility was for use outside of the server.  is it just a convenience thing for holding the tenantId and datastoreId pair?

fhir-persistence-blob/src/test/java/com/ibm/fhir/persistence/blob/async/ReactorTests.java
```diff
@@ -0,0 +1,147 @@
...
+
+import reactor.core.publisher.BaseSubscriber;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+/**
+ * Tests using Flux and Mono which are used in the Azure SDK for async
+ * handling.
+ */
+public class ReactorTests {
```
lmsurpre, line 26: just a playground for this stuff or tests we actually want?

fhir-persistence-blob/src/main/java/com/ibm/fhir/persistence/blob/FHIRPayloadPersistenceBlobImpl.java
```diff
@@ -71,25 +75,55 @@ public PayloadPersistenceResponse storePayload(String resourceType, int resource
             int version, String resourcePayloadKey, List<String> elements) throws FHIRPersistenceException {
 
         logger.fine(() -> "readResource " + rowResourceTypeName + "[" + resourceTypeId + "]/" + logicalId + "/_history/" + version);
-        final BlobPropertyGroupAdapter config = getConfigAdapter();
         BlobReadPayload cmd = new BlobReadPayload(resourceTypeId, logicalId, version, resourcePayloadKey, elements, !PAYLOAD_COMPRESSED);
-        return cmd.run(resourceType, getBlobContainerClient());
+        try {
+            // synchronous read
+            return cmd.run(resourceType, getBlobManagedContainer()).get();
```
lmsurpre, line 81: is this synchronous readResource flavor still being used somewhere?

fhir-persistence-blob/src/main/java/com/ibm/fhir/persistence/blob/BlobStorePayload.java
```diff
@@ -45,20 +62,22 @@ public BlobStorePayload(int resourceTypeId, String logicalId, int version, Strin
...
+        }
+        final String blobPath = BlobPayloadSupport.getPayloadPath(resourceTypeId, logicalId, version, resourcePayloadKey);
+        logger.fine(() -> "Payload storage path: " + blobPath);
+        BlobAsyncClient bc = client.getClient().getBlobAsyncClient(blobPath);
+
+        // Reactor pipeline to first call create then perform the upload.
+        return bc.getAppendBlobAsyncClient()
+                .create(true)
+                .flatMap(item -> {
+                    return bc.getAppendBlobAsyncClient().appendBlockWithResponse(Flux.just(ioStream.wrap()), ioStream.size(), null, null);
```
lmsurpre, line 78: this bit looks a bit foreign to me...

fhir-persistence-blob/src/main/java/com/ibm/fhir/persistence/blob/BlobStorePayload.java
```diff
@@ -1,27 +1,44 @@
...
+import reactor.core.publisher.Flux;
+
 /**
  * DAO command to store the configured payload in the Azure blob
  */
 public class BlobStorePayload {
+    private static final Logger logger = Logger.getLogger(BlobStorePayload.class.getName());
+    
+    // The maximum size of a single append blob block call
+    private static final int MAX_APPEND_BLOB_BLOCK_SIZE = AppendBlobAsyncClient.MAX_APPEND_BLOCK_BYTES;
```
lmsurpre, line 27: I looked this up and it looks to be 4MB currently

fhir-persistence-blob/src/main/java/com/ibm/fhir/persistence/blob/BlobStorePayload.java
```diff
@@ -45,20 +62,22 @@ public BlobStorePayload(int resourceTypeId, String logicalId, int version, Strin
...
-        blobNameBuilder.append(resourcePayloadKey);
-        BlobClient bc = client.getClient().getBlobClient(blobNameBuilder.toString());
+    public CompletableFuture<PayloadPersistenceResult> run(BlobManagedContainer client) throws FHIRPersistenceException {
         
-        BlobParallelUploadOptions uploadOptions = new BlobParallelUploadOptions(BinaryData.fromBytes(ioStream.getRawBuffer()));
-        Response<BlockBlobItem> response = bc.uploadWithResponse(uploadOptions, client.getProperties().getTimeout(), null);
-        
-        return response;
+        if (ioStream.size() > MAX_APPEND_BLOB_BLOCK_SIZE) {
+            throw new FHIRPersistenceException("Resource payload size cannot exceed " + MAX_APPEND_BLOB_BLOCK_SIZE + " bytes");
```
lmsurpre, line 68: does this mean we can only support offload for resources <4 MiB?

fhir-persistence-blob/src/main/java/com/ibm/fhir/persistence/blob/BlobResourceScanner.java
```diff
@@ -0,0 +1,101 @@
...
+import com.azure.storage.blob.models.BlobItem;
+import com.ibm.fhir.persistence.jdbc.dao.api.ResourceRecord;
+
+/**
+ * Scans resource payload records which have been stored in the blob
+ * store. Used by the reconciliation process to identify records
+ * existing in the blob store which do not have a corresponding
+ * record in the RDBMS.
+ */
+public class BlobResourceScanner {
```
lmsurpre, line 26: wondering if this one should live in fhir-persistence-blog-app since its for the reconciliation process

fhir-persistence-blob/src/main/java/com/ibm/fhir/persistence/blob/BlobPayloadSupport.java
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+ 
+package com.ibm.fhir.persistence.blob;
+
+import com.ibm.fhir.persistence.jdbc.dao.api.ResourceRecord;
+
+/**
+ *
```
lmsurpre, line 12: javadoc

fhir-persistence-blob-app/src/main/java/com/ibm/fhir/persistence/blob/app/ReadBlobValue.java
```diff
@@ -0,0 +1,61 @@
...
+        FHIRRequestContext.set(new FHIRRequestContext(tenantId, dsId));
+
+        // Check to see if the container already exists, and if not, then 
+        // issue the create container command and wait for the response
+        BlobManagedContainer bmc = BlobContainerManager.getSessionForTenantDatasource();
+        try {
+            BlobAsyncClient bac = bmc.getClient().getBlobAsyncClient(blobName);
+            BinaryData result = bac.downloadContent().block();
+            String blobValue = new String(result.toBytes(), StandardCharsets.UTF_8);
+            System.out.println(blobValue);
```
lmsurpre, line 55: probably fine for a utility, but wanted to make sure you left this on purpose

fhir-persistence-blob-app/src/main/java/com/ibm/fhir/persistence/blob/app/PayloadReconciliation.java
```diff
@@ -0,0 +1,186 @@
...
+
+    // Tracking how many resource versions we process
+    private long totalProcessed = 0;
+    
+    // Simple map cache of resource type id to name
+    private final Map<Integer,String> resourceTypeMap = new HashMap<>();
+
+    // Stop scanning after we hit this number of seconds
+    private final int maxScanSeconds;
+    /**
```
lmsurpre, line 60: super minor
```suggestion

    /**
```

fhir-persistence-blob-app/src/main/java/com/ibm/fhir/persistence/blob/app/PayloadReconciliation.java
```diff
@@ -0,0 +1,186 @@
...
+                        logger.fine(getLogRecord(record, "CHECK"));
+                    }
+                    
+                    // Check that we have the resource in the RDBMS configured for
+                    // this tenant
+                    try {
+                        ResourceExistsDAO dao = new ResourceExistsDAO(this.resourceTypeMap, 
+                            record.getResourceTypeId(), record.getLogicalId(), record.getVersion(), 
+                            record.getResourcePayloadKey());
+                        if (dao.run(c)) {
```
lmsurpre, line 123: i assume that deleting one-at-a-time is fine since we think this should be a very rare condition?

fhir-persistence-blob-app/README.md
punktilious: Oof!

fhir-persistence-blob-app/README.md
punktilious: done

fhir-persistence-blob-app/README.md
punktilious: also added a link to the docs

fhir-persistence-blob-app/README.md
punktilious: thanks!

fhir-persistence-blob-app/README.md
punktilious: The utility now makes it easy to map between the two, so it really becomes an implementation detail. These keys are part of an index space in the blob service, so it makes sense to keep them small by avoiding the repetition of large strings.

fhir-persistence-blob-app/README.md
punktilious: Almost. There is one issue: https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#blob-names

> Avoid blob names that end with a dot (.), a forward slash (/), or a sequence or combination of the two. No path segments should end with a dot (.).

FHIR resource names can end with a dot (although in practice this is unlikely out of convention). Some sort of escape encoding will be required to support this. The escape character can be one that isn't a valid FHIR id character.

fhir-persistence-blob-app/src/main/java/com/ibm/fhir/persistence/blob/app/Main.java
```diff
@@ -0,0 +1,229 @@
...
+                }
+                break;
+            case "--reconcile":
+                this.reconcile = true;
+                break;
+            case "--create-container":
+                this.createContainer = true;
+                break;
+            case "--confirm":
+                this.dryRun = false;
```
tbieste, line 106: Since `--confirm` and `--dry-run` both affect the same value, it would be good to explain what happens if both are specified (especially if order matters)

fhir-persistence-blob-app/src/main/java/com/ibm/fhir/persistence/blob/app/PayloadReconciliation.java
```diff
@@ -0,0 +1,186 @@
...
+    // Stop scanning after we hit this number of seconds
+    private final int maxScanSeconds;
+    /**
+     * Public constructor
+     * @param tenantId
+     * @param dsId
+     * @param dbProperties
+     * @param dbType
+     * @param dryRun
+     */
```
tbieste, line 67: javadoc: Several methods in this class have params not matching those method's javadoc params.

fhir-persistence-blob-app/src/main/java/com/ibm/fhir/persistence/blob/app/PayloadReconciliation.java
```diff
@@ -0,0 +1,186 @@
...
+                    
+                    // Check that we have the resource in the RDBMS configured for
+                    // this tenant
+                    try {
+                        ResourceExistsDAO dao = new ResourceExistsDAO(this.resourceTypeMap, 
+                            record.getResourceTypeId(), record.getLogicalId(), record.getVersion(), 
+                            record.getResourcePayloadKey());
+                        if (dao.run(c)) {
+                            // Found the record, so log it
+                            logger.info(getLogRecord(record, "OK"));
```
tbieste, line 125: Could OK be logged at lower log level, so only ORPHAN records are logged, if desired?

fhir-persistence-blob/src/main/java/com/ibm/fhir/persistence/blob/BlobDeletePayload.java
```diff
@@ -44,23 +44,17 @@ public BlobDeletePayload(int resourceTypeId, String logicalId, int version, Stri
...
-        blobNameBuilder.append(resourcePayloadKey);
-        BlobClient bc = client.getClient().getBlobClient(blobNameBuilder.toString());
-
+        final String blobPath = BlobPayloadSupport.getPayloadPath(resourceTypeId, logicalId, version, resourcePayloadKey);
+        
+        BlobAsyncClient bc = client.getClient().getBlobAsyncClient(blobPath);
         try {
-            Response<Void> response = bc.deleteWithResponse(DeleteSnapshotsOptionType.INCLUDE, null, client.getProperties().getTimeout(), null);
-            return response;
+            // TODO return future so we can avoid waiting for response here
```
tbieste, line 51: Is the intent to address this TODO now, or in the future?

fhir-persistence-blob/src/main/java/com/ibm/fhir/persistence/blob/BlobPropertyGroupAdapter.java
```diff
@@ -24,7 +24,7 @@
     public static final String PROP_CONNECTION_STRING = "connectionString";
```
tbieste, line 24: Update copyright date.

fhir-persistence-blob/src/main/java/com/ibm/fhir/persistence/blob/BlobResourceScanner.java
```diff
@@ -0,0 +1,101 @@
...
+
+    // A function used to process a page of blob items
+    private final Consumer<List<ResourceRecord>> pageHandler;
+
+    /**
+     * Public constructor
+     * @param marker the Azure Blob marker token from which to start the scan
+     * @param pageHandler
+     */
+    public BlobResourceScanner(String continuationToken, Consumer<List<ResourceRecord>> pageHandler) {
```
tbieste, line 40: java: Some methods in this class have params that do not match the javadoc params.

fhir-persistence-cassandra/src/main/java/com/ibm/fhir/persistence/cassandra/reconcile/CqlScanResources.java
```diff
@@ -24,6 +24,7 @@
 import com.datastax.oss.driver.api.querybuilder.select.Select;
 import com.datastax.oss.driver.api.querybuilder.select.Selector;
 import com.ibm.fhir.persistence.exception.FHIRPersistenceException;
+import com.ibm.fhir.persistence.jdbc.dao.api.ResourceRecord;
```
tbieste, line 27: Update copyright date.

fhir-persistence/src/main/java/com/ibm/fhir/persistence/payload/FHIRPayloadPersistence.java
```diff
@@ -6,9 +6,12 @@
```
tbieste, line 6: Update copyright date.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -1452,32 +1458,114 @@ public MultiResourceResult history(FHIRPersistenceContext context, Class<? exten
...
-                ResourceResult<? extends Resource> resourceResult = convertResourceDTOToResourceResult(resourceDTO, resourceType, elements, includeResourceData);
-                
-                // Check to make sure we got a Resource if we asked for it and expect there to be one
+            if (this.payloadPersistence != null) {
+                // With offloading, we can use async reads to fetch the payloads in parallel, making this
+                // significantly quicker for large bundles (search, history etc)
+                fetchPayloadsForDTOList(resourceResults, resourceDTOList, resourceType, elements, includeResourceData);
+            } else {
+                for (com.ibm.fhir.persistence.jdbc.dto.Resource resourceDTO : resourceDTOList) {
+                    // TODO Linear fetch of a large number of resources will extend response times. Need
```
tbieste, line 1467: Is this TODO intended to be addressed now or in the future?

fhir-persistence-blob-app/README.md
punktilious: Added encode/decode for the offload blob path so any `.` is replaced with a `#` (the latter not being a valid character for FHIR logical identifiers, making it a simple replace without any complicated escaping logic.

fhir-persistence-blob-app/src/main/java/com/ibm/fhir/persistence/blob/app/CreateContainer.java
punktilious: Yeah, we need to use it for multi-tenant/datasource support because I'm using the same class to manage connections to the blob storage service - which hides the logic to make sure you use the correct connection.

fhir-persistence-blob/src/test/java/com/ibm/fhir/persistence/blob/async/ReactorTests.java
punktilious: Playground...but I want to keep this because it's useful from an educational perspective as well as making sure if you change anything, you can check the fundamentals work as you expect.

fhir-persistence-blob/src/main/java/com/ibm/fhir/persistence/blob/FHIRPayloadPersistenceBlobImpl.java
punktilious: Yes, it's used from convertResourceDTO which gets called when we need the resource right away. For example, UPDATE requires that we read the current value of the resource so that we can emit the necessary before-update events. Same for PATCH etc. This makes these interactions take a little longer (because now we have to read from the database, then read the blob) but overall it shouldn't impact scalability. There are some tricks that we could do in the future. We could potentially read-ahead the resource payload before we hit the RDBMS, then check the resourcePayloadKey matches. But it actually requires two reads to the blob service (the first one being a listBlobs call) so I don't think we'd gain much in the end (the RDBMS read itself is very quick anyway).

fhir-persistence-blob/src/main/java/com/ibm/fhir/persistence/blob/BlobStorePayload.java
punktilious: Yeah, the async API stuff takes a bit of getting used to. I'll add a comment to explain what's going on here.

fhir-persistence-blob/src/main/java/com/ibm/fhir/persistence/blob/BlobStorePayload.java
punktilious: I've added a note on this limitation to the README.

fhir-persistence-blob/src/main/java/com/ibm/fhir/persistence/blob/BlobStorePayload.java
punktilious: Correct. Note added to README

fhir-persistence-blob/src/main/java/com/ibm/fhir/persistence/blob/BlobPayloadSupport.java
punktilious: fixed

fhir-persistence-blob/src/main/java/com/ibm/fhir/persistence/blob/BlobResourceScanner.java
punktilious: moved

fhir-persistence-blob-app/src/main/java/com/ibm/fhir/persistence/blob/app/PayloadReconciliation.java
punktilious: updated anyway

fhir-persistence-blob-app/src/main/java/com/ibm/fhir/persistence/blob/app/PayloadReconciliation.java
punktilious: I was thinking we'd do this at some point in the future, but now I'm thinking it's probably worth getting this into the next release.

fhir-persistence-blob-app/src/main/java/com/ibm/fhir/persistence/blob/app/Main.java
punktilious: I'll add to the README. I added `--confirm` as an extra requirement because I wanted the `--dry-run` to be the default (because reconciliation can delete stuff).

punktilious: Perhaps I'll rename `--confirm` to `--no-dry-run` which might make it a little easier. Last one wins.

punktilious: Changed to --no-dry-run and updated the README to describe the behavior.

fhir-persistence-blob-app/src/main/java/com/ibm/fhir/persistence/blob/app/PayloadReconciliation.java
punktilious: fixed

fhir-persistence-blob-app/src/main/java/com/ibm/fhir/persistence/blob/app/PayloadReconciliation.java
punktilious: thanks for the reminder - had thought about this, but was using info for dev.

fhir-persistence-blob/src/main/java/com/ibm/fhir/persistence/blob/BlobDeletePayload.java
punktilious: it's a future enhancement thing, should we need it based on performance results/use cases etc.

fhir-persistence-blob/src/main/java/com/ibm/fhir/persistence/blob/BlobPropertyGroupAdapter.java
punktilious: thanks

fhir-persistence-blob/src/main/java/com/ibm/fhir/persistence/blob/BlobResourceScanner.java
punktilious: fixed

fhir-persistence-cassandra/src/main/java/com/ibm/fhir/persistence/cassandra/reconcile/CqlScanResources.java
punktilious: ok

fhir-persistence/src/main/java/com/ibm/fhir/persistence/payload/FHIRPayloadPersistence.java
punktilious: ok

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
punktilious: comment no longer applied. removed

fhir-persistence-blob-app/src/main/java/com/ibm/fhir/persistence/blob/app/PayloadReconciliation.java
punktilious: done; definitely quicker

fhir-persistence-blob/src/test/java/com/ibm/fhir/persistence/blob/async/ReactorTests.java
lmsurpre: works for me

fhir-persistence-blob-app/src/main/java/com/ibm/fhir/persistence/blob/app/CreateContainer.java
lmsurpre: got it

fhir-persistence-blob-app/README.md
lmsurpre: actually, maybe its worth mentioning that in the README here...just in case anyone needs to work with the resources in azure blob directly?

fhir-persistence-blob-app/README.md
```diff
@@ -0,0 +1,258 @@
...
+4. Configure payload offloading in each tenant-specific `fhir-server-config.json` file;
+5. Create the container using the Azure Blob user interface, the Azure Blob API or the IBM FHIR Server fhir-persistence-blob-app-*-cli.jar tool;
+6. (Optional) Periodically run the reconciliation tool to remove any orphan resource records from the offload datastore.
+
+Take the following restrictions into account:
+
+1. Payload offloading is enabled at the server level. Offloading cannot be enabled/disabled on a per-tenant basis. If you want to support offloading for just one tenant, use a different IBM FHIR Server instance;
+2. Payload offloading must be configured prior to ingesting any resource data;
+3. Payload offloading must not be disabled after resource data has been ingested;
+4. When payload offloading is enabled, FHIR resources are limited to 4 MiB in size when rendered as uncompressed JSON. If this limit is too small, please raise an issue in the IBM FHIR Server [repository](https://github.com/IBM/FHIR/issues) with a description of your use case.
```
lmsurpre, line 21: I'm thinking we should have an open issue to track this, even if we decide not to support it for our 5.0

fhir-persistence-blob/src/main/java/com/ibm/fhir/persistence/blob/BlobStorePayload.java
lmsurpre: ok.  so if the resource payload does exceed this, I presume we'll fail the transaction with a 500 error?
if we're treating this as a "documented limitation", maybe we'd want a 400 error instead with an OperationOutcome that has an Issue with a type either "too-costly" or "business-rule"?

fhir-persistence-blob-app/README.md
punktilious: done

fhir-persistence-blob/src/main/java/com/ibm/fhir/persistence/blob/BlobStorePayload.java
punktilious: Now returns a 413 resource too large

fhir-persistence-blob-app/README.md
punktilious: issue #3414

lmsurpre (approved):
Excited to see this in action. LGTM

tbieste (approved):

Issue #2549, opened 2021-06-24
BulkData Import Operation: Switch S3 and File to use Byte Specific read locations


prb112:
**Is your feature request related to a problem? Please describe.**
BulkData Import Operation: Switch S3 and File to use Byte Specific read locations

Bulk Import for S3, File use a read of the object with a skip of a number of lines previously read. 
E.g. GetObject with BufferedReader skip lines
E.g. Files.bufferedReader with skip lines

There is a more efficient way to start reading the Object with a Range / Byte location
This should improve the performance/read speed of the Import operation

**Describe the solution you'd like**
Change the S3Provider to use https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/GetObjectRequest.html#withRange-long-long-
And the FileProvider to use Range

**Describe alternatives you've considered**
doing nothing and leaving the existing approach.

**Acceptance Criteria**

1.  
   GIVEN an Import from File,S3
    AND a decent file 34M AllergyIntollerance exists at the location
   WHEN an Import is executed
    AND the import is executed with the 34M file
   THEN the import should succeed and read fewer lines during the import

- Use the logging to confirm that there are fewer lines read

**Additional context**
n/a

prb112:
I shared with the team that this is a risky PR to merge. There are a lot of edge cases when considering single, double byte processing and using a sliding window to retrieve the data.

PR #3384, opened 2022-02-23
BulkData Import Operation: Switch S3 and File to use Byte Specific read locations #2549

prb112:
- *FILE* Provider
- Convert to use SeekableByteChannel with FileProvider
- Enables skipping to a byte versus read and discard buffer approach
- Use an upper bound read BLOCK to 512K
- Improved Read Performance from 33 Resources per second to 66 per
second in local tests
- Increase Import READ from a block of 20 to 50 (matrixes with block size)

- *S3* Provider
- Switch to retrieve bytes by range
- Added tests to test CountInputStreamReader to test
	- Blocks of 32K buffered while using readLine
	- Multiple Lines
	- Stripping CR from the Input
       - Establish Drain process for S3ObjectInputStream before closing it
	- Implement Range based retrieval in the S3Provider
	- Added a Test Provider Main to suck content down from S3
	- And ImportOperationTest for a decent sized resource and test data (patient-large.ndjson - 2 patient resources)


lmsurpre:
@prb112 where did [fhir-server-test/src/test/resources/testdata/import-operation/patient-large.ndjson](https://github.com/IBM/FHIR/pull/3384/files#diff-2feff63e2b0f30792517e10497a405d8859378257485dcd4e40ab1810e48631c) come from?
is that a synthea-generated thing or something we came up with?

prb112:
> @prb112 where did patient-large.ndjson come from? is that a synthea-generated thing or something we came up with?

I built it off one of our examples, I needed something larger than 512K

lmsurpre (approved):
LGTM, though I think it needs substantial testing

prb112:
> LGTM, though I think it needs substantial testing

Absolutely.. it should be run through a regression bucket. Note, it does follow the same pattern as Azure

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/context/BatchContextAdapter.java
```diff
@@ -133,6 +133,7 @@ public BulkDataContext getStepContextForImportChunkReader() {
         context(ctx);
```
tbieste, line 133: Update copyright date.

prb112: I have it in my other open PR. I'm going to defer this, as it'll fix when that is merged.

prb112: PR https://github.com/IBM/FHIR/pull/3393 is the one that has it

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/FileProvider.java
```diff
@@ -41,9 +44,10 @@
 
     private static final Logger logger = Logger.getLogger(FileProvider.class.getName());
 
+    private static final long READ_BLOCK_OPT = 524288L;
```
punktilious, line 47: Any rationale for picking this size? If so, might be worth a comment

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/FileProvider.java
```diff
@@ -79,43 +83,104 @@ public long getSize(String workItem) throws FHIRException {
...
-                if (br == null) {
-                    br = Files.newBufferedReader(Paths.get(getFilePath(workItem)));
-                }
-                // Imports must only skip lines if the size is higher, not equal to
-                // otherwise we start skipping the last line to skip.
-                for (int i = 0; i < numOfLinesToSkip; i++) {
-                    line++;
-                    br.readLine(); // We know the file has at least this number.
-                }
+            try (RandomAccessFile raf = new RandomAccessFile(Paths.get(getFilePath(workItem)).toFile(), "r")) {
```
punktilious, line 86: Random access might cause trouble depending on how the file-system is provided.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/FileProvider.java
```diff
@@ -79,43 +83,104 @@ public long getSize(String workItem) throws FHIRException {
...
+            if (r == -1) {
+                // End-of-stream
+                return null;
+            }
+            boolean read = true;
+            StringBuilder builder = new StringBuilder();
+            int lineLength = 0;
+
+            // Protect against attacks with a max line length (and max size we support in the db).
+            while (read && lineLength < MAX_LENGTH_PER_LINE) {
```
punktilious, line 158: shouldn't there be an IOException if we hit MAX_LENGTH_PER_LINE?

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/FileProvider.java
```diff
@@ -79,43 +83,104 @@ public long getSize(String workItem) throws FHIRException {
...
+                } else if (r == LF) {
+                    // \n  case
+                    length++;
+                    read = false;
+                } else {
+                    builder.append((char) r);
+                    length++;
+                    // We only need to account for line length here.
+                    lineLength++;
+                    r = this.read();
```
punktilious, line 170: Could be very slow unless the underlying InputStream is buffered.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/FileProvider.java
```diff
@@ -79,43 +83,104 @@ public long getSize(String workItem) throws FHIRException {
...
                 }
             }
         } catch (Exception e) {
             throw new FHIRException("Unable to read from Local File", e);
         }
     }
 
+    /**
+     * This specialized class enables a BufferedReader.readLine like implementation
+     * without the buffering as there is a local file with direct read and seek.
```
punktilious, line 129: You still need to use buffering to avoid hitting the file-system layer too many times

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/FileProvider.java
```diff
@@ -79,43 +83,104 @@ public long getSize(String workItem) throws FHIRException {
...
+     */
+    private static class CountInputStreamReader extends InputStreamReader {
```
punktilious, line 131: What about multi-byte characters? I think this needs to be based of a Reader, not an InputStream.

punktilious, line 131: I think this is broken for multibyte characters. The seek is based on a byte offset on the stream, but you are counting the number of characters read by the reader. The two are not equivalent.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/FileProvider.java
```diff
@@ -79,43 +83,104 @@ public long getSize(String workItem) throws FHIRException {
...
+                            continueRead = false;
+                        }
                     }
-                    resourceStr = br.readLine();
-                }
-            } finally {
-                if (br != null) {
-                    br.close();
+                    // The Channel enables us to skip the future seek.
+                    this.transientUserData.setCurrentBytes(this.transientUserData.getCurrentBytes() + reader.getLength());
```
punktilious, line 119: reader length is the number of characters read, not the number of bytes read

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
```diff
@@ -186,11 +203,23 @@ public ListObjectsV2Result getListObject(String prefix, String continuationToken
...
+
+    /**
+     * Get Size enables multiple use case beyond the default of getSize
+     * @param c
+     * @param bn
+     * @param workItem
+     * @return
+     * @throws FHIRException
+     */
+    public long getSize(AmazonS3 c, String bn, String workItem) throws FHIRException {
```
punktilious, line 217: What units for size? Bytes or characters?

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/FileProvider.java
prb112: Yes, that's right, it does speed up the import in this case. The downside of it is actually the same for us.  The prior code used read-to-line which is inefficient.  Some file systems use this approach of scan to location.  I think it's a nominal difference, and in most of the FileProvider usecases it's using the local file system on a developer machine.  

Let me know if there further downstream actions/preferences.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/FileProvider.java
prb112: Yes, I tried this with a few difference sizes.  Let me add an implNote

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/FileProvider.java
prb112: This code is based on the BufferedReader. It follows a similar pattern.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/FileProvider.java
prb112: I'll make an implnote, I chose not to do that, rather deferring to the downstream invalid parse.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/S3Provider.java
prb112: bytes - added implNote

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/FileProvider.java
prb112: Do we really support DBCS? I've tried with unicode 4byte and 3byte and 2byte.  It works well.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/FileProvider.java
prb112: I'll clarify

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/FileProvider.java
prb112: I added BufferedInputStream inbetween the Inputstream and Count

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/FileProvider.java
prb112: I'll add a custom test to show it works properly

prb112:
I rewrote the processing of multiple bytes and added a fairly complicated test related to it. 
If the PR passes, I'd probably recommend carefully accepting the PR post next release. 
It requires extensive testing, and I will not be available to support any necessary updates.

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/provider/impl/FileProvider.java
prb112: Added the tests for Files/S3 - looks great in my testing of the stream readers.

lmsurpre:
We decided to hold off on this one for now.  Might reopen after our 4.11

Issue #3385, opened 2022-02-23
ParametersUtil should be more test-friendly


lmsurpre:
**Is your feature request related to a problem? Please describe.**
Today, ParametersUtil is a set of static util methods.  When the class is first loaded, we iterate over all the tenant configurations in the configured "configDirectory" and apply search parameter filtering for each one.

This works fine for the server, but it makes this hard to use in unit tests.
For example, `fhir-search` and `fhir-persistence` each have their own set of "test tenants", but ParametersUtil will only iterate over one of the two (whichever one runs first in the current JVM).

To work around the issue, I updated the class to support loading config from newly added tenant configs "on-the-fly".
However, for existing tenants, it will use the cached copy of the config without any regard to an updated "configDirectory" setting.
Therefor, we have to ensure that there is no overlap in the test tenants across the different modules, which is error prone.

**Describe the solution you'd like**
1. Refactor ParametersUtil and SearchUtil to avoid reliance on static initialization.
2. Inject the ParametersUtil instance into the SearchUtil instance (e.g. from our ServletContextListener or via CDI if we get to #2687)

**Describe alternatives you've considered**

**Acceptance Criteria**

1.  
   GIVEN [a precondition]
    AND [another precondition]
   WHEN [test step]
    AND [test step]
   THEN [verification step]
    AND [verification step]

**Additional context**
The other part of this issue is that our config is all dependent on a particular directory layout on disk.  We have #1318 for that part of it.

Issue #1318, opened 2020-07-06
Support multiple FHIRConfiguration providers


punktilious:
**Is your feature request related to a problem? Please describe.**
The FHIR server configuration is read from JSON files stored on a local file-system. For container/cloud environments, it would be useful to support other (possibly pluggable) strategies for obtaining this configuration (from JDBC or REDIS for example).

**Describe the solution you'd like**
Refactor FHIRConfiguration so that reading from a file-system is one type of "configuration provider". Support other types of provider, allowing them to be plugged in (perhaps using ServiceLoader).

Currently configurations are JSON and typically hand-crafted. Some tooling would be needed to support these alternate data-sources. One useful tool would be something to load an existing tree of fhir-server-config.json files into a database. If the database schema were to support versioning, it would make rollback much easier, as well as rolling update/continuous availability scenarios where a change might only be propagated to a subset of nodes.

**Describe alternatives you've considered**
In a container environment, the configuration can be provided by mounting a volume.

**Additional context**
None.

lmsurpre:
A potential consideration for if/when we get to this one:  https://github.com/IBM/FHIR/issues/1865

lmsurpre:
One thought I had on this one is that we could leverage microprofile-config.
It would be interesting to see if we could shim our current config / ConfigHelper stuff to their "ConfigSource" abstraction.
One thing we'd need to work out is how to handle per-tenant config.

lmsurpre:
Note:  we do have at least one related interface already in fhir-config, though I think its only used from the JDBC PL impl (and only in a couple spots):
```
/**
 * Allows access to the FHIR server configuration to be hidden behind an adapter,
 * decoupling the consumer from file-based FHIRConfiguration stuff.
 */
public interface FHIRConfigProvider {

    /**
     * Get the named PropertyGroup
     * @param pgName
     * @return
     */
    PropertyGroup getPropertyGroup(String pgName);
}
```

PR #3387, opened 2022-02-24
issue #3385 - ParametersUtil and CompartmentUtil init changes

lmsurpre:
I made the constructors public and now we initialize the underlying maps
on construction instead of during static initialization.

Now that SearchUtil is not static, we must pass the instance down to the JAX-RS resources, the FHIRRestHelper, and the FHIRPersistence impl.
This is a breaking change for anyone with their own FHIRPersistenceFactory or FHIRPersistence implementation.

We also have numerous custom operation implementations that depend on SearchUtil, so now
the AbstractOperation.doInvoke takes an additional SearchUtil parameter.
This is a breaking change for anyone with their own extended operation implementation.


fhir-search/src/main/java/com/ibm/fhir/search/compartment/CompartmentCache.java
```diff
@@ -14,11 +14,13 @@
 import java.util.stream.Collectors;
 
 /**
- * Compartment Cache is a localized class to cache the compartment information and provide helper methods to add to the
- * cache.
+ * Information about a single compartment type.
```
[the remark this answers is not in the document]
lmsurpre: i don't think copyright update is *required* for comment-only change, but I went ahead and updated it anyway

fhir-search/src/main/java/com/ibm/fhir/search/compartment/ResourceCompartmentCache.java
```diff
@@ -15,12 +15,13 @@
 import java.util.stream.Collectors;
 
 /**
- * Resource Compartment Cache is a localized class to cache the compartment information for
- * a specific resource type
+ * Information about a specific resource type and which compartments it can be within.
```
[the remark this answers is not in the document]
lmsurpre: i don't think copyright update is *required* for comment-only change, but I went ahead and updated it anyway

build/derby-bootstrap.sh
```diff
@@ -26,13 +26,13 @@ java -jar ${WORKSPACE}/fhir-persistence-schema/target/fhir-persistence-schema-*-
   --update-schema
 java -jar ${WORKSPACE}/fhir-persistence-schema/target/fhir-persistence-schema-*-cli.jar \
   --db-type derby --prop db.database=${DB_LOC}/profile --prop db.create=Y \
-  --prop resourceTypes=Patient,Group,Practitioner,PractitionerRole,Person,RelatedPerson,Organization,Location,Observation,MedicationAdministration,StructureDefinition,ElementDefinition,CodeSystem,ValueSet,Encounter,Condition,MedicationRequest,Coverage,ServiceRequest,CarePlan,CareTeam,Claim,DiagnosticReport,ExplanationOfBenefit,Immunization,Procedure,Medication,Provenance,Consent \
+  --prop resourceTypes=Patient,Group,Practitioner,PractitionerRole,Person,RelatedPerson,Organization,Location,Observation,MedicationAdministration,StructureDefinition,ElementDefinition,CompartmentDefinition,CodeSystem,ValueSet,Encounter,Condition,MedicationRequest,Coverage,ServiceRequest,CarePlan,CareTeam,Claim,DiagnosticReport,ExplanationOfBenefit,Immunization,Procedure,Medication,Provenance,Consent \
```
punktilious, line 29: Do you think it's worth factoring out the common set of resource types into a variable?

punktilious: I also wonder if we should allow the resource types list to come from a file, which we could then use in multiple places. Minor.

build/migration/bin/2_previous-integration-test.sh
```diff
@@ -1,19 +1,19 @@
...
 ###############################################################################
 
 set -ex
 
 run_tests(){
-    # The integration tests may be overriden completely, or fall through to the default. 
+    # The integration tests may be overriden completely, or fall through to the default.
     migration="${1}"
 
     if [ ! -z "${migration}" ] && [ -f "build/migration/${migration}/2_previous-integration-test.sh" ]
```
punktilious, line 15: I know it's not part of this PR, but we should try and use `[[ ... ]]` for tests because they are then evaluated within the shell.

build/migration/bin/2_previous-integration-test.sh
```diff
@@ -34,10 +34,34 @@ run_tests(){
...
+    fhir_base_url=https://localhost:9443/fhir-server/api/v4
+    problem_types=CompartmentDefinition
+    for r in $(curl -k -u fhiruser:change-password "${fhir_base_url}?_type=${problem_types}&_count=1000" | jq -r '.entry[].fullUrl'); do
+      echo "erasing ${r}"
+      curl -i -k -u fhiruser:change-password -H "Content-Type: application/json" ${r}'/$erase' -d '
+        {
+          "resourceType": "Parameters",
+          "parameter": [
+            {
+              "name": "patient",
```
punktilious, line 52: what relationship does patient `test` have with CompartmentDefinition? Could use a comment to indicate how this value relates to the code.

punktilious: How are errors handled?

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/patient/PatientExportPartitionMapper.java
```diff
@@ -36,6 +36,8 @@
     @Inject
     JobContext jobCtx;
 
+    private static final CompartmentUtil compartmentHelper = new CompartmentUtil();
```
punktilious, line 39: Not a proper singleton pattern, but I assume it doesn't matter if multiple copies exist.

fhir-search/src/test/java/com/ibm/fhir/search/parameters/ParametersUtilTest.java
```diff
@@ -39,14 +31,17 @@
...
+    CompartmentUtil compartmentHelper = new CompartmentUtil();
+    ParametersUtil parametersHelper = new ParametersUtil(compartmentHelper);
+
     @Test
     public void testGetAllSearchParameters() throws IOException {
         // Tests JSON
         Set<SearchParameter> params = ParametersUtil.getAllSearchParameters();
         assertNotNull(params);
         // Intentionally the data is captured in the bytearray output stream.
         try (ByteArrayOutputStream outBA = new ByteArrayOutputStream(); PrintStream out = new PrintStream(outBA, true);) {
```
punktilious, line 43: should specify the charset encoding here, otherwise this could be affected by the system default charset encoding, which could cause the test to fail.

fhir-search/src/test/java/com/ibm/fhir/search/reference/ReferenceUtilTest.java
```diff
@@ -196,49 +197,49 @@ public void testGetBaseUrlFromBundleInvalid2() throws FHIRSearchException {
     }
 
     @Test
-    public void testGetServiceBaseUrl() throws FHIRSearchException {
+    public void testGetServiceBaseUrl() throws FHIRException {
         final String fullUrl = "https://localhost:9443/fhir-server/api/v4/Observation/17546b5a5a9-872ecfe4-cb5e-4f8c-a381-5b13df536f87";
-        FHIRRequestContext context = new FHIRRequestContext();
+        FHIRRequestContext context = new FHIRRequestContext("default");
```
punktilious, line 202: TENANT_DEFAULT constant?

fhir-server-test/src/test/java/com/ibm/fhir/server/test/operation/EraseOperationTest.java
```diff
@@ -194,8 +194,15 @@ private void eraseResource(String resourceType, String logicalId, boolean error,
      * @param reason should include the reason?
      * @param reasonMsg the string to provide
      */
-    private void eraseResource(String resourceType, String logicalId, boolean error, String msg, boolean patient, boolean reason, String reasonMsg) {
-        Entity<Parameters> entity = Entity.entity(generateParameters(patient, reason, reasonMsg, Optional.empty()), FHIRMediaType.APPLICATION_FHIR_JSON);
+    private void eraseResource(String resourceType, String logicalId, boolean error, String msg,
+            boolean patient, boolean reason, String reasonMsg) {
+        Parameters requestBody = generateParameters(patient, reason, reasonMsg, Optional.empty());
+        if (DEBUG) {
+            System.out.println("Invoking $erase on " + resourceType + "/" + logicalId +
```
punktilious, line 201: single quotes around the values.

build/derby-bootstrap.sh
lmsurpre: yes.  everytime i touch these I think that.  i've just never made it a priority...

fhir-server/src/test/java/com/ibm/fhir/server/test/CapabilitiesTest.java
```diff
@@ -24,13 +24,15 @@
...
 import com.ibm.fhir.model.resource.CapabilityStatement.Rest.Resource.Interaction;
 import com.ibm.fhir.model.type.code.ResourceType;
+import com.ibm.fhir.search.util.SearchUtil;
 import com.ibm.fhir.server.resources.Capabilities;
 
 public class CapabilitiesTest {
 
     @BeforeClass
     void setup() {
         FHIRConfiguration.setConfigHome("target/test-classes");
```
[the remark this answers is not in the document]
lmsurpre: i think our tests are split about 50/50 between these two.  in most (all?) cases I think they should be equivalent.  was trying not to bit off too much in this already large pr

cql/fhir-cql-server/src/test/java/com/ibm/fhir/cql/engine/server/retrieve/ServerFHIRRetrieveProviderTest.java
```diff
@@ -50,6 +52,11 @@
     private ServerFHIRRetrieveProvider provider;
     private FHIRResourceHelpers helpers;
 
+    @BeforeClass
+    public void initializeSearchUtil() {
```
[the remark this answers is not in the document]
lmsurpre: agree, but we already have that issue due to statics in FHIRUtil.  i decided to stop at just ParametersUtil and CompartmentUtil for this PR.  changing SearchUtil to avoid statics will be a large undertaking and I'm not sure its worth it at this time

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/patient/PatientExportPartitionMapper.java
lmsurpre: I debated whether to use a singleton pattern for CompartmentUtil or not.  The initialization logic does depend on the state of the registry and so if we did make it a singleton then I think we'd want a way to "re-initialize" it with the latest from the registry

fhir-persistence/src/test/java/com/ibm/fhir/persistence/test/common/AbstractPersistenceTest.java
```diff
@@ -113,15 +113,13 @@ public void setUp() throws Exception {
         // Note: this assumes that the concrete test classes will be in a project that is peer to the fhir-persistence module
         // TODO: it would be better for our unit tests if we could load config files from the classpath
         FHIRConfiguration.setConfigHome("../fhir-persistence/target/test-classes");
```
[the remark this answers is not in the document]
lmsurpre: I think we'd still want to set the configHome on each unit tests that need to read fhir-server-config from disk, so I'm not sure it would buy us much.  Maybe it would be better to clear it at the end of each test that sets it, to ensure each test is actually setting it?

Anyway, I agree that a pluggable FHIRConfiguration strategy is probably the right answer here...we have https://github.com/IBM/FHIR/issues/1318 for that

build/migration/bin/2_previous-integration-test.sh
lmsurpre: > what relationship does patient test have with CompartmentDefinition? Could use a comment to indicate how this value relates to the code.

none.  i thought this parameter was required but apparently its only required for resource types that can be in the patient compartment.  i could remove it, but if we ever want to delete resource types that *are* in the patient compartment we'd need to add it back in

lmsurpre: > How are errors handled?

they're not.  it would pass this step and move on to our e2e tests and those might fail.  that would be confusing, but not catastrophic.
i think i could update the curl command to return an error code if the request was unsuccessful...that would lead to a much more clear error in the ci job logs

lmsurpre: ok, i pushed an update that includes the `--fail` flag.  combined with the `set -ex` at the beginning of the script, that should ensure we exit if one of these requests fail (which they shouldn't)

build/derby-bootstrap.sh
lmsurpre: ok, i tried my hand at this.  now we have build/common/set_tenant1_datastore_vars.sh which exports the variables and I call that from each workflow instead of repeating the values each time

the one exception to this is our windows build which is using powershell...i didn't bother modifying it there.  instead, i just left a note in the set_tenant1_datastore_vars.sh script that says you should update the .ps1 script with each edit

fhir-server/src/test/java/com/ibm/fhir/server/test/CapabilitiesTest.java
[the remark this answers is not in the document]
punktilious: not too worried, as long as it works the same way

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/export/patient/PatientExportPartitionMapper.java
punktilious: Yeah, just want to make sure that we're not concerned about this being a strict singleton.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/operation/EraseOperationTest.java
lmsurpre: its just test output, but i went ahead and added single quotes around resourceType/logicalId (but not around the request body)

fhir-search/src/test/java/com/ibm/fhir/search/reference/ReferenceUtilTest.java
lmsurpre: done

fhir-search/src/test/java/com/ibm/fhir/search/parameters/ParametersUtilTest.java
lmsurpre: done

fhir-search/src/main/java/com/ibm/fhir/search/compartment/CompartmentUtil.java
```diff
@@ -76,11 +79,15 @@ private CompartmentUtil() {
      * @param compMap map of compartment name to CompartmentCache
      * @param resourceCompMap map of resource type name to ResourceCompartmentCache
      */
-    public static final void buildMaps(Map<String, CompartmentCache> compMap, Map<String, ResourceCompartmentCache> resourceCompMap) {
+    public final void buildMaps(Map<String, CompartmentCache> compMap, Map<String, ResourceCompartmentCache> resourceCompMap) {
```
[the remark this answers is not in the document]
lmsurpre: done

fhir-server/src/test/java/com/ibm/fhir/server/test/CapabilitiesTest.java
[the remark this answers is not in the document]
lmsurpre: ok, in at least one case it needs to be `target/test-classes` because we do some manipulation there (e.g. renaming/moving the file) which we wouldn't want to do under src.  so i'm leaning toward making them all `target/test-classes` for consistency

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
```diff
@@ -131,27 +132,27 @@
     // The functionality is split into a new class.
     private static final Sort sort = new Sort();
 
+    private static CompartmentUtil compartmentHelper;
```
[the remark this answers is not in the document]
lmsurpre: I added this note at the class level:
> Initialization is *not* thread safe, so care must be taken to avoid calling initialization while other threads are using the class.

Not very satisfying though.  Can you think of a better way without incurring the pain of changing every single spot that is using SearchUtil statically?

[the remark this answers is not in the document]
punktilious: Yeah you can do the inner static class pattern to make the singleton initialization thread-safe...
```
    private static class Holder {
        private static final CompartmentUtil INSTANCE = new CompartmentUtil();
    }

   private CompartmentUtil getCompartmentHelper() {
       return Holder.INSTANCE;
```

cql/fhir-cql-server/src/test/java/com/ibm/fhir/cql/engine/server/retrieve/ServerFHIRRetrieveProviderTest.java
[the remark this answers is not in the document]
lmsurpre: after much consternation, I've now added a commit that converts most SearchUtil methods from static to instance-level.  this addresses the race condition concern, but forced numerous other changes/decisions and so a new review is needed

fhir-search/src/main/java/com/ibm/fhir/search/util/SearchUtil.java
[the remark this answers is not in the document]
lmsurpre: we needed a way to re-initialize the CompartmentUtil for each different module on the system.  the only way i could find to do that in a thread-safe way was to "bite the bullet" and refactor SearchUtil...it is no longer static (nor a singleton instance) and so that should address the concern here.  however, this forced numerous other changes/decisions and so a new review is needed

cql/fhir-cql/src/main/java/com/ibm/fhir/cql/engine/searchparam/SearchParameterResolver.java
```diff
@@ -37,8 +43,8 @@ public SearchParameter getSearchParameterDefinition(String resourceType, String
             path = "";
         }
 
-        // XXX should this use the registry (all parameters) or the filtered set of search parameters for a given tenant?
-        Map<String, SearchParameter> params = SearchUtil.getSearchParameters(resourceType);
+        // should this use the registry (all parameters) or the filtered set of search parameters for a given tenant?
```
punktilious, line 46: Is this an open question, or is it documenting the behavior?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -221,6 +220,9 @@
     // When set, use this interface to persist the payload object. Can be null.
     private final FHIRPayloadPersistence payloadPersistence;
 
+    // A helper for processing search requests
```
punktilious, line 223: tenant-specific?

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -309,6 +312,7 @@ public FHIRPersistenceJDBCImpl(Properties configProps, IConnectionProvider cp, F
 
         this.cache = cache;
         this.payloadPersistence = null;
+        this.searchHelper = new SearchUtil();
```
punktilious, line 315: I was wondering if you think that SearchUtil should be an interface...in which case this helper would need to be injected.

punktilious: Perhaps that's a second level of refactor

cql/fhir-cql/src/main/java/com/ibm/fhir/cql/engine/searchparam/SearchParameterResolver.java
lmsurpre: i think its an open question, but unrelated to the PR.  i don't think its too important

cql/operation/fhir-operation-cpg/src/main/java/com/ibm/fhir/operation/cpg/AbstractCqlOperation.java
```diff
@@ -137,34 +138,36 @@ protected Issue checkUnsupportedParameter(ParameterMap paramMap, String paramNam
...
      * @param primaryLibrary
      *            Library resource that is the entry point for the evaluation
+     * @param searchHelper
+     *            A helper class for working with FHIR search parameters
      * @param libraries
      *            List of all necessary library resources. The first entry in the list is the primary library.
      * @return Parameters describing the evaluation result
      */
-    protected Parameters doEvaluation(FHIRResourceHelpers resourceHelper, ParameterMap paramMap, List<Library> libraries) {
+    protected Parameters doEvaluation(FHIRResourceHelpers resourceHelper, ParameterMap paramMap, SearchUtil searchHelper, List<Library> libraries) {
```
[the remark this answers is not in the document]
lmsurpre: I kind of want to rename SearchUtil, CompartmentUtil, *and* ParametersUtil from "util" to "helper" to reflect the change from static utilities to an instance-based class, but I held off because its already such a big changeset.  Happy to do that change if we agree everything else looks good.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
lmsurpre: yeah, it gets injected for the running server.  this particular constructor is only used from tests and so I didn't both to have them all inject it

lmsurpre: I did debate making an interface for SearchUtil and put it in fhir-server-spi (similar to the one for FHIRRestHelper), but that was more to try avoiding a dependency on fhir-search within our custom operation modules.
In the end, I didn't bother because those modules will mark all their dependencies as "provided" anyway.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
lmsurpre: nope.  a single instance covers all tenants

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/FHIRPersistenceJDBCFactory.java
```diff
@@ -22,12 +23,12 @@
     private final FHIRPersistenceJDBCTenantCache tenantCache = new FHIRPersistenceJDBCTenantCache();
 
     @Override
-    public FHIRPersistence getInstance() throws FHIRPersistenceException {
+    public FHIRPersistence getInstance(SearchUtil searchHelper) throws FHIRPersistenceException {
```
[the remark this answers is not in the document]
lmsurpre: its not really.  we end up passing the same searchHelper on each call of getInstance()

cql/fhir-cql/src/main/java/com/ibm/fhir/cql/engine/searchparam/SearchParameterResolver.java
```diff
@@ -21,6 +21,12 @@
 
 public class SearchParameterResolver {
 
+    private final SearchUtil searchHelper;
```
tbieste, line 24: Update copyright year.

fhir-persistence/src/main/java/com/ibm/fhir/persistence/FHIRPersistenceFactory.java
```diff
@@ -8,6 +8,7 @@
 
 import com.ibm.fhir.persistence.exception.FHIRPersistenceException;
 import com.ibm.fhir.persistence.payload.FHIRPayloadPersistence;
+import com.ibm.fhir.search.util.SearchUtil;
```
tbieste, line 11: Update copyright year.

fhir-server/src/test/java/com/ibm/fhir/server/test/BundleValidationTest.java
```diff
@@ -39,6 +39,7 @@
 import com.ibm.fhir.model.type.code.NarrativeStatus;
 import com.ibm.fhir.model.type.code.SearchEntryMode;
 import com.ibm.fhir.persistence.FHIRPersistence;
+import com.ibm.fhir.search.util.SearchUtil;
```
tbieste, line 42: Update copyright year.

operation/fhir-operation-validate/src/main/java/com/ibm/fhir/operation/validate/ValidateOperation.java
```diff
@@ -27,6 +27,7 @@
 import com.ibm.fhir.model.type.code.IssueType;
 import com.ibm.fhir.model.type.code.NarrativeStatus;
 import com.ibm.fhir.registry.FHIRRegistry;
+import com.ibm.fhir.search.util.SearchUtil;
```
tbieste, line 30: Update copyright year.

lmsurpre: done

fhir-persistence/src/main/java/com/ibm/fhir/persistence/FHIRPersistenceFactory.java
lmsurpre: done

fhir-bulkdata-webapp/src/main/java/com/ibm/fhir/bulkdata/jbatch/control/JobControlContextListener.java
```diff
@@ -24,7 +24,6 @@
 import com.ibm.fhir.model.util.FHIRUtil;
```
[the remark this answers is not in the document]
lmsurpre: done

cql/operation/fhir-operation-cpg/src/main/java/com/ibm/fhir/operation/cpg/AbstractCqlOperation.java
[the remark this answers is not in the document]
lmsurpre: i pushed a commit with the big rename

PR #3388, opened 2022-02-24
Issue with QA on #3083

prb112:
- added a check for result = empty and added supportedResourceTypes
- added small fix to FilePreflight to this commit

### DIFF
--- a/operation/fhir-operation-bulkdata/src/test/java/com/ibm/fhir/operation/bulkdata/util/BulkDataExportUtilTest.java
+++ b/operation/fhir-operation-bulkdata/src/test/java/com/ibm/fhir/operation/bulkdata/util/BulkDataExportUtilTest.java
@@ -429,7 +429,7 @@ public class BulkDataExportUtilTest {
         BulkDataExportUtil util = new BulkDataExportUtil();
         List<String> result = util.checkAndValidateTypes(ps, null);
         assertNotNull(result);
-        assertTrue(result.isEmpty());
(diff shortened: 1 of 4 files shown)


operation/fhir-operation-bulkdata/src/test/java/com/ibm/fhir/operation/bulkdata/util/BulkDataExportUtilTest.java
```diff
@@ -429,7 +429,7 @@ public void testCheckAndValidateTypesWithExtraComma() throws FHIROperationExcept
         BulkDataExportUtil util = new BulkDataExportUtil();
         List<String> result = util.checkAndValidateTypes(ps, null);
         assertNotNull(result);
-        assertTrue(result.isEmpty());
+        assertFalse(result.isEmpty());
```
lmsurpre, line 432: I guess this is fine at this level (BulkDataExportUtil), but I hope an invalid parameter value like this is caught when we first accept the export request :-)

lmsurpre (approved):
LGTM

operation/fhir-operation-bulkdata/src/test/java/com/ibm/fhir/operation/bulkdata/util/BulkDataExportUtilTest.java
prb112: It's for the case of absence of the parameter.  Now we return a subset instead of all of them.

Issue #2300, opened 2021-04-28
AuthzPolicyEnforcement should prevent access to the whole-system history api


lmsurpre:
**Is your feature request related to a problem? Please describe.**
The `fhir-smart` AuthzPolicyEnforcementPersistenceInterceptor can be used to provide user-scoped authorization to the the FHIR Server's APIs.

For example, it scopes all searches to the patient compartment that matches the patient context for a given session.

However, one thing it doesn't prevent access to is the whole-system history endpoint.
This endpoint provides an overview of all the create/update/delete interactions serviced on the server.
Therefor, a use with access to the api would be able to determine all of the resource ids on the server.

**Describe the solution you'd like**
Currently the whole-system-history implementation doesn't execute any beforeX or afterX hooks in our PersistenceInterceptor framework.  Since this is the mechanism used by AuthzPolicyEnforcementPersistenceInterceptor to restrict access, we should first add these hooks (or use the existing beforeHistory and afterHistory ones).

Next, we should either filter the list of results somehow or, more likely, just prohibit this interaction altogether.

**Describe alternatives you've considered**
We already have support for restricting access to the whole-system history endpoint and that is to filter the "history" interaction on the "Resource" resource in the fhir-server-config.json (as described at https://ibm.github.io/FHIR/guides/FHIRServerUsersGuide#412-fhir-rest-api ).
This might be sufficient, but it requires the operator to override this setting on each configured resource type.

Example:
```
"resources": {
    "open": false,
    "Condition": {
        "interactions": ["read", "vread", "history", "search"]
    },
    "Observation": {
        "interactions": ["read", "vread", "history", "search"]
    },
    "Patient": {
        "interactions": ["read", "vread", "search"]
    }
}
```

**Acceptance Criteria**
1.  
   GIVEN a deployment with fhir-smart installed
    AND a request scoped to a particular patient
   WHEN they invoke the whole-system history endpoint
   THEN they should not see the resource ids for any resources to which they do not have access

**Additional context**
Relates to #2026 which rounds out our support for proper "whole-system history" behavior.

**QA suggestions**
1) Test system-level history:  Requires `user/*.read` or `system/*.read` authority
  Example: https://localhost:9443/fhir-server/api/v4/_history
2) Test system-level history (with `_type` parameter):  Requires `user/<resource_type>.read` or `system/<resource_type>.read` authority to each resource type in the `_type` parameter.
[...]

PR #3411, opened 2022-02-28
Issue #2300 - Add calls to beforeHistory and afterHistory

tbieste:
Issue #2300 - Add calls to beforeHistory and afterHistory

1. System-level history: Requires `user/*.read` or `system/*.read` authority

    Example: https://localhost:9443/fhir-server/api/v4/_history

1. System-level history (with `_type` parameter): Requires `user/<resource_type>.read` or `system/<resource_type>.read` authority to each resource type in the `_type` parameter.

    Example: https://localhost:9443/fhir-server/api/v4/_history?_type=Patient,Observation

1. Type-level history: Requires `user/<resource_type>.read` or `system/<resource_type>.read` authority to the resource type.

    Example: https://localhost:9443/fhir-server/api/v4/Patient/_history

### DIFF
--- a/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
@@ -205,7 +205,7 @@ public class FHIRRestHelper implements FHIRResourceHelpers {
         try {
             // Prepare the persistence event
             FHIRPersistenceEvent event =
-                    new FHIRPersistenceEvent(resource, buildPersistenceEventProperties(type, null, null, null));
+                    new FHIRPersistenceEvent(resource, buildPersistenceEventProperties(type, null, null, null, null));
 
             // Run the meta phase to handle ifNoneExist and update the resource meta-data
             response = doCreateMeta(event, warnings, type, resource, ifNoneExist);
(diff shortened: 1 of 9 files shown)


fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -3061,28 +3066,32 @@ public Bundle doHistory(MultivaluedMap<String, String> queryParameters, String r
...
-                
+
                 if (resources.size() != records.size()) {
                     // very unlikely...unless we overlap with a patient erase operation
                     throw new FHIRPersistenceException("Could not read all required resource records");
                 }
             }
+
+            // Invoke the 'afterHistory' interceptor methods.
+            getInterceptorMgr().fireAfterHistoryEvent(event);
```
lmsurpre, line 3093: we need to decide what the contract should be for the event object for afterHistory.
shouldn't it include the resources that we're returning?
but what if we're only returning the ids (and not the contents)?

we'll have the same decision to make on afterSearch now that we support omitting the resource content

tbieste: Updated to include the resource on the afterHistory call.  Opened #3423 for handling `Prefer: return=minimal`.

lmsurpre (approved):

Issue #3423, opened 2022-03-02
AuthzPolicyEnforcement handling history and search with 'Perfer: return=minimal'


tbieste:
**Is your feature request related to a problem? Please describe.**
The fhir-smart AuthzPolicyEnforcementPersistenceInterceptor can be used to provide user-scoped authorization to the FHIR Server's APIs.  However, when the `Prefer: return=minimal` HTTP header is used with search or history, the interceptor does not have the resource available, so it produces a NullPointerException.

**Describe the solution you'd like**
The intercepter should handle no resource being available to it, by either denying accesses that cannot be checked, or to obtain the resource on its own, so the necessary checking can be done.

**Describe alternatives you've considered**
Only allow user/system access (and not patient access), so the resource does not need to be retrieved.  The point of `Prefer: return=minimal` is to make the API as fast as possible.  Having to return all the resources will likely be slower than just having the FHIR server retrieve all the resources in the first place.

**Acceptance Criteria**

1.  GIVEN a deployment with fhir-smart installed
    AND a request scoped to a particular patient
   WHEN they invoke the history endpoint with `Prefer: return=minimal` HTTP header
   THEN the results are returned if the user has the necessary access

1.  GIVEN a deployment with fhir-smart installed
    AND a request scoped to a particular patient
   WHEN they invoke the search endpoint with `Prefer: return=minimal` HTTP header
   THEN the results are returned if the user has the necessary access

**Additional context**
Related to issue #2300

tbieste:
After thinking about the performance implications, decided to go with:
Only allow user/system access (and not patient access), so the resource does not need to be retrieved. The point of Prefer: return=minimal is to make the API as fast as possible. Having to retrieve all the resources may end up being slower than just having the FHIR server retrieve all the resources in the first place.

PR #3427, opened 2022-03-02
Issue #3423 - Update afterHistory and afterSearch

tbieste:
Issue #3423 - Update afterHistory and afterSearch

### DIFF
--- a/docs/src/pages/guides/FHIRServerUsersGuide.md
+++ b/docs/src/pages/guides/FHIRServerUsersGuide.md
@@ -2760,6 +2760,8 @@ This component uses the IBM FHIR Server's PersistenceInterceptor feature to auto
 
 Additionally, before returning resources to the client, the `fhir-smart` component performs authorization policy enforcement based on the list of SMART scopes included in the token's `scope` claim and the list of patient compartments in the `patient_id` claim.
 
+When the HTTP header `Prefer: return=minimal` is specified on a search or history request, only minimal resource metadata is retrieved. In those cases, either `user` or `system` SMART scopes must be used, since the resource data necessary to enforce access via `patient` SMART scopes is not available.
+
 For an example of using the IBM FHIR Server together with a SMART-enabled Keycloak authorization server, please see the data-access pattern at https://github.com/Alvearie/health-patterns/tree/main/data-access.
 
 ## 5.4 Custom HTTP Headers
--- a/docs/src/pages/guides/FHIRServerUsersGuide.md
+++ b/docs/src/pages/guides/FHIRServerUsersGuide.md
@@ -2762,7 +2762,7 @@ Additionally, before returning resources to the client, the `fhir-smart` compone
 
 When the HTTP header `Prefer: return=minimal` is specified on a search or history request, only minimal resource metadata is retrieved. In those cases, either `user` or `system` SMART scopes must be used, since the resource data necessary to enforce access via `patient` SMART scopes is not available.
 
-For an example of using the IBM FHIR Server together with a SMART-enabled Keycloak authorization server, please see the data-access pattern at https://github.com/Alvearie/health-patterns/tree/main/data-access.
+For an example of using the IBM FHIR Server together with a SMART-enabled Keycloak authorization server, please see the data-access pattern at https://github.com/LinuxForHealth/health-patterns/tree/main/data-access.
 
 ## 5.4 Custom HTTP Headers
 IBM FHIR Server Supports the following custom HTTP Headers:
--- a/fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
+++ b/fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
@@ -748,7 +748,15 @@ public class AuthzPolicyEnforcementPersistenceInterceptor implements FHIRPersist
         }
 
         if (approvedScopeMap.containsKey(ContextType.PATIENT)) {
-            if (resource != null) {
+            if (resource == null) {
+                // For `patient` scopes, if no resource was retrieved, which happens with history/search
+                // operations with 'Prefer: return=minimal' HTTP header, there is no way to check if the
+                // resource meets the criteria of being in the Patient compartment.
+                if (log.isLoggable(Level.FINE)) {
+                    log.fine("Unable to determine if " + requiredPermission.value() + " permission for '" + resourceType + "/" + resourceId +
+                        "' is granted via scope " + approvedScopeMap.get(ContextType.PATIENT) + " due to resource not retrieved");
+                }
+            } else {
                 if (resource instanceof Provenance) {
                     // Addressed for issue #1881, Provenance is a special-case:  a Patient-compartment resource type that
                     // we want to allow access to if and only if the patient has access to one or more resources that it targets.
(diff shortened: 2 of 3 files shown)


fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
```diff
@@ -717,50 +741,60 @@ private boolean isAllowed(Resource resource, Set<String> contextIds, Permission
...
+                        return true;
+                    }
                 }
-            }
 
-            // Else, see if the target resource belongs to the Patient compartment of the in-context patient
-            return isInCompartment(resource, CompartmentType.PATIENT, contextIds);
+                // Else, see if the target resource belongs to the Patient compartment of the in-context patient
+                return isInCompartment(resource, CompartmentType.PATIENT, contextIds);
+            } else {
```
lmsurpre, line 789: style-only comment:  personally i prefer the smaller block first so that you don't have to remember what the else on line 789 was lined up with above

it could even be made as a guard clause with its own early exit:
```
if (resource == null) {
  log.fine(...);
  return false;
}
```

then you could remove one level of indentation

lmsurpre: although i suppose some folks prefer a single exit point of the function at the bottom (e.g.in case someone adds some other type of checking after the `approvedScopeMap.containsKey(ContextType.PATIENT` block

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
```diff
@@ -926,4 +960,40 @@ private String getAccessToken() throws FHIRPersistenceInterceptorException {
...
+     * Gets the resource type of the resource represented in the bundle entry.
+     * @param entry the bundle entry
+     * @return the resource type, or null
+     */
+    private String getResourceType(Bundle.Entry entry) {
+        Resource resource = entry.getResource();
+        if (resource != null) {
+            return resource.getClass().getSimpleName();
+        } else if (entry.getFullUrl() != null && entry.getFullUrl().getValue() != null) {
+            String[] urlParts = entry.getFullUrl().getValue().split("/");
```
lmsurpre, line 974: this presumes that we'll never start using absolute URLs in our fullUrls, which I think it a bit suspect.
could we make it resilient to that?

lmsurpre: if not, maybe just update the javadoc to be very clear about this assumption?

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
```diff
@@ -926,4 +960,40 @@ private String getAccessToken() throws FHIRPersistenceInterceptorException {
...
+     * Gets the resource ID of the resource represented in the bundle entry.
+     * @param entry the bundle entry
+     * @return the resource ID, or null
+     */
+    private String getResourceId(Bundle.Entry entry) {
+        Resource resource = entry.getResource();
+        if (resource != null) {
+            return resource.getId();
+        } else if (entry.getFullUrl() != null && entry.getFullUrl().getValue() != null) {
+            String[] urlParts = entry.getFullUrl().getValue().split("/");
```
lmsurpre, line 992: same comment as above

fhir-smart/src/test/java/com/ibm/fhir/smart/test/AuthzPolicyEnforcementTest.java
```diff
@@ -67,8 +68,10 @@
 
 public class AuthzPolicyEnforcementTest {
     private static final String PATIENT_ID =     "11111111-1111-1111-1111-111111111111";
-    private static final String OBSERVATION_ID = "11111111-1111-1111-1111-111111111111";
-    private static final String CONDITION_ID =   "11111111-1111-1111-1111-111111111111";
+    private static final String OBSERVATION_ID = "11111111-1111-1111-1111-111111111112";
```
lmsurpre, line 71: not sure why i chose to use a GUID-like value here, but I do think having a couple different resource types with the same id is an interesting test to ensure we're not using the id alone for any access decisions

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
tbieste: Yeah, you exactly followed my thought process.  I originally had the early exit, and then thought about if someone wanted to add more logic after it.

fhir-smart/src/test/java/com/ibm/fhir/smart/test/AuthzPolicyEnforcementTest.java
tbieste: Updated.

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
tbieste: Updated.

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
tbieste: Doesn't this code work for both literal and absolute URLs?  Or am I missing something?

lmsurpre: i think I just misread it as `urlParts[2]` ... we never echo parameters back in the fullUrl and we no longer have fullUrl that populate `[type]/[id]/_history/[vid]`, so I agree this should work

docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -2760,6 +2760,8 @@ This component uses the IBM FHIR Server's PersistenceInterceptor feature to auto
 
 Additionally, before returning resources to the client, the `fhir-smart` component performs authorization policy enforcement based on the list of SMART scopes included in the token's `scope` claim and the list of patient compartments in the `patient_id` claim.
 
+When the HTTP header `Prefer: return=minimal` is specified on a search or history request, only minimal resource metadata is retrieved. In those cases, either `user` or `system` SMART scopes must be used, since the resource data necessary to enforce access via `patient` SMART scopes is not available.
+
 For an example of using the IBM FHIR Server together with a SMART-enabled Keycloak authorization server, please see the data-access pattern at https://github.com/Alvearie/health-patterns/tree/main/data-access.
```
lmsurpre, line 2765: unrelated to your PR, but we should update this to https://github.com/LinuxForHealth/health-patterns/tree/main/data-access

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
lmsurpre: thanks.  even without the early exit, i think this version is a lot easier for my brain

lmsurpre (approved):
LGTM

docs/src/pages/guides/FHIRServerUsersGuide.md
tbieste: I can update that and set the ci-skip label.

Issue #3450, opened 2022-03-10
Patch of a deleted resource causes Null Pointer


d0roppe:
**Describe the bug**
Deleting a resource and then issuing a patch of that resource causes a 500 error code and a null pointer exception
```
{
	"resourceType": "OperationOutcome",
	"issue": [
		{
			"severity": "fatal",
			"code": "exception",
			"details": {
				"text": "NullPointerException: resourceOrElement"
			}
		}
	]
}
```

```
[INFO    ] Received request: tenantId:[default] dsId:[default] user:[fhiruser] method:[PATCH] uri:[https://localhost:9443/fhir-server/api/v4/Patient/sniff-test--BDE-check-P1] originalUri:[https://copper.minions-r.us/fhir-server/api/v4/Group/sniff-test--BDE-check-R1/$export]
[INFO    ] Transaction failed - afterCompletion(status = 4)
[ERROR   ] An unexpected exception occurred while processing the request
resourceOrElement
[INFO    ] Completed request[0.06406286 secs]: tenantId:[default] dsId:[default] user:[fhiruser] method:[PATCH] uri:[https://localhost:9443/fhir-server/api/v4/Patient/sniff-test--BDE-check-P1] originalUri:[https://copper.minions-r.us/fhir-server/api/v4/Group/sniff-test--BDE-check-R1/$export] status:[500]
```

**Environment**
fhir-main

**To Reproduce**

**Expected behavior**
a null pointer should not be the reason for this not working.
**Additional context**

tbieste:
@d0roppe Do you have the input used on the PATCH operation?

d0roppe:
Verified that now this scenario gets a 404.

PR #3451, opened 2022-03-10
Issue #3450 - Handle deleted resource on PATCH

tbieste:
Issue #3450 - Handle deleted resource on PATCH

### DIFF
--- a/fhir-server-test/src/test/java/com/ibm/fhir/server/test/FHIRPatchTest.java
+++ b/fhir-server-test/src/test/java/com/ibm/fhir/server/test/FHIRPatchTest.java
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright IBM Corp. 2019, 2021
+ * (C) Copyright IBM Corp. 2019, 2022
  *
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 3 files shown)


fhir-server-test/src/test/java/com/ibm/fhir/server/test/FHIRPatchTest.java
```diff
@@ -1002,6 +1003,61 @@ public void testBundleWithBadPatches() throws Exception {
...
+        JsonArray array = Json.createPatchBuilder()
+                .add("/name/0/given/1", "Jack")
+                .build().toJsonArray();
+
+        // Call the 'patch` API
+        Entity<JsonArray> patchEntity = Entity.entity(array, FHIRMediaType.APPLICATION_JSON_PATCH);
+        Response response = target.path("Patient/" + patient.getId())
+                .request(FHIRMediaType.APPLICATION_FHIR_JSON)
+                .method("PATCH", patchEntity, Response.class);
+        assertResponse(response, Response.Status.METHOD_NOT_ALLOWED.getStatusCode());
```
lmsurpre, line 1022: this seems like a slightly odd error response to me.  if the resource doesn't exist, maybe it should be a 404?

tbieste: For PATCH, the existing code returned 405 is the resource did not exist.  I thought that was a bit weird, but kept it since that was the current behavior.
I think it may be for consistency with how "update" works (returning 405 when the resource doesn't exist:
https://www.hl7.org/fhir/http.html#rejecting-updates

lmsurpre: That says:
> 405 Method Not allowed - the resource did not exist prior to the update, and the server does not allow client defined ids

So that is appropriate if we do not support "create-on-update" (configurable in fhir-server-config).

To me, that is quite different from being unable to patch a resource because it doesn't exist because there is no configuration in which that would make sense.

tbieste: Makes sense.  I'll update it to 404 Not Found.

lmsurpre (approved):
LGTM

PR #3452, opened 2022-03-11
Update per QA discussion on success versus failures and counting

prb112:

### DIFF
--- a/fhir-server/src/main/java/com/ibm/fhir/server/util/RestAuditLogger.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/util/RestAuditLogger.java
@@ -407,58 +407,61 @@ public class RestAuditLogger {
                                             || requestEntry.getRequest().getUrl().getValue().contains("/%24");
                     String action = "E";
                     HTTPVerb requestMethod = requestEntry.getRequest().getMethod();
-                    switch (HTTPVerb.Value.from(requestMethod.getValue())) {
-                    case GET:
-                        if (operation) {
+                    // Avoids counting successful changes when there is a failure by the client or the server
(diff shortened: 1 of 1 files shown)


lmsurpre:
https://github.com/IBM/FHIR/blob/main/fhir-audit/src/main/java/com/ibm/fhir/audit/beans/Batch.java#L131-L149
* resources_read  -> reads
* resources_created -> creates
* resources_updated -> updates
* resources_deleted -> deletes
* resources_executed -> executions

Although I think the whole count section was basically just a temporary measure until we can get to #3455

d0roppe:
for Audit entries, I would remove the count section all together.  It does not add any information, and only confuses the entry when the task failed.  Above this `batch` section the `action` shows what was done, the `api_parameters/request` shows to what, and `api_parameters/request_status` shows if it worked or not. 
So for a single entry per batch request the `batch` section in the message should be removed.

lmsurpre:
OK, lets close this one and ensure we have issues for any changes needed.

For batch bundles, we currently have two different types of audit messages:
1. an audit entry for that request with an action code of E and counts for everything in the bundle
2. an audit entry for each entry from within that batch with a useless count section that will only ever have a value of 1 for one of the 5 options

Dag and I discussed and we think the best option would be to drop message 1 and remove the counts from the messages for 2.

For transaction bundles, we think the best option is to replace the counts with information about the actual resources.

lmsurpre:
> Dag and I discussed and we think the best option would be to drop message 1 and remove the counts from the messages for 2.

https://github.com/IBM/FHIR/issues/3493 

> For transaction bundles, we think the best option is to replace the counts with information about the actual resources.

https://github.com/IBM/FHIR/issues/3455#issuecomment-1074590511

Issue #3390, opened 2022-02-24
Whole system interactions should honor resources config when open=false


d0roppe:
**Describe the bug**
with fhirServer/resource configurations in fhir-server-config.json with resources listed, and open=false.  Whole system search should only work on the resources listed

server log contains exception

**Environment**
FHIR main

lmsurpre:
this work was already implemented in our `r4b-rebased` branch as part of #3242 ...what we need to do is backport that to our main branch

d0roppe:
Now with `"open": = false` for the resources configuration, a whole system search now gets appended with `_type=` for the resource types noted in the configuratiopn.  So now the whole system search works as expected with the `"open": = false`

Issue #3319, opened 2022-02-09
Unify handling of `_type` parameter across system-search, system-history, $everything and $export


lmsurpre:
**Is your feature request related to a problem? Please describe.**
Currently, our system-history implementations accepts multiple instances of the `_type` parameter and combine them into a single list.

Conversely, our system-search implementation takes just the first one.

In a recent version of the bulkdata spec, they relaxed the cardinality and so multiple instances of this parameter should be supported there as well.  We have a separate issue, https://github.com/IBM/FHIR/issues/3081, for that one.

**Describe the solution you'd like**
across these 4 endpoints, we should have common behavior for this parameter

**Describe alternatives you've considered**

**Acceptance Criteria**

1.  
   WHEN a search like `GET [base]?_type=Patient&_type=Condition`
   THEN the values of the _type parameters are combined
    AND the result matches `GET [base]?_type=Patient,Condition`

And verify this same behavior when:
1. system-history is invoked with multiple _type query params
2. $everything is invoked with multiple _type query params
3. $export is invoked via GET with multiple _type query params
4. $export is invoked via POST with multiple _type params in the Parameters body

And finally, verify that when $export is invoked via POST, none of the query parameters are used (only the Parameters from the body).

**Additional context**
We should probably document the behavior in Conformance.md as well, since this differs from how we handle other "return parameters" (#3318 )

lmsurpre:
The $export operation has some special handling for the _types parameter where it checks in both the request url AND the Parameters object.  I'd like to try normalizing the two so that downstream code can look in a single spot for request parameters.  I think we already do something like that for extended operations that are invoked via GET.

lmsurpre:
Based on the last comment, I added this line to the acceptance criteria:
> And finally, verify that when $export is invoked via POST, none of the query parameters are used (only the Parameters from the body).

d0roppe:
Verified that specifying _type=<resourceType1>,<resourceType2> or _type=<resourceType1>&_type=<resourceType2> are handled the same for Search, History, $everything, and Bulk Data Export as requested in the Acceptance Criteria.

Issue #3091, opened 2021-12-07
/Patient/_search doesn't work for SMART App Launch enabled FHIR deployment


mcbride-p:
**Describe the bug**
With the Health Data Access deployment pattern here -> https://github.com/Alvearie/health-patterns/tree/main/data-access the following endpoint doesn't return the expected list of patient resources;

https://localhost:9443/fhir-server/api/v4/Patient/_search 

**Environment**
Which version of IBM FHIR Server?

**To Reproduce**
Deploy the Health Data Access deployment pattern as per above link and execute the following endpoint via a REST client;

https://localhost:9443/fhir-server/api/v4/Patient/_search 

Get empty bundle response like the following;

```
{
    "resourceType": "Bundle",
    "id": "21983671-d746-4264-8877-7975b6a78687",
    "type": "searchset",
    "total": 0,
    "link": [
        {
            "relation": "self",
            "url": "https://localhost:9443/fhir-server/api/v4/Patient/_search?_count=10&_page=1"
        }
    ]
}
```

**Expected behavior**

The above endpoint should return a bundle resource containing the patient resources persisted on server like the following;

```
{
  "resourceType": "Bundle",
  "id": "60e1b068-7b05-4b5e-8dc6-2cdea8de48c3",
  "type": "searchset",
  "total": 16,
  "link": [{
    "relation": "self",
    "url": "https://paul-dev.wh-health-patterns.dev.watson-health.ibm.com/fhir-server/api/v4/Patient/_search?_count=10&_page=1"
  }, {
    "relation": "next",
    "url": "https://paul-dev.wh-health-patterns.dev.watson-health.ibm.com/fhir-server/api/v4/Patient/_search?_count=10&_page=2"
  }],
  "entry": [{
    "fullUrl": "https://paul-dev.wh-health-patterns.dev.watson-health.ibm.com/fhir-server/api/v4/Patient/17d5b7def74-2a8b2f51-7e69-4415-81f3-194f5b4105a7",
    "resource": {
      "resourceType": "Patient",
      "id": "17d5b7def74-2a8b2f51-7e69-4415-81f3-194f5b4105a7",
      "meta": {
        "versionId": "1",
        "lastUpdated": "2021-11-26T09:05:19.774053Z"
      },
      "text": {
        "status": "generated",
[...]

PR #3473, opened 2022-03-15
issue #3390 - whole-system interactions honor resources config

lmsurpre:
* Introduce the ResourceTypeName enum in fhir-core and ResourcesConfigAdapter in fhir-config
* Use ResourcesConfigAdapter from whole-system search, whole-system history, $everything, and $export

I also started on #3319 - system-search now supports multiple instances of the `_type` parameter

Also a small fix for https://github.com/IBM/FHIR/issues/3091

### DIFF
--- a/fhir-config/src/main/java/com/ibm/fhir/config/ResourcesConfigAdapter.java
+++ b/fhir-config/src/main/java/com/ibm/fhir/config/ResourcesConfigAdapter.java
@@ -1,0 +1,114 @@
+/*
+ * (C) Copyright IBM Corp. 2022
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.ibm.fhir.config;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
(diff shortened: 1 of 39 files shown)


tbieste (approved):

fhir-config/src/main/java/com/ibm/fhir/config/ResourcesConfigAdapter.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+
+import com.ibm.fhir.config.PropertyGroup.PropertyEntry;
+import com.ibm.fhir.core.ResourceTypeName;
+
+/**
+ * An abstraction for the ibm-fhir-server fhirServer/resources property group
```
lmsurpre, line 23: indicate that we should create a new instance of this for each request (at least).  otherwise the property lookups won't be dynamic when retrieved through here

PR #3479, opened 2022-03-18
Issue #3450 - Return 404 on PATCH if resource does not exist

tbieste:
Issue #3450 - Return 404 on PATCH if resource does not exist


lmsurpre (approved):
LGTM

PR #3481, opened 2022-03-18
Issue 3478 fixes for offloading with Azure Blob

punktilious:


lmsurpre (approved):

fhir-persistence-blob/src/main/java/com/ibm/fhir/persistence/blob/FHIRPayloadPersistenceBlobImpl.java
```diff
@@ -117,17 +117,46 @@ public PayloadPersistenceResponse storePayload(String resourceType, int resource
...
+        if (e instanceof BlobStorageException) {
+            BlobStorageException bse = (BlobStorageException)e;
+            if (bse.getStatusCode() == 404) {
+                // Blob doesn't exist.
+                return true;
+            } else {
+                return false;
+            }
+        } else {
+            return false;
```
lmsurpre, line 158: I think you could remove both else clauses if you just return false after the if block

PR #3482, opened 2022-03-18
issue #3319 - normalize handling of _type parameters

lmsurpre:
Previously, I updated whole-system search to support combining multiple
instances of the `_type` parameter.  Now we will handle multiple '_type'
parameters in the extended operations $everything and $export as well.

I updated the search section in Conformance.md to indicate this change
and added a note in the extended operations section to indicate that
query parameters will be ignored for operations invoked via POST.

### DIFF
--- a/fhir-server-spi/src/main/java/com/ibm/fhir/server/spi/operation/FHIROperationUtil.java
+++ b/fhir-server-spi/src/main/java/com/ibm/fhir/server/spi/operation/FHIROperationUtil.java
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright IBM Corp. 2016, 2021
+ * (C) Copyright IBM Corp. 2016, 2022
  *
  * SPDX-License-Identifier: Apache-2.0
  */
--- a/fhir-server-spi/src/main/java/com/ibm/fhir/server/spi/operation/FHIROperationUtil.java
+++ b/fhir-server-spi/src/main/java/com/ibm/fhir/server/spi/operation/FHIROperationUtil.java
@@ -173,7 +173,7 @@ public final class FHIROperationUtil {
                         .severity(IssueSeverity.ERROR)
                         .code(IssueType.INVALID)
                         .details(CodeableConcept.builder()
-                                .text("Unable to process query parameters")
+                                .text(msg)
                                 .build())
                         .build());
         }
--- a/fhir-server-spi/src/main/java/com/ibm/fhir/server/spi/operation/FHIROperationUtil.java
+++ b/fhir-server-spi/src/main/java/com/ibm/fhir/server/spi/operation/FHIROperationUtil.java
@@ -26,6 +26,7 @@ import com.ibm.fhir.model.resource.Parameters.Parameter;
 import com.ibm.fhir.model.resource.Resource;
 import com.ibm.fhir.model.type.Canonical;
 import com.ibm.fhir.model.type.Code;
+import com.ibm.fhir.model.type.CodeableConcept;
 import com.ibm.fhir.model.type.Date;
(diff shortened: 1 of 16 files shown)


fhir-server-spi/src/main/java/com/ibm/fhir/server/spi/operation/FHIROperationUtil.java
```diff
@@ -72,103 +73,124 @@ public static void init() {
...
+                throw operationException;
+            }
+        } catch (NullPointerException | IllegalStateException e) {
+            String msg = "Unable to parse query parameter '" + paramName + "' to its expected type: '" + typeName + "'";
+            throw new FHIROperationException(msg, e)
+                    .withIssue(Issue.builder()
+                        .severity(IssueSeverity.ERROR)
+                        .code(IssueType.INVALID)
+                        .details(CodeableConcept.builder()
+                                .text("Unable to process query parameters")
```
lmsurpre, line 176:
```suggestion
                                .text(msg)
```

fhir-server-spi/src/main/java/com/ibm/fhir/server/spi/operation/FHIROperationUtil.java
```diff
@@ -72,103 +73,124 @@ public static void init() {
...
 
-            List<Issue> issues = new ArrayList<>();
-            Issue.Builder builder = Issue.builder();
-            builder.code(IssueType.INVALID);
-            builder.diagnostics(string("Unable to process query parameters"));
-            builder.severity(IssueSeverity.ERROR);
-            issues.add(builder.build());
+        Parameter.Builder parameterBuilder = Parameter.builder().name(paramName);
+        try {
+            if ("boolean".equals(typeName)) {
```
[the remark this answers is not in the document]
lmsurpre: in this PR I mostly just split this long if else block into its own method and tightened up the error handling a bit (and added tests).  personally i think its ok, but agree there might be cleaner approaches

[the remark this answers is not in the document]
lmsurpre: one thing we could do without much trouble is change it into a switch over the FHIRAllTypes.Value enum.

the only reason I didn't do that is that this enum has A LOT of values and we really only care about these 16 or so primitive ones.  so i think it would be about the same in practice.

punktilious (approved):
LGTM

fhir-server-spi/src/main/java/com/ibm/fhir/server/spi/operation/FHIROperationUtil.java
```diff
@@ -26,6 +26,7 @@
 import com.ibm.fhir.model.resource.Resource;
```
tbieste, line 26: Update copyright year.

operation/fhir-operation-bulkdata/src/main/java/com/ibm/fhir/operation/bulkdata/model/url/DownloadUrl.java
```diff
@@ -42,20 +42,19 @@
     private String objectKey = null;
```
tbieste, line 42: Update copyright year.

tbieste (approved):

fhir-server-spi/src/main/java/com/ibm/fhir/server/spi/operation/FHIROperationUtil.java
lmsurpre: fixed

Issue #3414, opened 2022-03-01
Address 4MB size limit for Azure Blob payload offload


punktilious:
**Is your feature request related to a problem? Please describe.**
The current payload offload for Azure Blob has a limit of 4MB. Extend the implementation to support larger payloads.

**Describe the solution you'd like**
A larger limit.

**Acceptance Criteria**

1.  
   GIVEN a server configured to use Azure blob payload offload
    AND a resource larger than 4MB
   WHEN the resource is posted (update/create)
   THEN the server will successfully store the resource in the Azure Blob store

**Additional context**
The Azure Blob SDK can support larger uploads using multiple calls.

punktilious:
New limit will be 5000MB

d0roppe:
Tested with 4 meg and 70 meg resource, worked just fine.  Closing issue

PR #3496, opened 2022-03-22
issue #3414 support larger payload offload using Azure Blob

punktilious:


lmsurpre (approved):
LGTM

Issue #3495, opened 2022-03-22
Modify fhir-smart behavior for "user/" scope requests


lmsurpre:
**Is your feature request related to a problem? Please describe.**
In the current implementation, we require the access token to have a `patient_id` claim for basically any search for resource types that can exist in the patient compartment.

However, for access tokens with `user/` or `system/` scopes, that feels a bit inconsistent with our `isAllowed` implementation which basically allows access to any resources when the access token contains a scope with a valid permission for the current interaction (e.g. `user/Patient.read` for searching patient resources).

This leads to confusing behavior like what is reported at https://chat.fhir.org/#narrow/stream/212434-ibm/topic/Custom.20Sort.20Parameters/near/275813474 where a search of the Basic endpoint with a scope like `user/Basic.read` ends in 0 results but a whole-system search with the same scope can retrieve these resources just fine.

**Describe the solution you'd like**
1. Only require the `patient_id` claim for `patient/` scopes (and only when a corresponding `user/` or `system/` scope doesn't cover the interaction).
2. Only scope searches to the patient compartment when the `patient/` scope is the only one that allows the interaction.

**Describe alternatives you've considered**
more clearly document that the `patient_id` claim is required even for `user/` scopes and ensure that these are enforced properly as part of whole-system search.

**Acceptance Criteria**

1.  
   GIVEN a server with multiple patient resources (a, b, c)
    AND a valid access token with patient_id=a and scope=`launch/patient patient/*.read`
   WHEN a search request would return a resource type that is in the the patient compartment
   THEN only resources in the compartment of patient a are returned

2.  
   GIVEN a server with multiple patient resources (a, b, c)
    AND a valid access token with scopes that include `user/*.read` (with or without a patient_id claim)
   WHEN a search request would return a resource type that is in the the patient compartment
   THEN all resources that meet the search criteria are returned without regard to any patient_id claim/context

3.  
   GIVEN a server with multiple patient resources (a, b, c)
    AND a valid access token with scopes that include `system/*.read` (with or without a patient_id claim)
   WHEN a search request would return a resource type that is in the the patient compartment
   THEN all resources that meet the search criteria are returned without regard to any patient_id claim/context

**Additional context**
See https://chat.fhir.org/#narrow/stream/212434-ibm/topic/GET.20Patients and https://chat.fhir.org/#narrow/stream/212434-ibm/topic/fhir-smart/near/264451208 for related discussion.

d0roppe:
[...]

PR #3498, opened 2022-03-23
issue #3495 - more selectively scope searches to patient compartment

lmsurpre:
Specifically, avoid adding the implicit compartment parameter(s) when
the interaction is allowed by one or more scopes with the 'user/' or
'system/' prefix ("context type")

### DIFF
--- a/fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
+++ b/fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
@@ -34,6 +34,7 @@ import com.ibm.fhir.model.resource.SearchParameter;
 import com.ibm.fhir.model.type.Canonical;
 import com.ibm.fhir.model.type.Reference;
 import com.ibm.fhir.model.type.code.CompartmentType;
+import com.ibm.fhir.model.type.code.HTTPVerb;
 import com.ibm.fhir.model.type.code.IssueType;
 import com.ibm.fhir.model.type.code.ResourceType;
 import com.ibm.fhir.model.util.FHIRUtil;
@@ -443,7 +444,7 @@ public class AuthzPolicyEnforcementPersistenceInterceptor implements FHIRPersist
                 break;
             case "Encounter":
                 Resource r = executeRead(event.getPersistenceImpl(), ModelSupport.getResourceType(compartment), compartmentId);
-                if (!isInCompartment(r, CompartmentType.PATIENT, patientIdFromToken)) {
+                if (!isInCompartment(compartment, compartmentId, r, CompartmentType.PATIENT, patientIdFromToken)) {
                     String msg = "Interaction with 'Encounter/" + compartmentId + "' is not permitted for patient context " + patientIdFromToken;
                     throw new FHIRPersistenceInterceptorException(msg).withIssue(FHIRUtil.buildOperationOutcomeIssue(msg, IssueType.FORBIDDEN));
                 }
@@ -546,6 +547,12 @@ public class AuthzPolicyEnforcementPersistenceInterceptor implements FHIRPersist
                 Resource resource = entry.getResource();
                 enforceDirectProvenanceAccess(event, resource, patientIdFromToken, scopesFromToken);
                 if (resourceType != null && resourceId != null) {
+                    if (resource == null && entry.getRequest() != null && entry.getRequest().getMethod().getValueAsEnum() == HTTPVerb.Value.DELETE) {
+                        // explicitly allow DELETE entries
+                        // this is safe because beforeHistory prohibits system and type-level history request for patient-compartment resource types
+                        // that are only covered by 'patient' scoped access tokens (and not 'user' or 'system' scopes)
+                        continue;
+                    }
                     enforce(resourceType, resourceId, resource, patientIdFromToken, Permission.READ, scopesFromToken);
                 } else {
                     throw new FHIRPersistenceInterceptorException("Unable to enforce authorization for history interaction "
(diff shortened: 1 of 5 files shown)


fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
```diff
@@ -695,51 +713,50 @@ private void enforce(String resourceType, String resourceId, Resource resource,
...
-            return true;
-        }
 
-        if (approvedScopeMap.containsKey(ContextType.USER)) {
-            // For `user` scopes, we grant access to all resources of the requested type.
-            // Implementers that use these scopes are encouraged to layer on their own permissions model beyond this.
-            if (log.isLoggable(Level.FINE)) {
-                log.fine(requiredPermission.value() + " permission for '" + resourceType + "/" + resourceId +
-                    "' is granted via scope " + approvedScopeMap.get(ContextType.USER));
+            Optional<Scope> approvingScope = grantedScopes.get(context).stream()
```
punktilious, line 732: comment here would be useful on what this section is intending to achieve

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
```diff
@@ -695,51 +713,50 @@ private void enforce(String resourceType, String resourceId, Resource resource,
...
+                    .filter(s -> s.getResourceType() == ResourceType.Value.RESOURCE || s.getResourceType().value().equals(resourceType))
+                    .filter(s -> hasPermission(s.getPermission(), requiredPermission))
+                    .findAny();
+            if (approvingScope.isPresent()) {
+                if (log.isLoggable(Level.FINE)) {
```
punktilious, line 737: I noticed that in other places we're not logging which scope resulted in the request being approved. Perhaps approvals should be logged more broadly with FINER?

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
```diff
@@ -695,51 +713,50 @@ private void enforce(String resourceType, String resourceId, Resource resource,
...
 
-        if (approvedScopeMap.containsKey(ContextType.PATIENT)) {
+        if (grantedScopes.containsKey(ContextType.PATIENT)) {
+            Optional<Scope> approvingScope = grantedScopes.get(ContextType.PATIENT).stream()
+                    .filter(s -> s.getResourceType() == ResourceType.Value.RESOURCE || s.getResourceType().value().equals(resourceType))
+                    .filter(s -> hasPermission(s.getPermission(), requiredPermission))
+                    .findAny();
+
             if (resource == null) {
                 // For `patient` scopes, if no resource was retrieved, which happens with history/search
```
punktilious, line 752: what about deleted resources? The resource value will be null for those too

punktilious (commented):
Also in the file, `therefor` should be `therefore`

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
lmsurpre: will add a comment like this:
> // look for any scope that approves the current interaction

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
lmsurpre: we don't call this from `afterDelete`, but we do call it from `afterHistory` and so I agree this is worth testing...I'll give it a try and maybe add a test

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
lmsurpre: We don't have a good "single spot" to add this logging, but I could add it to the "isApprovedByScopes" method which is used from a number of thee `beforeX` methods.  I think log level FINE is ok for these and would be consistent with the other spots.

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
lmsurpre: this wasn't really from this PR, but instead from https://github.com/IBM/FHIR/issues/3423
however, i confirmed that, with those changes, we now reject a history interaction if any one of the resource versions in the history was a delete...i'll see if i can address that in this PR.  if not, i'll open a separate issue for it

tbieste (approved):

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
lmsurpre: added

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
lmsurpre: added

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
lmsurpre: I made some changes to try addressing this one.  please have another look

punktilious (approved):
LGTM

PR #3504, opened 2022-03-24
issue #3390 - add implicit search/history _type scoping

lmsurpre:
When the fhirServer/resources config restricts search or history on one
or more resource types, we now add an implicit _type parameter to the
request to filter it down to just the resource types that support the
given interaction.


fhir-server-test/src/test/java/com/ibm/fhir/server/test/SystemHistoryTest.java
```diff
@@ -464,10 +465,36 @@ public void testSystemHistoryWithTypePatientAndOrderDESC() throws Exception {
...
             }
         } while (count > 0);
-        
+
         assertTrue(found, "Patient id in history");
     }
 
+    @Test
+    public void testSystemHistoryWithNoType_tenant1() throws Exception {
+        WebTarget target = getWebTarget();
```
tbieste, line 474: Since this test is similar to SearchAllTest:testSearchAllUsingId_tenant1, I think adding a similar comment like "tenant1 is configured to support history on only a subset of resource types" would make it clear why the `_type` parameter should exist in the self link.

tbieste (approved):

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SystemHistoryTest.java
lmsurpre: done

punktilious (approved):
LGTM

Issue #3501, opened 2022-03-24
Instance-level history fullUrl is wrong when resource id begins with resource type name


lmsurpre:
**Describe the bug**
While testing instance-level history, I noticed that the Bundle.entry.fullUrl we are returning are duplicating the resource type in their paths when the resource id starts with the resource type name (e.g. "Patient1")

Looks like this is the same as #2965 except for fullUrls (instead of location URIs)

**Environment**
main

**To Reproduce**
1. create a resource on the server with an id that begins with a resource type name (e.g. Patient/Patient1)
2. retrieve the instance-level history (e.g. Patient/Patient1/_history)

Note that the fullUrl contains an extra type in the path (e.g. https://localhost:9443/fhir-server/api/v4/Patient/Patient/Patient1)

**Expected behavior**
The fullUrl should be the correct URL to the resource

**Additional context**
I also noticed a couple other differences between our instance-level history and the new system and type-level history endpoints.
1. instance-level history returns a fullUrl that includes the full base url of the server whereas our new system and type-level history responses which include only a relative URL.
2. instance-level history includes (Response.status, Response.etag, and Response.lastModified) whereas the system and type-level history responses include (Response.status, Response.location, and Response.lastModified)
3. instance-level history sets Response.status to "200" for creates; system and type-level history set these to "201 Created"

For number 1, I actually think the instance-level interaction is more correct... @punktilious how costly would it be to return full url instead of relative url in our system and type-level history responses?

For number 2, I think we should add Response.location to the instance-level response.  Maybe a separate ticket?

For number 3, I think we should modify the instance-level response to use 201 for creates to match.  We should also be consistent as to whether we include just the numerical HTTP response code ("200") or also include the response code title ("200 OK").  Separate issue?

lmsurpre:
pointed to account for the changes in "additional context" as well

lmsurpre:
For number 3 (`200` vs `200 OK`) I check our current batch processing code and we use the number without the string there, so lets normalize on that (i.e. change the system-level history to match)

d0roppe:
verified that the _history full url is now correct with a couple of different resources. Closing issue

PR #3506, opened 2022-03-24
issue #3501 - apply fix for #2965 more universally

lmsurpre:
We had duplicate logic in two classes:
* FHIRResource.getRequestBaseUri
* FHIRRestHelper.getRequestBaseUri

For #2965, we fixed the version in FHIRResource but missed the one in
FHIRRestHelper.

Now I combined the two implementations into FHIRRestHelper, made it
static, and removed the version from FHIRResource.

I also added commits for the discrepencies between system-level and
instance-level history identified in #3501


punktilious (approved):
LGTM

Issue #3509, opened 2022-03-25
fhir-bucket tries using COSClient before its initialized


lmsurpre:
**Describe the bug**
I was seeing the following error from fhir-bucket after attempting a new load:
> SEVERE: Error during COS scan: null

After digging in a bit, I found that its actually caused by a NullPointer.
The CosScanner gets initialized with a null COSClient and then tries to use that.

**Environment**
Which version of IBM FHIR Server?

**To Reproduce**
1. upload some resources to COS
2. install a fresh deployment with a fresh db
3. run fhir-bucket with `--incremental` and some cos-properties

**Expected behavior**
It works (or at least provides a clearer error message)

**Additional context**

PR #3510, opened 2022-03-25
issue #3509 - initialize cosClient before its first use

lmsurpre:
and log the full exception if there's an issue during COS scanning

### DIFF
--- a/fhir-bucket/src/main/java/com/ibm/fhir/bucket/app/Main.java
+++ b/fhir-bucket/src/main/java/com/ibm/fhir/bucket/app/Main.java
@@ -611,6 +611,10 @@ public class Main {
             }
         }
 
+        if (cosProperties != null && cosProperties.size() > 0) {
+            cosClient = new COSClient(cosProperties);
+        }
+
         // We constrain the number of concurrent tasks which are inflight, so a breathable
         // pool works nicely because it cannot grow unbounded
         this.commonPool = Executors.newCachedThreadPool();
@@ -993,11 +997,7 @@ public class Main {
             if (!this.cosBucketList.isEmpty()) {
                 // Process using COS as our source data repository
                 if (cosClient == null) {
-                    if (cosProperties != null && cosProperties.size() > 0) {
-                        cosClient = new COSClient(cosProperties);
-                    } else {
-                        throw new IllegalArgumentException("COS configuration required");
-                    }
+                    throw new IllegalArgumentException("COS configuration required");
                 }
 
                 final IResourceEntryProcessor resourceEntryProcessor;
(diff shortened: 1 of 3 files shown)


fhir-bucket/src/main/java/com/ibm/fhir/bucket/app/Main.java
```diff
@@ -992,12 +992,14 @@ protected void scanAndLoad() {
 
             if (!this.cosBucketList.isEmpty()) {
                 // Process using COS as our source data repository
-                if (cosProperties != null && cosProperties.size() > 0) {
-                    cosClient = new COSClient(cosProperties);
-                } else {
-                    throw new IllegalArgumentException("COS configuration required");
+                if (cosClient == null) {
```
punktilious, line 995: I wonder if we should have the COSClient initialization just in one place. The issue is how we decide whether or not we need it. Perhaps a method that creates it on demand if it doesn't exist?

fhir-bucket/src/main/java/com/ibm/fhir/bucket/scanner/CosScanner.java
```diff
@@ -149,7 +153,7 @@ public void mainLoop() {
                 }
             } catch (Exception x) {
                 // Just catch and log so we don't break the main loop
-                logger.severe("Error during COS scan: " + x.getMessage());
+                logger.log(Level.SEVERE, "Error during COS scan", x);
```
punktilious, line 156: I think I was finding the stack traces annoying :-)

fhir-bucket/src/main/java/com/ibm/fhir/bucket/app/Main.java
lmsurpre: my we just add it to the current `configure()` method an instantiate it whenever we have some `cosProperties` passed to us?

in main, configure() is called right before process(), and that is where we set up the db stuff already...

fhir-bucket/src/main/java/com/ibm/fhir/bucket/scanner/CosScanner.java
lmsurpre: hopefully we've fixed the cases where we see this over and over and want to ignore it... without the stack trace its really hard to know what went wrong

fhir-bucket/src/main/java/com/ibm/fhir/bucket/app/Main.java
lmsurpre: ok, i pushed a commit that does this...let me know what you think

punktilious (approved):
LGTM

Issue #3270, opened 2022-02-02
value set 'http://hl7.org/fhir/us/core/ValueSet/us-core-usps-state' is empty or could not be expanded


lmsurpre:
**Is your feature request related to a problem? Please describe.**
In the US Core IG, they have a ValueSet for US Postal Service states:  http://hl7.org/fhir/us/core/STU4/ValueSet-us-core-usps-state.json.html

Unfortunately, the codesystem they pull from isn't defined in FHIR:
```
  "compose" : {
    "include" : [
      {
        "system" : "https://www.usps.com/"
      }
    ]
  }
```

Our current validation will therefor result in the following warning:
```
    {
        "severity": "warning",
        "code": "incomplete",
        "expression": ["Patient.address[0].state"],
        "details": {"text": "Membership check was not performed: value set 'http://hl7.org/fhir/us/core/ValueSet/us-core-usps-state' is empty or could not be expanded"}
    }
```

**Describe the solution you'd like**
An expanded version of this valueset available in the registry so that we can use it to validate

**Describe alternatives you've considered**
document how to use the remote term service provider for this codesystem

**Acceptance Criteria**

1.  
   GIVEN our us-core jar is in userlib
    AND a patient resource that claims conformance to us-core-patient and has an address with a state element that has a code from http://hl7.org/fhir/us/core/ValueSet/us-core-usps-state'
   WHEN you create/update (with `Prefer: return=OperationOutcome`) or $validate the resource
   THEN there are no informational or warning issues about being unable to validate the code

1.  
   GIVEN our us-core jar is in userlib
    AND a patient resource that claims conformance to us-core-patient and has an address with a state element that has a value which **is not** from http://hl7.org/fhir/us/core/ValueSet/us-core-usps-state'
   WHEN you create/update (with `Prefer: return=OperationOutcome`) or $validate the resource
   THEN there is an informational issue about being unable to validate the code

**Additional context**

lmsurpre:
consider looking to package a set of pre-expanded valuesets because there are other valuesets like 'http://hl7.org/fhir/us/core/STU4/ValueSet-us-core-condition-code.html' that are even more complicated to expand (requires SNOMED CT support)

d0roppe:
Verified that this now does not issue any warning if you use a valid USPS Code for the state

PR #3523, opened 2022-03-28
issue #3270 - package expanded us-core-usps-state ValueSet

lmsurpre:
The ValueSet in US Core does not include the actual codes. Rather than
add a CodeSystem resource for the codes, we replace the packaged
ValueSet with a pre-expanded version that lists all the codes.


punktilious (approved):
LGTM

Issue #3507, opened 2022-03-25
"Undelete" action is logged as an update in the changes table but a create in audit


lmsurpre:
**Describe the bug**
We're somewhat inconsistent in the way we handle the "undelete" case (an update after a delete).

Our server returns a 201 (Created) status and we now log this as a create in the audit log (as of https://github.com/IBM/FHIR/issues/2471).

This matches what the spec says (from https://www.hl7.org/fhir/http.html#delete)
> If the interaction is successful, the server SHALL return either a 200 OK HTTP status code if the resource was updated, or a 201 Created status code if the resource was created (or brought back to life/re-created)

Prior to https://github.com/IBM/FHIR/issues/3293 we were calling the `beforeUpdate` and `afterUpdate` methods for this case.
Therefor, the notification events we emitted also called this an `update`.
But now that the "prevResource" is null for this "undelete" case, we now call `beforeCreate` and `afterCreate` and therefor this interaction will lead to a `create` action code in the notification event going forward.
Furthermore, we also call FHIRPersistence.create (rather than FHIRPersistence.update).

However, when we insert into the RESOURCE_CHANGE_LOG table, we're still setting the changeType to U (and not C).
This leads to the _history api using the status code `200` for these requests, even though the actual status code was a 201.

**Environment**
main

**To Reproduce**
1. create a resource
2. delete it
3. update it (bringing it "back to life")
4. invoke system or type-level history

Note that the entry for step 3 indicates that the result was 200 OK even though the actual interaction resulted in a 201 Created

**Expected behavior**
The history api returns a status code that accurately reflects the actual status code for the "undelete" interaction.

**Additional context**
Originally identified at https://github.com/IBM/FHIR/issues/2988#issuecomment-1020241412

lmsurpre:
I think it will require an update to the stored procedures (and to DerbyResourceDAO)

lmsurpre:
For system and type-level history interactions, we can easily start writing those to the CHANGE_LOG table as creates ('C') rather than updates ('U').

However, for instance-level history its not so simple.  To do it properly (where the Bundle.entry.response element always reflects the proper HTTP method and response status of that specific interaction) I think we'd need to reimplement this instance-level history on top of the CHANGE_LOG table. For now, I'm thinking to just fix the system and type-level interactions... maybe we can open a separate issue for the instance-level history?

d0roppe:
From testing this with the latest code, I am still seeing code 200 on a recreate for the _history

[...]

PR #3524, opened 2022-03-28
issue #3507 - treat "undelete" updates like create-on-update in changes

lmsurpre:
Update db2 and postgresql add_any_resource.sql stored procs (and
DerbyResourceDAO.storeResource) to write a 'C' instead of a 'U' for this
case. This way, system and type-level history interactions can properly
report the response status as a 201 instead of a 200.

Update FHIRRestHelper.createHistoryBundle to handle this same case for
the instance-level history. Note, however, that there are ongoing
limitations for the accuracy of the instance-level history entries as
outlined at https://github.com/IBM/FHIR/issues/3507#issuecomment-1081157116

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/derby/DerbyResourceDAO.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/derby/DerbyResourceDAO.java
@@ -203,9 +203,9 @@ public class DerbyResourceDAO extends ResourceDAOImpl {
      * @return the resource_id for the entry we created
      * @throws Exception
      */
-    public long storeResource(String tablePrefix, List<ExtractedParameterValue> parameters, 
+    public long storeResource(String tablePrefix, List<ExtractedParameterValue> parameters,
             String p_logical_id, InputStream p_payload, Timestamp p_last_updated, boolean p_is_deleted,
-            String p_source_key, Integer p_version, String p_parameterHashB64, Connection conn, 
+            String p_source_key, Integer p_version, String p_parameterHashB64, Connection conn,
             ParameterDAO parameterDao, Integer ifNoneMatch, String resourcePayloadKey,
             AtomicInteger outInteractionStatus, AtomicInteger outIfNoneMatchVersion) throws Exception {
 
@@ -452,7 +452,7 @@ public class DerbyResourceDAO extends ResourceDAOImpl {
             stmt.setLong(1, v_resource_id);
             stmt.setLong(2, v_logical_resource_id);
             stmt.setInt(3, p_version);
-            
+
             if (p_payload != null) {
                 stmt.setBinaryStream(4, p_payload);
             } else {
@@ -521,7 +521,7 @@ public class DerbyResourceDAO extends ResourceDAOImpl {
 
         // Finally, write a record to RESOURCE_CHANGE_LOG which records each event
         // related to resources changes (issue-1955)
-        String changeType = p_is_deleted ? "D" : v_new_resource ? "C" :  "U";
+        String changeType = p_is_deleted ? "D" : (v_new_resource || v_currently_deleted) ? "C" : "U";
         String INSERT_CHANGE_LOG = "INSERT INTO resource_change_log(resource_id, change_tstamp, resource_type_id, logical_resource_id, version_id, change_type)"
                 + " VALUES (?,?,?,?,?,?)";
         try (PreparedStatement ps = conn.prepareStatement(INSERT_CHANGE_LOG)) {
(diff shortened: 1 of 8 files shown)


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/derby/DerbyResourceDAO.java
```diff
@@ -203,9 +203,9 @@ public Resource  insert(Resource resource, List<ExtractedParameterValue> paramet
      * @return the resource_id for the entry we created
      * @throws Exception
      */
-    public long storeResource(String tablePrefix, List<ExtractedParameterValue> parameters, 
+    public long storeResource(String tablePrefix, List<ExtractedParameterValue> parameters,
```
punktilious, line 206: There's also PostgresResourceNoProcDAO which I try to keep in-sync - it is handy sometimes to use instead of the stored proc when investigating issues. It replicates the logic of the stored procedure using discrete statements.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -2317,8 +2317,8 @@ private void validateReferenceParams(List<QueryParameter> chainQueryParameters,
     /**
      * Creates a bundle that will hold the results of a history operation.
      *
-     * @param resources
-     *            the list of resources to include in the bundle
+     * @param resourcesResults
+     *            the list of resource results to include in the bundle, sorted with the more recent changes first
```
punktilious, line 2321: This depends on the _sort request, surely?

lmsurpre: oh, did you add _sort support for instance-level history as well?

punktilious: oh, I thought this bit was common for the history calls...in this case the sorting is fixed according to the SQL_HISTORY_FROM_DATETIME query in ResourceDAOImpl

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/postgres/PostgresResourceNoProcDAO.java
```diff
@@ -385,27 +390,50 @@ public long storeResource(String tablePrefix, List<ExtractedParameterValue> para
             stmt.setLong(1, v_resource_id);
             stmt.setLong(2, v_logical_resource_id);
             stmt.setInt(3, p_version);
-            stmt.setBinaryStream(4, p_payload);
+
+            if (p_payload != null) {
+                stmt.setBinaryStream(4, p_payload);
+            } else {
+                // payload offloaded to another data store
+                stmt.setNull(4, Types.BLOB);
```
punktilious, line 398: I think for PostgreSQL the  BYTEA type should be `Types.BINARY`. Not sure if this is handled differently than BLOB.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/postgres/PostgresResourceNoProcDAO.java
```diff
@@ -219,15 +222,34 @@ public long storeResource(String tablePrefix, List<ExtractedParameterValue> para
             throw new IllegalStateException("resource type not found: " + v_resource_type);
         }
 
+        // Get a value for the new resource_id we'll be storing later
+        final String sql2 = "SELECT nextval('fhir_sequence')";
```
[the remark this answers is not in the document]
lmsurpre: I just copied the SQL from line 263.  Let me know if you'd like it changed and, if so, I can do it in both places.

I was also thinking we could have a constant for this SQL (and most of the other SQL in this method) but decided not to worry about that.

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/postgres/PostgresResourceNoProcDAO.java
lmsurpre: I changed to Types.BINARY to match PostgresResourceDAO which is exercised

fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/postgres/PostgresResourceNoProcDAO.java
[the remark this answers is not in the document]
punktilious: I think the IDatabaseAdapter provides it, but not really worth using it here. Shouldn't be an issue.

Issue #3536, opened 2022-03-30
Clarify fallback behavior for `fhirServer/resources` config


lmsurpre:
**Is your feature request related to a problem? Please describe.**
In https://github.com/IBM/FHIR/pull/3473 I added an abstraction layer for working with the `fhirServer/resources` property group.
Currently, this layer uses the `fhirServer/resources` config in whole.
If that doesn't exist in the tenant, it falls back to that property in the "default" tenant.

Where this gets a little confusing is if any of the code uses FHIRConfigHelper to read a sub-property *under* `fhirServer/resources`.

**Describe the solution you'd like**
1a. Ensure that all the places that access properties under `fhirServer/resources` are using the new abstraction layer; or
1b. ensure they do not support falling back

2. document that `fhirServer/resources` is considered as a whole (i.e. we won't fall back to the "default" tenant for specific subproperties)

**Describe alternatives you've considered**
Support tenant config fallback behavior for each property within the `fhirServer/resources` property group.
That would let you set a system-wide default like 
1. ("read, search") on some set of resource types
2. restricted search parameters for some set of resource types

and then selectively override those for specific resource types on a per-tenant basis

This feel quite powerful to me but I think it really complicates our testing because the properties under `fhirServer/resources` all sort of work together; the actual system behavior for a given resource type is determined by the combination "open", "Resource", and the resource-specific config.

**Acceptance Criteria**

1.  
   GIVEN [a precondition]
    AND [another precondition]
   WHEN [test step]
    AND [test step]
   THEN [verification step]
    AND [verification step]

**Additional context**

d0roppe:
Does not seem to be picking up default config correctly if no resource section is specified in the tenant config.

lmsurpre:
i merged a small fix.  specifically during search parameter loading, it was failing to fall back to the `fhirServer/resources` config in the default tenant.  i've addressed that now to match the documentation.

d0roppe:
Verified this is now working as expected, closing issue

PR #3544, opened 2022-03-31
issue #3536 - use ResourcesConfigAdapter from validateInteraction

lmsurpre:
Previously, validateInteraction was using FHIRConfigHelper to look up
sub-properties of fhirServer/resources which, due to fallback behavior,
may conflict with the configured interactions for the tenant.

Because validateInteraction is invoked by most interactions, I moved
ResourcesConfigAdapter initialization to the FHIRRestHelper constructor.
This should help a tad in bundle processing (no need to process this same
config over and over within the scope of a single request).

To avoid a potentially null ResourcesConfigAdapter, I updated
fhir-config to remove some unneccessary "throws Exception" cases...all
stemming from a possible UnsupportedEncodingException for UTF-8 from
FHIRUtilities.decode (a case we should never hit).

### DIFF
--- a/docs/src/pages/guides/FHIRServerUsersGuide.md
+++ b/docs/src/pages/guides/FHIRServerUsersGuide.md
@@ -274,9 +274,9 @@ For example, the fhir-server-config snippet from above would have a correspondin
              password="change-password"
              currentSchema="fhirdata"
          />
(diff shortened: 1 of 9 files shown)


punktilious (approved):
LGTM

PR #3561, opened 2022-04-05
issue #3390 - align ResourceTypeName with the R4 ResourceType enum

lmsurpre:
I could have sworn I did this before, but its definitely out-of-sync in
main. This changeset re-aligns the two so that:
1. we don't end up trying to use a resource type that doesn't exit in R4
2. we don't miss any resource types that existed in R4 but are being
removed for R4B


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -2891,7 +2891,14 @@ public ResourcePayload fetchResourcePayloads(Class<? extends Resource> resourceT
             if (resourceTypeNames != null && resourceTypeNames.size() > 0) {
                 // convert the list of type names to the corresponding list of resourceTypeId values
                 // the REST layer already has checked these names are valid, so no need to worry about failures
-                resourceTypeIds = resourceTypeNames.stream().map(n -> cache.getResourceTypeCache().getId(n)).collect(Collectors.toList());
+                resourceTypeIds = resourceTypeNames.stream()
+                        .map(n -> cache.getResourceTypeCache().getId(n))
+                        .filter(Objects::nonNull)
+                        .collect(Collectors.toList());
+
+                if (resourceTypeIds.isEmpty()) {
```
punktilious, line 2899: resourceTypeIds.size != resourceTypeNames.size()

punktilious (approved):
LGTM

Issue #3565, opened 2022-04-06
docker run should support control+c to exit process


lmsurpre:
**Is your feature request related to a problem? Please describe.**
1. `docker run -p 9443:9443 ibmcom/ibm-fhir-server`
2. control+c

Note that nothing happens.

I thought that was 'normal' docker behavior (you need to pass `-t` to be able to exit).

However, after some further reading, it turns out that this behavior comes from the fact our main process (pid=1) on the container is a shell script.

From from https://engineeringblog.yelp.com/2016/01/dumb-init-an-init-for-docker.html#process-behavior-inside-docker-containers
> Having a shell as PID 1 actually makes signaling your process almost impossible. Signals sent to the shell won’t be forwarded to the subprocess, and the shell won’t exit until your process does. The only way to kill your container is by sending it SIGKILL (or if your process happens to die).

> For this reason, you should always try to avoid spawning a shell. If you can’t easily avoid that (for example, if you want to spawn two processes), you should [exec](http://manpages.debian.org/cgi-bin/man.cgi?query=exec&apropos=0&sektion=0&manpath=Debian+8+jessie&format=html&locale=en) your last process so that it replaces the shell.
```

**Describe the solution you'd like**
update our bootstrap.sh script to use `exec` when it calls the liberty script

**Describe alternatives you've considered**


**Acceptance Criteria**

1.  
   GIVEN a running ibm-fhir-server container running on the foreground with no psuedo-tty (e.g. from `docker run -p 9443:9443 ibmcom/ibm-fhir-server`)
   WHEN you press control+c
   THEN it should kill the container


**Additional context**

lmsurpre:
note:  Control+C actually sends a SIGINT, not a SIGEXIT or SIGQUIT

d0roppe:
Verified that the CTL-c does now cause the Docker container to stop Closing issue

PR #3567, opened 2022-04-06
issue #3565 - use exec in docker container launch scripts

lmsurpre:
this way you can exit the containers via control+c when they are launch
on the foreground with a command like `docker run`


punktilious (approved):
LGTM

PR #3571, opened 2022-04-08
R4b rebased 2

lmsurpre:
r4b-rebased but rebased onto main after 4.11.0


lmsurpre:
force pushing this to `r4b-rebased` now that the tests are passing

Issue #3576, opened 2022-04-14
POST requests without an ID cause a NPE with fhir-smart enabled


SibrenTalens:
**Describe the bug**

A `POST` request for a new resource (as tested, `Observation`) without an `id` causes a `NullPointerException` because of [this](https://github.com/IBM/FHIR/blob/4.11.0/fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java#L823) `Objects.requireNonNull` check in `fhir-smart`. When an `id` is present, it is ignored [in accordance to the specifications](https://www.hl7.org/fhir/http.html#create) and the resource is created as usual with a server generated ID.

**Environment**
`4.11.0`, using the Docker image [`docker.io/ibmcom/ibm-fhir-server`](https://hub.docker.com/r/ibmcom/ibm-fhir-server) with `fhir-smart` enabled.

**To Reproduce**
Perform a create by sending a `POST` request with a new resource to an endpoint without an `id` present in the request body. This will cause a `NullPointerException` and a `500 Internal Server Error` response without the resource being created.

**Expected behavior**
A `201 Created` response for POST requests without an `id` key present, and the resource being created with an ID generated by the server. 

**Additional context**
Request headers:
```
POST /fhir-server/api/v4/Observation HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate, br
Accept-Language: en-US,en;q=0.9
Authorization: Bearer <redacted>
Connection: keep-alive
Content-Length: 458
Content-Type: application/json
Host: localhost:9080
Origin: http://localhost:3000
Prefer: return=representation
Referer: http://localhost:3000/
```

Request JSON:
```json
{
  "resourceType":"Observation",
  "status":"final",
  "effectiveDateTime":"2022-04-14T12:38:18+02:00",
  "subject":{
    "reference":"Patient/001467ad114848ad824fc6878ef9b267"
  },
  "code":{
    "coding":[
      {
        "code":"29463-7",
        "system":"http://loinc.org"
      }
    ]
  },
  "category":[
    {
      "coding":[
        {
          "code":"vital-signs",
          "display":"Vital Signs",
          "system":"http://hl7.org/fhir/observation-category"
        }
      ]
    }
  ],
  "note":[
    
  ],
  "valueQuantity":{
    "unit":"kg",
    "system":"http://unitsofmeasure.org",
    "value":123
  },
  "component":[
    
  ]
}
```

Logs:
```
[...]

PR #3580, opened 2022-04-19
issue #3576 - remove unnecessary guard against null resourceId

lmsurpre:
Now that our persistence layer requires resources to have proper
metadata (id, versionId, and lastUpdated) prior to calling create/update/delete, I thought that resources
will always have an id before we call the persistence interceptors.

However, it turns out that we call the beforeCreate from the REST layer
BEFORE setting this metadata on the resource.
After a brief discussion, we decided to leave that as-is and update the
interceptor to allow a null resourceId.

### DIFF
--- a/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
@@ -195,7 +195,7 @@ public class FHIRRestHelper implements FHIRResourceHelpers {
 
     @Override
     public FHIRRestOperationResponse doCreate(String type, Resource resource, String ifNoneExist,
-        boolean doValidation) throws Exception {
+            boolean doValidation) throws Exception {
 
         // Validate that interaction is allowed for given resource type
         validateInteraction(Interaction.CREATE, type);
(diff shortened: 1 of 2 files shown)


fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -195,7 +195,7 @@ public FHIRRestHelper(FHIRPersistence persistence, SearchHelper searchHelper) {
 
     @Override
     public FHIRRestOperationResponse doCreate(String type, Resource resource, String ifNoneExist,
-        boolean doValidation) throws Exception {
+            boolean doValidation) throws Exception {
```
lmsurpre, line 198: unrelated formatting changes i had in my workspace

fhir-smart/src/main/java/com/ibm/fhir/smart/AuthzPolicyEnforcementPersistenceInterceptor.java
```diff
@@ -954,7 +957,7 @@ private boolean isInCompartment(String resourceType, String resourceId, Resource
             for (String searchParmCode : inclusionCriteria) {
                 try {
                     SearchParameter inclusionParm = searchHelper.getSearchParameter(resourceType, searchParmCode);
-                    if (inclusionParm != null & inclusionParm.getExpression() != null) {
```
punktilious, line 957: ouch

punktilious (approved):
LGTM

Issue #3569, opened 2022-04-06
Add demo client for system-system synchronization


punktilious:
**Is your feature request related to a problem? Please describe.**
The whole-system history API can be used for system-to-system synchronization. Create a client application to demonstrate how to read the change history of a server and feed these changes to a downstream system.

The IBM FHIR Server implements an optimization to return only the metadata of the change using the `Prefer: return=minimal` optional header. This can speed up the flow rate because resource values can be read from the upstream system in parallel. Resource values can also be sent to the downstream system in parallel, as long as changes related to a particular resourceType/logicalId are serialized (processed by the same queue/thread).

**Describe the solution you'd like**
An example Java implementation of the above.

**Describe alternatives you've considered**
N/A

**Acceptance Criteria**

1.  
   GIVEN [a precondition]
    AND [another precondition]
   WHEN [test step]
    AND [test step]
   THEN [verification step]
    AND [verification step]

**Additional context**

PR #3584, opened 2022-04-20
issue #3569 fhir-flow project demonstration system-system sync

punktilious:


fhir-bucket/src/main/java/com/ibm/fhir/bucket/app/Main.java
```diff
@@ -199,9 +200,19 @@ public class Main {
     // How many payer scenario requests do we want to make at a time.
     private int concurrentPayerRequests = 0;
 
+    // How many $bench requests to we want to make at a time.
```
lmsurpre, line 203: only if making other changes
```suggestion
    // How many $bench requests do we want to make at a time.
```

fhir-bucket/src/main/java/com/ibm/fhir/bucket/app/Main.java
```diff
@@ -935,6 +975,7 @@ public void process() {
             if (this.targetBucket == null || this.targetBucket.length() == 0) {
                 // Set up the client we use to send requests to the FHIR server
                 fhirClient = new FHIRBucketClient(new ClientPropertyAdapter(fhirClientProperties));
+                fhirClient.addHeader("Prefer", "return=representation");
```
lmsurpre, line 978: I think this will cause the server to include the resource content in the response of each create/udpate...do we really want that?

lmsurpre: I see we used to add this on line 186 of FHIRBucketClient, so I guess its not really a change

fhir-bucket/src/main/java/com/ibm/fhir/bucket/client/FHIRBucketClient.java
```diff
@@ -489,15 +591,21 @@ public Object getUserToken(HttpContext context) {
...
         try {
-            try (InputStream instream = entity.getContent()) {
-                Resource resource = FHIRParser.parser(Format.JSON).parse(new InputStreamReader(instream, StandardCharsets.UTF_8));
-                sr.setResource(resource);
-            } catch (FHIRParserException e) {
-                throw new IllegalStateException(e);
-            } finally {
+            if (parseResource) {
+                try (InputStream instream = entity.getContent()) {
+                    Resource resource = FHIRParser.parser(Format.JSON).parse(new InputStreamReader(instream, StandardCharsets.UTF_8));
```
lmsurpre, line 599: i don't think you need to wrap the instream...our parser will do that for you
```suggestion
                    Resource resource = FHIRParser.parser(Format.JSON).parse(instream);
```

lmsurpre: oops, i guess this was there before as well, probalby fine either way

fhir-bucket/src/main/java/com/ibm/fhir/bucket/app/Main.java
lmsurpre: could use the new Headers.PREFER_HEADER constant

fhir-flow/src/main/java/com/ibm/fhir/flow/api/ResourceIdentifierVersion.java
```diff
@@ -0,0 +1,87 @@
...
+
+import com.ibm.fhir.model.type.Uri;
+import com.ibm.fhir.model.util.ModelSupport;
+
+/**
+ * A resource type, id and version tuple
+ */
+public class ResourceIdentifierVersion extends ResourceIdentifier {
+    // the resource meta.versionId value
+    private final int version;
```
lmsurpre, line 19: do we rely on this being an integer?
for our server it always will be, but its not technically a spec requirement (i.e. other servers could potentially use opaque version ids).  wasn't sure if we only care about our server or wanted to be more general.

fhir-flow/src/main/java/com/ibm/fhir/flow/impl/UpstreamFHIRHistoryReader.java
```diff
@@ -0,0 +1,291 @@
...
+     * @param bundle
+     */
+    private List<ITrackerTicket> addTrackerValues(Bundle bundle) {
+
+        List<Long> requestIds = new ArrayList<>(bundle.getEntry().size());
+        for (Bundle.Entry entry: bundle.getEntry()) {
+            // The entry.id field is database resourceId which is used as the PK for the
+            // changes table. It just needs to be unique and match the change sequence. It
+            // does not need to be contiguous
+            requestIds.add(Long.parseLong(entry.getId()));
```
lmsurpre, line 287: just wanted to flag this as another case where its dependent on our server's unique behavior.  probably fine, but thought it was worth mentioning.

fhir-server-webapp/pom.xml
```diff
@@ -79,6 +79,11 @@
             <artifactId>fhir-operation-versions</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>fhir-operation-bench</artifactId>
```
lmsurpre, line 84: i'm not sure if we should have this in our server by default.  maybe keep in fhir-parent but remove from here?  do any e2e tests in the CI pipeline depend on it?

operation/fhir-operation-bench/src/main/java/com/ibm/fhir/operation/bench/BenchOperation.java
```diff
@@ -0,0 +1,303 @@
...
+            InputOutputByteStream ioStream = new InputOutputByteStream(DATA_BUFFER_INITIAL_SIZE);
+
+            // Serialize and compress the Resource
+            try (GZIPOutputStream zipStream = new GZIPOutputStream(ioStream.outputStream())) {
+                FHIRGenerator.generator(Format.JSON, false).generate(resource, zipStream);
+                zipStream.finish();
+
+                // Now parse the generated resource again
+                try (GZIPInputStream inStream = new GZIPInputStream(ioStream.inputStream())) {
+                    FHIRParser.parser(Format.JSON).parse(new InputStreamReader(inStream, StandardCharsets.UTF_8));
```
lmsurpre, line 239: no need to wrap the inStream (our Parser does that automatically if you pass an InputStream)

lmsurpre (changes requested):
The only comment that really needs to be addressed is the one about removing $bench from the default webapp.

fhir-bucket/src/main/java/com/ibm/fhir/bucket/app/Main.java
punktilious: fixed

fhir-bucket/src/main/java/com/ibm/fhir/bucket/app/Main.java
punktilious: yeah it was just a slight refactor so I wanted to keep it the same. Now using the Headers constant.

fhir-flow/src/main/java/com/ibm/fhir/flow/api/ResourceIdentifierVersion.java
punktilious: I'm so used to think of version as an integer...in this case it doesn't actually matter to us because we only use the version for VREAD, so I guess we should keep it as a string.

fhir-flow/src/main/java/com/ibm/fhir/flow/impl/UpstreamFHIRHistoryReader.java
punktilious: yeah, perhaps this is something we should bring up with the spec when dealing with Bundles and multiple pages of results. Right now, the id only needs to be unique within a Bundle...but for pagination I think there are additional uniqueness/sequence semantics which could be useful like above.

fhir-server-webapp/pom.xml
punktilious: Right. We certainly wouldn't want it exposed by default. I'll take it out for now.

operation/fhir-operation-bench/src/main/java/com/ibm/fhir/operation/bench/BenchOperation.java
punktilious: ok

fhir-bucket/src/main/java/com/ibm/fhir/bucket/client/FHIRBucketClient.java
punktilious: done

fhir-flow/src/main/java/com/ibm/fhir/flow/impl/UpstreamFHIRHistoryReader.java
lmsurpre: I don't think HL7 will be interested in further constraining the paging behavior.
They base theirs on RFC 5005 and they keep it rather opaque...everyone does it a little different.

The ids we set there have no defined meaning int the spec...those fields are really for "internal linking"
https://www.hl7.org/fhir/element-definitions.html#Element.id but in our case its kind of handy to emit this extra info without needing to define an extension for it with some defined semantics.  IMHO clients really shouldn't depend on it unless we document it.

lmsurpre (approved):

PR #3585, opened 2022-04-20
issue #1412 - only rebuild the last release

lmsurpre:
The rebuild process has been failing for the prior releases, but I think it might work starting with 4.11.0


punktilious (approved):
LGTM

PR #3586, opened 2022-04-20
issue #3576 - remove Resource.id passed via POST

lmsurpre:
prior to calling doCreate on the interceptors


punktilious (approved):
LGTM. Hopefully not a common occurrence due to having to rebuild the resource just to null out the id.

Issue #3592, opened 2022-04-22
fhir-smart should allow read access to Patient resources in compartment of the patient context


lmsurpre:
**Is your feature request related to a problem? Please describe.**
Currently, fhir-smart controls "direct" (read, vread, delete, and history-instance) access to the Patient resource based on the resource id.
For almost every other resource type, its based on compartment membership instead.

**Describe the solution you'd like**
With a `patient/Patient.read` scope, a user should be able to access the Patient resources that are linked to the in-context patient.

**Describe alternatives you've considered**

**Acceptance Criteria**

1.  
   GIVEN a Patient resource on the server with a link to a specific patient (e.g. Patient/123)
    AND a valid access token with a patient_id that includes that patient id (e.g. 123)
   WHEN the user performs a direct interaction (read, vread, delete, and history-instance) with that resource
   THEN the interaction is permitted

1.  
   GIVEN a valid access token with a patient_id
   WHEN the user attempts to performs a direct interaction (read, vread, delete, and history-instance) with any Patient resource (existent or not) 
   THEN the interaction is rejected with an HTTP 403 (Forbidden)

**Additional context**
I think the reason we originally prohibited any interaction with any Patient resource with any other id was because we didn’t want to leak whether or not that other patient id was a valid id.  If the patient doesn't exist it still needs to be a 403 (not a 404).

d0roppe:
Verified this is working as expected with SMART on FHIR.  Closing issue

PR #3595, opened 2022-04-24
issue #3592 - use compartment membership for patient resource type

lmsurpre:
Previously, for scopes granting patient-scoped Patient resource access
(patient/Patient.read or patient/Patient.write), the beforeRead, beforeVread, and beforeHistory
methods prevented any interaction with any Patient resource unless that
resource had one of the ids passed in the `patient_id` claim of the
access token.

Now we attempt to perform those interactions and rely on the afterRead, 
afterVread, and afterHistory methods to ensure that this is part of the 
in-context patient(s)' record.
To avoid leaking "does patient id xyz exist" info, we reject reading of
non-existent patient resources with a 403 (unlike the usual 404)...this
also prevents 'create-on-update' behavior for this resource type.

To accommodate the change, FHIRRestHelper.doRead is updated so that
`afterRead` is invoked prior to throwing the
FHIRPersistenceResourceNotFoundException that becomes a 404.

To avoid leaking the existence (or nonexistence) of a patient resource id 
in the case of a read for a deleted resource, I changed the persistence layer to
avoid throwing.  That was needed to ensure that the `afterRead` gets
called for this case.  This partially addresses #194.

After that change, the 'includeDeleted' flag in our FHIRRestHelper doRead
methods didn't really make sense and so I removed that. To keep the refactor
manageable, the FHIRRestHelper will still throw (conditionally) in the case of
a null resource in the SingleResourceResult from the persistence layer.

Finally, since we already had this signature change to some FHIRRestHelper methods,
I also addressed #3606 by removing the "contextResource" parameter from
doRead and doSearch.

### DIFF
--- a/cql/fhir-cql-server/src/main/java/com/ibm/fhir/cql/engine/server/retrieve/ServerFHIRRetrieveProvider.java
+++ b/cql/fhir-cql-server/src/main/java/com/ibm/fhir/cql/engine/server/retrieve/ServerFHIRRetrieveProvider.java
@@ -94,7 +94,7 @@ public class ServerFHIRRetrieveProvider extends SearchParameterFHIRRetrieveProvi
         try {
             if (queryParameters.containsKey("_id")) {
                 String id = queryParameters.getFirst("_id");
-                SingleResourceResult<?> result = resourceHelpers.doRead(dataType, id, true, null);
+                SingleResourceResult<?> result = resourceHelpers.doRead(dataType, id, true);
                 if (result.isSuccess()) {
                     resource = result.getResource();
                 }
--- a/cql/fhir-cql-server/src/main/java/com/ibm/fhir/cql/engine/server/retrieve/ServerFHIRRetrieveProvider.java
+++ b/cql/fhir-cql-server/src/main/java/com/ibm/fhir/cql/engine/server/retrieve/ServerFHIRRetrieveProvider.java
@@ -94,7 +94,7 @@ public class ServerFHIRRetrieveProvider extends SearchParameterFHIRRetrieveProvi
         try {
             if (queryParameters.containsKey("_id")) {
                 String id = queryParameters.getFirst("_id");
-                SingleResourceResult<?> result = resourceHelpers.doRead(dataType, id, true);
+                SingleResourceResult<?> result = resourceHelpers.doRead(dataType, id, FHIRResourceHelpers.THROW_EXC_ON_MISSING);
                 if (result.isSuccess()) {
                     resource = result.getResource();
                 }
--- a/cql/fhir-cql-server/src/main/java/com/ibm/fhir/cql/engine/server/retrieve/ServerFHIRRetrieveProvider.java
+++ b/cql/fhir-cql-server/src/main/java/com/ibm/fhir/cql/engine/server/retrieve/ServerFHIRRetrieveProvider.java
@@ -94,7 +94,7 @@ public class ServerFHIRRetrieveProvider extends SearchParameterFHIRRetrieveProvi
         try {
             if (queryParameters.containsKey("_id")) {
                 String id = queryParameters.getFirst("_id");
-                SingleResourceResult<?> result = resourceHelpers.doRead(dataType, id, FHIRResourceHelpers.THROW_EXC_ON_MISSING);
+                SingleResourceResult<?> result = resourceHelpers.doRead(dataType, id);
                 if (result.isSuccess()) {
                     resource = result.getResource();
                 }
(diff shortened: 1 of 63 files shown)


fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -1111,7 +1107,8 @@ public SingleResourceResult<? extends Resource> doRead(String type, String id, b
      * @return a {@link SingleResourceResult} containing the Resource
      * @throws Exception
      */
-    private SingleResourceResult<? extends Resource> doRead(String type, String id, boolean throwExcOnNull, boolean includeDeleted,
+    // TODO remove contextResource?  it doesn't seem to ever be populated...
```
lmsurpre, line 1110: @punktilious do you happen to know what this "contextResource" parameter is for?  
We use it when we construct the persistence event for doRead (line 1146 below), but I havn't found any callers that set it to anything other than null...

lmsurpre: it seems to be a remnant for our DSTU2 server...I see it going back more than 2 years...

cql/operation/fhir-operation-cqf/src/main/java/com/ibm/fhir/operation/cqf/MeasureCollectDataOperation.java
```diff
@@ -108,7 +108,7 @@ protected void resolveReferences(Resource resource, Parameters.Builder parameter
 
                         if( ! resourceMap.containsKey(id) ) {
                             try {
-                                SingleResourceResult<? extends Resource> readResult = resourceHelper.doRead(type, id, /*throwExOnNull=*/true, /*includeDeleted=*/false, /*contextResource=*/null);
+                                SingleResourceResult<? extends Resource> readResult = resourceHelper.doRead(type, id, /*throwExOnNull=*/true, /*contextResource=*/null);
```
punktilious, line 111: Didn't you get rid of contextResource?

lmsurpre: fixed by removing the trailing param.  found one other case like this and fixed that too.  if it were any value other than null it would be a compile error.

lmsurpre: did a search for "contextResource" to confirm I got them all now

cql/fhir-cql-server/src/main/java/com/ibm/fhir/cql/engine/server/retrieve/ServerFHIRRetrieveProvider.java
```diff
@@ -95,12 +94,12 @@ private Resource executeQuery(String dataType, MultivaluedMap<String, String> qu
         try {
             if (queryParameters.containsKey("_id")) {
                 String id = queryParameters.getFirst("_id");
-                SingleResourceResult<?> result = resourceHelpers.doRead(dataType, id, true, false, null);
+                SingleResourceResult<?> result = resourceHelpers.doRead(dataType, id, true);
```
punktilious, line 97: FHIRResourceHelpers has constants for things like DO_VALIDATION. I think a boolean constant for THROW_EXC_ON_NULL would improve readability a bit. But not essential.

lmsurpre: I checked all the callers of the simpler 3-arg `doRead` and I found that they all pass false...what do you think about removing the 3rd arg (forcing anyone that wants to control that behavior into the 4-arg flavor that also takes queryParameters)?

lmsurpre: And actually the 4-arg flavor only has to callers... our JAX-RS Read resource and our FHIRRestInteractionVisitorPersist (called from bundle processing and set to true whether the bundle entry is a read)

lmsurpre: OK, now I added a commit to remove the 'throwExcOnNull' param from the simpler doRead version

punktilious (approved):
LGTM. We'll need to QA the offload stuff again to make sure this hasn't impacted that.

PR #3605, opened 2022-04-26
add 4.11.1 to rebuild list

lmsurpre:
And reduce the schedule from daily to weekly (on tuesdays)


.github/workflows/rebuild.yml
```diff
@@ -4,7 +4,7 @@ on:
   schedule:
     # 7am (UTC) each day
```
lmsurpre, line 5:
```suggestion
    # 7am (UTC) each tuesday
```

Issue #3614, opened 2022-05-03
Make skippable updates the default


lmsurpre:
**Is your feature request related to a problem? Please describe.**
In #2263 we added opt-in support for skipping updates when the the resource content matches the existing resource content.
This seems to be working well, but most clients don't know to opt in for this.

**Describe the solution you'd like**
There seems to already be broad support for this idea in the community (e.g. https://chat.fhir.org/#narrow/stream/179166-implementers/topic/Handling.20of.20Updates/near/270802907 ) and so I think we could remove the client opt-in and just make this the normal behavior for 5.0.

I'm thinking that we retire the custom `X-FHIR-UPDATE-IF-MODIFIED` header (the name was confusing anyway) and always skip the update in this case, but I'd also be fine with renaming the header to something like `X-FHIR-FORCE-UPDATE` (with a default of false).

**Describe alternatives you've considered**
Make it configurable via a server config property instead.

**Acceptance Criteria**

1.  
   GIVEN a patch or update
    AND the updated resource matches the existing one (sans Resource.id, meta.versionId, and meta.lastUpdated)
   WHEN HTTP header `X-FHIR-FORCE-UPDATE` is absent or set to "false"
   THEN the update is skipped

1.  
   GIVEN a patch or update
    AND the updated resource matches the existing one (sans Resource.id, meta.versionId, and meta.lastUpdated)
   WHEN HTTP header `X-FHIR-FORCE-UPDATE` is set to "true"
   THEN the update is performed

**Additional context**

d0roppe:
Verified this is now the new default, closing issue.

PR #3615, opened 2022-05-03
issue #3614 - replace "updateIfModified" with "forceUpdate"

lmsurpre:
Replaced the old "X-FHIR-UPDATE-IF-MODIFIED" header with a new
"X-FHIR-FORCE-UPDATE" one. The default is still false, but now that
means that no-op updates are now skipped by default.

To force the server to perform a no-op update, users must now explicitly
set this X-FHIR-FORCE-UPDATE header to true.


docs/src/pages/guides/FHIRPerformanceGuide.md
```diff
@@ -484,11 +484,11 @@ Avoiding these unnecessary updates is important for two reasons:
...
 1. each update must perform a search which can be more costly than simply performing read before the update
 2. conditional requests require intricate locking techniques to avoid race conditions and the currently-implemented approach has [significant limitations](https://github.com/IBM/FHIR/issues/2051)
 
-Instead, IBM FHIR Server version 4.7.1 introduces support for a server-enabled optimization to avoid performing unnecessary updates. When users pass the `X-FHIR-UPDATE-IF-MODIFIED` header with a value of `true`, the server will perform a comparison of the resource contents from the update with the contents of the resource in the database.
+Instead, IBM FHIR Server version 4.7.1 introduces support for a server-enabled optimization to avoid performing unnecessary updates. The server will only create a new resource version if the resource contents from the update differ from the contents of the resource in the database.
```
punktilious, line 491: should we keep the mention of 4.7.1 here? It's old...perhaps just say "The IBM FHIR Server supports  a server.enabled optimization..."

docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -2764,7 +2764,7 @@ IBM FHIR Server Supports the following custom HTTP Headers:
...
-|`X-FHIR-UPDATE-IF-MODIFIED`|When set to true, for update and patch requests, the server will perform a resource comparison and only perform the update if the contents of the resource have changed. For all other values, the update will be executed as normal.|
+|`X-FHIR-FORCE-UPDATE`|By default, when the server receives an update request, it performs a resource comparison and only performs the update if the contents of the resource have changed. Set this header to true to skip that check and force the update.|
```
punktilious, line 2767: Perhaps we should be clear that by "forcing" an update, we mean performing the standard update interaction which creates a new version of the resource. We're not saying this is an in-place update. Just thinking it might be worth clarifying.

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SystemHistoryTest.java
```diff
@@ -186,10 +187,11 @@ public void testSystemHistoryWithTypePatient() throws Exception {
         int found = 0;
         int count = 10;
         do {
-            Response historyResponse =
-                    target.path(requestPath)
+            Response historyResponse = target.path(requestPath)
                     .queryParam("_count", Integer.toString(count))
-                    .request().get(Response.class);
+                    .request()
+                    .header("X-FHIR-FORCE-UPDATE", "true")
```
punktilious, line 193: why add this to a history GET?

docs/src/pages/guides/FHIRPerformanceGuide.md
lmsurpre: I was thinking of cleaning all such references out when we scrub the docs for the impending rename

docs/src/pages/guides/FHIRServerUsersGuide.md
lmsurpre: maybe something like this?
```suggestion
|`X-FHIR-FORCE-UPDATE`|By default, when the server receives an update request, it performs a resource comparison and only performs the update if the contents of the resource have changed. Set this header to true to skip that check and perform a normal update (i.e. create a new version for this resource id).|
```

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SystemHistoryTest.java
lmsurpre: I added it to a number of tests to force the old behavior...I agree it doesn't make sense for history requests.  Let's remove it.
```suggestion
```

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -821,19 +819,9 @@ public FHIRRestOperationResponse doPatchOrUpdatePersist(FHIRPersistenceEvent eve
...
-                    ior.setLocationURI(FHIRUtil.buildLocationURI(type, id, result.getIfNoneMatchVersion()));
-                    Boolean ifNoneMatchNotModified = FHIRConfigHelper.getBooleanProperty(
-                        FHIRConfiguration.PROPERTY_IF_NONE_MATCH_RETURNS_NOT_MODIFIED, Boolean.FALSE);
-                    if (ifNoneMatchNotModified != null && ifNoneMatchNotModified) {
-                        // Don't treat as an error
-                        ior.setStatus(Response.Status.NOT_MODIFIED);
-                    } else {
-                        throw new FHIRPersistenceIfNoneMatchException("IfNoneMatch precondition failed.");
-                    }
+                    // this should have been handled before the db interaction; this is just a double check
```
punktilious, line 822: It's to avoid a race condition - because this check was integrated into the SELECT FOR UPDATE locking we do, which means we correctly handle the if-none-match case where a resource has been deleted and two threads are trying to create it again.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -872,6 +860,20 @@ public FHIRRestOperationResponse doPatchOrUpdatePersist(FHIRPersistenceEvent eve
         }
     }
 
+    private void handleIfNoneMatchExisted(String type, String id, FHIRRestOperationResponse ior,
+            Integer ifNoneMatchVersion) throws FHIRPersistenceIfNoneMatchException {
+        ior.setResource(null); // the resource shouldn't be needed for either case (302 or 412)
```
punktilious, line 865: I think you mean 304. I guess this works now because we're using the ResourceResult to help build the response.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
lmsurpre: for sure its needed in the case above (`if (createOnUpdate)`), but here we are in the else block which means that we found a resource when we did the read, which means we should have caught it when we called doUpdateMeta

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
lmsurpre: oops, yes, 304
```suggestion
        ior.setResource(null); // the resource shouldn't be needed for either case (304 or 412)
```

docs/src/pages/guides/FHIRPerformanceGuide.md
punktilious: makes sense

fhir-server-test/src/test/java/com/ibm/fhir/server/test/SystemHistoryTest.java
punktilious: agreed

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
punktilious: OK. I didn't get the full context from the diff...I just thought a fuller comment would be helpful

punktilious (approved):
LGTM

PR #3618, opened 2022-05-04
minor tweak to search response with omitted resource content

lmsurpre:
With this change, the search response entries will include resource
version info and a lastModified value that could be useful for clients.


punktilious (approved):
LGTM

Issue #3619, opened 2022-05-05
Populate "path" property in FHIRPathNode implementations representing system values


jarrettranderson:
**Is your feature request related to a problem? Please describe.**
We have a requirement to remove the "id" property from certain elements of a resource.  Consistent with other planned operations, we would find the targets using a FHIRPath expression and then remove them using FHIRPathUtil.delete().  FHIRPathUtil.delete() requires the path to the target, but the path() method returns null for FHIRPathStringValue objects returned from the FHIRPath evaluation.

**Describe the solution you'd like**
The path() method in FHIRPathStringValue objects and probably other FHIRPathSystemValue implementations should return the path to the node like it does for other implementations of FHIRPathNode.

**Describe alternatives you've considered**
For this particular scenario, we could use FHIRPath to find all the Element type objects in the tree and check each of them to see if the "id" value is populated.  We would then update the parent object in a different manner.

**Acceptance Criteria**
1.  
   GIVEN [a FHIRPath expression that targets "id" or another system value field, such as "descendants().select(id)"]
     AND [EvaluationContext evaluationContext = new EvaluationContext(some-resource)]
   WHEN [FHIRPathEvaluator.evaluator().evaluate(evaluationContext, fhir-path-expression]
   THEN [for each object returned, path() is non-null]

**Additional context**

lmsurpre:
Info for executing the fhir-path cli jar can be found at https://github.com/IBM/FHIR/tree/main/fhir-path

lmsurpre:
Per the above I posted the following resource:
```
{
  "resourceType": "Patient",
  "id": "{{patient}}",
  "language": "en",
	"text": {
		"status": "empty",
		"div": "<div xmlns=\"http://www.w3.org/1999/xhtml\">Lee (1234)</div>"
	},
  "identifier": [{
    "use": "usual",
		"_use": {
			"id": "my-primitive-extension",
			"extension": [{
				"url": "http://example.com/primitive-extension",
				"valueString": "test1"
			}]
		},
		"system": "https://example.com",
		"value": "1234"
  }],
	"name": [{
		"given": ["Lee"],
		"_given": [{
			"id": "my-repeat-primitive-extension",
			"extension": [{
				"url": "http://example.com/primitive-extension",
				"valueString": "test2"
			}]
		}]
	}],
  "active": true
}
```

Then I removed the `Patient.identifier.use.id` element via a patch like this:
```
{
	"resourceType": "Parameters",
[...]

Issue #2153, opened 2021-03-26
Make it more obvious how to obtain primitive values from FHIRPathSystemValue


lmsurpre:
**Is your feature request related to a problem? Please describe.**
I wrote some code to extract all the resource ids from a bundle via fhirpath, but it took me longer than it should have to figure out how to get the primitive value out of the FHIRPathNode:
```
        Bundle bundle = load("patients.json");
        Collection<FHIRPathNode> nodes = FHIRPathEvaluator.evaluator().evaluate(bundle,
                "Bundle.entry.resource.id");

        List<Member> members = nodes.stream()
                .filter(node -> node.isSystemValue())
                .map(node -> string("Patient/" + node.howToGetTheValue...
```

Eventually I found the correct answer:
`node.asSystemValue().asStringValue().string()`

What tripped me up is that we have a `FHIRPathNode.getValue()` which I thought would work, but it turns out that actually returns the FHIRPathNode's value (child) when it holds one, not the primitive value of current FHIRPathNode when you already have a FHIRPathSystemValue.

Having this method on instances of FHIRPathSystemValue doesn't make much sense, because FHIRPathSystemValue itself is a terminal node and the `getValue()` just gets in the way of the actual accessors which don't use `get` at all.

**Describe the solution you'd like**
1. override the `getValue()` javadoc for FHIRPathSystemValue to indicate that it will always return null.
2. in the concrete subtypes for FHIRPathSystemValue, add a `@see` link to the actually getters for the given primitive type

**Describe alternatives you've considered**

**Acceptance Criteria**

**Additional context**

PR #3620, opened 2022-05-05
issue #3619 - populate 'path' field for SystemValueNodes in context tree

lmsurpre:
Previously we had "interesting" behavior for fields like Element.id that
are of type System.String (rather than FHIR.String).  Now these nodes
will become children in the EvaluationContext tree.

This allows users to remove fields like Resource.id and Element.id via
FHIRPath and it even allows you to remove the value of primitive FHIR
elements (e.g. Patient.active.value) in cases where those primitives
already have an extension (otherwise they need to have a value).

I also updated the javadoc (relates to #2153) because the user was
confused about the FHIRPathNode API and I thought it could use further
clarification.


JohnTimm (approved):
LGTM

Issue #3372, opened 2022-02-21
documentation for transforming exported ndjson to parquet with spark


lmsurpre:
> I think we should remove this feature and replace it with documentation that clearly shows how to convert the exported NDJSON to Parquet using spark (maybe a blog post?)

this piece of it is not done yet and I think we should do it either for this one or in a new related task

_Originally posted by @lmsurpre in https://github.com/IBM/FHIR/issues/3156#issuecomment-1025874476_

punktilious:
I read through the post again and it seems reasonable. I think this one should be OK to just close.

PR #3626, opened 2022-05-10
issue #3372 - document transforming ndjson export to parquet

lmsurpre:
I also added a byline to the existing blog posts so that readers will know who wrote it and when.


punktilious:
I read through the post again and it seems reasonable. I think this one should be OK to just close.

Issue #3646, opened 2022-05-17
Implement the FHIRPath repeat function


lmsurpre:
**Is your feature request related to a problem? Please describe.**
While adding support for US Core 5.0, I found that their QuestionnaireResponse profile inherits some complex FHIRPath constraints from http://hl7.org/fhir/uv/sdc/STU3/StructureDefinition-sdc-questionnaireresponse.html 

One of those constraints, sdcqr-2, uses the FHIRPath 'repeat' function which is not supported in our FHIRPath evaluator.

**Describe the solution you'd like**
Add support for `repeat` as defined in the FHIRPath specification:  https://hl7.org/fhirpath/#repeatprojection-expression-collection

**Describe alternatives you've considered**
disable or override this constraint with a custom java implementation

**Acceptance Criteria**

1.  
   GIVEN a QuestionnaireResponse that conforms to http://hl7.org/fhir/us/core/StructureDefinition/us-core-questionnaireresponse
   WHEN validated against http://hl7.org/fhir/us/core/StructureDefinition/us-core-questionnaireresponse
   THEN validation should complete with no exceptions

**Additional context**

punktilious:
I downloaded all the JSON example resources from http://hl7.org/fhir/us/core/StructureDefinition-us-core-questionnaireresponse.html and was able to load them successfully using the latest fhir-is-us-core jar:


```
Examples:
# QuestionnaireResponse-AHC-HRSN-screening-example.json
# QuestionnaireResponse-glascow-coma-score.json
# QuestionnaireResponse-hunger-vital-sign-example.json
# QuestionnaireResponse-PRAPARE-example.json


Results:

HTTP/2 201
location: https://localhost:9443/fhir-server/api/v4/QuestionnaireResponse/AHC-HRSN-screening-example/_history/1
etag: W/"1"
last-modified: Thu, 11 Aug 2022 10:30:39 GMT
content-type: application/fhir+json
date: Thu, 11 Aug 2022 10:30:40 GMT
content-language: en-GB
content-length: 488

[...]

Issue #3641, opened 2022-05-16
Support for US Core 5.0


lmsurpre:
US Core 5.0.0 was recently published:  https://hl7.org/fhir/us/core/
We should add that to our prepackaged support in fhir-ig-us-core

lmsurpre:
Hit #3646 which is currently blocking this upgrade.  Since the constraint in question is just a "best practice" one (and not an "validity" one), I'm thinking to use our SPI to avoid firing this constraint for now.  Then we can circle back up and implement the `repeat` function sometime later.

punktilious:
QA for #3646 demonstrates support for US Core 5.0.0, so closing this too.

PR #3647, opened 2022-05-18
issue #3641 - package US Core 5.0.0 artifacts

lmsurpre:
And refine our approach for src/test/java tools to avoid the need to
have a different set of the same classes for each version of the IG.

Initially we were failing to validate the QuestionnaireResponse examples
due to a new constraint that uses an unsupported FHIRPath function.
I added a workaround and opened #3646 to get proper support added.

Additionally, I found that some of the US Core ValueSet resources are no
longer listing the actual code values...it just says they are composed of
value sets that come from the US NIH's VSAC system (value set authority center).
To handle it, I added a fledgling VSACRegistryResourceProvider and then
delegated to our ValueSetSupport.expand logic; now the packaged
ValueSets for 5.0.0 include these pre-expanded ValueSets.


punktilious (approved):
LGTM

PR #3648, opened 2022-05-18
issue #3646 - FHIRPath repeat

lmsurpre:
The implementation delegates to select(args) but then uses that result to
set the context and call select(args) again in a loop until there's no
new nodes to add.

This allows us to remove the workaround that was put in place for
us-core-questionnaire constraint `sdcqr-2`.


conformance/fhir-ig-us-core/src/test/java/com/ibm/fhir/ig/us/core/tool/ResourceProcessor.java
```diff
@@ -36,8 +43,33 @@ public static void main(String[] args) throws Exception {
             throw new IllegalArgumentException("Please specify one or more version qualifiers (e.g. '4.0.0 5.0.0')");
         }
 
-        for (String packageVersion : args) {
-            updateInPlace(packageVersion.replace(".", ""));
+        for (String version : args) {
+            String packageVersion = version.replace(".", "");
```
punktilious, line 47: will this ever end up with unintended collisions? I guess as long as there are always 3 numbers in version, it should be OK

lmsurpre: yeah, its just a tool we use to process the conformance artifacts and so I thought the heuristic would be ok here

JohnTimm (commented):
LGTM

fhir-path/src/main/java/com/ibm/fhir/path/evaluator/FHIRPathEvaluator.java
```diff
@@ -1180,6 +1194,9 @@ public Collection<FHIRPathNode> visitFunction(FHIRPathParser.FunctionContext ctx
             case "ofType":
                 result = ofType(arguments);
                 break;
+            case "repeat":
```
punktilious, line 1197: Which of these options is most likely? If repeat isn't likely to be used, then perhaps we can stick it later in the switch. Should perform a little better, I think (assuming Java implements this as a linear series of comparisons....I remember discussing with John ages ago, but can't remember what he found).

lmsurpre: I'm pretty sure its not linear comparisons.

fhir-path/src/main/java/com/ibm/fhir/path/evaluator/FHIRPathEvaluator.java
```diff
@@ -470,6 +470,20 @@ private void pushContext(Collection<FHIRPathNode> context) {
             }
         }
 
+        private Collection<FHIRPathNode> repeat(List<ExpressionContext> arguments) {
```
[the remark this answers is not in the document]
lmsurpre: none of the other implementations here had comments and so I just sort of followed suit...

agree it would be nice to at least point to the definition of these operations in some javadoc though...will try adding it for a handful of them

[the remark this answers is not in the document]
lmsurpre: done

fhir-path/src/main/java/com/ibm/fhir/path/evaluator/FHIRPathEvaluator.java
lmsurpre: from https://docs.oracle.com/javase/8/docs/technotes/guides/language/strings-switch.html
> The Java compiler generates generally more efficient bytecode from switch statements that use String objects than from chained if-then-else statements.

Issue #3661, opened 2022-05-23
Validate extensions against the version referenced in a profile instead of the latest version


lmsurpre:
**Is your feature request related to a problem? Please describe.**
In PDEX US Drug Formulary 1.1.0, the following extension values have changed from type string to type uri:
* http://hl7.org/fhir/us/davinci-drug-formulary/StructureDefinition/usdf-MarketingURL-extension
* http://hl7.org/fhir/us/davinci-drug-formulary/StructureDefinition/usdf-SummaryURL-extension
* http://hl7.org/fhir/us/davinci-drug-formulary/StructureDefinition/usdf-FormularyURL-extension
* http://hl7.org/fhir/us/davinci-drug-formulary/StructureDefinition/usdf-EmailPlanContact-extension

To further complicate matters, the code display names in many of the CodeSystems have changed as well:

* http://hl7.org/fhir/us/davinci-drug-formulary/CodeSystem/usdf-DrugTierCS
* http://hl7.org/fhir/us/davinci-drug-formulary/CodeSystem/usdf-CoinsuranceOptionCS
* http://hl7.org/fhir/us/davinci-drug-formulary/CodeSystem/usdf-CopayOptionCS
* http://hl7.org/fhir/us/davinci-drug-formulary/CodeSystem/usdf-PharmacyTypeCS

These CodeSystem are referenced from http://hl7.org/fhir/us/davinci-drug-formulary/StructureDefinition/usdf-DrugTierDefinition-extension and some are required.

These breaking changes mean that a valid 1.0.1 instance will be invalid in version 1.1.0 and, because we check extensions separate from profile validation, such instances will fail our validation just by having the 1.1.0 versions in the registry.

**Describe the solution you'd like**
Two different ideas:
A. Instead of just using the version url to look up the extension definition in the registry, look for all versions of the extension definition and consider the instance valid if its conformant to at least one of those.  Optionally, we could add an additional constraint to check conformance against a specific version of the extension that is required by the profile.
B. Add some smarts to the Validator so that it only adds a `conformTo` constraint when that particular element doesn't already have one from the profile.

In either case, we'd benefit from adding a `conformsTo` call in a constraint that comes back from ConstraintGenerator.generateExtensionConstraint.  For option A, this would only make sense to add if the profile called for a specific version of the extension.  For option B, it could be done either way (either all the time or just when a specific version is mentioned by the profile).

**Describe alternatives you've considered**

**Acceptance Criteria**

1.  
[...]

PR #3662, opened 2022-05-25
issue #3661 - Only generate instance-specific constraints when needed

lmsurpre:
1. ConstraintGenerator.generateExtensionConstraint will now include a
`conformsTo` clause on all extension constraints
2. ConstraintGenerator.generate will assign such extension constraints a
special constraint id that uses the form:  `prefix~path<url>`
3. FHIRValidator will look for these special constraints and avoid
generating instance-specific extension constraints if they overlap with
an existing profile constraint


JohnTimm (commented):
LGTM

punktilious (approved):
LGTM

Issue #1565, opened 2020-10-07
Duplicate examples under fhir-examples json/ibm/minimal


lmsurpre:
**Is your feature request related to a problem? Please describe.**
Robin noticed that all of the minimal "Parameters" examples are duplicates.

Currently, we generate one instance per "choice type".  However, in the case of the minimal Parameters examples, these choice types aren't even required in the resource and so they are not present, leading to 50 copies of the same resource.

**Describe the solution you'd like**
The minimal example generator should be smarter.  It should only generate permutations of the resource for choice elements with a min cardinality of 1 or greater.

I think we could add a "minimal" flag to the private DataCreatorBase.getMaxChoiceCount helper method and then pass a value of true when its called from the MinimalDataCreator.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

Patil2099:
I would love to work on this issue. Can I work on this? @lmsurpre

prb112:
@Patil2099 we'd be happy to have you work on it.  Please let us know if you have any questions.

prb112:
Posted a proposal after looking at the SHA256SUMs and seeing the broad number of files impacted.

lmsurpre:
I snuck this one into my work on R4B.

lmsurpre:
Its much better now, but here are still a couple resource type examples with identical content. Most notably, we have 49 minimal StructureMap examples that are all the same.  At the very least we could special-case this resource type and get down to 1.

punktilious:
Some minimal examples are still the same e.g.
```
diff GuidanceResponse-1.json GuidanceResponse-2.json
```

Need some guidance on acceptance criteria.

lmsurpre:
Ideally we'd investigate and remove those, but if its just ones and twos here and there I'm not too worried about it...probably not worth our time.

PR #3667, opened 2022-05-27
issue #1565 - fix minimal generator and rm redundant StructureMap examples

lmsurpre:
The logic fix in the minimal data generator also resulted in a few more
example instances of resource types that have a required choice field.


fhir-examples-generator/src/main/java/com/ibm/fhir/examples/MinimalDataCreator.java
```diff
@@ -100,12 +99,20 @@ private Builder<?> addData(Builder<?> builder, int choiceIndicator, String refer
         return builder;
     }
 
+    /**
+     * For the minimal data creator, we only count required elements
+     */
     @Override
     protected int getMaxChoiceCount(Class<?> resourceOrElementClass, int maxChoiceCount, int levelsDeep) {
```
punktilious, line 106: Is levelsDeep no longer required now?

lmsurpre: If you don't mind, I think I'll just do C for now since its quick+easy and this code is only used to generate examples

fhir-examples-generator/src/main/java/com/ibm/fhir/examples/MinimalDataCreator.java
```diff
@@ -100,12 +99,20 @@ private Builder<?> addData(Builder<?> builder, int choiceIndicator, String refer
         return builder;
     }
 
+    /**
+     * For the minimal data creator, we only count required elements
+     */
```
lmsurpre, line 104:
```suggestion
    /**
     * @implNote For the minimal data creator, we only count required elements.
     * @implNote `levelsDeep` isn't actually used here but we could add a check for it just to be safe
     */
```

fhir-examples-generator/src/main/java/com/ibm/fhir/examples/MinimalDataCreator.java
punktilious: Sounds reasonable

punktilious (approved):
LGTM

Issue #3665, opened 2022-05-26
If display value is invalid exception message states it is an incorrect code 


henryhbk:
If the display value for a CodeableConcept does not match the codeset, the error thrown is misleading:

```
{
            "severity": "information",
            "code": "code-invalid",
            "details": {
                "text": "appointment-5: A code in this element should be from the specified value set 'http://hl7.org/fhir/ValueSet/c80-practice-codes' if possible"
            },
            "expression": [
                "Appointment.specialty[0]"
            ]
        },
```

In that case I had: "419192003", "http://hl7.org/fhir/ValueSet/c80-practice-codes",
			"**General Internal Medicine**"

![2022-05-26_12-02-56](https://user-images.githubusercontent.com/32741745/170531264-73d2b60e-9c1f-4464-bca3-5573b23fd6f9.png)

where the correct display value was just "Internal Medicine". A better error message would state that the validation was failing on the **display** rather than the **code itself** not being from the codeset.

duplicatable with the values above (or any change to the _display_ of a CodeableConcept)

issue discussed with @lmsurpre

lmsurpre:
I missed it when we talked before, but the *system* is whats wrong in your example.
The "http://hl7.org/fhir/ValueSet/c80-practice-codes" valueset consists of codes from SNOMED CT.

Therefor, the error is right to say that the code you used doesn't exist in http://hl7.org/fhir/ValueSet/c80-practice-codes ...just for a different reason than I assumed.

With that said, I'm pretty sure I hit this same situation from an invalid code display value before and so I confirmed the following:  after fixing the system to "http://snomed.info/sct", I still get the appointment-5 constraint failure.
After some digging, I found that this is due to the fact we don't have SNOMED CT configured on the server by default (that would require configuring either our GraphTermServiceProvider or our RemoteTermServiceProvider).
However, in such cases I think it would be better to call this constraint met and log an *informational* message to indicate that we didn't actually check the display value.

punktilious:
Acceptance criteria?

lmsurpre:
QA:
1. Add a ValueSet to the registry which enumerates concept from a made up CodeSystem
   * this can be provided via either a test RegistryResourceProvider or by setting `serverRegistryResourceProviderEnabled` to true and posting the ValueSet to the server
2. Define a profile that has a binding to this valueset with strength=required
   * this can be done with our ProfileBuilder class under fhir-profile/src/test
[...]

PR #3669, opened 2022-05-28
issue #3665 - handle coding in valueset when we can't validate display

lmsurpre:
Previously, FHIRTermServer.validateCode(ValueSet, CodeableConcept)
returned a false ValidationOutcome in cases where the code was a valid
member of the ValueSet but includes a display value that we couldn't
validate (e.g. due to missing the corresponding CodeSystem).

Now, we will return true in the ValidationOutcome for this case, but we
include a message alongside it to indicate that we didn't actually
validate the display value. From the FHIRPath MemberOfFunction, we
translate that into an informational OperationOutcomeIssue and add it to
the EvaluationContext.

For example:
> Unable to validate display value 'Diastolic Blood Pressure' for code
'8462-4' in system 'http://loinc.org'

### DIFF
--- a/fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
+++ b/fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
@@ -246,7 +246,7 @@ public class MemberOfFunction extends FHIRPathAbstractFunction {
         if (Boolean.FALSE.equals(outcome.getResult())) {
             generateIssue(outcome, evaluationContext, elementNode, strength);
             return false;
-        } else if (outcome.getMessage() != null){
+        } else if (outcome.getMessage() != null) {
             // this uses inside knowledge about the reason an outcome might come back true but with a message;
             // is there a cleaner way to get this supplemental detail back from the validateCode?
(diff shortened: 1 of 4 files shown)


fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
```diff
@@ -242,9 +242,14 @@ private boolean validateCode(FHIRTermService service, ValueSet valueSet, Codeabl
         } else {
             outcome = service.validateCode(valueSet, codeableConcept);
         }
+
         if (Boolean.FALSE.equals(outcome.getResult())) {
             generateIssue(outcome, evaluationContext, elementNode, strength);
             return false;
+        } else if (outcome.getMessage() != null){
```
punktilious, line 249: null) {

punktilious (approved):
LGTM, minor code formatting

fhir-path/src/main/java/com/ibm/fhir/path/function/MemberOfFunction.java
lmsurpre:
```suggestion
        } else if (outcome.getMessage() != null) {
```

Issue #3242, opened 2022-01-24
Scope whole-system interactions to the intersection of R4 and R4B by default


lmsurpre:
**Is your feature request related to a problem? Please describe.**
Building on the support added in #3137 ...

For example, for whole-system history:
* IBM FHIR Server 5.0 is deployed and instances of the new resource types are added to it
* An existing R4 client invokes whole-system history and, when they get to the page with the new resource type, they are unable to parse the response.

Similar scenarios exist for whole-system search and $export (and maybe $everything as well?)

**Describe the solution you'd like**
Introduce a fhir-server-config property that controls the server behavior when no fhirVersion is specified by the client.  The default value for this property should result in the server only returning 4.0.1 resource types (but still not ones removed in 4.3.0).

Regardless of the setting, if the client passes an Accept header with a FHIR mime type that includes `fhirVersion=4.3`, and the new resource types are not filtered out in the fhir-server-config `resources` section, then include the new resource types introduced in R4B.

This is consistent with the CapabilityStatement behavior implemented for #3192 except that we should update that to support changing the default if the user wants something other than 4.0.1 to be the default behavior.

**Describe alternatives you've considered**
Force the IBM FHIR Server operator to explicitly list the set of supported resource types in their fhir-server-config and use exactly that set when the client doesn't pass any `fhirVersion` in their Accept header.
If they only want to server 4.0 resource types, then only list 4.0 resource types and set `open=false`.

If the client explicitly asks for `fhirVersion=4.0` we could still omit all the resource types introduced in R4B.

If we went with this alternative, we'd need to update the CapabilityStatement behavior to match because, currently, that defaults to omitting the new R4B resource types (unless fhirVersion=4.3 is passed)

**Acceptance Criteria**

1.  
   GIVEN a 5.0 IBM FHIR Server with resource type instances of types added in R4B (e.g. Ingredient)
   WHEN a whole-system *history* query is sent with no explicit _type parameter 
   AND the client sets the Accept header to `application/fhir+[xml/json];fhirVersion=4.3`
   THEN the history response contains the R4B resource types

1.  
   GIVEN a 5.0 IBM FHIR Server with resource type instances of types added in R4B (e.g. Ingredient)
   WHEN  a whole-system *search* query is sent with no explicit _type parameter
   AND the client sets the Accept header to `application/fhir+[xml/json];fhirVersion=4.3`
   THEN the search response contains the R4B resource types

1.  
[...]

PR #3685, opened 2022-06-03
issue #3242 - use the fhirVersion from the import/export request

lmsurpre:
In the case of export, validate any `_type` parameters against the
resource types for the configured fhirVersion. If none was specified,
use an appropriate list for that version.

In the case of import, validate the input parameters against the
supported types for the fhirVersion.

### DIFF
--- a/operation/fhir-operation-everything/src/main/java/com/ibm/fhir/operation/everything/EverythingOperation.java
+++ b/operation/fhir-operation-everything/src/main/java/com/ibm/fhir/operation/everything/EverythingOperation.java
@@ -46,7 +46,6 @@ import com.ibm.fhir.model.type.Uri;
 import com.ibm.fhir.model.type.code.BundleType;
 import com.ibm.fhir.model.type.code.HTTPVerb;
 import com.ibm.fhir.model.type.code.IssueType;
-import com.ibm.fhir.model.type.code.SearchEntryMode;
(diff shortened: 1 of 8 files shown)


operation/fhir-operation-everything/src/main/java/com/ibm/fhir/operation/everything/EverythingOperation.java
```diff
@@ -205,7 +204,7 @@ protected Parameters doInvoke(FHIROperationContext operationContext, Class<? ext
         MultivaluedMap<String, String> queryParametersWithoutDates = new MultivaluedHashMap<String,String>(queryParameters);
         boolean startOrEndProvided = queryParametersWithoutDates.remove(DATE_QUERY_PARAMETER) != null;
 
-        List<String> defaultResourceTypes = new ArrayList<String>(0);
+        List<String> defaultResourceTypes = new ArrayList<String>();
```
punktilious, line 207: there's no need to assign any value to defaultResourceTypes here. The value gets assigned in the next block, or an exception is thrown

lmsurpre:
```suggestion
        List<String> defaultResourceTypes;
```

punktilious (approved):
LGTM

PR #3694, opened 2022-06-07
Support additional extraResources types from Patient/$everything

lmsurpre:
Instead of a big if/else with all the manually preconfigured search parameters, we now look up the applicable search parameters for the given resource type for this tenant and then process those into the set of includes.

It should work mostly the same, except that:
1. we now support "includeTypes" for all resource types; previously it only supported "Medication", "Location", "Practitioner", and "Organization
2. for the 4 types that were previously supported, we include a few additional includes that were missed in the manually computed list
3. we fix a single issue with an include value from ExplanationOfBenefit (it used to add "insurer" but this is not a valid search parameter in the core specification)

### DIFF
--- a/fhir-search/src/main/java/com/ibm/fhir/search/parameters/IncludesUtil.java
+++ b/fhir-search/src/main/java/com/ibm/fhir/search/parameters/IncludesUtil.java
@@ -7,8 +7,10 @@ package com.ibm.fhir.search.parameters;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -23,7 +25,7 @@ public class IncludesUtil {
     private static final Logger LOG = java.util.logging.Logger.getLogger(IncludesUtil.class.getName());
 
     /**
-     * Computes the list of supported include parameter values that target resource types from the passed resourcesToInclude list.
+     * Computes the list of supported include parameter values that target resource types from the passed set of resourceTypesToInclude.
      *
      * @param baseResourceType
      * @param resourceTypesToInclude
@@ -31,11 +33,13 @@ public class IncludesUtil {
      * @return The list of values to use for the _include search parameter.
      * @throws Exception
      */
-    public static List<String> computeIncludesParamValues(String baseResourceType, List<String> resourceTypesToInclude, SearchHelper searchHelper)
+    public static List<String> computeIncludesParamValues(String baseResourceType, Set<String> resourceTypesToInclude, SearchHelper searchHelper)
             throws Exception {
         List<String> result = new ArrayList<>();
 
-        List<String> allowedIncludes = FHIRConfigHelper.getSearchPropertyRestrictions(baseResourceType, FHIRConfigHelper.SEARCH_PROPERTY_TYPE_INCLUDE);
+        List<String> allowedIncludesList = FHIRConfigHelper.getSearchPropertyRestrictions(baseResourceType, FHIRConfigHelper.SEARCH_PROPERTY_TYPE_INCLUDE);
+        Set<String> allowedIncludes = (allowedIncludesList == null) ? null : new HashSet<>(allowedIncludesList);
+
         Map<ResourceType, List<String>> codesByType = IncludesUtil.getSearchCodesByType(baseResourceType, searchHelper);
 
         for (Map.Entry<ResourceType, List<String>> entry : codesByType.entrySet()) {
@@ -86,13 +90,13 @@ public class IncludesUtil {
     /**
      * @param compartmentMemberType the resource type that is the source of the includes
      * @param subResourceType the target resource type of the includes; null
-     * @param allowedIncludes the list of include parameter values to allow
+     * @param allowedIncludes the set of include parameter values to allow
      * @param codes the search parameter codes to include on
      * @return A list of parameter values to use for the _include query parameter.
      * @throws Exception
      */
     private static List<String> computeIncludes(String compartmentMemberType, ResourceType subResourceType,
-            List<String> allowedIncludes, List<String> codes) throws Exception {
+            Set<String> allowedIncludes, List<String> codes) throws Exception {
         List<String> includes = new ArrayList<>();
 
         for (String code : codes) {
--- a/fhir-search/src/main/java/com/ibm/fhir/search/parameters/IncludesUtil.java
+++ b/fhir-search/src/main/java/com/ibm/fhir/search/parameters/IncludesUtil.java
@@ -1,0 +1,116 @@
+/*
+ * (C) Copyright IBM Corp. 2022
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.ibm.fhir.search.parameters;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.logging.Level;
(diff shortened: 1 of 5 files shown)


fhir-search/src/main/java/com/ibm/fhir/search/parameters/IncludesUtil.java
```diff
@@ -0,0 +1,116 @@
...
+     */
+    public static List<String> computeIncludesParamValues(String baseResourceType, List<String> resourceTypesToInclude, SearchHelper searchHelper)
+            throws Exception {
+        List<String> result = new ArrayList<>();
+
+        List<String> allowedIncludes = FHIRConfigHelper.getSearchPropertyRestrictions(baseResourceType, FHIRConfigHelper.SEARCH_PROPERTY_TYPE_INCLUDE);
+        Map<ResourceType, List<String>> codesByType = IncludesUtil.getSearchCodesByType(baseResourceType, searchHelper);
+
+        for (Map.Entry<ResourceType, List<String>> entry : codesByType.entrySet()) {
+            if (resourceTypesToInclude.contains(entry.getKey().value())) {
```
punktilious, line 42: Multiple calls to List.contains here. Would perform better if resourceTypesToInclude were a Set, which also makes more sense from a semantic point of view too.

fhir-search/src/main/java/com/ibm/fhir/search/parameters/IncludesUtil.java
```diff
@@ -0,0 +1,116 @@
...
+                continue;
+            }
+
+            String expression = sp.getExpression().getValue();
+            for (ResourceTypeCode target : sp.getTarget()) {
+                // The search parameter target types come from the possible types of the target reference
+                // and don't take into account any "is X" filters from within the expression,
+                // so we use this heuristic to avoid adding the code when expression filtering applies
+                // and does not include a filter clause for this particular target type.
+                if (expression.contains(".is(") || expression.contains(" is ")) {
```
punktilious, line 72: Is this brittle?

fhir-search/src/main/java/com/ibm/fhir/search/parameters/IncludesUtil.java
```diff
@@ -0,0 +1,116 @@
...
+            List<String> allowedIncludes, List<String> codes) throws Exception {
+        List<String> includes = new ArrayList<>();
+
+        for (String code : codes) {
+            // Need to make sure the search parameter has not been excluded
+            String parameterName = compartmentMemberType + ":" + code + ":" + subResourceType.value();
+            String simplifiedParameterName = compartmentMemberType + ":" + code;
+
+            boolean isAllowed = allowedIncludes == null ||
+                    allowedIncludes.contains(parameterName) ||
```
punktilious, line 104: Calling List#contains within a loop. Might be more appropriate to use a Set<String> here.

operation/fhir-operation-everything/src/main/java/com/ibm/fhir/operation/everything/EverythingOperation.java
```diff
@@ -765,8 +457,7 @@ private void readsOfAdditionalAssociatedResources(String compartmentMemberType,
                     String type = referenceValue.getTargetResourceType();
                     String value = referenceValue.getValue();
                     String resourceId = type + "/" + value;
-                    if (!resourceIds.contains(resourceId) && extraResources.contains(type) && (type.equals(ResourceType.LOCATION.value()) || type.equals(ResourceType.MEDICATION.value())
-                            || type.equals(ResourceType.ORGANIZATION.value()) || type.equals(ResourceType.PRACTITIONER.value()))) {
+                    if (!resourceIds.contains(resourceId) && extraResources.contains(type)) {
```
punktilious, line 460: Call to List#contains. Slow.

operation/fhir-operation-everything/src/main/java/com/ibm/fhir/operation/everything/EverythingOperation.java
```diff
@@ -744,12 +434,14 @@ private void addSearchParameterIfNotExcluded(String compartmentMemberType, Strin
      * @param compartmentMemberType the type of resource currently drilling down on
      * @param newEntries a list of Bundle response entries that include compartmentMemberType instances found via search
      * @param allEntries a list of all entries that have been found so far (minus duplicates)
+     * @param resourceIds the set of external resource ids ("resourceType/id") to be included in the result (to avoid duplicates)
      * @param resourceHelper an instance of FHIRResourceHelpers to use to make the calls
      * @param extraResources a list of "extra" resource types configured for this server
```
punktilious, line 439: I think extraResourceTypes would be a better parameter name

operation/fhir-operation-everything/src/main/java/com/ibm/fhir/operation/everything/EverythingOperation.java
```diff
@@ -807,7 +498,7 @@ private void readsOfAdditionalAssociatedResources(String compartmentMemberType,
      * @param retrieveAdditionalResources whether the server config has specified to retrieve additional resources
      */
     private void processResults(Bundle results, String compartmentMemberType, List<Entry> newEntries,
-                List<Entry> allEntries, List<String> resourceIds, FHIRResourceHelpers resourceHelper, List<String> extraResources,
+                List<Entry> allEntries, Set<String> resourceIds, FHIRResourceHelpers resourceHelper, List<String> extraResources,
```
punktilious, line 501: Perhaps `Set<String> extraResourceTypes`

fhir-search/src/main/java/com/ibm/fhir/search/parameters/IncludesUtil.java
lmsurpre: its definitely "heuristic".  IMHO the conformance artifacts should account for this when they list the allowed "target" types for a search parameter of type reference.

lmsurpre: a potential alternative would be for us to patch the official resources in order to make those right...then we wouldn't need to do this.  but it wouldn't necessarily help with user-packaged IGs that may have the same issue

punktilious: I don't think it's a blocker...just struck me that the test could hit false positives...but I don't know enough about the context.

lmsurpre: ok.  i think its pretty safe.  theoretically someone could construct a FHIRPath to confuse us (like by combining an “all” or “iff” or “not” or something like that with these “is” clauses), but in practice no one does that for search parameters.

Its almost always just a simple path and sometimes they have an is at the end to filter to a particular type.

fhir-search/src/main/java/com/ibm/fhir/search/parameters/IncludesUtil.java
lmsurpre: done

operation/fhir-operation-everything/src/main/java/com/ibm/fhir/operation/everything/EverythingOperation.java
lmsurpre: done

operation/fhir-operation-everything/src/main/java/com/ibm/fhir/operation/everything/EverythingOperation.java
lmsurpre: done

PR #3705, opened 2022-06-13
convert BPConstraintGeneratorTest main into a test

lmsurpre:

### DIFF
--- a/fhir-profile/src/test/java/com/ibm/fhir/profile/test/BPConstraintGeneratorTest.java
+++ b/fhir-profile/src/test/java/com/ibm/fhir/profile/test/BPConstraintGeneratorTest.java
@@ -1,10 +1,12 @@
 /*
- * (C) Copyright IBM Corp. 2021
+ * (C) Copyright IBM Corp. 2021, 2022
  *
  * SPDX-License-Identifier: Apache-2.0
  */
 
 package com.ibm.fhir.profile.test;
+
+import static org.testng.Assert.assertEquals;
 
 import java.util.List;
 
@@ -18,7 +20,7 @@ import com.ibm.fhir.profile.ProfileSupport;
 import com.ibm.fhir.registry.FHIRRegistry;
 
 public class BPConstraintGeneratorTest {
-    private static final boolean DEBUG = true;
+    private static final boolean DEBUG = false;
 
     @BeforeClass
     public void before() {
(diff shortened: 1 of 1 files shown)


fhir-profile/src/test/java/com/ibm/fhir/profile/test/BPConstraintGeneratorTest.java
```diff
@@ -21,34 +18,23 @@
 import com.ibm.fhir.registry.FHIRRegistry;
 
 public class BPConstraintGeneratorTest {
+    private static final boolean DEBUG = true;
```
punktilious, line 21: Shouldn't we try to reduce the amount of output in the build log unless there's an error?

fhir-profile/src/test/java/com/ibm/fhir/profile/test/BPConstraintGeneratorTest.java
```diff
@@ -21,34 +18,23 @@
...
-        logger.addHandler(handler);
+    @Test
+    public void generateConstraints() throws Exception {
         StructureDefinition profile = ProfileSupport.getProfile("http://hl7.org/fhir/StructureDefinition/bp");
         ConstraintGenerator generator = new ConstraintGenerator(profile);
         List<Constraint> constraints = generator.generate();
-        constraints.forEach(System.out::println);
+        for (Constraint constraint : constraints) {
+            if (DEBUG) {
+                System.out.println(constraint.id() + "\t" + constraint.expression());
```
punktilious, line 36: I think we should have some asserts here. What is actually being tested?

lmsurpre: sure.  the main goal with this PR is to make it run without errors.  previously it was just a main method and, when you ran it, it threw an exception.
we can't verify that the constraints compile since we don't have our FHIRPathEvaluator here.
maybe we just assert the number of constraints?

fhir-profile/src/test/java/com/ibm/fhir/profile/test/BPConstraintGeneratorTest.java
lmsurpre: yeah, it used to always print.  i made it conditional.  fine with turning this to false now.
```suggestion
    private static final boolean DEBUG = false;
```

lmsurpre: done

fhir-profile/src/test/java/com/ibm/fhir/profile/test/BPConstraintGeneratorTest.java
lmsurpre: i added the assert for the number of constraints

punktilious (approved):
LGTM

PR #3734, opened 2022-06-22
Fix fhirVersion in the CapabilityStatement

lmsurpre:
Use "4.3.0" instead of "4.3.0-cibuild"


punktilious:
Why were we using -cibuild in the first place?

punktilious (approved):
LGTM

lmsurpre:
> Why were we using -cibuild in the first place?

Because when I initially implemented this, 4.3.0 didn't exist yet (we generate FHIRVersion from the FHIR spec artifacts and it didn't exist there yet since it wasn't published yet).

Issue #3735, opened 2022-06-22
$validate operation returns 200 OK when it should return a 500 Server Error


lmsurpre:
**Describe the bug**
When you invoke $validate at the resource-type level (not the instance level) and an unexpected server error occurs, the current implementation returns a 200 OK with an OperationOutcome description like "An error occurred during validation".

**Environment**
main

**To Reproduce**
1. register a profile with known bad FHIRPath or something similar to force the server to throw during validation (instead of returning an issue list)
2. invoke $validate at the resource type level

Note the generic error message and the fact that we are missing the exception info in the logs.

**Expected behavior**
per http://hl7.org/fhir/operation-resource-validate.html
> This operation returns a 200 OK whether or not the resource is valid. A 4xx or 5xx error means that the validation itself could not be performed, and it is unknown whether the resource is valid or not.

I think this should be a 5xx.
However, it will be good to confirm that the failure to parse the incoming resource continues to fail with either a 200 OK (with OperationOutcome indicating the failure) or with a 4xx response code.

**Additional context**

PrasannaHegde1:
tried $validate operation at the resource type level with below request having us-core-patient profile which does not have one of the required fields 'Patient.gender'.  
Request: POST /fhir-server/api/v4/Patient
Requesy body : 
````json
{
  "resourceType": "Patient",
    "meta": {
        "tag": [{
            "system": "http://example.com",
            "code": "text",
            "display": "Text"
        }],
        "profile": ["http://hl7.org/fhir/us/core/StructureDefinition/us-core-patient"]
    },
  "language": "en",
    "text": {
        "status": "empty",
        "div": "<div xmlns=\"http://www.w3.org/1999/xhtml\">Lee (123)</div>"
    },
  "identifier": [{
    "use": "usual",
        "system": "https://example.com",
        "value": "1234"
  }],
    "name": [{
        "given": ["David"]
    }],
    
  "active": true
}
````
Response : 200 OK. 
```json
{
    "resourceType": "OperationOutcome",
    "id": "Error",
    "text": {
        "status": "additional",
        "div": "<div xmlns=\"http://www.w3.org/1999/xhtml\"><p>ERROR</p></div>"
    },
    "issue": [
        {
            "severity": "error",
            "code": "invariant",
            "details": {
                "text": "generated-us-core-patient-4: Constraint violation: gender.exists()"
            },
            "diagnostics": "Caused by: [[expression: gender.exists(), result: false, location: Patient]]",
[...]

PR #3736, opened 2022-06-22
issue #3735 - remove invoke(String,String,Resource) special processing

lmsurpre:
For some reason, our exception handling for
`invoke(String,String,Resource)` differs from the exception handling in
all the other invoke methods.

I assume that was done so that the $invoke operation properly returns a
200 OK in cases where the data is invalid.
But this is not the right level for that logic and it can prevent us
from seeing important server issues in the logs.
I double-checked and the ValidateOperation is
already returning a valid OperationOutcome response for successful
validation of invalid resources, so I think we just need to remove the
special handling at the Operation JAX-RS resource.

Additionally, we found "interesting" liberty behavior that was
preventing us from seeing the exception stack trace even when FINE
logging is enabled; that log message is now level INFO so that the stack
trace gets printed (but it will still only be printed when the FINE
level is enabled).

Finally, I snuck in a small pom.xml fixup for fhir-server.

### DIFF
--- a/fhir-server/pom.xml
+++ b/fhir-server/pom.xml
@@ -67,6 +67,11 @@
         </dependency>
         <dependency>
             <groupId>${project.groupId}</groupId>
+            <artifactId>fhir-profile</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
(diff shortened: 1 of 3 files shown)


punktilious (approved):
LGTM

Issue #3737, opened 2022-06-22
reindex operation should lock logical_resources_ident with select for update 


punktilious:
**Describe the bug**
The reindex operation is vulnerable to a race condition if the resource is updated after the reindex read is performed.

**Environment**
Which version of IBM FHIR Server?

**To Reproduce**

Concurrent resource updates at the same time as concurrent reindex operations on overlapping resources.

**Expected behavior**
The reindex operation must lock the logical_resource_ident record before inserting the new parameters.

**Additional context**
The select for update needs to use the resource_type_id and logical_id (the PK for logical_resource_ident). Because the reindex operation is driven by logical_resource_id, there must be a read against logical_resources first to get these values. The logical_resource_ident record can then be locked and the resource payload read without risk of a race condition.

lmsurpre:
I set the TRACE_SPEC and I confirm that I see the log message about locking for each resource that gets re-indexed using both *server-driven reindex* () and *client-driven reindex* (where we pass a list of index ids).

One thing I noted is that we perform the locking irregardless of whether:
a. the tstamp is later (server-driven reindex); or
b. the parameters hash matches (both client and server-driven reindex)

@punktilious if the above is what we expect, then I think this is working.

PR #3740, opened 2022-06-23
issue #3737 reindex must lock logical_resource_ident

punktilious:

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/ReindexResourceDAO.java
+++ b/fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/ReindexResourceDAO.java
@@ -22,6 +22,7 @@ import javax.transaction.TransactionSynchronizationRegistry;
 
 import com.ibm.fhir.database.utils.api.IDatabaseTranslator;
 import com.ibm.fhir.database.utils.common.CalendarHelper;
+import com.ibm.fhir.database.utils.common.PreparedStatementHelper;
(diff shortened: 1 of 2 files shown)


fhir-persistence-jdbc/src/main/java/com/ibm/fhir/persistence/jdbc/dao/ReindexResourceDAO.java
```diff
@@ -267,6 +319,17 @@ protected ResourceIndexRecord getNextResource(SecureRandom random, Instant reind
...
         } while (offsetRange > 0 && result == null);
 
+        if (result != null) {
+            // Since V0027 we lock resources using LOGICAL_RESOURCE_IDENT. There is a small chance of
+            // deadlock here if reindex is processing the same resources which are also being updated
+            // as part of ingestion. This is because ingestion locks LOGICAL_RESOURCE_IDENT before
+            // updating LOGICAL_RESOURCES, whereas here we updated LOGICAL_RESOURCES before locking
+            // LOGICAL_RESOURCE_IDENT. Eliminating this would require a significant rework. The
+            // recommendation going forward is to always use the client-driven process, because it
+            // is more efficient.
```
lmsurpre, line 329: I wonder if we should deprecate (or remove) the server-driven one...

Personally I do like the simplicity of being able to just call `[base]/[resourceType]/$reindex`.
In fact I was sort of wondering if we could support `[base]/[resourceType]/$reindex?timestamp=[isoDateTime]` to make it even easier... (today we don't support that on POST requests, only GET ones...and we force reindex to use POST)

lmsurpre (approved):
LGTM

Issue #3712, opened 2022-06-15
Coding display validation should consider all possible designations


lmsurpre:
**Is your feature request related to a problem? Please describe.**
From https://hl7.org/fhir/datatypes.html#codesystem
> If the 'display' element is populated, the string used in display SHALL be one of the display strings defined for that code by the code system (code systems may define multiple display strings for a single code). If one of the available display strings is labeled as preferred, it SHOULD be used.

However, CodeSystem.concept.display is `0..1`.  Does it means you can use any of the "designations" for the Coding.display?
Currently, our implementation only checks against CodeSystem.concept.display and not the corresponding designations.

This is in `FHIRTermService.validateDisplay`.

**Describe the solution you'd like**
Instead of checking against just the CodeSystem.concept.display, check against any existing designation values as well.

**Describe alternatives you've considered**

**Acceptance Criteria**

1.  
   GIVEN a CodeSystem with a concept that has one or more designations
    AND a resource instance that has a Coding from that concept in this CodeSystem
   WHEN you validate that instance
   THEN it passes validation

**Additional context**
I opened https://jira.hl7.org/browse/FHIR-37603 to get some clarification from the spec authors.

lmsurpre:
Performance question:  will doing the designationValues.contains dramatically increase the time to validate large instances?

lmsurpre:
QA:
1. find or create a CodeSystem that has a concept which has one or more "designations" (that don't match that concept's "display")
2. find or create a StructureDefinition (base resourceType or profile) that has a binding to a valueset which includes that concept
3. create an instance of that resourceType/profile and use the value from the designation

codeableConcept: {
  coding: [
{
  system: ""
  code: ""
  display: "THIS MUST MATCH THE DESIGNATION"
}
]
...
```
Before this change, the server would say thats invalid.  Now that should succeed.

PrasannaHegde1:
The Patient resource type has a binding with CodeSystem for "Medical Record Number". 
"Medical Record Number" has display value "Medical Record Number" and designation value "Krankenaktennummer"
Tried to create an instance of Patient resource type with below request :
````json
{
    "resourceType" : "Patient",
    "active" : true,
    "identifier": [		
       {
			"type": {
				"coding": [
					{
						"system": "http://terminology.hl7.org/CodeSystem/v2-0203",
						"code": "MR",
						"display": "Krankenaktennummer"
					}
				],
				"text": "Medical Record Number"
			},
[...]

PR #3771, opened 2022-07-11
issue #3712 - check all CodeSystem concept designations

lmsurpre:
1. Split `validateDisplay` into two flavors; now
validateCode(CodeSystem, Coding, ValidationParameters) will use the
passed CodeSystem to determine case sensitivity (instead of looking it
up via the registry)

2. When the Coding.display doesn't match the lookup display, we now look
for a match within the CodesSystem.concept.designation values before
calling it invalid.

3. Added a designation to a couple sample code systems and added
testValidateDisplay tests for both case-sensitive and case-insensitive
behavior.

### DIFF
--- a/term/fhir-term/src/main/java/com/ibm/fhir/term/service/FHIRTermService.java
+++ b/term/fhir-term/src/main/java/com/ibm/fhir/term/service/FHIRTermService.java
@@ -435,9 +435,8 @@ public class FHIRTermService {
      *     the coding to lookup
      * @param parameters
      *     the lookup parameters
-     *
      * @return
-     *     the outcome of the lookup
+     *     the outcome of the lookup; null if the passed CodeSystem is not applicable to the system and version of the coding
      */
     public LookupOutcome lookup(CodeSystem codeSystem, Coding coding, LookupParameters parameters) {
         if (!LookupParameters.EMPTY.equals(parameters)) {
@@ -828,7 +827,7 @@ public class FHIRTermService {
         }
         LookupOutcome outcome = lookup(codeSystem, coding, LookupParameters.EMPTY);
         if (outcome != null) {
-            return validateDisplay(null, coding, outcome);
+            return validateDisplay(coding, outcome, CodeSystemSupport.isCaseSensitive(codeSystem));
         } else {
             StringBuilder message = new StringBuilder("Code '");
             if (coding != null && coding.getCode() != null) {
@@ -976,7 +975,7 @@ public class FHIRTermService {
             if (result) {
                 LookupOutcome outcome = lookup(coding);
                 if (outcome != null || coding.getDisplay() == null) {
-                    return validateDisplay(null, coding, outcome);
+                    return validateDisplay(coding, outcome);
                 }
 
                 // lookup outcome was null and we have a non-null display, so include a message in the response
@@ -1054,7 +1053,7 @@ public class FHIRTermService {
         if (result) {
             LookupOutcome outcome = lookup(coding);
             if (outcome != null || coding.getDisplay() == null) {
-                return validateDisplay(null, coding, outcome);
+                return validateDisplay(coding, outcome);
             }
 
             // lookup outcome was null and we have a non-null display, so include a message in the response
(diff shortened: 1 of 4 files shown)


term/fhir-term/src/main/java/com/ibm/fhir/term/service/FHIRTermService.java
```diff
@@ -1167,29 +1166,77 @@ private List<FHIRTermServiceProvider> loadProviders() {
...
         }
 
-        java.lang.String system = null;
-        if (coding.getSystem() != null) {
-            system = coding.getSystem().getValue();
-        } else if (codeSystem != null && codeSystem.getUrl() != null) {
-            system = codeSystem.getUrl().getValue();
+        java.lang.String displayToValidate = isCaseSensitive ? coding.getDisplay().getValue() : normalize(coding.getDisplay().getValue());
+
+        boolean result = false;
```
punktilious, line 1219: Minor. No need to assign a default value. But I like to avoid assigning a default because it helps to make sure a value gets assigned in all possible branches of the subsequent code.

term/fhir-term/src/main/java/com/ibm/fhir/term/service/FHIRTermService.java
```diff
@@ -1167,29 +1166,77 @@ private List<FHIRTermServiceProvider> loadProviders() {
...
+        if (isCaseSensitive) {
+            result = lookupOutcome.getDisplay().getValue().equals(displayToValidate);
+            if (result == false) {
```
punktilious, line 1222: !result

term/fhir-term/src/main/java/com/ibm/fhir/term/service/FHIRTermService.java
```diff
@@ -1167,29 +1166,77 @@ private List<FHIRTermServiceProvider> loadProviders() {
...
+                result = lookupOutcome.getDesignation().stream()
+                        .anyMatch(d -> d.getValue() != null && displayToValidate.equals(d.getValue().getValue()));
+            }
+        } else {
+            result = normalize(lookupOutcome.getDisplay().getValue()).equals(displayToValidate);
+            if (result == false) {
+                result = lookupOutcome.getDesignation().stream()
```
punktilious, line 1229: comment as to what this is doing would be useful

term/fhir-term/src/main/java/com/ibm/fhir/term/service/FHIRTermService.java
lmsurpre: so i usually assign false to be explicit / help me remember it'll have a value

term/fhir-term/src/main/java/com/ibm/fhir/term/service/FHIRTermService.java
lmsurpre:
```suggestion
            if (!result) {
```

term/fhir-term/src/main/java/com/ibm/fhir/term/service/FHIRTermService.java
```diff
@@ -1167,29 +1166,77 @@ private List<FHIRTermServiceProvider> loadProviders() {
...
+        boolean result = false;
+        if (isCaseSensitive) {
+            result = lookupOutcome.getDisplay().getValue().equals(displayToValidate);
+            if (result == false) {
+                result = lookupOutcome.getDesignation().stream()
+                        .anyMatch(d -> d.getValue() != null && displayToValidate.equals(d.getValue().getValue()));
+            }
+        } else {
+            result = normalize(lookupOutcome.getDisplay().getValue()).equals(displayToValidate);
+            if (result == false) {
```
lmsurpre, line 1228:
```suggestion
            if (!result) {
```

term/fhir-term/src/main/java/com/ibm/fhir/term/service/FHIRTermService.java
lmsurpre:
```suggestion
                // call it valid if any of the normalized designations matches the normalized display value
                result = lookupOutcome.getDesignation().stream()
```

term/fhir-term/src/main/java/com/ibm/fhir/term/service/FHIRTermService.java
```diff
@@ -1167,29 +1166,77 @@ private List<FHIRTermServiceProvider> loadProviders() {
...
-            system = coding.getSystem().getValue();
-        } else if (codeSystem != null && codeSystem.getUrl() != null) {
-            system = codeSystem.getUrl().getValue();
+        java.lang.String displayToValidate = isCaseSensitive ? coding.getDisplay().getValue() : normalize(coding.getDisplay().getValue());
+
+        boolean result = false;
+        if (isCaseSensitive) {
+            result = lookupOutcome.getDisplay().getValue().equals(displayToValidate);
+            if (result == false) {
+                result = lookupOutcome.getDesignation().stream()
```
lmsurpre, line 1223:
```suggestion
                // call it valid if any of the designations match the display value
                result = lookupOutcome.getDesignation().stream()
```

term/fhir-term/src/main/java/com/ibm/fhir/term/service/FHIRTermService.java
punktilious: If they are member variables they are assigned false by default, but variables in a method do not get assigned a default value.

punktilious (approved):
LGTM

PR #3778, opened 2022-07-13
Use string constants

lmsurpre:
I was poking around these classes and noticed we define these constants
but then we didn't actual use them. Very minor.


punktilious (approved):
LGTM

Issue #3503, opened 2022-03-24
Smart app launch and $everything operation should handle 403 error better


d0roppe:
**Describe the bug**
With the $everything operation, and having SMART app launch configured, a 403 error turns into a 500 error back to the caller. 

**Environment**
FHIR-main

**To Reproduce**
Call with /Patient/<id>/$everything with Oauth2 through SMART App launch a 403 error causes a 500 error back to the caller

**Expected behavior**
the 403 error should flow back to the caller

lmsurpre:
with the changes from #3779, the expected response status for this case is now a 403 (not a 500) and the error message should be like this:
```Error retrieving $everything related resources of type 'Appointment' for Patient/002e2eac-c802-41fc-8b38-6d45b67a38d0; caused by read permission for 'Appointment' is not granted by any of the provided scopes: [[patient/Patient.*]]```

lmsurpre:
During QA, we found that we still emitted a generic 500 error in the case that the user was missing read access to their patient resource (e.g. missing scope `patient/Patient.read`).

lmsurpre:
After the fix, we confirmed some additional cases to ensure we're not leaking information, and everything looks OK to us.

PR #3779, opened 2022-07-14
issue #3503 - special case underlying $everything "Forbidden" exceptions

lmsurpre:
In addition to the change, I added a new e2e test package under
fhir-server-test to confirm the fix. The test will conditionally execute
based on whether the server advertises its
.well-known/smart-configuration (which we only serve when its
configured). There's a ton of cases we could add, but for now I just
wanted to cover the specific case in #3503. Previously, we had only unit
tests for the fhir-smart interceptor.

I also removed an unused private method from the fhir-smart interceptor.

Finally, I snuck in a couple very minor updates:
* removed an unused test-jar dependency from fhir-core-r4b
* a single word change to a specific AuthzPolicyEnforcement message to
hint that only "SMART" scopes are echoed back to the client (and not
other scopes that may have been passed)

### DIFF
--- a/conformance/fhir-core-r4b/pom.xml
+++ b/conformance/fhir-core-r4b/pom.xml
@@ -26,13 +26,6 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>${project.groupId}</groupId>
-            <artifactId>fhir-model</artifactId>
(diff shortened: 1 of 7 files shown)


punktilious (approved):
LGTM

PR #3780, opened 2022-07-14
Log one long message at level info instead of many smaller ones

lmsurpre:
Very minor change to (very slightly) reduce noise in the startup log


punktilious (approved):
LGTM

Issue #3655, opened 2022-05-20
CQL-related integration test failures


d0roppe:
**Describe the bug**
When running the integration tests locally and adding the ability to run the CPG and CQF tests, there are failures in 2 tests
com.ibm.fhir.server.test.cpg.ServerCqlOperationTest
  testUsefulDetailIncludedInMissingPatientResponse 
and
com.ibm.fhir.server.test.cqf.ServerEvaluateMeasureOperationTest
  testEvaluatePatientMeasureMissingPatient

Both tests had 1 failure and was expecting a 500 but got a 200.

**Environment**
main

**To Reproduce**
run the full integration tets with the proper files copied into userlib
fhir-operation-test-5.0.0-SNAPSHOT.jar
userlib
  fhir-ig-carin-bb-5.0.0-SNAPSHOT.jar
  fhir-ig-davinci-hrex-4.11.0-SNAPSHOT.jar
  fhir-ig-davinci-hrex-5.0.0-SNAPSHOT.jar
  fhir-ig-davinci-pdex-formulary-4.11.0-SNAPSHOT.jar
  fhir-ig-davinci-pdex-formulary-5.0.0-SNAPSHOT.jar
  fhir-ig-davinci-pdex-plan-net-5.0.0-SNAPSHOT.jar
  fhir-ig-mcode-5.0.0-SNAPSHOT.jar
  fhir-ig-spl-4.11.0-SNAPSHOT.jar
  fhir-ig-spl-5.0.0-SNAPSHOT.jar
  fhir-ig-us-core-5.0.0-SNAPSHOT.jar
  fhir-operation-cpg-5.0.0-SNAPSHOT-shaded.jar
  fhir-operation-cqf-5.0.0-SNAPSHOT-shaded.jar
  fhir-operation-term-cache-5.0.0-SNAPSHOT.jar
  fhir-operation-test-5.0.0-SNAPSHOT.jar
**Expected behavior**
Expected the tests to pass

**Additional context**

d0roppe:
Verified the test now passes with the IG's loaded.

PR #3786, opened 2022-07-18
issue #3655 - throw from ServerFHIRRetrieveProvider for failed reads

lmsurpre:
In my opinion, neither the old behavior (500 server error) nor the new
behavior (200 OK with a null expression result) are ideal.
However, since we didn't intend to change it, I've restored the previous
behavior for now.
The most relevant advantage is we'll now get a reasonable error message
when the user tries to evaluate expressions on resources that don't
exist.


punktilious (approved):
LGTM

Issue #3494, opened 2022-03-22
install scripts should install to defaultServer to match the container image


lmsurpre:
**Is your feature request related to a problem? Please describe.**
For https://github.com/IBM/FHIR/issues/923 I moved the server from `wlp/usr/servers/fhir-server` to `wlp/usr/servers/defaultServer` but only for the docker image.

For the zip installer, it still installs to fhir-server.  At the time this felt a bit safer to me...only changing what we actually needed to change.  However, now it feels needlessly different.

**Describe the solution you'd like**
Update install.sh and install.bat to install the fhir-server to the liberty "defaultServer" instead of using the custom server name fhir-server.

**Describe alternatives you've considered**

**Acceptance Criteria**

1.  
   GIVEN [a precondition]
    AND [another precondition]
   WHEN [test step]
    AND [test step]
   THEN [verification step]
    AND [verification step]

**Additional context**
Will require lots of automation changes and should be clearly highlighted in the release notes for anyone using the zip installer.

lmsurpre:
similar but different #3728

PR #3789, opened 2022-07-19
issue #3494 - always install to defaultServer

lmsurpre:
Previously we had two different install script flavors, each with a
different server name:
* install.sh/install.bat installed a fresh Liberty server with our
application configured under `wlp/usr/servers/fhir-server`
* install-fhir.sh installed to an existing Liberty server's
`wlp/usr/servers/defaultServer`

The latter is what we used for our Linux container image.

Now, with these changes:
* install.sh/install.bat will install to defaultServer to match the
existing container layout.
* install.sh will install on the defaultServer of an existing liberty installation

This allows us to use install.sh from our Dockerfile and delete our install-fhir.sh flavor.

I also made the following tweaks:
* Modified script output for correctness and clarity
* Removed the needless `server-runtime` directory from the distribution
zip; the liberty-runtime zip is now top-level in the zip
* In the container image, our tools and scripts are now installed to
`/opt/fhir-server` instead of `/opt/ibm-fhir-server`
* Updated container image labels


fhir-install/Dockerfile
```diff
@@ -9,30 +9,29 @@ FROM openliberty/open-liberty:22.0.0.7-kernel-slim-java11-openj9-ubi as base
...
 # The following labels are required:
-LABEL name='IBM FHIR Server'
-LABEL vendor='IBM'
-LABEL version="$FHIR_VERSION"
-LABEL summary="Image for IBM FHIR Server with OpenJ9 and UBI 8"
-LABEL description="The IBM FHIR Server is a modular Java implementation of version 4 of the HL7 FHIR specification with a focus on performance and configurability."
+LABEL name='LinuxForHealth FHIR Server'
+LABEL version="$FHIR_SERVER_VERSION"
+LABEL summary="LinuxForHealth FHIR Server with OpenJ9 and UBI 8"
+LABEL description="The LinuxForHealth FHIR Server is a modular Java implementation of version 4 of the HL7 FHIR specification with a focus on performance and configurability."
```
punktilious, line 34: should we be more specific about the FHIR versions we support?

fhir-install/src/main/resources/scripts/install.bat
```diff
@@ -90,21 +90,25 @@ if errorlevel 1 (
...
+echo 3. Deploy the database schema via the fhir-persistence-schema cli jar under %BASEDIR%\tools
+echo    and grant necessary permissions.
+echo 4. Modify the Liberty server config (server.xml) by adding/removing/modifying the XML snippets under
+echo    %WLP_ROOT%\usr\servers\defaultServer\configDropins to configure datasource definitions, 
+echo    TLS configuration (keystores), webapp security, and more.
+echo 5. Modify the FHIR server config (fhir-server-config.json) under
+echo    %WLP_ROOT%\usr\servers\defaultServer\config to configure the persistence, resource endpoints,
+echo    and related FHIR server features.
+echo You can start and stop the server with these commands:
+echo    %WLP_ROOT%\bin\server.bat start
```
punktilious, line 110: align the text for install.sh too?

fhir-parent/pom.xml
```diff
@@ -29,11 +29,16 @@
...
-            <organization>IBM Watson Health</organization>
-            <organizationUrl>http://www.ibm.com</organizationUrl>
+            <name>Lee Surprenant</name>
+            <email>lmsurpre@merative.com</email>
+            <organization>Merative</organization>
+            <organizationUrl>https://merative.com</organizationUrl>
+        </developer>
+        <developer>
+            <name>Robin Arnold</name>
+            <email>robin.arnold@merative.com</email>
```
punktilious, line 39: rarnold@merative.com

lmsurpre: spoke with robin and he decided to stick with the robin.arnold one

fhir-install/Dockerfile
punktilious: Yeah, I'm never sure where it might show up in searches. Thought it might be useful to mention both.

lmsurpre: I still think I prefer just keeping it at "version 4"... if they came out with a 4.3.1 or a 4.4.0 we'd support those too.

If we did add it, I think most folks know 4.3.0 as "R4B" and so maybe it would become something like this:
```suggestion
LABEL description="The LinuxForHealth FHIR Server is a modular Java implementation of HL7 FHIR versions R4 and R4B with a focus on performance and configurability."
```

lmsurpre: spoke with robin and he felt stronger than me on this one...merging the suggestion now

punktilious (approved):
LGTM

PR #3790, opened 2022-07-20
document parsing from and generating to java.lang.String

lmsurpre:
pretty basic info but i've had this question a few times and so thought it might be worth mentioning in the docs

also made a few more model guide tweaks/additions while I was in there (partially due to review comment)

### DIFF
--- a/docs/src/pages/guides/FHIRModelGuide.mdx
+++ b/docs/src/pages/guides/FHIRModelGuide.mdx
@@ -76,14 +76,17 @@ bodyWeight = bodyWeight.toBuilder()
 ## Parsing a Resource from an InputStream or Reader
 
 ```
-// Parse from InputStream
+// Parse JSON from InputStream
 InputStream in = getInputStream("JSON/bodyweight.json");
 Observation observation = FHIRParser.parser(Format.JSON).parse(in);
 
-// Parse from Reader
+// Parse JSON from Reader
 Reader reader = getReader("JSON/bodyweight.json");
 Observation observation = FHIRParser.parser(Format.JSON).parse(reader);
 ```
+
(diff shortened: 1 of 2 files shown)


docs/src/pages/guides/FHIRModelGuide.mdx
```diff
@@ -76,15 +76,18 @@ bodyWeight = bodyWeight.toBuilder()
 ## Parsing a Resource from an InputStream or Reader
 
 ```
-// Parse from InputStream
+// Parse JSON from InputStream
```
punktilious, line 79: Is it also worth mentioning how to configure javax.ws.rs.client.Client using providers e.g.:

```
        ClientBuilder cb =
                ClientBuilder.newBuilder()
                    .register(new FHIRProvider(RuntimeType.CLIENT))
                    .register(new FHIRJsonProvider(RuntimeType.CLIENT))
                    .register(new FHIRJsonPatchProvider(RuntimeType.CLIENT));
```

lmsurpre: I'm not sure if that makes sense here or not.  I suppose we could add a section at the bottom.

Side-note:  I bet most clients just need the FHIRProvider.  FHIRJsonProvider just lets you read and write from JsonObject (which exposes a capability that our parser has to do the same)...I'm not actually sure where its used though.  FHIRJsonPatchProvider is only relevant if your issuing JSON patches.

punktilious (approved):
LGTM

docs/src/pages/guides/FHIRModelGuide.mdx
lmsurpre: ok, i added a section for it

Issue #3676, opened 2022-05-31
Drop support for Db2


lmsurpre:
**Is your feature request related to a problem? Please describe.**
Given the cost/benefit of supporting a unique multi-tenant schema design for Db2, I think it would be in our best interest to drop this support (so we can focus elsewhere).

**Describe the solution you'd like**
Unless other users/contributors step forward in support of Db2, we should remove it wholesale.
This includes at least:
* removing the db2 driver (jcc) from the pom.xml and the built installer/container images
* datasource server.xml and configDropins related to Db2
* removing db2 from the CI pipeline (replace it with postgresql and/or use chance to conslidate the overall number of tests)
* any db2-specific code in fhir-database-util, fhir-persistence-schema, and fhir-persistence jdbc
  * anything mentioning "multi-tenant schema" could also be removed
* project documentation

**Describe alternatives you've considered**
Continue to support db2, but drop the multi-tenant schema part of it (so it works like all our other db types).

**Acceptance Criteria**

1.  
   GIVEN [a precondition]
    AND [another precondition]
   WHEN [test step]
    AND [test step]
   THEN [verification step]
    AND [verification step]

**Additional context**
Please contact us or comment here if you are dependent on Db2 to continue being supported out-of-the box.

lmsurpre:
linking to a review comment that hints that we may be able to remove some related classes when we remove Db2: 
 https://github.com/LinuxForHealth/FHIR/pull/3754#discussion_r924225874

lmsurpre:
No specific QA done; will be tested along with everything else

PR #3794, opened 2022-07-22
issue #3676 - Replace db2 with postgres in containerized e2e tests

lmsurpre:


build/audit/kafka/pre-integration-test.sh
```diff
@@ -76,7 +73,7 @@ bringup(){
     Docker container status:"
     docker ps -a
 
-    containerId=$(docker ps -a | grep kafka_fhir-server_1 | cut -d ' ' -f 1)
+    containerId=$(docker ps -a | grep fhir-server | cut -d ' ' -f 1)
```
punktilious, line 76: I'm assuming this is intentionally less selective?

build/common/gather-logs.sh
```diff
@@ -29,7 +29,7 @@ package_logs(){
     echo $(date) > ${it_results}/runtime.txt
 
     # Look for the FHIR Server Container
-    containerId=$(docker ps -a | grep ibm-fhir-server | cut -d ' ' -f 1)
+    containerId=$(docker ps -a | grep fhir-server | cut -d ' ' -f 1)
```
punktilious, line 32: Could there ever be more than one fhir-server active at a time? Just want to check that the name `fhir-server` is now being used for any FHIR server container, and that there's only one active at a time so we don't need to call them different things.

build/docker/postgres/init/db.sql
```diff
@@ -0,0 +1,32 @@
...
+CREATE SCHEMA IF NOT EXISTS fhir_oauth;
+CREATE SCHEMA IF NOT EXISTS fhir_jbatch;
+
+-- only for the fhiradmin user
+GRANT ALL ON schema fhirdata TO fhiradmin WITH GRANT OPTION;
+GRANT ALL ON schema fhir_admin TO fhiradmin WITH GRANT OPTION;
+GRANT ALL ON schema fhir_oauth TO fhiradmin WITH GRANT OPTION;
+GRANT ALL ON schema fhir_jbatch TO fhiradmin WITH GRANT OPTION;
+
+GRANT USAGE ON SCHEMA fhirdata to fhirserver;
```
punktilious, line 25: strictly speaking the create schema and related grants for the `fhirserver` user should be performed by the CLI tool.

build/migration/bin/4_current-migrate.sh
```diff
@@ -19,7 +19,7 @@ run_migrate(){
     mvn -T2C -B install --file fhir-parent -DskipTests -P include-fhir-igs,integration --no-transfer-progress
 
     cd fhir-install
-    docker build -t ibmcom/ibm-fhir-server:snapshot .
+    docker build --build-arg VERBOSE=false -t ibmcom/ibm-fhir-server:snapshot .
```
punktilious, line 22: Do we know how long these old images will be available?

build/docker/postgres/init/db.sql
lmsurpre: lets open this one as issue

punktilious (approved):
Reviewed with Lee. LGTM

build/common/gather-logs.sh
lmsurpre: not in our CI (at least not that i know of).  i suppose its possible if you're running things locally and you have a different fhir-server image running.  but i don't expect that to be common

build/migration/bin/4_current-migrate.sh
lmsurpre: no, we don't.  i don't think they'll go away any time soon.
that said, in this particular case, we're actually building an ibmcom/ibm-fhir-server image from source.
in all the other spots, I updated the image name.  But specifically for the migration tests, the `ibmcom/ibm-fhir-server` name is actually in our source code under fhir-install/pom.xml and so I had to chose between:
A. patching that file in the older version that we checkout so that it uses linuxforhealth/fhir-server ; or
B. continuing to using ibmcom/ibm-fhir-server for the migration tests

For now, I chose B.  We should be able to change it once we stop testing migrations from 4.x

Issue #3564, opened 2022-04-05
Move payload offloading documentation to the User's Guide


lmsurpre:
**Is your feature request related to a problem? Please describe.**
The current documentation for the 'payload offloading" feature is in a README at https://github.com/IBM/FHIR/tree/main/fhir-persistence-blob-app

As a user, I would expect to find that information in the user's guide where we document other server configuration options.

In my opinion, the README for `https://github.com/IBM/FHIR/tree/main/fhir-persistence-blob-app` should be about how to run the reconciliation app, not about how to configure the server for payload offloading.

**Describe the solution you'd like**
1. add a subsection under [3.3 Persistence Layer Configuration](https://ibm.github.io/FHIR/guides/FHIRServerUsersGuide#33-persistence-layer-configuration)
1. add the new `payload` config properties to the tables in section 5.1

**Describe alternatives you've considered**
have a dedicated guide for payload config outside of the user's guide

**Additional context**

PR #3795, opened 2022-07-22
issue #3564 move payload offloading configuration to FSUG

punktilious:


docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -483,6 +483,127 @@ The TransactionManager controls the timeout of database queries.
...
+
+1. Payload offloading is enabled at the server level. Offloading cannot be enabled/disabled on a per-tenant basis. If you want to support offloading for just one tenant, use a different IBM FHIR Server instance;
+2. Payload offloading must be configured prior to ingesting any resource data;
+3. Payload offloading must not be disabled after resource data has been ingested;
+4. When payload offloading is enabled, FHIR resources are limited to 5000 MiB in size when rendered as uncompressed JSON. The practical limit is likely to be lower due to memory usage and typical socket and transaction timeouts.
+
+
+#### 3.3.2.2 Pick a Container Name
+
+Each tenant/datasource combination requires its own container. Use a name that can be easily identified as belonging to the tenant and datasource, subject to the Azure Blob service naming restrictions.
```
lmsurpre, line 511: a link to said naming restrictions might be nice...i recall there being at least one gotcha

docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -483,6 +483,127 @@ The TransactionManager controls the timeout of database queries.
...
+    "fhirServer": {
+        ...
+        "persistence": {
+            "factoryClassname": "com.ibm.fhir.persistence.blob.FHIRPersistenceJDBCBlobFactory",
+            "datasources": {
+                "default": {
+                    ...
+                }
+            },
+            "payload": {
```
lmsurpre, line 541: I think we should add this config property to the appendix tables in section 5.1.
If we want, we can have a very high-level description there that points back to here.

docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -483,6 +483,127 @@ The TransactionManager controls the timeout of database queries.
...
+                        "containerName": "default-default"
+                    }
+                }
+            }
+        }
+    }
+}
+```
+
+Container names allowed by Azure Blob are more restrictive than LinuxForHealth FHIR Server tenant and datasource names (for example they must be lower case and `_` is not allowed - see the [official documentation](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blobs-introduction#containers) for more details). For this reason, the container name for each tenant and datasource must be specified in the `fhirServer/persistence/payload/connectionProperties/containerName` property.
```
lmsurpre, line 556: Ah, now here is the content I was expecting to see in 3.3.2.2.
I also see the benefit of having it here since it explains why containerName is required.
Maybe just leave it as is and I'll mark my other comment as resolved too.

docs/src/pages/guides/FHIRServerUsersGuide.md
lmsurpre: i found it below...should be OK

docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -483,6 +483,127 @@ The TransactionManager controls the timeout of database queries.
...
+```
+java -jar fhir-persistence-blob-app-*-cli.jar \
+    --fhir-config-dir /path/to/wlp/usr/servers/defaultServer \
+    --tenant-id <tenant-id> \
+    --ds-id <ds-id> \
+    --create-container \
+    --no-dry-run
+```
+
+The container name will be read from the tenant's fhir-server-config.json configuration, so it is important to complete that configuration step before running this command.
```
lmsurpre, line 602: not for this PR, but we should consider adding fhir-persistence-blob-app-*-cli.jar alongside the schema jar in our zip installation package.  i think that will put it at /opt/fhir-server/tools in our main image

docs/src/pages/guides/FHIRServerUsersGuide.md
lmsurpre: i see them there now.

lmsurpre (approved):
LGTM

PR #3813, opened 2022-07-26
tweak postgres settings

lmsurpre:
Our other postgres compose environments use an init script that modifies a lot of postgres settings as follows:
```
    sed -i -e"s/^max_connections = 100.*$/max_connections = 100/" /db/data/postgresql.conf
    sed -i -e"s/^max_prepared_transactions = 100.*$/max_prepared_transactions = 100/" /db/data/postgresql.conf
    sed -i -e"s/^max_locks_per_transaction = .*$/max_locks_per_transaction = 128/" /db/data/postgresql.conf
    sed -i -e"s/^shared_buffers =.*$/shared_buffers = 4GB/" /db/data/postgresql.conf
    sed -i -e"s/^#effective_cache_size = 128MB.*$/effective_cache_size = 2GB/" /db/data/postgresql.conf
    sed -i -e"s/^#work_mem = 1MB.*$/work_mem = 1MB/" /db/data/postgresql.conf
    sed -i -e"s/^#maintenance_work_mem = 16MB.*$/maintenance_work_mem = 512MB/" /db/data/postgresql.conf
    sed -i -e"s/^#checkpoint_segments = .*$/checkpoint_segments = 32/" /db/data/postgresql.conf
    sed -i -e"s/^#checkpoint_completion_target = 0.5.*$/checkpoint_completion_target = 0.7/" /db/data/postgresql.conf
    sed -i -e"s/^#wal_buffers =.*$/wal_buffers = 16MB/" /db/data/postgresql.conf
    sed -i -e"s/^#default_statistics_target = 100.*$/default_statistics_target = 100/" /db/data/postgresql.conf
```

For the new CI work, I took a more minimalistic approach...starting with the base postgresql image and only modifying what I needed to...mostly just setting a larger `shared_buffers`.
However, I found advice online about
1. setting shm_size higher than shared_buffers; and
2. increasing max_wal_size when you increase shared_buffers

So this PR adjusts the config accordingly.  If it works well, I'd like to try this config in our other compose environments (including `/demo`) as well.

### DIFF
--- a/build/docker/docker-compose.yml
+++ b/build/docker/docker-compose.yml
@@ -7,6 +7,8 @@ volumes:
 services:
   postgres:
     image: postgres:15beta2
+    # https://dba.stackexchange.com/q/275378/46472
+    shm_size: 2GB
     volumes:
       - type: volume
         source: pgdata
@@ -24,8 +26,11 @@ services:
       interval: 5s
       timeout: 4s
       retries: 3
+    # "Larger settings for shared_buffers usually require a corresponding increase in max_wal_size, 
+    # to spread out the process of writing large quantities of new or changed data over a longer period of time."
     command: "
-      -c shared_buffers=2GB
+      -c shared_buffers=1GB
+      -c max_wal_size=2GB
       -c max_connections=100
       -c max_prepared_transactions=100
       -c max_locks_per_transaction=128


build/docker/docker-compose.yml
```diff
@@ -24,8 +26,11 @@ services:
       interval: 5s
       timeout: 4s
       retries: 3
+    # "Larger settings for shared_buffers usually require a corresponding increase in max_wal_size, 
+    # to spread out the process of writing large quantities of new or changed data over a longer period of time."
     command: "
-      -c shared_buffers=2GB
+      -c shared_buffers=1GB
```
punktilious, line 32: What value do you currently have for wal_buffers? We've found that bumping this up to 16MB improves performance (in our ingestion workload).

punktilious:
The max_wal_size you need really depends on the workload and is a balance between throughput and recovery time. If you have a very large max_wal_size, it can delay checkpoints which can help to reduce IO if the same blocks are updated multiple times. But for our FHIR ingestion workloads, we're pretty much always writing new blocks, so we don't gain much by delaying checkpoints (apart from bursts of performance, where it might seem as though throughput has improved even though only for a little bit). 

You'll get a warning in the PostgreSQL logs like this if PostgreSQL thinks it is checkpointing too frequently:
```
2022-07-27 08:39:36.878 UTC [11] LOG:  checkpoints are occurring too frequently (29 seconds apart)
2022-07-27 08:39:36.878 UTC [11] HINT:  Consider increasing the configuration parameter "max_wal_size".
```
See the values for checkpoint_warning, checkpoint_timeout and checkpoint_completion_target.

build/docker/docker-compose.yml
lmsurpre: i'm not setting that currently, but could easily add it.  however, based on https://postgresqlco.nf/doc/en/param/wal_buffers/
> The default setting of -1 selects a size equal to 1/32nd (about 3%) of [shared_buffers](https://postgresqlco.nf/doc/en/param/shared_buffers/)

0.03 * 1 GB = 60 MB, so a setting of 16 MB would actually be reducing it.
The default setting of -1 selects a size equal to 1/32nd (about 3%) of [shared_buffers](https://postgresqlco.nf/doc/en/param/shared_buffers/)

lmsurpre:
> The max_wal_size you need really depends on the workload and is a balance between throughput and recovery time. If you have a very large max_wal_size, it can delay checkpoints which can help to reduce IO if the same blocks are updated multiple times. But for our FHIR ingestion workloads, we're pretty much always writing new blocks, so we don't gain much by delaying checkpoints (apart from bursts of performance, where it might seem as though throughput has improved even though only for a little bit).

Yeah, I added it just based on this line in the postgres docs, not any real issues:
> Larger settings for shared_buffers usually require a corresponding increase in [max_wal_size](https://postgresqlco.nf/doc/en/param/max_wal_size/), in order to spread out the process of writing large quantities of new or changed data over a longer period of time.

I could easily take back out if you want, but I figured in both CI and demo we don't care too much about "recovery time".
The default was 1GB.

build/docker/docker-compose.yml
punktilious: From the docs: "The default setting of -1 selects a size equal to 1/32nd (about 3%) of [shared_buffers](https://www.postgresql.org/docs/current/runtime-config-resource.html#GUC-SHARED-BUFFERS), but not less than 64kB nor more than the size of one WAL segment, typically 16MB."

So for larger sizes of shared_buffers, we'll get the 16MB. We were seeing smaller values on some of the database instances we had provisioned, and so setting it manually helped.

Net: in real world scenarios shared_buffers will be sufficiently large enough (>512MB) to ensure wal_buffers hits 16MB.

punktilious (approved):
LGTM

PR #3814, opened 2022-07-26
copy the new postgres ci config to the demo env

lmsurpre:


demo/docker-compose.yml
```diff
@@ -1,36 +1,42 @@
...
+      POSTGRES_DB: fhirdb
+      POSTGRES_USER: fhiradmin
+      POSTGRES_PASSWORD: change-password
+      PGDATA: /var/lib/postgresql/data/pgdata
     healthcheck:
-      start_period: 21s
       test: ["CMD-SHELL", "pg_isready -d fhirdb -U fhiradmin"]
-      interval: 10s
-      timeout: 5s
+      interval: 5s
```
punktilious, line 24: I suppose this is OK for a new database. If the database were restarting and needed to run recovery it could take longer to start up. The short setting is probably fine for a CI environment, but for a demo environment where the database may be restarted (perhaps after a crash), this might be an issue.

demo/docker-compose.yml
```diff
@@ -1,36 +1,42 @@
...
+      timeout: 4s
       retries: 3
+    command: "
+      -c shared_buffers=2GB
```
punktilious, line 28: make sure the container is also allocated sufficient memory

demo/docker-compose.yml
lmsurpre: what i found:  the demo env uses a persistent volume, so after the initial setup the postgres image was routinely starting in less than 5 seconds.  however, because our fhir service delays its startup until the db is healthy, we were waiting needlessly.
however, we can add a `start_period` back in to ensure that an initial slow startup doesn't cause use any trouble (while still allowing it to become "healthy" in 5 seconds if its ready).
```suggestion
      start_period: 22s
      interval: 5s
```

demo/docker-compose.yml
lmsurpre: do you know what happens if shared_buffers is set higher than the alotted memory?

lmsurpre: I could drop this back down to 1 GB like I have in our CI pipeline, but I was trying a slightly higher value since we used to set it to 4 GB.

punktilious: The database will halt if it can't allocate sufficient shared memory. The default configuration for PostgreSQL is to use mmap to attach to anonymous shared memory using /dev/shm which on Linux is mounted as a tmpfs and is sized by specifying the `--shm-size` argument to docker run (or docker-compose equivalent).

If you bash into the container you can see the mount:
```
docker exec -ti your-container-name bash
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
bash-5.0# mount | grep shm
shm on /dev/shm type tmpfs (rw,context="system_u:object_r:container_file_t:s0:c640,c766",nosuid,nodev,noexec,relatime,size=3906252k,uid=1000,gid=1001)
```
In this case, the container was started with `--shm-size=4g` which corresponds to `size=3906252k` (4000/1024).

However, it looks like PostgreSQL isn't using /dev/shm for its shared buffers. It uses shared memory segments from /dev/shm during parallel query (think our large search queries). The amount of memory it uses will depend on the number of parallel query workers and the work_mem.

punktilious: In terms of accommodating shared_buffers, it depends if the container has been configured with a memory limit. If not, then it depends how much (virtual) memory is available on the host.

demo/docker-compose.yml
```diff
@@ -1,36 +1,42 @@
...
+    driver: local
 services:
   postgres:
-    build:
-      context: postgres
-      dockerfile: Dockerfile
-    shm_size: 256MB
+    image: postgres:15beta2
+    # https://dba.stackexchange.com/q/275378/46472
+    shm_size: 3GB
```
lmsurpre, line 9:
```suggestion
    shm_size: 256MB
```

Based on https://dba.stackexchange.com/a/275385/46472 I thought this needed to be higher than shared_buffers, but Robin set me straight.

punktilious (approved):
LGTM

Issue #3787, opened 2022-07-19
Interceptor data sent back to the caller


albvegli:
Using the interceptor I need to add more information to the returned afterSearch (bundle) data.
I have generated a new bundle but cannot send it to the caller. Everything I've tried to do doesn't change the packet sent to the caller.
How can I update the packet sent back to the caller?

punktilious:
@albvegli thank you for opening this issue. Unfortunately the current code does not return the updated resource if it is changed by the interceptor. In FHIRRestHelper:

```
            // Invoke the 'afterSearch' interceptor methods.
            getInterceptorMgr().fireAfterSearchEvent(event);

            // Commit our transaction if we started one before.
            txn.commit();
            txn = null;

            return bundle;
```
We'll add this to our backlog. The team needs to discuss this to see if there are any implications for supporting this behavior. To make sure we properly consider your scenario, could you provide more detail on what changes your interceptor needs to make to the bundle?

punktilious:
Thanks. It's an interesting point you raise about interceptors invoking interactions which then cause nested calls to the interceptor.  It seems like it would be possible to track this in the request context, which would make it a lot easier for an implementer (you) to understand where you are and react appropriately. I'll discuss with the team and get back to you.

lmsurpre:
@albvegli how are you performing the search from within your interceptor.

I would have expected your interceptor to use `event.getPersistenceImpl()` to get the FHIRPersistence implementation from the FHIRPersistenceEvent.

Its a little tricky to set up the FHIRSearchContext, but it should be possible to invoke "FHIRPersistence.search" from your interceptor and avoid going all the way back out through the REST layer.  If done this way, there is no "re-entry" because the interceptors are invoked outside of the persistence layer.

Let us know.

albvegli:
Yes, I’m usig event.getPersistenceImpl().

Patient pat = null;
try {
  FHIRRestHelper helper = new FHIRRestHelper(event.getPersistenceImpl(),searchHelper); 
  Resource resource = helper.doRead("Patient", id, false, false, null, null).getResource();
  pat = (Patient) resource;
} catch (Exception e) {
  e.printStackTrace();
}

I have now stopped developing this feature due to the inability to return additional information. If I need it later I will try to follow your advice.
Thank you.

punktilious:
As @lmsurpre mentioned, you can use the persistence layer to perform the read directly:
```
    /**
[...]

PR #3815, opened 2022-07-26
issue #3787 allow afterSearch interceptor to modify resource

punktilious:

Allows persistence interceptors to modify the resource when processing the following events:

- afterRead
- afterVread
- afterSearch
- afterHistory

Currently the interceptor is trusted to not do anything which would invalidate the resource for the given interaction.

### DIFF
--- a/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
+++ b/fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
@@ -1401,6 +1401,13 @@ public class FHIRRestHelper implements FHIRResourceHelpers {
             // Invoke the 'afterSearch' interceptor methods.
             getInterceptorMgr().fireAfterSearchEvent(event);
 
+            // Interceptors might want to change the response bundle, so make sure pick up the new value.
+            // Protect against an interceptor returning something other than a Bundle
+            Resource afterSearchEventValue = event.getFhirResource();
+            if (afterSearchEventValue.is(Bundle.class)) {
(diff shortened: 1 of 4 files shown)


fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -1401,6 +1401,13 @@ public Bundle doSearch(String type, String compartment, String compartmentId,
             // Invoke the 'afterSearch' interceptor methods.
             getInterceptorMgr().fireAfterSearchEvent(event);
 
+            // Interceptors might want to change the response bundle, so make sure pick up the new value.
+            // Protect against an interceptor returning something other than a Bundle
```
lmsurpre, line 1405: ClassCastException might actually be ok in that case...it would trigger a 500 server error which seems appropriate for an interceptor thats setting a response resource which could not be cast to the expected response object.
It make me wonder if we should have an else here that emits a warning.
Hopefully no one would do that though, so probably doesn't matter much.

punktilious: Yeah, was having similar thoughts on how to handle this. We probably also want to protect against the resource type being changed (e.g. a Patient read returning an Observation). I was also thinking we might want to verify that the resource id wasn't changed (easy for read/vread...a bit harder if we wanted to check everything inside a search/history response Bundle).

lmsurpre (approved):
LGTM

Issue #3805, opened 2022-07-25
FHIRPath evaluator should throw when string function is invoked on non-string


lmsurpre:
**Describe the bug**
http://hl7.org/fhirpath/N1/#string-manipulation says this about the String functions in this section:
> If there is more than one item, or an item that is not a String, the evaluation of the expression will end and signal an error to the calling environment.

However, users report that our engine returns empty in this case instead of throwing.

**Environment**
4.11.1

**To Reproduce**
See https://chat.fhir.org/#narrow/stream/179266-fhirpath/topic/error.20handling

**Expected behavior**
"the evaluation of the expression will end and signal an error to the calling environment"

**Additional context**

lmsurpre:
I tried the happy path on various data types including:

- [x] Resource.id, Element.id, and Extension.url (System.String)
- [x] Extension.value.as(FHIR.string)
- [x] Extension.value.as(FHIR.uri)
- [x] Extension.value.as(FHIR.code)
- [x] Extension.value.as(FHIR.oid)
- [x] Extension.value.as(FHIR.uuid)
- [x] Extension.value.as(FHIR.markdown)
- [x] Extension.value.as(FHIR.base64Binary)

Interestingly, I found that the implementation for oid and uuid also match values with the prefix (e.g. "id" will always match the "oid" or "uuid" in the prefix, regardless of the actual unique values).
Similarly, I found that the base64Binary variant matches the encoded base64Binary string, not the decoded value.
I have no clue if that is 'correct' or not, but it seems reasonable and so I"m moving on.

I also confirmed the negative cases for
* more than one item in the input collection
* an input collection of size 1 where the type is not string

PR #3819, opened 2022-07-27
issue #3805 - fix FHIRPath evaluator string function error handling

PrasannaHegde1:


fhir-path/src/main/java/com/ibm/fhir/path/function/ToStringFunction.java
```diff
@@ -38,9 +38,7 @@ public int getMaxArity() {
 
     @Override
     public Collection<FHIRPathNode> apply(EvaluationContext evaluationContext, Collection<FHIRPathNode> context, List<Collection<FHIRPathNode>> arguments) {
-        if (!isSingleton(context)) {
-            return empty();
-        }
+        checkStringValue(context);
```
lmsurpre, line 41: toString doesn't exist under http://hl7.org/fhirpath/N1/#string-manipulation and so shouldn't use this check.  

instead this one falls under section 5.5 which says this:
> The functions in this section operate on collections with a single item. If there is more than one item, the evaluation of the expression will end and signal an error to the calling environment.

fhir-path/src/main/java/com/ibm/fhir/path/function/ContainsFunction.java
```diff
@@ -36,9 +35,7 @@ public int getMaxArity() {
     
     @Override
     public Collection<FHIRPathNode> apply(EvaluationContext evaluationContext, Collection<FHIRPathNode> context, List<Collection<FHIRPathNode>> arguments) {
-        if (!hasStringValue(context)) {
-            return empty();
-        }
+        checkStringValue(context);
```
lmsurpre, line 38: Most (all?) of the function definitions say that if the input collection is empty, then the result must be empty as well (not an error).  Please make sure we're handling this case.

fhir-path/src/main/java/com/ibm/fhir/path/function/ToStringFunction.java
```diff
@@ -38,7 +39,10 @@ public int getMaxArity() {
 
     @Override
     public Collection<FHIRPathNode> apply(EvaluationContext evaluationContext, Collection<FHIRPathNode> context, List<Collection<FHIRPathNode>> arguments) {
-        checkSingleton(context);
+        if(!isSingleton(context)) {
```
lmsurpre, line 41: I think this one should return empty if the input collection is empty, but I don't think it does with this

fhir-path/src/main/java/com/ibm/fhir/path/function/FHIRPathStringAbstractFunction.java
```diff
@@ -0,0 +1,53 @@
...
+ * Abstract class which contains common apply logic for all abstract string manipulation functions
+ */
+public abstract class FHIRPathStringAbstractFunction extends FHIRPathAbstractFunction {
+    
+    /**
+     * @implSpec Common apply logic for all abstract string manipulation functions. Delegates to doApply()
+     */
+    @Override
+    public Collection<FHIRPathNode> apply(EvaluationContext evaluationContext, Collection<FHIRPathNode> context, List<Collection<FHIRPathNode>> arguments) {
+        if(context.isEmpty()) {
```
lmsurpre, line 31: minor style thing:  in this project we always put a space between `if` and `(` to distinguish from function calls (and same for other control flow keywords `for`, `while`, etc.)

fhir-path/src/main/java/com/ibm/fhir/path/function/FHIRPathStringAbstractFunction.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.Collection;
+import java.util.List;
+
+import com.ibm.fhir.path.FHIRPathNode;
+import com.ibm.fhir.path.FHIRPathType;
+import com.ibm.fhir.path.evaluator.FHIRPathEvaluator.EvaluationContext;
+
+/**
+ * Abstract class which contains common apply logic for all abstract string manipulation functions
```
lmsurpre, line 22: nice

fhir-path/src/main/java/com/ibm/fhir/path/function/ContainsFunction.java
lmsurpre: looks good in the new common parent; I left a separate comment for the toString function

fhir-path/src/main/java/com/ibm/fhir/path/util/FHIRPathUtil.java
```diff
@@ -1238,4 +1238,21 @@ public static com.ibm.fhir.model.type.String getDisplay(FHIRPathTree tree, FHIRP
...
         return null;
     }
+    
+    /**
+     * returns true if the node is a String or if FHIR element can be automatically converted to System.String 
+     * https://www.hl7.org/fhir/fhirpath.html#types
+     * @param node
+     * @return boolean
+     */
+    public static boolean isStringSubType(FHIRPathNode node) {
```
lmsurpre, line 1248: One thing I didn't think of on the call... we should be clear about whether this should return true or false for FHIRPathNode of type SYSTEM_STRING

Options:
A. add this type to the list of types we return true for; or
B. change the method signature to `isFHIRStringSubType(FHIRPathElementNode)` to make it clear this only operates over FHIRPathNode that wrap FHIR elements (not primitive values)

fhir-path/src/main/java/com/ibm/fhir/path/function/FHIRPathStringAbstractFunction.java
```diff
@@ -0,0 +1,59 @@
...
+        if (!isSingleton(context)) {
+            throw new IllegalArgumentException("Input collection must not contain more than one item, but found " + context.size());
+            
+        } else if (!hasStringValue(context)) {
+            FHIRPathNode node = getSingleton(context);
+            if (isStringSubType(node)) {
+                logger.fine("collection item does not have a value, returning empty");
+                return empty();
+            }
+            throw new IllegalArgumentException("Input collection item must be of type String, but found '"+node.type().getName()+ "'");
```
lmsurpre, line 48: very minor style thing
```suggestion
            throw new IllegalArgumentException("Input collection item must be of type String, but found '" + node.type().getName() + "'");
```

PrasannaHegde1: fixed the above review comment

fhir-path/src/main/java/com/ibm/fhir/path/util/FHIRPathUtil.java
PrasannaHegde1: Handled the above review comment with Option A.

lmsurpre (approved):
LGTM

Issue #3420, opened 2022-03-02
Test payload offloading with azurite as part of github action CI 


lmsurpre:
**Is your feature request related to a problem? Please describe.**
Today we have no end-to-end test coverage for the payload-offload feature with azure-blob

**Describe the solution you'd like**
As part of our CI:
1. spin up an azurite container
2. configure a fhir server instance with payload offload enabled
3. run the full suite for end-to-end tests

**Describe alternatives you've considered**

**Acceptance Criteria**

1.  
   make sure it runs as part of CI

**Additional context**
Just like https://github.com/ibm/fhir/issues/3239 but for azure-blob

punktilious:
I don't think that payload offloading is actually happening because the configuration in `fhir-server-config-postgresql-azurite.json` specifies the standard persistence factory:
```
        "factoryClassname": "org.linuxforhealth.fhir.persistence.jdbc.FHIRPersistenceJDBCFactory",
```

For offloading, the persistence factory needs to be:

```
        "factoryClassname": "com.ibm.fhir.persistence.blob.FHIRPersistenceJDBCBlobFactory",
```

punktilious:
Confirmed that we are now using an azurite payload configuration in one of the integration tests:

```
[8/24/22, 10:31:43:227 UTC] 00000028 FHIRRestServl I   Received request: tenantId:[default] dsId:[default] user:[fhiruser] method:[POST] uri:[https://localhost:9443/fhir-server/api/v4/ActivityDefinition]
[8/24/22, 10:31:43:231 UTC] 0000003b FHIRRestServl I   Received request: tenantId:[default] dsId:[default] user:[fhiruser] method:[POST] uri:[https://localhost:9443/fhir-server/api/v4/AdverseEvent]
[8/24/22, 10:31:43:247 UTC] 00000037 FHIRRestServl I   Received request: tenantId:[default] dsId:[default] user:[fhiruser] method:[POST] uri:[https://localhost:9443/fhir-server/api/v4/AllergyIntolerance]
[8/24/22, 10:31:43:256 UTC] 0000002b FHIRRestServl I   Received request: tenantId:[default] dsId:[default] user:[fhiruser] method:[POST] uri:[https://localhost:9443/fhir-server/api/v4/ActivityDefinition]
[...]

PR #3829, opened 2022-08-01
issue #3420 - initial pass at adding azurite env to ci

lmsurpre:
1. Refactored pre and post docker integration test scripts to take an argument
that identifies which environment to setup/teardown. Almost everything is shared
but each environment has a single "configure.sh" script to do environment-specific config.
2. Renamed the current "docker" directory to "minio"
3. Introduced a new environment flavor for "azurite"
4. Updated the integration github action workflow to invoke these scripts with an environment argument
from the "environment" test matrix (currently "minio" and "azurite")


build/post-integration-test-docker.sh
```diff
@@ -21,9 +32,9 @@ rm -rf ${it_results} 2>/dev/null
 mkdir -p ${it_results}/server-logs
 mkdir -p ${it_results}/fhir-server-test
 
-containerId=$(docker ps -a | grep fhir | cut -d ' ' -f 1)
-if [[ -z "${containerId}" ]]; then
-    echo "Warning: Could not find fhir container!!!"
+containerId=$(docker ps -a | grep ${1}[-_]fhir-server | cut -d ' ' -f 1)
```
punktilious, line 35: some common functions would be good for this kind of thing (I'm assuming this isn't the only place we need something like this.

build/post-integration-test-docker.sh
```diff
@@ -35,12 +46,16 @@ else
...
+fi
+
+# Source the tenant1 datastore variables; not strictly needed here
+# but avoids "variable is not set" docker compose warnings
+. ${WORKSPACE}/build/common/set_tenant1_datastore_vars.sh
 
 echo "Bringing down the fhir server docker container(s)..."
-cd ${DIR}/docker
-docker-compose down
+cd ${DIR}/${1}
```
punktilious, line 58: might be worth capturing `${1}` in a variable and checking its value before using it. Also, I notice our scripts aren't quoting any of the path arguments, so this assumes we won't have any spaces anywhere. That's probably a reasonable restriction though.

fhir-install/Dockerfile
```diff
@@ -18,7 +18,6 @@ RUN unzip -qq /tmp/fhir-server-distribution.zip -d /tmp && \
     mv /tmp/fhir-server-dist/tools /opt/fhir-server/tools
 COPY src/main/docker/ibm-fhir-server/bootstrap.properties /opt/ol/wlp/usr/servers/defaultServer/
```
punktilious, line 19: still ibm-fhir-server?

build/post-integration-test-docker.sh
lmsurpre: agree

build/post-integration-test-docker.sh
lmsurpre: any specific recommendation for "checking its value before using it"?

we do check that its nonnull/nonempty at the top:
```
if [ -z ${1} ]; then
  echo "This script requires an argument that points to one of the integration folders"
  exit 1
fi
```

maybe we should prevent `..` in there as well?  anything else?

lmsurpre: I ended up adding this test `[ -n "$(find ${DIR} -maxdepth 1 -name ${1} -type d)" ]` which is possibly over-restrictive but should be a bit safer

build/post-integration-test-docker.sh
lmsurpre: I added it under common with a filename of `gather-pretest-logs.sh`.  I also renamed the existing `gather-logs.sh` to `gather-posttest-logs.sh` to avoid confusion

build/post-integration-test-docker.sh
punktilious: You shouldn't need to make that a find. I think you can just check it exists and is a directory (but it looks like the find is doing something similar).
````
if [ ! -d "${DIR}" ]; then
   ...
   exit
fi
````

Comment (lmsurpre) [reply build/post-integration-test-docker.sh]: the find ensure that its a subdirectory of ${DIR} that we find.
it could have been 
```
if [ ! -d "${DIR}/${1}" ]; then
   ...
   exit
fi
```

but then ${1} could have `..`

like i said, probably overkill, but i think its ok

fhir-install/Dockerfile
lmsurpre: I added a commit to this PR for this one.
I left the `ibm-fhir-*` names in just two places:
1. the migration tests (so that we can continue using a mix of old and new); and
2. the security workflow (because I havn't touched this one much and it probably needs some dedicated attention)

We've decided to use GitHub Packages (ghcr) for our docker images moving forward,
so we'll have more release automation work to make that happen.  I'll open a separate issue for that part of it.

lmsurpre: this is now https://github.com/LinuxForHealth/FHIR/issues/3845

punktilious (approved):
LGTM

Issue #3831, opened 2022-08-02
reindex uses individual parameter table delete statements


punktilious:
**Describe the bug**
The reindex operation deletes all the resource parameter values before inserting them again. Currently, the delete is performed in the ParameterTableSupport class which executes individual delete statements for every parameter table.

This incurs a significant number of round-trips with the database (and hence overhead). For PostgreSQL, the delete should be performed using the stored procedure call delete_resource_parameters.  This replaces 16 database calls with 1.

**Environment**
Which version of IBM FHIR Server? main 5.0.0-SNAPSHOT

**To Reproduce**
1. Set up a PostgreSQL instance
2. Load data
3. Perform reindex

**Expected behavior**
Reindex should be performed efficiently.

lmsurpre:
Robin deployed to our cloud account and reindex throughput improved by almost 2x in this environment.

PR #3832, opened 2022-08-02
issue #3831 use stored procedure to delete parameters for reindex

punktilious:


lmsurpre (approved):
LGTM

Issue #3304, opened 2022-02-08
Integration test  com.ibm.fhir.server.test.examples needs to be updated to monitor for error messages


d0roppe:
**Is your feature request related to a problem? Please describe.**
In running this integration test  com.ibm.fhir.server.test.examples, if it gets some errors it should be more aware of that and fail if they occur 

```
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:836)
Feb 07, 2022 3:05:04 PM com.ibm.fhir.model.spec.test.R4ExamplesDriver processIndex
INFO: Overall success rate = 16/629 = 2%. Took 130415 ms
PASSED: processExamples

===============================================
    Default test
    Tests run: 1, Failures: 0, Skips: 0
===============================================


===============================================
Default suite
Total tests run: 1, Passes: 1, Failures: 0, Skips: 0
===============================================
```

**Describe the solution you'd like**
if the test encounters error messages in data setup, it should fail

**Describe alternatives you've considered**

**Acceptance Criteria**

Run the test and if errors occurred it should not pass. 

**Additional context**

lmsurpre:
Seeing this again:
```
WARNING: Response body: {"resourceType":"OperationOutcome","id":"ac-12-0-4-6e13de61-83fb-429b-8a14-bd53354aa837","issue":[{"severity":"error","code":"not-found","details":{"text":"The requested resource type 'Evidence' is not found"}}]}
Exception in thread "pool-2-thread-5" java.lang.AssertionError: expected:<201> but was:<404>
	at org.testng.AssertJUnit.fail(AssertJUnit.java:65)
	at org.testng.AssertJUnit.failNotEquals(AssertJUnit.java:467)
	at org.testng.AssertJUnit.assertEquals(AssertJUnit.java:88)
	at org.testng.AssertJUnit.assertEquals(AssertJUnit.java:318)
	at org.testng.AssertJUnit.assertEquals(AssertJUnit.java:328)
	at com.ibm.fhir.server.test.FHIRServerTestBase.assertResponse(FHIRServerTestBase.java:588)
	at com.ibm.fhir.server.test.examples.ExampleRequestProcessor.process(ExampleRequestProcessor.java:86)
[...]

PR #3836, opened 2022-08-02
issue #3304 - favor Exceptions over Errors in ExampleRequestProcessor

lmsurpre:
The method signature indicates that Exception is thrown, but the
implementation was throwing AssertionErrors instead.
This lead us to miss these in our exception handling and, since they
were within an ExecutorPool, it never got bubbled up to the test thread.

After this change, 3 examples from our MINIMAL set were failing due to 
the server defaulting to fhirVersion 4.0 and so we now explicitly set the 
fhirVersion to 4.3 in ExampleRequestProcessor.

I made some other minor changes (mostly style preference stuff) along
the way and included those in the commit.

### DIFF
--- a/fhir-model/src/test/java/com/ibm/fhir/model/spec/test/R4ExamplesDriver.java
+++ b/fhir-model/src/test/java/com/ibm/fhir/model/spec/test/R4ExamplesDriver.java
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corp. 2019, 2021
+ * (C) Copyright IBM Corp. 2019, 2022
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -151,9 +151,6 @@ public class R4ExamplesDriver {
             if (testCount.get() > 0) {
                 long elapsed = (System.nanoTime() - start) / 1000000;
 
-                // Just for formatting
-                System.out.println();
-
                 // We count overall success if we successfully process the resource,
                 // or if we got an expected exception earlier on
                 logger.info("Overall success rate = " + successCount + "/" + testCount + " = "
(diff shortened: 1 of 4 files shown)


fhir-model/src/test/java/com/ibm/fhir/model/spec/test/R4ExamplesDriver.java
```diff
@@ -261,17 +262,15 @@ public void processExample(String file, Format format, Expectation expectation)
...
-                    firstException = error;
-                }
-                throw error;
+                throw new ExampleProcessorException(file, expectation, actual);
             } else {
                 // validate and process the example
                 actual = processExample(file, resource, expectation);
             }
         } catch (ExampleProcessorException e) {
+            if (firstException == null) {
```
punktilious, line 271: What is the purpose of changing the logic here?

fhir-model/src/test/java/com/ibm/fhir/model/spec/test/R4ExamplesDriver.java
```diff
@@ -180,56 +177,60 @@ public void processExample(String file, Expectation expectation) throws ExampleP
...
-            // large index to process
-            lock.lock();
-            while (this.currentlySubmittedCount == this.maxInflight) {
-                try {
-                    this.inflightCondition.await(1000, TimeUnit.MILLISECONDS);
-                }
-                catch (InterruptedException x) {
-                    // NOP
-                }
+        if (pool == null) {
```
punktilious, line 180: why the switch? Now there are multiple exits from the method.

punktilious:
If you're going to change an exception variable from e to x you'll be changing it in a lot of places - I use x all the time.

fhir-model/src/test/java/com/ibm/fhir/model/spec/test/R4ExamplesDriver.java
lmsurpre: just a preference thing.  with the long if block i find it hard to remember the context when we get to that bottom else.  i can change it back to how it was or i could just handle this case first, remove the return, and put the rest into an else

fhir-model/src/test/java/com/ibm/fhir/model/spec/test/R4ExamplesDriver.java
lmsurpre: I wasn't sure why we weren't getting firstException set.
I realized that if processExample (line 268) threw an ExampleProcessorException then we'd never set it.
So I took the logic from the old lines 266-268 and moved them down here instead so that we apply them no matter where ExampleProcessorException is thrown from.

lmsurpre: It turned out not to be our issue, but I thought the change was good and so I left it.

punktilious (approved):
LGTM

lmsurpre:
> If you're going to change an exception variable from e to x you'll be changing it in a lot of places - I use x all the time.

That one wasn't really purposeful.  At first I changed the catch to catch "Error" and so I changed the x to and e since x didn't make sense to me for Error.  But it does make sense for Exception (although I do generally use 'e' for that).
Anyway, I changed it back to exception but never changed the variable name back to x.  Should be fine, but don't expect a bunch of PRs changing `x` to `e` :-)

PR #3840, opened 2022-08-03
remove TODO in FHIRPathType

lmsurpre:
I double-checked and the new data type (CodeableReference) and
resources are already listed.


punktilious (approved):
I checked too. LGTM.

Issue #2964, opened 2021-11-09
Implement instance-level $validate and enforce mode binding


lmsurpre:
**Is your feature request related to a problem? Please describe.**
In #2905 we added partial support for the `mode` input parameter on the `$validate` operation.
Specifically, when this parameter is set to `create` or `update`, we will validate the passed resource by using the corresponding fhir-server-config settings under `resources/<resourceType>/profiles`.

However, https://www.hl7.org/fhir/operation-resource-validate.html defines `mode` with a required binding to https://www.hl7.org/fhir/valueset-resource-validation-mode.html whereas our implementation allows any string that would be a valid code.

Specifically, in addition to `create` and `update` modes, it defines:
* a `profile` mode that it says should only apply to a resource "nominated by id, not provided as a parameter"
* a `delete` mode that only makes sense against a previously-ingested resource

But our implementation doesn't even support validation against existing resources in the db.
If you try invoking it at this level, it fails with a 400 Bad Request error that says:
> "Input parameter 'resource' is required for the $validate operation"

**Describe the solution you'd like**
1. Support (and advertise support for) resource-instance level $validate (at `[base]/[resourceType]/[id]/$validate`)
2. Ensure that only valid "mode" codes are accepted by the $validate operation... if an invalid code is sent then return a 400 Bad Request with a reasonable error
  * Ensure `delete` is only ever invoked at the resource-instance level (although I have no clue when we'd say that a given delete is "invalid"...I guess never?)
  * I don't understand why `profile` wouldn't be used when validating a passed resource against a particular profile.  I think this matches our current behavior if the 'profile' parameter is passed in and that seems fine to me.

**Describe alternatives you've considered**

**Acceptance Criteria**

1.  
   GIVEN [a precondition]
    AND [another precondition]
   WHEN [test step]
    AND [test step]
   THEN [verification step]
    AND [verification step]

**Additional context**

punktilious:
I found a couple of possible issues:
1. Minor. If a resource is delete, if you try to delete it again you get 200 OK with an OperationOutcome indicating that the resource is already deleted (the operation is idempotent). However, validate mode=delete will give you a 404 for a resource which is currently deleted. So that's a slight discrepancy.
2. The validate mode create/update/delete appear to ignore the allowed interactions configuration. For example, I think if we disable delete, then validate mode=delete should also say that the resource cannot be deleted.
[...]

PR #3841, opened 2022-08-03
issue #2964 - Support resource-instance level validate operation and …

PrasannaHegde1:
…enforce mode binding


fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
```diff
@@ -3400,4 +3400,14 @@ public PayloadPersistenceResponse storePayload(Resource resource, String logical
         // Delegate to the persistence layer. Result will be null if offloading is not supported
         return persistence.storePayload(resource, logicalId, newVersionNumber, resourcePayloadKey);
     }
+    
+    @Override
+    public List<Issue> validateDeleteResource(String type, String id) throws FHIROperationException {
```
lmsurpre, line 3405: I was expecting to see logic like this within ValidateOperation.java (rather than a new method on FHIRRestHelper).  Do you expect it to be used from anywhere else?

operation/fhir-operation-validate/src/main/java/com/ibm/fhir/operation/validate/ValidateOperation.java
```diff
@@ -46,14 +47,26 @@ protected Parameters doInvoke(FHIROperationContext operationContext, Class<? ext
             String versionId, Parameters parameters, FHIRResourceHelpers resourceHelper, SearchHelper searchHelper) throws FHIROperationException {
         try {
             Parameter resourceParameter = getParameter(parameters, "resource");
-            if (resourceParameter == null) {
+            String resourceTypeName = resourceType.getSimpleName();
+            Resource resource;
+            // @implNote if $validate is on resource-instance level, fetch the resource from database.
```
lmsurpre, line 52: no need to use "@implNote" here...thats more for use in javadocs 
it allows you to add some extra context to the javadoc without overriding the parent's javadoc.

operation/fhir-operation-validate/src/main/java/com/ibm/fhir/operation/validate/ValidateOperation.java
```diff
@@ -46,14 +47,26 @@ protected Parameters doInvoke(FHIROperationContext operationContext, Class<? ext
...
+                resource = resourceParameter.getResource();
             }
-            Resource resource = resourceParameter.getResource() ;
-
+            
             List<OperationOutcome.Issue> issues;
             Parameter profileParameter = getParameter(parameters, "profile");
             Parameter modeParameter = getParameter(parameters, "mode");
+            
+            // @implNote validation for only valid "mode" codes are accepted by the $validate operation
```
lmsurpre, line 68: same comment

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
lmsurpre: Its a little ugly (requires a cast) but it is possible to get the FHIRPersistence implementation straight from your operation impl by using something like this:
```
FHIRPersistence pl =
                    (FHIRPersistence) operationContext.getProperty(FHIROperationContext.PROPNAME_PERSISTENCE_IMPL);
```

operation/fhir-operation-validate/src/main/java/com/ibm/fhir/operation/validate/ValidateOperation.java
lmsurpre:
```suggestion
            // if $validate is on resource-instance level, fetch the resource from database.
```

operation/fhir-operation-validate/src/main/java/com/ibm/fhir/operation/validate/ValidateOperation.java
lmsurpre:
```suggestion
            // Ensure that the passed 'mode' parameter is valid.
```

operation/fhir-operation-validate/src/main/java/com/ibm/fhir/operation/validate/type/ModeType.java
```diff
@@ -0,0 +1,50 @@
+/*
+ * (C) Copyright IBM Corp. 2022
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.ibm.fhir.operation.validate.type;
+
+/**
+ * The Mode Type indicates a valid resource validation mode code.
+ * @see <a href="http://hl7.org/fhir/resource-validation-mode">resource validation mode</a>
```
lmsurpre, line 10: Somewhat minor, but because we implement a specific version of the FHIR specification, I generally use the version-specific URLs when linking to the spec from our code.
```suggestion
 * @see <a href="https://hl7.org/fhir/R4B/codesystem-resource-validation-mode.html">resource validation mode</a>
```

operation/fhir-operation-validate/src/main/java/com/ibm/fhir/operation/validate/type/ModeType.java
```diff
@@ -0,0 +1,50 @@
...
+        this.value = value;
+    }
+
+    /**
+     * method to validate a valid resource validation mode code.
+     *
+     * @param value 
+     *      A string that matches one of the allowed ModeType values
+     * @return 
+     *      The corresponding FHIRVersionParam or null if a null value was passed or if no matching resource validation mode code is found
```
lmsurpre, line 36:
```suggestion
     *      The corresponding ModeType or null if a null value was passed or if no matching resource validation mode code is found 
```

lmsurpre, line 36: In our other Enum classes that have a 'from' method, we throw an IllegalArgumentException if the input String does not match any of the enum values.

I'm not sure if that is the best choice or not, but I mention it because I do like consistency across the codebase.

fhir-server/src/main/java/com/ibm/fhir/server/util/FHIRRestHelper.java
PrasannaHegde1: this has been fixed. The logic has been moved to ValidateOperation.java as suggested.

operation/fhir-operation-validate/src/main/java/com/ibm/fhir/operation/validate/ValidateOperation.java
PrasannaHegde1: this has been fixed. removed "@implNote"

operation/fhir-operation-validate/src/main/java/com/ibm/fhir/operation/validate/ValidateOperation.java
PrasannaHegde1: this has been fixed. removed "@implNote"

operation/fhir-operation-validate/src/main/java/com/ibm/fhir/operation/validate/type/ModeType.java
PrasannaHegde1: This has been fixed by using the version-specific URL https://hl7.org/fhir/R4B/codesystem-resource-validation-mode.html

operation/fhir-operation-validate/src/main/java/com/ibm/fhir/operation/validate/type/ModeType.java
PrasannaHegde1: This has been fixed.

PrasannaHegde1: This has been fixed as suggested.  IllegalArgumentException is thrown if the input String is not null and does not match any of the allowed string values.

operation/fhir-operation-validate/src/test/java/com/ibm/fhir/operation/validate/ValidateOperationTest.java
```diff
@@ -268,4 +284,218 @@ public void testProfileParmProfileUnsupportedError() {
...
+        Parameters input = Parameters.builder()
+                .build();
+        
+        Issue expectedOutput = Issue.builder()
+                    .severity(IssueSeverity.INFORMATION)
+                    .code(IssueType.INFORMATIONAL)
+                    .details(CodeableConcept.builder()
+                        .text(string("All OK"))
+                        .build())
+                    .build();
```
lmsurpre, line 347: we're expecting the exception in this case, so maybe leave this out and end with the doInvoke call?

operation/fhir-operation-validate/src/test/resources/Patient.json
```diff
@@ -0,0 +1,174 @@
+{
+  "resourceType": "Patient",
```
lmsurpre, line 2: Is there anything special about this resource, or is it just a random Patient example?
Personally I prefer testing with minimal resources that have only the fields necessary for the specific test case.
I have a utility for creating those in the fhir-model tests jar.
Alternatively, if we don't want to introduce that dependency here, you could build a minimal Patient quite easilly (basically none of the fields are required for that type).

operation/fhir-operation-validate/src/test/java/com/ibm/fhir/operation/validate/ValidateOperationTest.java
```diff
@@ -268,4 +284,218 @@ public void testProfileParmProfileUnsupportedError() {
...
+     * Read a mock resource from the input JSON file.
+     *
+     * @param path the path to JSON file
+     * @return Resource - A FHIR Resource object representation
+     * @throws Exception
+     */
+    private static Resource getTestResource(String path) throws Exception {
+        try (InputStream is = ClassLoader.getSystemResourceAsStream(path)) {
+            return FHIRParser.parser(Format.JSON).parse(is);
+        }
```
lmsurpre, line 366: per my other comment, I think this could just be
```
return Patient.builder().id("test").build();
```

lmsurpre: or, if you take a dependency on fhir-model's test-jar (which unfortunately also means depending on org.skyscreamer:jsonassert at the moment) you could be
```
return TestUtil.getMinimalResource(Patience.class);
```


here is the dependency snippet for the pom.xml in case you chose to go this route:
```xml
        <dependency>
            <groupId>${project.groupId}</groupId>
            <artifactId>fhir-model</artifactId>
            <version>${project.version}</version>
            <type>test-jar</type>
        </dependency>
        <dependency>
            <groupId>org.skyscreamer</groupId>
            <scope>test</scope>
        </dependency>
```

operation/fhir-operation-validate/src/test/java/com/ibm/fhir/operation/validate/ValidateOperationTest.java
PrasannaHegde1: this has been fixed. Cleaned up unwanted code.

operation/fhir-operation-validate/src/test/resources/Patient.json
PrasannaHegde1: This has been fixed. Removed Patient.json and built up Patient resource using Patient.builder() with minimal fields as suggested.

operation/fhir-operation-validate/src/test/java/com/ibm/fhir/operation/validate/ValidateOperationTest.java
PrasannaHegde1: This has been fixed. Removed Patient.json and built up Patient resource using Patient.builder() with minimal fields as suggested in the previous review comment.

lmsurpre (approved):
LGTM

PR #3858, opened 2022-08-08
fix package name for TaggingInterceptor

lmsurpre:
and update it to add both "before" and "after" tags (used to manually
verify #3787

also some minor formatting (mostly autosquashing of trailing whitespace)
in FHIRPersistenceJDBCImpl


fhir-server-test/src/main/java/org/linuxforhealth/fhir/server/test/TaggingInterceptor.java
```diff
@@ -0,0 +1,113 @@
...
+import org.linuxforhealth.fhir.model.type.Code;
+import org.linuxforhealth.fhir.model.type.Coding;
+import org.linuxforhealth.fhir.model.type.Meta;
+import org.linuxforhealth.fhir.model.type.Uri;
+import org.linuxforhealth.fhir.persistence.context.FHIRPersistenceEvent;
+import org.linuxforhealth.fhir.server.spi.interceptor.FHIRPersistenceInterceptor;
+import org.linuxforhealth.fhir.server.spi.interceptor.FHIRPersistenceInterceptorException;
+
+/**
+ * A sample persistence interceptor that adds a tag to each resource before it gets persisted.
```
punktilious, line 22: Old description? Also adds a tag after reading the resource.

lmsurpre:
```suggestion
 * A sample persistence interceptor that adds a "before" tag to each resource before it gets persisted and an "after" tag when it is read.
```

punktilious (approved):
LGTM

PR #3863, opened 2022-08-09
issue #3420 - set persistence factoryClassname to the blob one

lmsurpre:
To actually test payload offloading in the CI.


punktilious (approved):
LGTM

Issue #3866, opened 2022-08-10
Azure blob offload cannot connect to older instance of azurite


punktilious:
**Describe the bug**
Running an older instance of azurite, after updating the Azure blob client libraries we can no longer connect:

```
Caused by: java.util.concurrent.ExecutionException: com.azure.storage.blob.models.BlobStorageException: Status code 400, "<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<Error>
  <Code>InvalidHeaderValue</Code>
  <Message>The value for one of the HTTP headers is not in the correct format.
RequestId:9ac1dcdd-f371-47b1-bad5-d04060811be4
Time:2022-08-09T23:06:16.790Z</Message>
  <HeaderName>x-ms-version</HeaderName>
  <HeaderValue>2021-06-08</HeaderValue>
</Error>"
```

Need to provide a way to specify the service version when constructing the Azure Blob client.

**Environment**
Which version of IBM FHIR Server? 5.0.0-SNAPSHOT

**To Reproduce**
Run older azurite container with newer client.

**Expected behavior**
The FHIR server should be able to connect.

**Additional context**
Mostly an issue with bit rot in local test environments. Updating the container would probably address this, but we should also allow configuration to control the service version used by the client, as offered by the blob client API.

punktilious:
Also found that the payload offload parse call was ignoring the elements variable so no subsetting was supported, causing integration test failures with the Azure e2e updated added to the CI pipeline.

lmsurpre:
I updated the demo env locally to use an old version of azurite (3.10.0).  After this the fhir-server image was crashing during startup.  I then updated the config to specify the specific API version supported by azurite 3.10.0 (`2020-04-08`) and it started working again.

I then proceeded to create a Patient resource (with payload still offloading enabled) and then read it back out with an `_elements=identifier` filter.  This is working as expected.

PR #3867, opened 2022-08-10
issue #3866 parse offload using elements to subset

punktilious:

### DIFF
--- a/fhir-persistence-blob/src/main/java/org/linuxforhealth/fhir/persistence/blob/BlobDeletePayload.java
+++ b/fhir-persistence-blob/src/main/java/org/linuxforhealth/fhir/persistence/blob/BlobDeletePayload.java
@@ -9,17 +9,19 @@ package org.linuxforhealth.fhir.persistence.blob;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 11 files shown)


fhir-persistence-blob/src/main/java/org/linuxforhealth/fhir/persistence/blob/BlobDeletePayload.java
```diff
@@ -9,17 +9,19 @@
...
+import com.azure.storage.blob.models.BlobItem;
 import com.azure.storage.blob.models.DeleteSnapshotsOptionType;
-import org.linuxforhealth.fhir.persistence.exception.FHIRPersistenceException;
 
 /**
  * DAO command to delete the configured Azure blob (holding a FHIR payload object)
  */
 public class BlobDeletePayload {
-    private static final Logger logger = Logger.getLogger(BlobReadPayload.class.getName());
+    private static final Logger logger = Logger.getLogger(BlobDeletePayload.class.getName());
```
lmsurpre, line 24: good catch

lmsurpre (approved):
Its still marked as draft, but I walked through the edits and it seems fine to me.

Issue #3874, opened 2022-08-11
Bulk export fails when payload offloading is configured


lmsurpre:
**Describe the bug**
```
[8/11/22, 17:55:29:709 UTC] 00000054 FHIRPersisten E   Unexpected error while processing token value records.
                                 java.lang.NullPointerException
	at java.base/java.util.zip.InflaterInputStream.<init>(Unknown Source)
	at java.base/java.util.zip.GZIPInputStream.<init>(Unknown Source)
	at java.base/java.util.zip.GZIPInputStream.<init>(Unknown Source)
	at org.linuxforhealth.fhir.persistence.jdbc.dao.impl.FetchResourcePayloadsDAO.run(FetchResourcePayloadsDAO.java:132)
	at org.linuxforhealth.fhir.persistence.jdbc.impl.FHIRPersistenceJDBCImpl.fetchResourcePayloads(FHIRPersistenceJDBCImpl.java:3168)
	at org.linuxforhealth.fhir.bulkdata.jbatch.export.fast.ResourcePayloadReader.readItem(ResourcePayloadReader.java:316)
	at com.ibm.jbatch.container.artifact.proxy.ItemReaderProxy.readItem(ItemReaderProxy.java:77)
	at com.ibm.jbatch.container.controller.impl.ChunkStepControllerImpl.readItem(ChunkStepControllerImpl.java:329)
	at com.ibm.jbatch.container.controller.impl.ChunkStepControllerImpl.readAndProcess(ChunkStepControllerImpl.java:245)
	at com.ibm.jbatch.container.controller.impl.ChunkStepControllerImpl.invokeChunk(ChunkStepControllerImpl.java:626)
	at com.ibm.jbatch.container.controller.impl.ChunkStepControllerImpl.invokeCoreStep(ChunkStepControllerImpl.java:795)
	at com.ibm.jbatch.container.controller.impl.BaseStepControllerImpl.execute(BaseStepControllerImpl.java:295)
	at com.ibm.jbatch.container.controller.impl.ExecutionTransitioner.doExecutionLoop(ExecutionTransitioner.java:118)
	at com.ibm.jbatch.container.controller.impl.WorkUnitThreadControllerImpl.executeCoreTransitionLoop(WorkUnitThreadControllerImpl.java:96)
	at com.ibm.jbatch.container.controller.impl.WorkUnitThreadControllerImpl.executeWorkUnit(WorkUnitThreadControllerImpl.java:178)
[...]

PR #3875, opened 2022-08-11
issue #3874 - use legacy export when payload offloading is configured

lmsurpre:
A cleaner approach would be to add a method to FHIRPersistence so that
implementations can indicate whether they support fetchResourcePayloads
which is what we use for the fast export. Then we could pass that into
BulkDataClient.submitExport where we decide which export implementation
to use.

However, ideally we'd just add this support...then we probably wouldn't
need those changes.

Therefor, for now, I opted for the quick and dirty fix.

### DIFF
--- a/fhir-config/src/main/java/org/linuxforhealth/fhir/config/FHIRConfigHelper.java
+++ b/fhir-config/src/main/java/org/linuxforhealth/fhir/config/FHIRConfigHelper.java
@@ -104,8 +104,8 @@ public class FHIRConfigHelper {
     }
 
     /**
-     * This generic function will perform the work of retrieving a property from either the tenant-specific config, or
-     * the default config, and then converting the resulting value to the appropriate type.
+     * This generic function will perform the work of retrieving a property from either the tenant-specific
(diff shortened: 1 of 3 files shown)


fhir-config/src/main/java/org/linuxforhealth/fhir/config/FHIRConfigHelper.java
```diff
@@ -104,8 +104,8 @@ private static JsonValue getPropertyFromTenantOrDefault(String propertyName) {
     }
 
     /**
-     * This generic function will perform the work of retrieving a property from either the tenant-specific config, or
-     * the default config, and then converting the resulting value to the appropriate type.
+     * This generic function will perform the work of retrieving a property from either the tenant-specific
+     * config or the default config, and then converting the resulting value to the appropriate type.
```
lmsurpre, line 108: just some style changes to reduce line lengths in this file...not related to rest of PR

punktilious (approved):
LGTM. Glad it was a simple fix.

Issue #3256, opened 2022-01-27
[fhir-client] Document/support setting the fhirVersion mime-type parameter on requests


lmsurpre:
**Is your feature request related to a problem? Please describe.**
From https://hl7.org/fhir/http.html#version-parameter
> This parameter can be used anywhere that the FHIR Mime type is used. When used in an HTTP request, the version parameter may be used on either the Content-Type header, or the Accept header, or both, and applies to the entire interaction

For R4B, via https://github.com/IBM/FHIR/issues/3192 and https://github.com/IBM/FHIR/issues/3242, we're gonna let this parameter drive server behavior, so we definitely need a way to set it from the client.

**Describe the solution you'd like**
1. if not supported already, add a config property to set the default value across all requests
2. document how to set it

**Describe alternatives you've considered**
Make it possible to set on a per-request basis

**Acceptance Criteria**

1.  
   GIVEN a client with `fhirclient.default.mimetype = application/fhir+json; fhirVersion=4.0`
   WHEN the request is made to the server
   THEN any applicable Accept or Content-Type header values have the fhirVersion mime type parameter set to 4.0

1.  
   GIVEN a client with `fhirclient.default.mimetype = application/fhir+json; fhirVersion=4.3` (the default) 
   WHEN the request is made to the server
   THEN any applicable Accept or Content-Type header values have the fhirVersion mime type parameter set to 4.3

**Additional context**

lmsurpre:
It turns out we already support setting the default FHIR version from fhir-client.
I'm not sure how commonly we'd need to set the fhirVersion on a per-request basis, and so I'm updating the description and acceptance criteria to remove that part of it.

If needed, that can be a separate issue.

punktilious:
Confirmed that when setting 
```
fhirclient.default.mimetype = application/fhir+json; fhirVersion=4.3
```

On the server side we see:
```
[25/08/2022, 11:31:33:934 UTC] 0000002a FHIRHttpServl 3   getHeaders("Content-Type") : [application/fhir+json;fhirversion=4.3]
[25/08/2022, 11:31:33:934 UTC] 0000002a FHIRHttpServl 3   getHeaders("Accept") : [application/fhir+json; fhirVersion=4.3]
```

and when setting
```
fhirclient.default.mimetype = application/fhir+json; fhirVersion=4.0
```

```
[...]

PR #3876, opened 2022-08-12
issue #3256 added unit tests to support QA of FHIRClient default mime…

punktilious:
…type

The two new unit tests check:
1. Using the default FHIRClient configuration, interactions related to an R4B resource type (Ingredient) are successful.
2. When overriding the mime type property to remove fhirVersion, interactions related to an R4B resource type are unsuccessful.
 
### DIFF
--- a/fhir-server-test/src/test/java/org/linuxforhealth/fhir/client/test/FHIRClientTest.java
+++ b/fhir-server-test/src/test/java/org/linuxforhealth/fhir/client/test/FHIRClientTest.java
@@ -13,12 +13,13 @@ import static org.testng.AssertJUnit.assertEquals;
 import static org.testng.AssertJUnit.assertNotNull;
 import static org.testng.AssertJUnit.assertTrue;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Properties;
+import java.util.UUID;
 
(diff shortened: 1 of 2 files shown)


fhir-server-test/src/test/java/org/linuxforhealth/fhir/client/test/FHIRClientTest.java
```diff
@@ -79,6 +84,67 @@ public void testFHIRClientCtorProperties2() throws Exception {
         assertNotNull(c);
         assertEquals(MIMETYPE_XML, c.getDefaultMimeType());
     }
+    
+    @Test
+    public void testFHIRClientMimeType() throws Exception {
+        // The default mimetype should specify FHIR version 4.3, so let's make
+        // sure that with the default settings we can create a 4.3-only resource
+        Ingredient ingredient = TestUtil.readLocalResource("Ingredient-1.json");
```
lmsurpre, line 92: we could use `TestUtil.getMinimalResource(Ingredient.class)` here instead and avoid the need for checking in this extra resource under src/test

punktilious: Fair point. But will skip this this time.

Issue #3859, opened 2022-08-09
When creating a new schema for R4B, the update schema still warns of tables that should not be used. 


d0roppe:
**Is your feature request related to a problem? Please describe.**
We should not create the tables that we don't want used when creating a new schema with R4B code. 

**Describe the solution you'd like**
when we update a new created schema for R4B we should not create tables and then warn not to use them. 
Here is an example of some of the messages in question.
```
2022-08-09 07:36:52.854 00000001 WARNING forhealth.fhir.schema.app.Main EffectEvidenceSynthesis tables [EffectEvidenceSynthesis_DATE_VALUES, EffectEvidenceSynthesis_LATLNG_VALUES, EffectEvidenceSynthesis_LOGICAL_RESOURCES, EffectEvidenceSynthesis_NUMBER_VALUES, EffectEvidenceSynthesis_QUANTITY_VALUES, EffectEvidenceSynthesis_RESOURCE_TOKEN_REFS, EffectEvidenceSynthesis_RESOURCES, EffectEvidenceSynthesis_STR_VALUES] will be dropped in a future release. No data should be written to these tables. If any data exists in these tables, that data should be exported (if desired) and deleted from these tables.
2022-08-09 07:36:52.854 00000001 WARNING forhealth.fhir.schema.app.Main MedicinalProduct tables [MedicinalProduct_DATE_VALUES, MedicinalProduct_LATLNG_VALUES, MedicinalProduct_LOGICAL_RESOURCES, MedicinalProduct_NUMBER_VALUES, MedicinalProduct_QUANTITY_VALUES, MedicinalProduct_RESOURCE_TOKEN_REFS, MedicinalProduct_RESOURCES, MedicinalProduct_STR_VALUES] will be dropped in a future release. No data should be written to these tables. If any data exists in these tables, that data should be exported (if desired) and deleted from these tables.
[...]

PR #3884, opened 2022-08-15
issue #3859 - only create tables for R4B resource types by default

lmsurpre:
And only warn about the deprecated (R4 resource type) tables when we're
migrating from an older schema version that has these.

I also renamed ResourceTypeHelper to ResourceTypeUtil to better reflect
project naming conventions (helpers can be instantiated, utils are
static).


PrasannaHegde1 (approved):
LGTM

Issue #3885, opened 2022-08-15
Update release process to post to the new sonatype OSS hosting


lmsurpre:
**Is your feature request related to a problem? Please describe.**
I opened the request for the "org.linuxforhealth" group id at https://issues.sonatype.org/browse/OSSRH-83433
and that should be all set now:

> Congratulations! Welcome to the Central Repository!
> org.linuxforhealth has been prepared, now user(s) lmsurprenant can:
> Publish snapshot and release artifacts to s01.oss.sonatype.org

Now its time to update our release automation.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Acceptance Criteria**

1.  
   GIVEN a release candidate
   WHEN you push a release tag (e.g. 5.0.0)
   THEN if the tagged version passes all of the tests, the java artifacts should be uploaded to s01.oss.sonatype.org
    AND can be published to maven central from there

**Additional context**
Should be done together with the updates to the linux container image release under #3845

punktilious:
@lmsurpre how do you propose we QA this one?

lmsurpre:
I confirm that this is working now.

However, there's one step in particular thats been flaky:  uploading the fhir-install zip to OSSRH

For RC1, it took 5 tries.
For RC2, it took 2 tries.

I opened https://issues.sonatype.org/browse/OSSRH-83767 for it, but if it continues to happen we might want to either:
A. add special retry logic to our release scripts so that we don't need to "start over" each time it fails to upload; or
B. stop putting this particular artifact in central (we'd still have it in GitHub Releases tab, just not maven central).

Notably, this zip installer has gotten quite large (almost 500 MB!).  Most of that is the liberty zip that we embed in it.
We definitely have options for slimming that down and I think that would make the upload to OSSRH much more reliable.

1. The biggest "bang-for-the-buck" option would be to remove the embedded liberty zip and instead have the install script download that at install time.  However, that makes it less convenient because you need internet access to install.

2. Another option I've wanted to do for a while is to split our bulkdata stuff

3. Another idea I had is to use the maven-liberty-plugin to package our webapp:  https://github.com/OpenLiberty/ci.maven/blob/main/docs/package.md#package
   Today we basically do that work ourselves.  Hopefully wouldn't be too hard to use this more official mechanism and then, hopefully, we'd only be including the bits that are needed at runtime (vs the liberty-runtime zip which probably includes more).

lmsurpre:
```$ du -sh *
[...]

Issue #3845, opened 2022-08-05
Release docker images to github packages (ghcr.io)


lmsurpre:
**Is your feature request related to a problem? Please describe.**
For IBM FHIR Server, we pushed our docker images to DockerHub under the organization `ibmcom`.

The LinuxForHealth project has its own organization on DockerHub at https://hub.docker.com/u/linuxforhealth.
However, DockerHub has some rather significant limitations for us and alternative image hosting is readily available.

I discussed it with @dixonwhitmire and we decided to use the GitHub Container Registry for our images going forward.
This will mean a longer image path (`ghcr.io/linuxforhealth/fhir-server` instead of just `linuxforhealth/fhir-server`) but we think its still reasonable.

**Describe the solution you'd like**
Update our automation

Question:  Will github packages let us update an image thats already been pushed.  We do that today to ensure that the base OS stays up-to-date with security fixes.  I assume it will allow that since its pretty common practice.

**Describe alternatives you've considered**
Continue pushing to DockerHub (under linuxforhealth)

**Acceptance Criteria**

1.  
   GIVEN [a precondition]
    AND [another precondition]
   WHEN [test step]
    AND [test step]
   THEN [verification step]
    AND [verification step]

**Additional context**
At least two workflows need updating:
* release.yml
* rebuild.yml

punktilious:
Confirmed that the docker image exists:

```
REPOSITORY                                   TAG                                      IMAGE ID      CREATED        SIZE
ghcr.io/linuxforhealth/fhir-server           latest                                   415b2d5018c8  11 days ago    1.26 GB
```
Pulled it, ran it with bootstrapping enabled and was able to POST a new resource.

PR #3886, opened 2022-08-15
issues #3845 and #3885 - update release process for LinuxForHealth

lmsurpre:
1. maven artifacts will now be staged at s01.oss.sonatype.org with a groupId of org.linuxforhealth.fhir
2. docker images will now be published to GitHub Packages under the LinuxForHealth organization (ghcr.io/LinuxForHealth)

Note: for the "rebuild" workflow, it checks out the previously tagged version of both the code AND the release automation, so 4.11.0 and 4.11.1 will continue to go to ibmcom/ibm-fhir-server on docker hub


.github/workflows/integration.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         mvn -B org.apache.maven.plugins:maven-dependency-plugin:3.1.2:go-offline -f fhir-parent -DexcludeReactor=true -Dmaven.wagon.http.retryHandler.count=3
         mvn -B org.apache.maven.plugins:maven-dependency-plugin:3.1.2:resolve-plugins -f fhir-parent -DexcludeReactor=true -Dmaven.wagon.http.retryHandler.count=3
         mvn -B install --file fhir-parent -DskipTests -P include-fhir-igs,integration --no-transfer-progress -Dmaven.wagon.http.retryHandler.count=3
-        mvn -B dockerfile:build -f fhir-install --no-transfer-progress -Dmaven.wagon.http.retryHandler.count=3 -Ddocker.buildArg.VERBOSE=false
+        docker build fhir-install --build-arg VERBOSE=false -t linuxforhealth/fhir-server
```
dixonwhitmire, line 34: Is the docker build process migrating from maven to the docker CLI? I didn't know if there were relevant portions of the POM(s) which could be removed. I did see this [snippet](https://github.com/LinuxForHealth/FHIR/pull/3886/files#diff-0b51d62d390dc0b76f0597ee8788fbb662104fbad227ba8aa5bd43d494fbe3f0L134) but of course I'm not familiar with the overall context, pun intended :)

.github/workflows/release.yml
```diff
@@ -29,14 +29,18 @@ jobs:
         distribution: 'temurin'
         java-version: ${{ matrix.java }}
         cache: 'maven'
+    - name: Log in to GitHub Container Registry
```
dixonwhitmire, line 32: These configurations align with my current understanding of the release workflow. We will also need to ensure we have these credentials backed up in a "secondary" vault, which we can chat about offline.

.github/workflows/integration.yml
lmsurpre: It's a step in that direction.  Today we have pretty even a mix of `docker build` and `mvn dockerfile:build` in our various workflows.  After this round of changes, we'll use the `docker build` flavor almost exclusively and we'll build to a local image named `linuxforhealth/fhir-server`.

The one place I left the maven-based docker build is in our release workflow.  In that one case we build to `ghcr.io/linuxforhealth/fhir-server`.  See https://github.com/LinuxForHealth/FHIR/blob/c28cb62b3b3188dad33f3086033c8bf5fa279048/build/release/bin/30_release/1_docker.sh#L29-L32

But we don't really *need* to use the dockerfile-maven-plugin there either.  Do you think we should change to use the docker cli there as well?

Additionally we could use the `ghcr.io/linuxforhealth/fhir-server` flavor all over...I just felt that keeping the registry prefix out of the name in the other workflows might help make it more clear that its built locally (and won't fall back to pulling from the registry if we ever fail to build it).

.github/workflows/release.yml
lmsurpre: OK.  I tried to follow your pattern here and I think you had just followed the recommendation of the GitHub docs.  The alternative to using the injected GITHUB_TOKEN is to use a Personal Access Token (PAT).
I'm not sure that either needs to be stored outside of GitHub but I'm willing to hear the argument.

P.S. a related topic for us to chat about is whether we should set "restrictive" permissions for our GITHUB_TOKEN at the org level.  https://docs.github.com/en/actions/security-guides/automatic-token-authentication

dixonwhitmire (approved):

PrasannaHegde1 (approved):
LGTM

Issue #3888, opened 2022-08-16
Integration tests are flaky on windows


lmsurpre:
**Describe the bug**
I've noticed our windows workflow failing at about 5 minutes on a few occasions now.
It seemed to start on https://github.com/LinuxForHealth/FHIR/pull/3887 but I don't see how that could be related.

**Environment**
main (CI only)

**To Reproduce**
1. run the CI a lot

**Expected behavior**
tests should reliably pass or fail

**Additional context**
GraphTermServiceProviderTimeLimitTest sets a time limit of 1 millisecond for cassandra, then performs a query and expects that query to time out.  Its reliable on my machine, but its occasionally failing in the CI for some reason.

JDBCSortTest is failing because the windows-19 environment was assigning the same timestamp (microsecond granularity) to 5 of the 6 resources that get created for this test.  It never used to do that.

JDBCWholeSystemSearchTest is failing for a similar reason...one of the test cases searched by lastUpdated time and was expecting a single match (but was getting 2).

lmsurpre:
The windows-2019 behavior is still pretty odd, but I've made changes to GraphTermServiceProviderTimeLimitTest and JDBCWholeSystemSearchTest to make them more robust.

For GraphTermServiceProviderTimeLimitTest, I just disabled it.

punktilious:
For Windows, it's possible the JVM isn't using the hi-resolution (aka multimedia) clock...in which case it probably runs at 100 ticks/second (10ms ticks). It could be a limitation of the virtualization environment.

lmsurpre:
Integration tests have been more stable on windows.  Considering this one done.

PR #3890, opened 2022-08-16
issue #3888 - work around timestamp-related issues in test environment

lmsurpre:
There must have been some environmental change to the github actions windows-2019 environment because today a couple tests started failing due to multiple resources being assigned the same timestamp (`Instant.now()` truncated to nearest microsecond).

JDBCSortTest creates 6 resources in a row and in this environment (and only this environment) a number of them were assigned the same lastUpdated time.  To address the issue, I added a 10ms pause between the creates.  I also kept some of the logging statements that I added to AbstractSortTest in case we need to further debug.

JDBCWholeSystemSearchTest was failing for a similar reason...one of the test cases searched by lastUpdated time and was expecting a single match (but was getting 2). I fixed it by checking the lastUpdated time in each response instead of just checking the count.

I'm not sure whether/how it is related but GraphTermServiceProviderTimeLimitTest began failing intermittently as well.
For this one, I'm not sure of the root cause...all I did is assign the result of the call to a variable and add an assert to hopefully get a better error message. We'll need to monitor it.  If it continues failing intermittently, we should probably just skip it (its already marked as "experimental" and its not worth chasing down the root cause at this time).
    
### DIFF
--- a/fhir-persistence/src/test/java/org/linuxforhealth/fhir/persistence/test/common/AbstractSortTest.java
+++ b/fhir-persistence/src/test/java/org/linuxforhealth/fhir/persistence/test/common/AbstractSortTest.java
@@ -51,7 +51,7 @@ import org.linuxforhealth.fhir.search.exception.FHIRSearchException;
  *
  */
 public abstract class AbstractSortTest extends AbstractPersistenceTest {
-    boolean DEBUG = true;
+    boolean DEBUG = false;
 
     Basic resource1a;
     Basic resource2a;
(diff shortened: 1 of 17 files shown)


fhir-persistence/src/test/java/org/linuxforhealth/fhir/persistence/test/common/AbstractSortTest.java
```diff
@@ -49,6 +51,8 @@
  *
  */
 public abstract class AbstractSortTest extends AbstractPersistenceTest {
+    boolean DEBUG = true;
```
lmsurpre, line 54:
```suggestion
    boolean DEBUG = false;
```

PrasannaHegde1 (approved):
LGTM

PR #3892, opened 2022-08-17
use GITHUB_TOKEN and set permissions

lmsurpre:
I also added `-Dsurefire.failIfNoSpecifiedTests=false` to the release script to avoid an issue I saw where the lack of tests in one module caused the entire release process to fail.

### DIFF
--- a/.github/workflows/javadocs.yml
+++ b/.github/workflows/javadocs.yml
@@ -4,6 +4,12 @@ on:
   push:
     tags:
         - '*'
+  # allows the workflow to be manually executed any time
+  workflow_dispatch:
+
+permissions:
+  pages: write
+  contents: write
 
 jobs:
   build:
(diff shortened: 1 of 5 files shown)


.github/workflows/javadocs.yml
```diff
@@ -53,9 +59,7 @@ jobs:
             cp -R fhir-parent/target/site/apidocs ../apidocs/
     - name: Commit and Add GH-Pages
       env:
-            GITHUB_TOKEN: ${{ secrets.DOCS_SITE_TOKEN }}
-            GITHUB_REPOSITORY: ${{ secrets.GITHUB_REPOSITORY }}
-            GITHUB_ACTOR: ${{ secrets.GITHUB_ACTOR }}
```
lmsurpre, line 58: according to their docs, github actions already set these environment variables (no need to access like a secret).  we should be able to confirm that as soon as this one gets merged :-)

dixonwhitmire (approved):
lgtm

PR #3896, opened 2022-08-18
Remove redundant step from javadocs workflow

lmsurpre:
It looks like we no longer need to trigger a build after pushing the change to gh-pages


PrasannaHegde1 (approved):
LGTM

PR #3900, opened 2022-08-19
make dockerhub rebuild workflow compatible with previously tagged scripts

lmsurpre:
GitHub executes the "current" rebuild.yml workflow, but I forgot that this workflow
checks out an older version of the repo and thus the release scripts that are used
are actually from that older version and the dockerhub one was named `1_dockerhub.sh`
and required the DOCKERHUB_USERNAME and DOCKERHUB_TOKEN env vars to be populated.

### DIFF
--- a/.github/workflows/rebuild.yml
+++ b/.github/workflows/rebuild.yml
@@ -70,11 +70,6 @@ jobs:
         distribution: 'temurin'
         java-version: ${{ matrix.java }}
         cache: 'maven'
-    - name: Log in to Docker Hub
-      uses: docker/login-action@v2
-      with:
-        username: ${{ secrets.DOCKERHUB_USERNAME }}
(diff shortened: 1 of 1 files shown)


dixonwhitmire (approved):

.github/workflows/rebuild.yml
```diff
@@ -91,6 +88,5 @@ jobs:
         bash build/release/20_test.sh
         # go straight to the dockerhub release instead of calling 30_release.sh
         # which uploads the build artifacts to Sonatype OSSRH in the tag release workflow
-        docker login -u "${DOCKERHUB_USERNAME}" -p "${DOCKERHUB_TOKEN}"
         bash build/release/bin/30_release/1_docker.sh
```
lmsurpre, line 91:
```suggestion
        bash build/release/bin/30_release/1_dockerhub.sh
```

PR #3901, opened 2022-08-21
dependency version bumps

lmsurpre:
1. lock in to specific versions for a few transitive dependencies that were flagged by sonatype
   * manage commons-configuration2 to 2.8.0
   * manage netty-handler to 4.1.79.Final (we already managed other netty modules but com.datastax.oss:java-driver-core specifically depends on the handler)
   * manage icu4j to 71.1

2. group the apache commons dependencies together in fhir-parent

3. add a version for the nexus-staging-maven-plugin plugin

4. update the fhir-examples plugin versions to match the plugin versions from fhir-parent


punktilious (approved):
LGTM

PR #3902, opened 2022-08-22
add ghcr.io prefix to build scripts for the other container images

lmsurpre:
In 5.0.0-RC1, only the linuxforhealth/fhir-server image was successfully pushed to
the GitHub container registry. I'm hoping that, with these changes, when we tag
RC2 it will push all of the desired images to the registry.

### DIFF
--- a/fhir-install/src/main/docker/fhir-bucket-tool/build.sh
+++ b/fhir-install/src/main/docker/fhir-bucket-tool/build.sh
@@ -27,13 +27,13 @@ mkdir -p target/
 cp ${WORKSPACE}/fhir-bucket/target/fhir-bucket-*cli.jar target/
 cp ${WORKSPACE}/LICENSE target/
 
-docker build --build-arg FHIR_VERSION=${BUILD_ID} -t fhir-bucket-tool:latest .
-DOCKER_IMAGE=$(docker images --filter=reference='fhir-bucket-tool:latest' --format "{{.ID}}")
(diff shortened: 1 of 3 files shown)


fhir-install/src/main/docker/fhir-bucket-tool/build.sh
```diff
@@ -27,13 +27,13 @@ mkdir -p target/
 cp ${WORKSPACE}/fhir-bucket/target/fhir-bucket-*cli.jar target/
 cp ${WORKSPACE}/LICENSE target/
 
-docker build --build-arg FHIR_VERSION=${BUILD_ID} -t fhir-bucket-tool:latest .
-DOCKER_IMAGE=$(docker images --filter=reference='fhir-bucket-tool:latest' --format "{{.ID}}")
+docker build --build-arg FHIR_VERSION=${BUILD_ID} -t linuxforhealth/fhir-bucket-tool:latest .
```
punktilious, line 30: Minor. Perhaps a comment or two would be useful to indicate why the tag names are slightly different for example:
```
linuxforhealth/fhir-bucket-tool:latest
vs
ghcr.io/linuxforhealth/fhir-bucket-tool:latest
```

lmsurpre: to be honest i think its a needless difference... there's no actual need to build them with one name and then tag them with the ghcr.io prefix.

i do follow a similar workflow locally, but we could probably save a couple lines here if we want.  but i'll save it for another time.

PR #3905, opened 2022-08-22
Update ProfileSupport.java

berkant-k:
Default cardinality of StructureDefinition.Version is 0..1. 
So it can be empty and this code could throw NullPointerException.
We need to check the result of getVersion methot not null.

Signed-off-by: berkant-k <33718658+berkant-k@users.noreply.github.com>

### DIFF
--- a/fhir-profile/src/main/java/org/linuxforhealth/fhir/profile/ProfileSupport.java
+++ b/fhir-profile/src/main/java/org/linuxforhealth/fhir/profile/ProfileSupport.java
@@ -280,8 +280,15 @@ public final class ProfileSupport {
 
     private static List<Constraint> getConstraints(StructureDefinition profile, Class<?> type) {
         String url = profile.getUrl().getValue();
-        String version = profile.getVersion().getValue();
-        CacheKey key = key(url + "|" + version);
+        CacheKey key;
+        if(profile.getVersion()!=null) {
+            key = key(url + "|" + profile.getVersion().getValue());
+        }
+        else
+        {
+            key = key(url);
+        }
+        
         Map<CacheKey, List<Constraint>> constraintCache = CacheManager.getCacheAsMap(CONSTRAINT_CACHE_NAME, CONSTRAINT_CACHE_CONFIG);
 
         try {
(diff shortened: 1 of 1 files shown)


fhir-profile/src/main/java/org/linuxforhealth/fhir/profile/ProfileSupport.java
```diff
@@ -280,8 +280,15 @@ public static List<Constraint> getConstraints(String url, Class<?> type) {
 
     private static List<Constraint> getConstraints(StructureDefinition profile, Class<?> type) {
         String url = profile.getUrl().getValue();
-        String version = profile.getVersion().getValue();
-        CacheKey key = key(url + "|" + version);
+        CacheKey key;
```
punktilious, line 283: My suggestion. Use final to ensure key is assigned once and once only. Also some updates to match the style for this repo:
```
        final CacheKey key;
        if (profile.getVersion() != null) {
            key = key(url + "|" + profile.getVersion().getValue());
        } else {
            key = key(url);
        }

```

fhir-profile/src/main/java/org/linuxforhealth/fhir/profile/ProfileSupport.java
```diff
@@ -280,8 +280,15 @@ public static List<Constraint> getConstraints(String url, Class<?> type) {
...
+        if(profile.getVersion()!=null) {
+            key = key(url + "|" + profile.getVersion().getValue());
+        }
+        else
+        {
+            key = key(url);
+        }
```
lmsurpre, line 290: turning robin's comment into a github suggestion:
```suggestion
        final CacheKey key;
        if(profile.getVersion()!=null) {
            key = key(url + "|" + profile.getVersion().getValue());
        } else {
            key = key(url);
        }
```

fhir-profile/src/main/java/org/linuxforhealth/fhir/profile/ProfileSupport.java
lmsurpre: I turned this into a github suggestion just below.  With this change, I'm happy to approve.  I'll apply it later if we don't here back.  Thanks @berkant-k !

fhir-profile/src/main/java/org/linuxforhealth/fhir/profile/ProfileSupport.java
lmsurpre: I also added the profile.getVersion().hasValue() check which covers an extremely rare edge case where the StructureDefinition has a version, but that version has an extension but no value

lmsurpre (approved):
LGTM

Issue #3911, opened 2022-08-23
Bulk operation transaction fails due to new MetricHandle imbalance


punktilious:
**Describe the bug**

[8/22/22, 21:45:29:291 UTC] 00000084 RegisteredSyn E   WTRN0074E: Exception caught from after_completion synchronization operation: java.lang.IllegalStateException: Metric being closed is '/M_JDBC_COMMIT/' but currentMetric is '/M_JDBC_COMMIT/M_JDBC_COMMIT/M_JDBC_COMMIT/'
        at org.linuxforhealth.fhir.config.FHIRRequestContext.endMetric(FHIRRequestContext.java:364)
        at org.linuxforhealth.fhir.config.MetricHandle.close(MetricHandle.java:82)
        at org.linuxforhealth.fhir.persistence.jdbc.impl.CacheTransactionSync.afterCompletion(CacheTransactionSync.java:85)
        at com.ibm.tx.jta.impl.RegisteredSyncs.coreDistributeAfter(RegisteredSyncs.java:301)
        at com.ibm.tx.jta.impl.RegisteredSyncs.distributeAfter(RegisteredSyncs.java:278)
        at com.ibm.tx.jta.embeddable.impl.EmbeddableTransactionImpl.distributeAfter(EmbeddableTransactionImpl.java:315)
        at com.ibm.tx.jta.impl.TransactionImpl.postCompletion(TransactionImpl.java:2484)
        at com.ibm.tx.jta.impl.TransactionImpl.internalCommit(TransactionImpl.java:1541)
        at com.ibm.tx.jta.impl.TransactionImpl.coreStage2CommitProcessing(TransactionImpl.java:875)
        at com.ibm.tx.jta.impl.TransactionImpl.stage2CommitProcessing(TransactionImpl.java:912)
        at com.ibm.tx.jta.impl.TransactionImpl.processCommit(TransactionImpl.java:773)
        at com.ibm.tx.jta.impl.TransactionImpl.commit(TransactionImpl.java:713)
        at com.ibm.tx.jta.impl.TranManagerImpl.commit(TranManagerImpl.java:153)
        at com.ibm.tx.jta.impl.TranManagerSet.commit(TranManagerSet.java:110)
        at com.ibm.tx.jta.impl.UserTransactionImpl.commit(UserTransactionImpl.java:145)
[...]

PR #3912, opened 2022-08-23
issue #3911 - fix transaction data registration for bulk data

punktilious:

Bulk data transactions are started and ended by the JavaBatch framework. To handle this, we have to tell our FHIRUserTransactionAdapter whenever it is part of a bulk data job. This is done using the `bulk` flag on the FHIRRequestContext. This used to work, but we were actually causing our transaction sync data to be registered multiple times because the bulk data code invoked FHIRTransactionHelper multiple times within a single JavaBatch transaction:

```
JavaBatch: start global transaction
BulkData: FHIRTransactionHelper txn = new FHIRTransactionHelper(fhirPersistence.getTransaction());
BulkData: txn.begin();
BulkData: txn.end();
...
BulkData: FHIRTransactionHelper txn = new FHIRTransactionHelper(fhirPersistence.getTransaction());
BulkData: txn.begin();
BulkData: txn.end();
...
JavaBatch: end global transaction
```

This was causing our M_JDBC_COMMIT metric to be created multiple times, with the first instance being closed first, which triggers the IllegalStateException seen in the logs.

For the fix, we have to add a flag to the local thread context to indicate whether the transaction has been configured. The cleanest way is to add this flag to the FHIRRequestContext (alongside the `bulk` flag). Then we can guard the configuration of the transaction so that it only happens once for a particular request in FHIRUserTransactionAdapter:

```
        } else if (isActive(status) && bulk) {
            // On starting a bulk transaction, we need to register a callback so that
            // the cache is informed when the transaction commits it can promote thread-local
            // ids to the shared caches.
            if (!ctx.isBulkTransactionConfigured()) {
                // Only register on the first encounter
                syncRegistry.registerInterposedSynchronization(new CacheTransactionSync(this.syncRegistry, this.cache, this.transactionDataKey, this.afterTransactionHandler));
                ctx.setBulkTransactionConfigured(true);
            }
```

This ensures that we only end up with a single CacheTransactionSync object per JavaBatch global transaction, and therefore create only a single M_JDBC_COMMIT metric.

There remains the issue why the bulk data tests were passing even though the jobs themselves were failing with this IllegalStateException. Raised issue #3913 to address this.

Also fixed an issue causing build/common/bulkdata-setup.sh to fail, which prevented the copying of bulkdata test artefacts to the correct test directory.


lmsurpre (approved):
LGTM

PR #3916, opened 2022-08-25
issue #3859 - track retired tables (and views) in the model so we can drop them

lmsurpre:
After the previous changes, dropping a schema that was created with a
4.x version of fhir-persistence-schema was failing because we never
dropped the tables that have been removed from the model.
With these changes, we flag the removed resource types as "retired" and
avoid their creation in new schemas that way (vs. leaving them out of
the model entirely).

We could potentially remove this logic if we ever get to
https://github.com/LinuxForHealth/FHIR/issues/3263

I also added `@Deprecated` annotations for old migration steps in our fhir-persistence-schema
Main which I'm thinking we can remove whenever we bump to 6.0 (so we'll continue to support 
any 4.x to 5.x migration, but not 4.x to 6.x).

### DIFF
--- a/build/migration/postgres/8_teardown.sh
+++ b/build/migration/postgres/8_teardown.sh
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+###############################################################################
+# (C) Copyright IBM Corp. 2021, 2022
+#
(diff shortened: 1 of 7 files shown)


punktilious (approved):
LGTM

PR #3918, opened 2022-08-25
issue-2964 - adding validations for interaction, mode paramter and in…

PrasannaHegde1:
…put resource

### DIFF
--- a/operation/fhir-operation-validate/src/main/java/org/linuxforhealth/fhir/operation/validate/ValidateOperation.java
+++ b/operation/fhir-operation-validate/src/main/java/org/linuxforhealth/fhir/operation/validate/ValidateOperation.java
@@ -34,6 +34,7 @@ import org.linuxforhealth.fhir.registry.FHIRRegistry;
 import org.linuxforhealth.fhir.search.util.SearchHelper;
 import org.linuxforhealth.fhir.server.spi.operation.AbstractOperation;
 import org.linuxforhealth.fhir.server.spi.operation.FHIROperationContext;
+import org.linuxforhealth.fhir.server.spi.operation.FHIROperationContext.Type;
 import org.linuxforhealth.fhir.server.spi.operation.FHIROperationUtil;
 import org.linuxforhealth.fhir.server.spi.operation.FHIRResourceHelpers;
 import org.linuxforhealth.fhir.validation.FHIRValidator;
@@ -51,49 +52,40 @@ public class ValidateOperation extends AbstractOperation {
         try {
             Parameter resourceParameter = getParameter(parameters, "resource");
             String resourceTypeName = resourceType != null ? resourceType.getSimpleName() : null;
-            Resource resource;
-            // If $validate is on resource-instance level, fetch the resource from database.
-            if (operationContext != null && FHIROperationContext.Type.INSTANCE.equals(operationContext.getType()) && resourceParameter == null) {
-                resource = resourceHelper.doRead(resourceTypeName, logicalId).getResource();
-                if (resource == null) {
-                    throw buildExceptionWithIssue(resourceTypeName + " with id '" + logicalId + "' was not found", IssueType.NOT_FOUND);
-                }
-            } else if (resourceParameter == null) {
-                throw buildExceptionWithIssue("Input parameter 'resource' is required for the $validate operation", IssueType.INVALID);
-            } else {
-                resource = resourceParameter.getResource();
-            }
+            Parameter modeParameter = getParameter(parameters, "mode");
+            ModeType modeType = getModeType(modeParameter);
+            // Validate the resource parameter
+            Resource resource = validateResource(operationContext, logicalId, resourceHelper, resourceParameter, resourceTypeName, modeType);
             
             List<OperationOutcome.Issue> issues;
             Parameter profileParameter = getParameter(parameters, "profile");
-            Parameter modeParameter = getParameter(parameters, "mode");
             
-            // Only valid "mode" codes are accepted by the $validate operation.
-            validateModeParameter(modeParameter);
+            FHIRPersistence persistence =
+                    (FHIRPersistence) operationContext.getProperty(FHIROperationContext.PROPNAME_PERSISTENCE_IMPL);
+            // Validate mode parameter.
+            validateModeParameter(modeParameter, resourceHelper, resourceTypeName, operationContext.getType(), persistence);
 
             if (profileParameter != null && profileParameter.getValue() != null) {
                 // If the 'profile' parameter is specified, always validate the resource against that profile.
                 Uri profileUri = profileParameter.getValue().as(Uri.class);
                 String profile = profileUri == null ? null : profileUri.getValue();
                 issues = FHIRValidator.validator().validate(resource, profile);
-            } else if (modeParameter != null && modeParameter.getValue() != null
-                    && ("create".equals(modeParameter.getValue().as(Code.class).getValue())
-                            || "update".equals(modeParameter.getValue().as(Code.class).getValue()))) {
+            } else if (modeType != null
+                    && ("create".equals(modeType.value())
+                            || "update".equals(modeType.value()))) {
                 // If the 'mode' parameter is specified and its value is 'create' or 'update', validate the resource
                 // against the FHIR server config's profile properties and the resource's asserted profiles.
                 issues = resourceHelper.validateResource(resource);
-            } else if (modeParameter != null 
-                    && modeParameter.getValue() != null
-                    && ModeType.DELETE.value().equals(modeParameter.getValue().as(Code.class).getValue())
+            } else if (modeType != null
+                    && ModeType.DELETE.value().equals(modeType.value())
                     && FHIROperationContext.Type.INSTANCE.equals(operationContext.getType())) {
                 // If the 'mode' parameter is specified and its value is 'delete' and delete is invoked at the resource-instance level,
                 // validate if the persistence layer implementation supports the "delete" operation
-                issues = validateDeleteResource(resourceTypeName, logicalId, operationContext);
+                issues = validateDeleteResource(resourceTypeName, logicalId, operationContext, persistence);
             } else {
                 // Standard validation against the resource's asserted profiles.
                 issues = FHIRValidator.validator().validate(resource);
-            }
-
+            } 
             return FHIROperationUtil.getOutputParameters(buildResourceValidOperationOutcome(issues));
         } catch (FHIROperationException e) {
             throw e;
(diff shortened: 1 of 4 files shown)


operation/fhir-operation-validate/src/main/java/org/linuxforhealth/fhir/operation/validate/ValidateOperation.java
```diff
@@ -131,20 +123,75 @@ private OperationOutcome buildResourceValidOperationOutcome(List<Issue> issues)
...
             } catch (IllegalArgumentException e) {
                 String msg = "'" + modeParameter.getValue().as(Code.class).getValue() + "' is not a valid resource validation mode";
-                throw buildExceptionWithIssue(msg, IssueType.VALUE);
+                throw buildExceptionWithIssue(IssueSeverity.ERROR, IssueType.INVALID, msg, null);
+            }
+            
+            // Validate an interaction for a specified resource type.
+            if (ModeType.CREATE.value().equals(modetype.value())  
+                    || ModeType.UPDATE.value().equals(modetype.value()) 
+                    || ModeType.DELETE.value().equals(modetype.value())) {
```
lmsurpre, line 153: you can use `==` with enums
```suggestion
            if (ModeType.CREATE == modetype)  
                    || ModeType.UPDATE == modetype) 
                    || ModeType.DELETE == modetype)) {
```

operation/fhir-operation-validate/src/test/java/org/linuxforhealth/fhir/operation/validate/ValidateOperationTest.java
```diff
@@ -404,13 +426,13 @@ public void testValidateOperationDeleteNotSupported() throws FHIROperationExcept
...
-                    .severity(IssueSeverity.WARNING)
+                    .severity(IssueSeverity.ERROR)
                     .code(IssueType.NOT_SUPPORTED)
                     .details(CodeableConcept.builder()
-                        .text(string("Resource deletion of type 'Patient' with id '1' is not supported."))
+                        .text(string("Resource deletion, of type 'Patient', with id '1', is not supported."))
                         .build())
                     .build();
-            
+            System.out.println(input.toString());
```
lmsurpre, line 435: don't forget to remove this

operation/fhir-operation-validate/src/main/java/org/linuxforhealth/fhir/operation/validate/ValidateOperation.java
PrasannaHegde1: this has been fixed.

operation/fhir-operation-validate/src/test/java/org/linuxforhealth/fhir/operation/validate/ValidateOperationTest.java
PrasannaHegde1: this has been fixed.

operation/fhir-operation-validate/src/main/java/org/linuxforhealth/fhir/operation/validate/ValidateOperation.java
```diff
@@ -131,20 +120,66 @@ private OperationOutcome buildResourceValidOperationOutcome(List<Issue> issues)
...
      *
      * @param modeParameter resource validation mode code to be validated
+     * @param resourceHelper Resource operation provider for loading related Library resources
+     * @param resourceType a valid resource type string
+     * @param operationContext the FHIROperationContext associated with the request
      * @throws FHIROperationException when the resource validation mode code is invalid
      */
-    private void validateModeParameter(Parameter modeParameter) throws FHIROperationException {
+    private void validateModeParameter(ModeType modetype, FHIRResourceHelpers resourceHelper, String resourceType, FHIROperationContext operationContext) throws FHIROperationException {
+        if (modetype != null) {
```
lmsurpre, line 136: this is a style-only comment, but I like to use "guard clauses" for stuff like this.
basically you'd add this here 
```
if (mode == null) {
    return;
}
```
and then unindent everything that follows.
Its just a personal preference thing.

operation/fhir-operation-validate/src/main/java/org/linuxforhealth/fhir/operation/validate/ValidateOperation.java
```diff
@@ -131,20 +120,66 @@ private OperationOutcome buildResourceValidOperationOutcome(List<Issue> issues)
...
+            // Validate an interaction for a specified resource type.
+            if ((modetype == ModeType.CREATE  
+                    || modetype == ModeType.UPDATE 
+                    || modetype == ModeType.DELETE)
+                    && resourceType != null) {
+                FHIRResourceHelpers.Interaction interaction = FHIRResourceHelpers.Interaction.from(modetype.value());
+                resourceHelper.validateInteraction(interaction, resourceType);
+            }
+            // Validate if persistence layer implementation supports update/create mode if mode = update/create.
```
lmsurpre, line 145: its a little confusing, but I think we should only perform this check (`isUpdateCreateEnabled`) if:
A. mode=create; or
B. mode=update but the resource doesn't exist yet

if the resource already exists, and mode=update, and the update interaction is allowed for this resource type, then the request is valid when the resource that is passed is valid (even if "isUpdateCreateEnabled" is false)

lmsurpre: if B is too complicated (because it requires reading of the resource), we could consider checking just A here

operation/fhir-operation-validate/src/main/java/org/linuxforhealth/fhir/operation/validate/ValidateOperation.java
```diff
@@ -163,9 +198,72 @@ public List<Issue> validateDeleteResource(String type, String id, FHIROperationC
...
+     */
+    protected FHIROperationException buildExceptionWithIssue(IssueSeverity severity, IssueType issueType, String msg, Throwable cause) throws FHIROperationException {
+        OperationOutcome.Issue ooi = FHIRUtil.buildOperationOutcomeIssue(severity, issueType, msg);
+        return new FHIROperationException(msg, cause).withIssue(ooi);
+    }
+    
+    /**
+     * Method to validate the resource parameter. The below validations are performed by this method.
+     * 1. if $validate is invoked at instance level and mode = create, check if the resource already exists.
+     * 2. resource parameter must be present unless the mode is "delete"
```
lmsurpre, line 224: Its probably worth noting here that we also allow the resource parameter to be absent when the mode is "profile".
I opened the follow issue against the specification so that we could reference it in our comment:  https://jira.hl7.org/browse/FHIR-37998
```suggestion
     * 2. resource parameter must be present unless the mode is "delete" (or "profile" per https://jira.hl7.org/browse/FHIR-37998)
```

operation/fhir-operation-validate/src/main/java/org/linuxforhealth/fhir/operation/validate/ValidateOperation.java
PrasannaHegde1: This has been fixed as suggested by adding below check and then unindent everything that follows.
 if (modeType == null) {
    return;
}

operation/fhir-operation-validate/src/main/java/org/linuxforhealth/fhir/operation/validate/ValidateOperation.java
PrasannaHegde1: This has been fixed as suggested.
we should only perform this check (isUpdateCreateEnabled) if
A. mode=create; or
B. mode=update but the resource doesn't exist yet

operation/fhir-operation-validate/src/main/java/org/linuxforhealth/fhir/operation/validate/ValidateOperation.java
PrasannaHegde1: This has been fixed as suggested.

operation/fhir-operation-validate/src/main/java/org/linuxforhealth/fhir/operation/validate/ValidateOperation.java
```diff
@@ -51,49 +51,38 @@ protected Parameters doInvoke(FHIROperationContext operationContext, Class<? ext
...
+                    && (modeType == ModeType.CREATE
+                            || modeType == ModeType.UPDATE)) {
                 // If the 'mode' parameter is specified and its value is 'create' or 'update', validate the resource
                 // against the FHIR server config's profile properties and the resource's asserted profiles.
                 issues = resourceHelper.validateResource(resource);
-            } else if (modeParameter != null 
-                    && modeParameter.getValue() != null
-                    && ModeType.DELETE.value().equals(modeParameter.getValue().as(Code.class).getValue())
-                    && FHIROperationContext.Type.INSTANCE.equals(operationContext.getType())) {
+            } else if (modeType != null
```
punktilious, line 76: Strictly speaking modeType != null isn't required here because if modeType == DELETE, we know it isn't null.

operation/fhir-operation-validate/src/main/java/org/linuxforhealth/fhir/operation/validate/ValidateOperation.java
```diff
@@ -131,20 +120,88 @@ private OperationOutcome buildResourceValidOperationOutcome(List<Issue> issues)
...
      *
      * @param modeParameter resource validation mode code to be validated
-     * @throws FHIROperationException when the resource validation mode code is invalid
+     * @param resourceHelper Resource operation provider for loading related Library resources
+     * @param resourceType a valid resource type string
+     * @param operationContext the FHIROperationContext associated with the request
+     * @throws Exception 
+     */
+    private void validateModeParameter(ModeType modetype, FHIRResourceHelpers resourceHelper, String resourceType, FHIROperationContext operationContext, String logicalId) throws Exception {
+        if(modetype == null) {
```
punktilious, line 137:
```
        Objects.requireNonNull(modeType, "modelType");
```

punktilious: I'm assuming that if validateModeParameter is called with a null value that is an error. The javadoc doesn't say if null is allowed.

operation/fhir-operation-validate/src/main/java/org/linuxforhealth/fhir/operation/validate/ValidateOperation.java
```diff
@@ -131,20 +120,88 @@ private OperationOutcome buildResourceValidOperationOutcome(List<Issue> issues)
...
+                || modetype == ModeType.UPDATE 
+                || modetype == ModeType.DELETE)
+                && resourceType != null) {
+            FHIRResourceHelpers.Interaction interaction = FHIRResourceHelpers.Interaction.from(modetype.value());
+            resourceHelper.validateInteraction(interaction, resourceType);
+        }
+        // Validate if persistence layer implementation supports update/create mode if mode = create.
+        if (operationContext != null && operationContext.getType() == FHIROperationContext.Type.INSTANCE 
+                && modetype == ModeType.CREATE) {
+            validateUpdateCreateEnabled(modetype, resourceType, operationContext);
```
punktilious, line 151: Hmm, the updateCreateEnabled flag is only relevant for UPDATE interactions. It indicates whether or not we can create a resource which doesn't yet exist.

operation/fhir-operation-validate/src/main/java/org/linuxforhealth/fhir/operation/validate/ValidateOperation.java
```diff
@@ -131,20 +120,88 @@ private OperationOutcome buildResourceValidOperationOutcome(List<Issue> issues)
...
+        }
+        // Validate if persistence layer implementation supports update/create mode if mode = update but the resource does not exist yet in the DB.
+        if (operationContext != null && operationContext.getType() == FHIROperationContext.Type.INSTANCE 
+                && modetype == ModeType.UPDATE) {
+            Resource existingResource = resourceHelper.doRead(resourceType, logicalId).getResource();
+            if(existingResource == null) {
```
punktilious, line 157:
```
    if (existingResource == null) {
       ...
    }
```
Space after if please.

operation/fhir-operation-validate/src/main/java/org/linuxforhealth/fhir/operation/validate/ValidateOperation.java
lmsurpre: it sort of depends what mode=create means when invoked at the instance level.
if we take that to mean "can we do a create-on-update"? then i think the above is fine

lmsurpre: if we take it to mean something else, then mode=create should not even be a valid option when invoked at the instance level

operation/fhir-operation-validate/src/main/java/org/linuxforhealth/fhir/operation/validate/ValidateOperation.java
punktilious: I'm OK either way.

lmsurpre:
```suggestion
        if (modetype == null) {
```

operation/fhir-operation-validate/src/main/java/org/linuxforhealth/fhir/operation/validate/ValidateOperation.java
lmsurpre:
```suggestion
            if (existingResource == null) {
```

lmsurpre (approved):
LGTM

operation/fhir-operation-validate/src/main/java/org/linuxforhealth/fhir/operation/validate/ValidateOperation.java
punktilious: Right, CREATE at the instance level is a bit odd. I'm OK leaving it as is.

operation/fhir-operation-validate/src/main/java/org/linuxforhealth/fhir/operation/validate/ValidateOperation.java
PrasannaHegde1: This has been fixed as suggested.

punktilious (approved):
LGTM

Issue #3910, opened 2022-08-23
Add retry mechanism to ExampleRequestProcessor


lmsurpre:
**Is your feature request related to a problem? Please describe.**
R4ExampleServerTest used to pass even when some of the resources failed to post.  That was fixed but, due to derby limitations, we're experiencing some resultant intermittent test failures in the CI.

**Describe the solution you'd like**
If we get a 500 server error, we should retry the interaction.
Can we still fail the test if we get too many 500 server errors?

**Describe alternatives you've considered**
Consider adding something more explicit about whether/when the client should retry.
For example, if we can return 503 instead of 500, we could include a `Retry-After` header in the response.

**Acceptance Criteria**

1.  
   GIVEN [a precondition]
    AND [another precondition]
   WHEN [test step]
    AND [test step]
   THEN [verification step]
    AND [verification step]

**Additional context**

lmsurpre:
one example of an intermittent issue we see only on derby:
```
[8/23/22, 3:00:38:567 UTC] 00000055 ir.persistence.params.database.PlainDerbyParamValueProcessor E failed: INSERT INTO common_token_values (code_system_id, token_value)      VALUES (?,?) 
java.sql.SQLTransactionRollbackException: A lock could not be obtained within the time requested
	at org.apache.derby.impl.jdbc.SQLExceptionFactory.getSQLException(SQLExceptionFactory.java:100)
	at org.apache.derby.impl.jdbc.Util.generateCsSQLException(Util.java:230)
	at org.apache.derby.impl.jdbc.TransactionResourceImpl.wrapInSQLException(TransactionResourceImpl.java:431)
	at org.apache.derby.impl.jdbc.TransactionResourceImpl.handleException(TransactionResourceImpl.java:360)
	at org.apache.derby.impl.jdbc.EmbedConnection.handleException(EmbedConnection.java:2405)
	at org.apache.derby.impl.jdbc.ConnectionChild.handleException(ConnectionChild.java:88)
	at org.apache.derby.impl.jdbc.EmbedStatement.executeStatement(EmbedStatement.java:1436)
	at org.apache.derby.impl.jdbc.EmbedPreparedStatement.executeStatement(EmbedPreparedStatement.java:1709)
	at org.apache.derby.impl.jdbc.EmbedPreparedStatement.executeLargeUpdate(EmbedPreparedStatement.java:320)
[...]

PR #3926, opened 2022-08-29
issue #3910 - add a simple retry loop

lmsurpre:
I added a simple retry loop with a hard-coded MAX_TRIES to 2
and no backoff between attempts.

I did not add any logic to fail the test if we see too many of these.


PrasannaHegde1 (approved):
LGTM

punktilious:
I know this has already been approved and merged, but it's worth stating that typically we use a random backoff in situations like this where the failure could be caused by a deadlock. By retrying immediately, it increases the chance we may hit yet another deadlock. This also attempts only a single retry. Based on similar failures in the past, 5 or even 10 attempts might be appropriate.

lmsurpre:
thanks robin.  for now its just a bandaid to avoid frustrating CI failures.  we can revisit if we see repeated failures with no better workaround.  in meantime, i captured your good idea about the `Retry-After` header in https://github.com/LinuxForHealth/FHIR/issues/3930

PR #3932, opened 2022-08-31
Update CommonUtil.java

berkant-k:
When logLevel set to FINE;  castings to URLClassLoader  was throwing  java.lang.ClassCastException.
To fix this issue logClassPath method changed

### DIFF
--- a/fhir-persistence-schema/src/main/java/org/linuxforhealth/fhir/schema/app/util/CommonUtil.java
+++ b/fhir-persistence-schema/src/main/java/org/linuxforhealth/fhir/schema/app/util/CommonUtil.java
@@ -62,10 +62,8 @@ public final class CommonUtil {
     public static void logClasspath(Logger logger) {
         if (logger.isLoggable(Level.FINE)) {
             logger.fine("CLASSPATH: ");
-            String[] classPaths = System.getProperty("java.class.path").split(File.pathSeparator);
-            for (String path : classPaths) {
-                logger.fine(path);
-            }           
+            String classPaths = System.getProperty("java.class.path");
+            logger.fine(path);            
         }
     }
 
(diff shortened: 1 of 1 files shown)


fhir-persistence-schema/src/main/java/org/linuxforhealth/fhir/schema/app/util/CommonUtil.java
```diff
@@ -62,11 +62,10 @@ public static void configureLogger() {
     public static void logClasspath(Logger logger) {
         if (logger.isLoggable(Level.FINE)) {
             logger.fine("CLASSPATH: ");
-            ClassLoader cl = ClassLoader.getSystemClassLoader();
-            URL[] classpath = ((URLClassLoader) cl).getURLs();
-            for (URL u : classpath) {
-                logger.fine("  " + u.getFile());
-            }
+            String[] classPaths = System.getProperty("java.class.path").split(File.pathSeparator);
```
lmsurpre, line 65: is `System.getProperty("java.class.path")` a reliable way to get the classPath string?

is it always delimited by File.pathSeparator?  I might have expected ":" for *nix, and ";" for windows ... is the `java.class.path` documented somewhere?

lmsurpre: why not just log the result of `System.getProperty("java.class.path")` directly?

berkant-k: Yes. You are right. I will edit the code.

lmsurpre (approved):
LGTM. Thanks @berkant-k!

PR #3952, opened 2022-09-06
Issue 3650 followon

lmsurpre:
1. add versioned references in all US Core 5.0.1 profiles
2. add versioned references for all CARIN for BB references to US Core extensions


punktilious (approved):
LGTM

Issue #3959, opened 2022-09-09
Polynomial regular expression used on uncontrolled data


lmsurpre:
operation/fhir-operation-bulkdata/src/main/java/org/linuxforhealth/fhir/operation/bulkdata/client/BulkDataClient.java is splitting with the following regular expression `//s*,//s*` (twice)...it would be better to split on the comma and then trim each item.

fhir-search/src/main/java/org/linuxforhealth/fhir/search/util/SearchHelper.java does the same.

lmsurpre:
we fixed this before merging the codescanning to the repo, so it doesn't even appear on the list

PR #3961, opened 2022-09-12
issue #3959 - avoid needless whitespace handling in String.split

lmsurpre:
To avoid a moderate-level warnings from CodeQL.

In BulkDataClient, we're parsing our own output...there's no reason to expect whitespace.

In SearchHelper.parseQueryParameters I don't think we need to handle type parameter values with whitespace in the list of types either. However, since we already had code to support that, I decided to add a call to trim() after removing whitespace handling from the split.

### DIFF
--- a/operation/fhir-operation-bulkdata/src/main/java/org/linuxforhealth/fhir/operation/bulkdata/client/BulkDataClient.java
+++ b/operation/fhir-operation-bulkdata/src/main/java/org/linuxforhealth/fhir/operation/bulkdata/client/BulkDataClient.java
@@ -548,12 +548,12 @@ public class BulkDataClient {
 
         // Export Jobs with data in the exitStatus field
         if (!"COMPLETED".equals(exitStatus) && !"bulkimportchunkjob".equals(executionResponse.getJobName())) {
-            List<String> resourceTypeInfs = Arrays.asList(exitStatus.split("\\s*:\\s*"));
+            List<String> resourceTypeInfs = Arrays.asList(exitStatus.split(":"));
             List<PollingLocationResponse.Output> outputList = new ArrayList<>();
             for (String resourceTypeInf : resourceTypeInfs) {
                 String resourceType = resourceTypeInf.substring(0, resourceTypeInf.indexOf("["));
                 String[] resourceCounts =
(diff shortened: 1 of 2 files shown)


operation/fhir-operation-bulkdata/src/main/java/org/linuxforhealth/fhir/operation/bulkdata/client/BulkDataClient.java
```diff
@@ -548,12 +548,12 @@ private void addExecutionResponse(PollingLocationResponse pollingLocationRespons
...
         // Export Jobs with data in the exitStatus field
         if (!"COMPLETED".equals(exitStatus) && !"bulkimportchunkjob".equals(executionResponse.getJobName())) {
-            List<String> resourceTypeInfs = Arrays.asList(exitStatus.split("\\s*:\\s*"));
+            List<String> resourceTypeInfs = Arrays.asList(exitStatus.split(":"));
             List<PollingLocationResponse.Output> outputList = new ArrayList<>();
             for (String resourceTypeInf : resourceTypeInfs) {
                 String resourceType = resourceTypeInf.substring(0, resourceTypeInf.indexOf("["));
                 String[] resourceCounts =
-                        resourceTypeInf.substring(resourceTypeInf.indexOf("[") + 1, resourceTypeInf.indexOf("]")).split("\\s*,\\s*");
+                        resourceTypeInf.substring(resourceTypeInf.indexOf("[") + 1, resourceTypeInf.indexOf("]")).split(",");
```
punktilious, line 556: I'm going to assume here that there's no trimming required, right?

punktilious (approved):
LGTM

operation/fhir-operation-bulkdata/src/main/java/org/linuxforhealth/fhir/operation/bulkdata/client/BulkDataClient.java
lmsurpre: yeah, i tried to cover that in the commit message:
> In BulkDataClient.addExecutionResponse we're parsing our own output...there's no reason to expect whitespace.

to test it, I did an export of multiple resource types and changed the config to ensure that we end up with multiple "objects" for at least one of those resource types

Issue #3958, opened 2022-09-09
Use better cipher for encrypting bulk data job ids


lmsurpre:
Currently we use in `AES/ECB/PKCS5Padding` in JobIdEncodingTransformer.  This is known to be weak, but its not a very sensitive value being encoded...its just a way to obfuscate the otherwise-numeric job id.

Still, it would be better to use a more secure cipher like "AES/GCM/NoPadding" (based on https://rules.sonarsource.com/java/tag/owasp/RSPEC-5542 )

lmsurpre:
Instead of using a cipher, I replaced the use of AES encryption with the use of [Hashids](https://hashids.org)

lmsurpre:
we fixed this before merging the codescanning to the repo, so it doesn't even appear on the list

PR #3969, opened 2022-09-13
issue #3958 - replace AES encryption with Hashids

lmsurpre:
The encryption was used to obscure the incrementing job id and not as a secure cryptographic function.
However, it was flagged in CodeQL as insecure and the use of "encryption" in the code made it less clear that repeatable encryption/decryption was expected/desired (and not a vulnerability).

Here I replaced the encryption with the use of a separate library called "Hashids" which is fit-for-purpose without implying imaginary security; its just a convenient way to convert between numeric values and short, unique string values for use in URLs.


operation/fhir-operation-bulkdata/src/main/java/org/linuxforhealth/fhir/operation/bulkdata/config/impl/AbstractSystemConfigurationImpl.java
```diff
@@ -107,7 +107,7 @@ private static final int defaultCoreMaxParititions() {
     }
 
     @Override
-    public String getCoreBatchIdEncryptionKey() {
+    public String getCoreBatchIdEncodingKey() {
         return FHIRConfigHelper.getStringProperty("fhirServer/bulkdata/core/batchIdEncryptionKey", null);
```
lmsurpre, line 111: I sort of wanted to change the config property name here to use "encoding" instead of "encryption" but that would break all the existing configs.  Think its worth it?

operation/fhir-operation-bulkdata/src/main/java/org/linuxforhealth/fhir/operation/bulkdata/model/transformer/JobIdEncodingTransformer.java
```diff
@@ -50,86 +40,54 @@ public static JobIdEncodingTransformer getInstance() {
...
-        }
+    private Hashids createHashids() {
+        String salt = ConfigurationFactory.getInstance().getCoreBatchIdEncodingKey();
 
-        if (secretKey == null) {
-            logger.warning("Failed to get encryption key, JavaBatch Job ids will not be encrypted!");
+        if (salt != null && !salt.isEmpty()) {
+            return new Hashids(salt, MIN_ENCODED_LENGTH);
+        } else {
+            return null;
```
punktilious, line 53: Should this be an exception?

operation/fhir-operation-bulkdata/src/main/java/org/linuxforhealth/fhir/operation/bulkdata/model/transformer/JobIdEncodingTransformer.java
```diff
@@ -50,86 +40,54 @@ public static JobIdEncodingTransformer getInstance() {
...
-            try {
-                byte[] keyBytes = encryptionKey.getBytes("UTF-8");
-                keyBytes = Arrays.copyOf(MessageDigest.getInstance("SHA-1").digest(keyBytes), 16);
-                secretKey = new SecretKeySpec(keyBytes, "AES");
-            } catch (Exception e) {
-                logger.log(Level.WARNING, "Fail to generate encryption key from config!", e);
-            }
-        }
+    private Hashids createHashids() {
+        String salt = ConfigurationFactory.getInstance().getCoreBatchIdEncodingKey();
```
punktilious, line 48: Shouldn't the tenantId also be included as part of the salt?

operation/fhir-operation-bulkdata/src/main/java/org/linuxforhealth/fhir/operation/bulkdata/model/transformer/JobIdEncodingTransformer.java
lmsurpre: currently we allow the `coreBatchIdEncryptionKey` to be omitted in which case we use the numeric job id.  i'm not sure if thats really necessary, but it does make testing slightly nicer (because the job ids are numeric instead of a cryptic string).

basically if this comes back null then there is no encoding/decoding (but not an error)

operation/fhir-operation-bulkdata/src/main/java/org/linuxforhealth/fhir/operation/bulkdata/model/transformer/JobIdEncodingTransformer.java
lmsurpre: we have tenant-level protection on the job stuff already (a job is for a single tenant so if the wrong tenant asks for it then they will be rejected).

however, I guess it wouldn't hurt to include the tenantid in the salt... I can add that if you want

operation/fhir-operation-bulkdata/src/main/java/org/linuxforhealth/fhir/operation/bulkdata/config/impl/AbstractSystemConfigurationImpl.java
punktilious: Yes. I think breaking the config is beneficial in this case because it will draw attention to the change.

operation/fhir-operation-bulkdata/src/main/java/org/linuxforhealth/fhir/operation/bulkdata/model/transformer/JobIdEncodingTransformer.java
punktilious: OK, I think it's fine at this time - no need to make the change bigger. I'm just not a fan of that (original) design because it doesn't really adhere to separation of concerns. My personal choice would be to skip the encode/decode call if not required by the config. It's a subtle point, but it means that if we were to use a different encoder/decoder implementation, that too would need to handle null keys...but this is being a bit purist. Still, always interesting to think about the ramifications of a particular API design.

operation/fhir-operation-bulkdata/src/main/java/org/linuxforhealth/fhir/operation/bulkdata/config/impl/AbstractSystemConfigurationImpl.java
lmsurpre: done

operation/fhir-operation-bulkdata/src/main/java/org/linuxforhealth/fhir/operation/bulkdata/model/transformer/JobIdEncodingTransformer.java
lmsurpre: done

operation/fhir-operation-bulkdata/src/main/java/org/linuxforhealth/fhir/operation/bulkdata/model/transformer/JobIdEncodingTransformer.java
```diff
@@ -50,86 +40,54 @@ public static JobIdEncodingTransformer getInstance() {
...
     /**
-     * decodes the job id.
-     *
-     * @implNote note a Cipher is used here, however it is not used to encrypt sensitive information rather encode the
-     *           jobId.
+     * decodes the job id
      *
      * @param encodedJobId
      * @return
+     * @throws IllegalArgumentException if the passed encodedJobId could not be decoded
```
punktilious, line 78: remove this old code?

lmsurpre: I chose to keep that part because we don't have a good way to know whether the jobId passed by the client is valid or not until we try decoding it.

lmsurpre: One thing we could (should?) do is look for invalid chars in the "encodedJobId" before we try decoding that...that would at least catch anything nefarious.
I can add that but I think we need to keep throwing if the encodedJobId doesn't decode as expected.

lmsurpre: actually I just checked the Hashids lib and its basically already doing this (right away when you call decode):
```
    String validChars = this.alphabet + this.guards + this.seps;
    for (int i = 0; i < hash.length(); i++) {
      if(validChars.indexOf(hash.charAt(i)) == -1) {
        return new long[0];
      }
    }
 ```
 
 so they chose to return an empty array.  in our case we're returning a String, but we could define a String of "-1" to be an error if we prefer.

punktilious: But is the KEY_MAP still required?

lmsurpre: oh, i thought i deleted this whole file... oops

lmsurpre: ok, its gone now.  good catch.

punktilious (approved):
LGTM

PR #3976, opened 2022-09-15
revert to setup-java@v2

lmsurpre:
Recently I bumped the version to setup-java@v3 and it broke. Something about checking out old tags must have changed between v2 and v3 but I don't have time to look into it right now. So, back to v2 we go.


PrasannaHegde1 (approved):
LGTM

PR #3986, opened 2022-09-19
Remove tests

berkant-k:
Tests should not expect certain number of of resources. Because fhir-registry can be extended  by adding new structure definitions, profiles etc. There can be more than expected resource count.  (For example observationProfiles.size()can be more than 17 )

### DIFF
--- a/conformance/fhir-core-r4b/src/test/java/org/linuxforhealth/fhir/core/r4b/test/FHIRRegistryTest.java
+++ b/conformance/fhir-core-r4b/src/test/java/org/linuxforhealth/fhir/core/r4b/test/FHIRRegistryTest.java
@@ -27,24 +27,6 @@ public class FHIRRegistryTest {
     }
 
     @Test
-    public void testGetResourcesByResourceType() {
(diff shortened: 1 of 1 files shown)


lmsurpre:
I do agree that these types of "assert x resources" are brittle and somewhat limited in value.
However, I've always kept them around because its a nice sanity check if you ever update the package.
The specific test you modified here is expected to be run in the context of the specific module they exist within (fhir-core-r4b) and thus serve to confirm that is the number of such resources form this module.

You should not be adding your own profiles to the fhir-core-r4b module, but rather contributing your resources to the Registry through your own FHIRRegistryResourceProvider as documented at https://linuxforhealth.github.io/FHIR/guides/FHIRValidationGuide#making-profiles-available-to-the-fhir-registry-component-fhirregistry

berkant-k:
I already implemented my own resources as documented. I did not touch rb4 registry. However rb4 tests are failed. I think it loads all registry providers and calculate counts.

berkant-k:
I will check what is the reason. It might be a duplicated registry provider. Thanks for your quick reply.

PR #3993, opened 2022-09-26
issue-3968 - validating the tenant Id passed in header in configured tenants list

PrasannaHegde1:
…tenants list

Signed-off-by: PrasannaHegde1 <phegde@merative.com>

### DIFF
--- a/fhir-server/src/main/java/org/linuxforhealth/fhir/server/filter/rest/FHIRRestServletFilter.java
+++ b/fhir-server/src/main/java/org/linuxforhealth/fhir/server/filter/rest/FHIRRestServletFilter.java
@@ -63,7 +63,7 @@ public class FHIRRestServletFilter extends HttpFilter {
     private static final String preferHandlingHeaderSectionName = "handling";
     private static final String preferReturnHeaderSectionName = "return";
     
-    private static List<String> configuredTenants = null;
+    private static Map<String, String> configuredTenants = null;
 
     private static String defaultTenantId = null;
     private static final HTTPReturnPreference defaultHttpReturnPref = HTTPReturnPreference.MINIMAL;
@@ -138,9 +138,10 @@ public class FHIRRestServletFilter extends HttpFilter {
             FHIRConfiguration.validateTenantId(tenantId);
             FHIRConfiguration.validateDatastoreId(dsId);
 
-            // Checks for Valid Tenant Configuration
+            // Check if the tenant Id is present in the list of configured tenants and return the tenant Id from the configured tenant list.
+            // This is done to avoid Uncontrolled user input data used in path expression
             
-            tenantId = fetchValidTenantId(tenantId);
+            tenantId = fetchValidConfiguredTenantId(tenantId);
             
             checkValidTenantConfiguration(tenantId);
 
@@ -531,8 +532,8 @@ public class FHIRRestServletFilter extends HttpFilter {
             log.info("Configured shard-key header name is: '" +  shardKeyHeaderName + "'");
             
             if (filterConfig.getServletContext().getAttribute(FHIRServletContextListener.FHIR_CONFIGURED_TENANTS) != null 
-                    && filterConfig.getServletContext().getAttribute(FHIRServletContextListener.FHIR_CONFIGURED_TENANTS) instanceof List<?>) {
-                configuredTenants = (List<String>) filterConfig.getServletContext().getAttribute(FHIRServletContextListener.FHIR_CONFIGURED_TENANTS);
+                    && filterConfig.getServletContext().getAttribute(FHIRServletContextListener.FHIR_CONFIGURED_TENANTS) instanceof Map<?, ?>) {
+                configuredTenants = (Map<String, String>) filterConfig.getServletContext().getAttribute(FHIRServletContextListener.FHIR_CONFIGURED_TENANTS);
             }
         } catch (Exception e) {
             throw new ServletException("Servlet filter initialization error.", e);
@@ -546,10 +547,10 @@ public class FHIRRestServletFilter extends HttpFilter {
      * @return tenantId If the tenant Id is present this method returns the value from the configured tenants list.
      * @throws Exception If the tenant Id is not present in the list configured tenants.
      */
-    private String fetchValidTenantId(String tenantId) throws Exception {
+    private String fetchValidConfiguredTenantId(String tenantId) throws Exception {
         try {
-            if (!configuredTenants.isEmpty() && configuredTenants.contains(tenantId)) {
-                return configuredTenants.get(configuredTenants.indexOf(tenantId));
+            if (configuredTenants != null && !configuredTenants.isEmpty() && configuredTenants.containsKey(tenantId)) {
+                return configuredTenants.get(tenantId);
             } else {
                 log.severe("Missing tenant configuration for '"  + tenantId + "'");
                 throw new FHIRException("Tenant configuration does not exist: " + tenantId);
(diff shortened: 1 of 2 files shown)


fhir-server/src/main/java/org/linuxforhealth/fhir/server/filter/rest/FHIRRestServletFilter.java
```diff
@@ -61,6 +62,8 @@ public class FHIRRestServletFilter extends HttpFilter {
     private static final String preferHeaderName = "Prefer";
     private static final String preferHandlingHeaderSectionName = "handling";
     private static final String preferReturnHeaderSectionName = "return";
+    
+    private static List<String> configuredTenants = null;
```
lmsurpre, line 66: Can we make this a set instead?

PrasannaHegde1: This has been fixed as suggested to unmodifiable concurrent hashmap.

fhir-server/src/main/java/org/linuxforhealth/fhir/server/filter/rest/FHIRRestServletFilter.java
```diff
@@ -135,7 +138,11 @@ public void doFilter(HttpServletRequest request, HttpServletResponse response, F
             FHIRConfiguration.validateTenantId(tenantId);
```
lmsurpre, line 138: Now that we're checking the tenantId against a preconfigured list of tenants, maybe we don't need to validate it here any more?

fhir-server/src/main/java/org/linuxforhealth/fhir/server/listener/FHIRServletContextListener.java
```diff
@@ -299,6 +304,17 @@ public void contextInitialized(ServletContextEvent event) {
...
             EventManager.serverReady(serviceManagerId);
+            
+            // Set the configured tenants list which will be used in FHIRRestServletFilter to check 
+            //   if the tenant Id is present in the list of configured tenants and return the tenant Id from the configured tenant list. 
+            // This is done to avoid Uncontrolled user input data used in path expression
+            List<String> configuredTenantsList = fhirConfiguration.getConfiguredTenants();
+            if (configuredTenantsList.isEmpty()) {
+                throw new IllegalStateException("No configured tenants were found");
+            }
+            configuredTenants = Collections.unmodifiableMap(configuredTenantsList.stream().collect(Collectors.toConcurrentMap(tenantId -> tenantId, tenantId -> tenantId)));
```
lmsurpre, line 315: I don't think it needs to be a concurrent map, but maybe there's something I'm missing?

lmsurpre (approved):

fhir-server/src/main/java/org/linuxforhealth/fhir/server/filter/rest/FHIRRestServletFilter.java
```diff
@@ -135,7 +138,11 @@ public void doFilter(HttpServletRequest request, HttpServletResponse response, F
...
             FHIRConfiguration.validateDatastoreId(dsId);
 
-            // Checks for Valid Tenant Configuration
+            // Check if the tenant Id is present in the list of configured tenants and return the tenant Id from the configured tenant list.
+            // This is done to avoid Uncontrolled user input data used in path expression
+            
+            tenantId = fetchValidConfiguredTenantId(tenantId);
+            
             checkValidTenantConfiguration(tenantId);
```
PrasannaHegde1, line 146: Can this validation be removed now?

fhir-server/src/main/java/org/linuxforhealth/fhir/server/filter/rest/FHIRRestServletFilter.java
PrasannaHegde1: This validation is for allowed characters for a tenant Id. Can this be removed?

fhir-server/src/main/java/org/linuxforhealth/fhir/server/filter/rest/FHIRRestServletFilter.java
lmsurpre: yes, I think we can remove this now that you've implemented fetchValidConfiguredTenantId instead

fhir-server/src/main/java/org/linuxforhealth/fhir/server/filter/rest/FHIRRestServletFilter.java
```diff
@@ -523,8 +530,38 @@ public void init(FilterConfig filterConfig) throws ServletException {
...
+    /**
+     * Checks if the tenant id is present in the list configured tenants. 
+     * If the tenant Id is present this method returns the value from the configured tenants list.
+     * @param tenantId the tenant for which the configured Tenants is going to be checked
+     * @return tenantId If the tenant Id is present this method returns the value from the configured tenants list.
+     * @throws Exception If the tenant Id is not present in the list configured tenants.
+     */
+    private String fetchValidConfiguredTenantId(String tenantId) throws Exception {
+        try {
+            if (configuredTenants != null && !configuredTenants.isEmpty() && configuredTenants.containsKey(tenantId)) {
```
punktilious, line 552: Doing 2 lookups into the Map (containsKey and get). Just do a get and check that the result is not null (because we never store null values in this map).

fhir-server/src/main/java/org/linuxforhealth/fhir/server/filter/rest/FHIRRestServletFilter.java
```diff
@@ -523,8 +530,38 @@ public void init(FilterConfig filterConfig) throws ServletException {
...
+                return configuredTenants.get(tenantId);
+            } else {
+                log.severe("Missing tenant configuration for '"  + tenantId + "'");
+                throw new FHIRException("Tenant configuration does not exist: " + tenantId);
+            }
+        } catch (FHIRException fe) {
+            throw fe;
+        } catch (Throwable t) {
+            String msg = "Unexpected error while valid the input tenant Id with the list of configured tenants.";
```
punktilious, line 561: validating

fhir-server/src/main/java/org/linuxforhealth/fhir/server/listener/FHIRServletContextListener.java
punktilious: The map is being initialized in one thread but used in other threads. Without explicit synchronization, it's safer to use a ConcurrentHashMap here (although the act of making it unmodifiable may offer the same protection....would be interesting to look at this).

fhir-server/src/main/java/org/linuxforhealth/fhir/server/filter/rest/FHIRRestServletFilter.java
lmsurpre:
```suggestion
            String msg = "Unexpected error while validating the input tenant Id with the list of configured tenants.";
```

PrasannaHegde1: This has been fixed as suggested.

fhir-server/src/main/java/org/linuxforhealth/fhir/server/filter/rest/FHIRRestServletFilter.java
PrasannaHegde1: This has been fixed as suggested. Removed the additional lookup from Map.

fhir-server/src/main/java/org/linuxforhealth/fhir/server/filter/rest/FHIRRestServletFilter.java
lmsurpre: I think it could be removed if we want, but it shouldn't hurt to keep it and could potentially provide a better error message than the generic "no config for this tenant".

PR #4008, opened 2022-10-09
Update FHIRPersistenceJDBCImpl.java

berkant-k:
searchHelper.getSearchParameter might return null, and constructure of JDBCParameterBuildingVisitor needs searchParameter. Added new control For checking null reference and create a useful log entry.

### DIFF
--- a/fhir-persistence-jdbc/src/main/java/org/linuxforhealth/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
+++ b/fhir-persistence-jdbc/src/main/java/org/linuxforhealth/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
@@ -1912,7 +1912,7 @@ public class FHIRPersistenceJDBCImpl implements FHIRPersistence, SchemaNameSuppl
                                 SearchParameter compSP = searchHelper.getSearchParameter(p.getResourceType(), component.getDefinition());
                                 if (compSP == null) {
                                     throw new NullPointerException(String.format("The 'compSP' parameter was " + 
-                                        specified to null [ResourceType=%s,Uri=%s]", p.getResourceType(), component.getDefinition()));
+                                        "specified to null [ResourceType=%s,Uri=%s]", p.getResourceType(), component.getDefinition()));
                                 }                                
                                 JDBCParameterBuildingVisitor parameterBuilder = new JDBCParameterBuildingVisitor(p.getResourceType(), compSP);
                                 FHIRPathNode node = nodes.iterator().next();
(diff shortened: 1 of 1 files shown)


fhir-persistence-jdbc/src/main/java/org/linuxforhealth/fhir/persistence/jdbc/impl/FHIRPersistenceJDBCImpl.java
```diff
@@ -1910,6 +1910,9 @@ private ExtractedSearchParameters extractSearchParameters(Resource fhirResource,
                                 // Alternatively, we could pull the search parameter from the FHIRRegistry so we can use versioned references.
                                 // However, that would bypass search parameter filtering and so we favor the SeachUtil method here instead.
                                 SearchParameter compSP = searchHelper.getSearchParameter(p.getResourceType(), component.getDefinition());
+                                if(compSP==null){
+                                    throw new NullPointerException(String.format("The 'compSP' parameter was specified to null [ResourceType=%s,Uri=%s]",p.getResourceType(),component.getDefinition()));
+                                }
```
lmsurpre, line 1915: style suggestion and unrelated comment fix
```suggestion
                                // However, that would bypass search parameter filtering and so we favor the SeachHelper method here instead.
                                SearchParameter compSP = searchHelper.getSearchParameter(p.getResourceType(), component.getDefinition());
                                if (compSP == null) {
                                    throw new NullPointerException(String.format("The 'compSP' parameter was " + 
                                        specified to null [ResourceType=%s,Uri=%s]", p.getResourceType(), component.getDefinition()));
                                }                                
```

lmsurpre:
Thanks  Berkant.  Can you share what is the current behavior if we hit this case (composite search parameter where it references a component that doesn't exist)?

berkant-k:
> Thanks Berkant. Can you share what is the current behavior if we hit this case (composite search parameter where it references a component that doesn't exist)?

Throws  nullpointerexception with out details like below.  We can add some details to log entry . 

```
[10/7/22, 11:23:56:321 TRT] 00000033 FHIRPersisten E   Unexpected error while performing a create operation  java.lang.NullPointerException
	at org.linuxforhealth.fhir.persistence.jdbc.util.JDBCParameterBuildingVisitor.<init>(JDBCParameterBuildingVisitor.java:116)
```

lmsurpre (approved):
I wrote up https://github.com/LinuxForHealth/FHIR/issues/4019 with what I believe would be the proper behavior, but I'm fine with this small PR which improves the error message associated with the NPE for this case.

Issue #4023, opened 2022-10-13
R4JDBCExamplesProcessor fails to test parameter flush to db


lmsurpre:
**Describe the bug**
Prasanna found https://github.com/LinuxForHealth/FHIR/issues/4021 by manually executing our fhir-server-test suite.
This was a bad regression and pointed to a fundamental gap in our testing.

For 5.0, the parameter processing was reworked so that the parameter inserts happen just before the transaction commits
This relies on the use of our FHIRPersistenceTransaction abstraction under fhir-persistence.

However, when I dug into this test it turns out it was using SimpleTransactionProvider from fhir-database-utils instead, and so we were never calling the IFHIRTransactionAdapterCallback which triggers the inserts.

**Environment**
Which version of LinuxForHealth FHIR Server? 5.0

**To Reproduce**
1. manufacture an error such as the one described in #4021
2. ensure the problem example is included in the index being executed
3. execute R4JDBCExamplesProcessor.process

Note that the test succeeds.  This is because the problematic search parameter was never actually written.

**Expected behavior**
The test should execute the IFHIRTransactionAdapterCallback and thus it should fail when there is a problem with this code.

**Additional context**

lmsurpre:
After the fix, the CI pipeline was failing as expected until the fix for #4021 landed.  I think that is sufficient to show that the persistence tests are now covering the parameters flush (at least on derby).

Issue #4021, opened 2022-10-13
Truncate String parameter value before persisting to the database


PrasannaHegde1:
**Describe the bug**
The string parameter value should be truncated before persisting to the database. 
The size of string columns in the search tables(_str_values) is limited to 1024 bytes. 

**Environment**
Which version of LinuxForHealth FHIR Server? 5.0

**To Reproduce**

1. Go to R4ExampleServerTest.main() method
2. Pass "json/spec/medicationstatement.profile.json" as the JSON example in driver.processExample(...) .
3. Run the main method. 
````
[13/10/22, 14:10:42:079 UTC] 00000034 ParameterTran E   Failed persisting parameter transaction data. Marking transaction for rollback
                                 org.linuxforhealth.fhir.persistence.exception.FHIRPersistenceException: pushBatch failed for 'StructureDefinition'  [probeId=c0-a8-1-f-c4963745-837e-4ee4-aba7-5a8e9c31ad09]
	at org.linuxforhealth.fhir.persistence.params.database.PlainBatchParameterProcessor.pushBatch(PlainBatchParameterProcessor.java:117)
	at org.linuxforhealth.fhir.persistence.params.database.PlainParamValueProcessor.pushBatch(PlainParamValueProcessor.java:110)
	at org.linuxforhealth.fhir.persistence.jdbc.impl.FHIRPersistenceJDBCImpl.flush(FHIRPersistenceJDBCImpl.java:3124)
	at org.linuxforhealth.fhir.persistence.jdbc.impl.FHIRPersistenceJDBCImpl.onCommit(FHIRPersistenceJDBCImpl.java:3155)
	at org.linuxforhealth.fhir.persistence.jdbc.impl.ParameterTransactionDataImpl.persist(ParameterTransactionDataImpl.java:70)
	at org.linuxforhealth.fhir.persistence.jdbc.dao.impl.TransactionDataImpl.lambda$persist$0(TransactionDataImpl.java:46)
	at java.base/java.util.HashMap$Values.forEach(HashMap.java:977)
	at org.linuxforhealth.fhir.persistence.jdbc.dao.impl.TransactionDataImpl.persist(TransactionDataImpl.java:46)
	at org.linuxforhealth.fhir.persistence.jdbc.impl.CacheTransactionSync.beforeCompletion(CacheTransactionSync.java:74)
	at com.ibm.tx.jta.impl.RegisteredSyncs.coreDistributeBefore(RegisteredSyncs.java:238)
[...]

PR #4024, opened 2022-10-13
issue #4023 - use FHIRPersistenceTransaction instead of ITransactionProvider

lmsurpre:
So that the callback for inserting search parameter values is invoked.

I'm expecting this to fail until #4021 is addressed... which actually means its working.

### DIFF
--- a/fhir-validation/src/main/java/org/linuxforhealth/fhir/validation/FHIRValidator.java
+++ b/fhir-validation/src/main/java/org/linuxforhealth/fhir/validation/FHIRValidator.java
@@ -330,10 +330,11 @@ public class FHIRValidator {
                 constraints.addAll(ProfileSupport.getConstraints(profiles, resourceType));
             }
 
-            // path (with array indices) -> Extension instance (which never have a url version suffix)
-            Map<String, Extension> pathToExtension = gatherInstanceExtensions(resourceNode);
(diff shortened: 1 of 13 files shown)


fhir-validation/src/main/java/org/linuxforhealth/fhir/validation/FHIRValidator.java
```diff
@@ -330,11 +330,12 @@ private void validate(FHIRPathResourceNode resourceNode) {
                 constraints.addAll(ProfileSupport.getConstraints(profiles, resourceType));
             }
 
-            // path (with array indices) -> Extension instance (which never have a url version suffix)
-            Map<String, Extension> pathToExtension = gatherInstanceExtensions(resourceNode);
             // path (with no array indices) -> Extension URLs (with no version suffix)
             Map<String, Set<String>> alreadyGeneratedExtensionConstraints = gatherExtensionConstraints(constraints);
 
+            // path (with array indices) -> Extension instance (which never have a url version suffix)
+            Map<String, Extension> pathToExtension = gatherInstanceExtensions(resourceNode);
```
lmsurpre, line 337: another minor formatting thing I had locally... the foreach on the next line references pathToExtension and so it made more sense for this to be the line which proceeds that

punktilious (approved):
LGTM

PR #4034, opened 2022-10-20
issue-3716 - validate securityContext for a Binary resource

PrasannaHegde1:
Signed-off-by: PrasannaHegde1 <phegde@merative.com>

### DIFF
--- a/docs/src/pages/guides/FHIRServerUsersGuide.md
+++ b/docs/src/pages/guides/FHIRServerUsersGuide.md
@@ -2323,7 +2323,7 @@ This section contains reference information about each of the configuration prop
 |`fhirServer/security/oauth/smart/enabled`|boolean|Whether or not the server is enabled for OAuth-based authentication/authorization|
 |`fhirServer/security/oauth/smart/scopes`|array|The list of SMART scopes to advertise in the `.well-known/smart-configuration endpoint|
 |`fhirServer/security/oauth/smart/capabilities`|array|The list of SMART capabilities to advertise in the `.well-known/smart-configuration endpoint|
-|`fhirServer/security/validateSecurityContext`|boolean| Whether or not the securityContext validation is enabled for Binary resources|
+|`fhirServer/security/validateBinarySecurityContext`|boolean| Whether or not the securityContext validation is enabled for Binary resources|
 |`fhirServer/audit/serviceClassName`|string|The audit service to use. Currently, org.linuxforhealth.fhir.audit.impl.NopService to indicate the logger service is disabled, and org.linuxforhealth.fhir.audit.impl.KafkaService to indicate using Kafka as a destination.|
 |`fhirServer/audit/serviceProperties/auditTopic`|string|The kafka topic to use for CADF audit logging service|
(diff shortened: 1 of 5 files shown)


fhir-smart/src/test/java/org/linuxforhealth/fhir/smart/test/AuthzPolicyEnforcementTest.java
```diff
@@ -212,6 +238,31 @@ public void testUpdate(String scopeString, List<String> contextIds, Set<Resource
...
+        }
+        
+        // Test update Binary Resource which has a securityContext. Should Fail since securityContext is not supported.
+        try {
+            properties.put(FHIRPersistenceEvent.PROPNAME_RESOURCE_TYPE, "Binary");
+            FHIRPersistenceEvent event = new FHIRPersistenceEvent(binaryWithSecurityContext, properties);
+            event.setPrevFhirResource(binaryWithSecurityContext);
+            interceptor.beforeUpdate(event);
+        } catch (FHIRPersistenceInterceptorException e) {
+            assertTrue(e.getMessage() != null);
```
punktilious, line 262: this test will still pass if the interceptor doesn't throw an exception. Need to add a flag to show that the exception happened.

lmsurpre: @PrasannaHegde1 maybe a line like`fail("Did not receive the expected FHIRPersistenceInterceptorException");` to the end of the try block?

docs/src/pages/guides/FHIRServerUsersGuide.md
```diff
@@ -2323,6 +2323,7 @@ This section contains reference information about each of the configuration prop
 |`fhirServer/security/oauth/smart/enabled`|boolean|Whether or not the server is enabled for OAuth-based authentication/authorization|
 |`fhirServer/security/oauth/smart/scopes`|array|The list of SMART scopes to advertise in the `.well-known/smart-configuration endpoint|
 |`fhirServer/security/oauth/smart/capabilities`|array|The list of SMART capabilities to advertise in the `.well-known/smart-configuration endpoint|
+|`fhirServer/security/validateSecurityContext`|boolean| Whether or not the securityContext validation is enabled for Binary resources|
```
lmsurpre, line 2326: I wonder if we should put the word "Binary" into the config property if its only used for Binary resources.
```suggestion
|`fhirServer/security/validateBinarySecurityContext`|boolean| Whether or not the securityContext validation is enabled for Binary resources|
```
we'd need to update it in all the places, but I do think this is better...

fhir-smart/src/test/java/org/linuxforhealth/fhir/smart/test/AuthzPolicyEnforcementTest.java
lmsurpre: i think the same is true around line 197 above

PrasannaHegde1: This has been fixed as suggested.

lmsurpre (approved):

PR #4066, opened 2022-11-09
Another `capabilities` change

lmsurpre:
The HL7 validator issues a warning on our current CapabilityStatement because `name` is intended for computer processing (so they recommend against whitespace but still allow it).

`CapabilityStatement.title` is the field intended for human consumption and so I think its cleaner to use that instead.

I also removed the no-longer-needed calls to the static `string()` factory...because thats what I do.

### DIFF
--- a/fhir-server/src/main/java/org/linuxforhealth/fhir/server/resources/Capabilities.java
+++ b/fhir-server/src/main/java/org/linuxforhealth/fhir/server/resources/Capabilities.java
@@ -230,12 +230,12 @@ public class Capabilities extends FHIRResource {
         TerminologyCapabilities.Implementation impl;
         if (customTerminologyImpl != null) {
             impl = TerminologyCapabilities.Implementation.builder()
-                    .description(string(buildDescription))
+                    .description(buildDescription)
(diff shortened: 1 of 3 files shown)


fhir-server/src/main/java/org/linuxforhealth/fhir/server/resources/Capabilities.java
```diff
@@ -568,13 +568,13 @@ private CapabilityStatement buildCapabilityStatement(FHIRVersionParam fhirVersio
                 .patchFormat(Code.of(FHIRMediaType.APPLICATION_JSON_PATCH),
                              Code.of(FHIRMediaType.APPLICATION_FHIR_JSON),
                              Code.of(FHIRMediaType.APPLICATION_FHIR_XML))
-                .version(string(buildInfo.getBuildVersion()))
-                .name(string(FHIR_SERVER_NAME))
+                .version(buildInfo.getBuildVersion())
+                .title(FHIR_SERVER_NAME)
```
lmsurpre, line 572: this is the main change

PrasannaHegde1 (approved):
LGTM

PR #4072, opened 2022-11-11
issue-3455 - adding resource info to transaction bundle audit message

PrasannaHegde1:


fhir-server-test/src/test/java/org/linuxforhealth/fhir/server/test/FHIRServerTestBase.java
```diff
@@ -547,13 +559,18 @@ protected boolean isSortingSupported() throws Exception {
...
-
+        boolean txnSupported = false;
         List<CapabilityStatement.Rest> restList = conf.getRest();
         if (restList != null) {
             CapabilityStatement.Rest rest = restList.get(0);
             if (rest != null) {
                 assertEquals(RestfulCapabilityMode.SERVER.getValue(), rest.getMode().getValue());
                 if (rest.getInteraction() != null) {
+                    for (org.linuxforhealth.fhir.model.resource.CapabilityStatement.Rest.Interaction interaction : rest.getInteraction()) {
+                        if (interaction.getCode().getValue().equals(SystemRestfulInteraction.TRANSACTION.getValue())) {
```
lmsurpre, line 570: just test code, but I think we should still prefer the enum flavor:
```suggestion
                        if (interaction.getCode().getValueAsEnum() == SystemRestfulInteraction.TRANSACTION) {
```

lmsurpre: i guess the current code already does the string comparison though (line 583), so maybe better to be consistent...

fhir-server-test/src/test/java/org/linuxforhealth/fhir/server/test/BundleTest.java
```diff
@@ -127,6 +155,14 @@ public void retrieveConfig() throws Exception {
 
         deleteSupported = isDeleteSupported();
         System.out.println("Delete operation supported?: " + deleteSupported.toString());
+        
+        kafkaAuditEnabled = isAuditLogSupported();
+        logger.info("kafkaAuditEnabled flag " +kafkaAuditEnabled);
```
lmsurpre, line 160: very minor, but we usually use spaces before and after the `+` operator
```suggestion
        logger.info("kafkaAuditEnabled flag " + kafkaAuditEnabled);
```
applies to a number of places in this PR

fhir-server-test/src/test/java/org/linuxforhealth/fhir/server/test/FHIRServerTestBase.java
```diff
@@ -890,4 +905,45 @@ private String getLink(Bundle bundle, String relation) {
...
+     */
+    protected boolean isAuditLogSupported() throws Exception {
+        Boolean auditLogSupported = Boolean.FALSE;
+        CapabilityStatement conf = retrieveConformanceStatement();
+        List<Extension> extensions = conf.getExtension();
+        String auditLogServiceName = null;
+        if (extensions != null) {
+            for (Extension extension : extensions) {
+                if (extension.getUrl().contains("auditLogServiceName")) {
+                    auditLogServiceName = extension.getValue().as(org.linuxforhealth.fhir.model.type.String.class).getValue();
```
lmsurpre, line 938: fyi we have a constant that we sometime statically import for this class:  `ModelSupport.FHIR_STRING`

fhir-server-test/src/test/resources/fhir-server-config.json
```diff
@@ -0,0 +1,233 @@
+{
+  "__comment": "FHIR Server configuration",
```
lmsurpre, line 2: i've been putting test configurations like this one under `fhir-server-webapp/src/test/liberty/config` ...does it make sense to move this one there and give it a name that indicates what its for?

fhir-server-test/src/test/resources/test.properties
```diff
@@ -60,3 +60,5 @@ test.bulkdata.export.s3.enabled = false
 # Enable/disable the Azure storage provider tests
 test.bulkdata.import.azure.enabled = false
 test.bulkdata.export.azure.enabled = false
+test.audit.kafka.connectionInfo = localhost:29094,localhost:49098
```
lmsurpre, line 63: it probably makes sense to separate these visually from the bulkdata properties
```suggestion

# Audit log testing parameters
test.audit.kafka.connectionInfo = localhost:29094,localhost:49098
```

fhir-server/src/main/java/org/linuxforhealth/fhir/server/util/RestAuditLogger.java
```diff
@@ -496,71 +497,29 @@ private static void logBundleBatch(AuditLogService auditLogSvc, HttpServletReque
...
-                    }
-                }
+            Iterator<Bundle.Entry> iter = requestBundle.getEntry().iterator();
+            for (Entry bundleEntry : responseBundle.getEntry()) {
+                Bundle.Entry requestEntry = iter.next();
+                entry = initLogEntry(AuditLogEventType.FHIR_BUNDLE);
+                populateAuditLogEntry(entry, request, bundleEntry.getResource(), startTime, endTime, responseStatus);
+                entry.setDescription("FHIR Bundle Transaction request");
+                entry.getContext().setAction(selectActionForBundleEntry(requestEntry));
+                auditLogSvc.logEntry(entry);
```
lmsurpre, line 516: why not just use bundleEntry here and remove the "iter" variable?

lmsurpre, line 516: why not just use bundleEntry here and remove the "iter" and "requestEntry" variables?

lmsurpre (approved):

fhir-server/src/main/java/org/linuxforhealth/fhir/server/util/RestAuditLogger.java
PrasannaHegde1: requestEntry is being used to get the request method of Bundle entry in                "entry.getContext().setAction(selectActionForBundleEntry(requestEntry));"

fhir-server-test/src/test/java/org/linuxforhealth/fhir/server/test/FHIRServerTestBase.java
PrasannaHegde1: This has been fixed as suggested. Removed redundant check in line 583 as well.

fhir-server-test/src/test/java/org/linuxforhealth/fhir/server/test/BundleTest.java
PrasannaHegde1: This has been fixed as suggested.

fhir-server-test/src/test/java/org/linuxforhealth/fhir/server/test/FHIRServerTestBase.java
PrasannaHegde1: This has been fixed as suggested.

fhir-server-test/src/test/resources/fhir-server-config.json
PrasannaHegde1: This has been fixed as suggested. Removed the fhir-server-config.json added to fhir-server-test  module and reused existing config "fhir-server-webapp/src/test/liberty/config/config/default/fhir-server-config-audit-cicd.json"

fhir-server-test/src/test/resources/test.properties
PrasannaHegde1: This has been fixed as suggested. Also made changes to fetch the kafka topic name from "fhir-server-webapp/src/test/liberty/config/config/default/fhir-server-config-audit-cicd.json"

fhir-server-test/src/test/java/org/linuxforhealth/fhir/server/test/FHIRServerTestBase.java
```diff
@@ -567,23 +568,13 @@ protected boolean isTransactionSupported() throws Exception {
                 assertEquals(RestfulCapabilityMode.SERVER.getValue(), rest.getMode().getValue());
                 if (rest.getInteraction() != null) {
                     for (org.linuxforhealth.fhir.model.resource.CapabilityStatement.Rest.Interaction interaction : rest.getInteraction()) {
-                        if (interaction.getCode().getValue().equals(SystemRestfulInteraction.TRANSACTION.getValue())) {
+                        if (interaction.getCode().getValueAsEnum() == SystemRestfulInteraction.TRANSACTION.getValueAsEnum()) {
```
lmsurpre, line 571: FWIW I was trying to recommend 
`interaction.getCode().getValueAsEnum() == SystemRestfulInteraction.Value.TRANSACTION` but I think this is fine

fhir-server-test/src/test/resources/test.properties
```diff
@@ -60,5 +60,6 @@ test.bulkdata.export.s3.enabled = false
 # Enable/disable the Azure storage provider tests
 test.bulkdata.import.azure.enabled = false
 test.bulkdata.export.azure.enabled = false
-test.audit.kafka.connectionInfo = localhost:29094,localhost:49098
-test.audit.kafka.topicName = FHIR_AUDIT
```
lmsurpre: `test.audit.kafka.topicName` isn't needed in this properties file any more?

lmsurpre: ah, i see we get it from the fhir-server-config now (via ConfigurationTranslator); great

lmsurpre:
I would have liked to "squash and merge" this, but prasanna has other changes on top of these ones, so I'm keeping all the commits

fhir-server-test/src/test/resources/fhir-server-config.json
lmsurpre: great.  its not great that the test depends on this other file in this other module (likely brittle) but I think I like that more than duplicating that file here.  we'll find out in the long run :-)

fhir-server/src/main/java/org/linuxforhealth/fhir/server/util/RestAuditLogger.java
lmsurpre: i think i missed the fact that requestEntry is from the requestBundle whereas the bundle entry is from the **responseBundle**

Issue #4101, opened 2022-11-28
fhir-smart not enforced for JSONPatch and FHIRPathPatch


jornvanwier:
**Describe the bug**

`fhir-smart` does appear to enforce JSONPatch and FHIRPathPatch requests. When using a JWT with `ff94628809f34970a7a8199bccf2f23c` as one of the `patient_id`s and a `"patient/*.read"` scope the following happens:

This PUT request gives a 403 as expected:
```
curl 'http://localhost:9080/fhir-server/api/v4/Patient/ff94628809f34970a7a8199bccf2f23c' -H "Authorization: Bearer $bearer" -H 'Content-Type: application/json' -d '{"resourceType": "Patient", "id": "ff94628809f34970a7a8199bccf2f23c", "gender":"female"}' -X PUT
```

However, if I do what is essentially the same request using JSONPatch, the change is allowed:
```
curl 'http://localhost:9080/fhir-server/api/v4/Patient/ff94628809f34970a7a8199bccf2f23c' -H "Authorization: Bearer $bearer" -H 'Content-Type: application/json-patch+json' -d '[{ "op": "add", "path": "/gender", "value": "female" }]' -X PATCH
```

With FHIRPathPatch the same can be observed.

I initially suspected that it was because `org.linuxforhealth.fhir.smart.AuthzPolicyEnforcementPersistenceInterceptor` does not have a `beforePatch` method, but after adding this method (with the same implementation as `beforeUpdate`) the problem persisted. 

Am I missing something, or is this a bug in `fhir-smart`?

**Environment**

I'm currently still using 4.11.1. Looking through the changelogs and relevant code I don't believe this has already been fixed in the newer versions. 

**To Reproduce**

1. Have a JWT with read-only access to a patient (I'm using `"patient/*.read"`)
2. Perform a JSONPatch or FHIRPathPatch request for data belonging to the current patient.
3. The request will succeed.

**Expected behavior**

The request should not succeed unless the JWT has a write scope for the resource.

PrasannaHegde1:
Generated a token with scope "patient/*.read" for patient_id 184f5967297-a894709d-f46c-4746-b4af-5d4498ea568e
1. The JSONPatch request for patient_id 184f5967297-a894709d-f46c-4746-b4af-5d4498ea568e resulted in a 403 error as expected.

````
https://localhost:9443/fhir-server/api/v4/Patient/1850f9523a4-c7c2cc41-6544-404b-843a-7620eac731d9
[...]

PR #4105, opened 2022-11-30
issue #4101 - call beforeX interceptor methods prior to checking if update is skippable

lmsurpre:
Previously, a skippable update (one that would not affect the contents of the resource being updated) would result in a success without invoking beforeUpdate or beforePatch interceptor methods.

For users of fhir-smart, this behavior is confusing because the user may not have write access to this resource type, but the server is "allowing" the write anyway (even though its really being optimized away).


PrasannaHegde1 (approved):
LGTM

Issue #4193, opened 2023-03-02
Schematool does not accept `-` in usernames


bauerjs1:
**Describe the bug**
When there are dashes (`-`) in the Postgres username, the schematool fails with the following exception:
```
Running create-schemas task
2023-03-02 10:01:21.542 00000001    INFO forhealth.fhir.schema.app.Main Opening connection to: jdbc:postgresql://pg-albatross.databases:5432/polar_fhir_ibm
2023-03-02 10:01:22.186 00000001    INFO utils.postgres.PostgresAdapter The schema 'FHIR_ADMIN' is created or already exists
2023-03-02 10:01:22.189 00000001    INFO utils.postgres.PostgresAdapter The schema 'FHIRDATA' is created or already exists
2023-03-02 10:01:22.193 00000001    INFO utils.postgres.PostgresAdapter The schema 'FHIR_OAUTH' is created or already exists
2023-03-02 10:01:22.197 00000001    INFO utils.postgres.PostgresAdapter The schema 'FHIR_JBATCH' is created or already exists
2023-03-02 10:01:22.202 00000001    INFO forhealth.fhir.schema.app.Main Processing took:   0.673 s
2023-03-02 10:01:22.203 00000001    INFO forhealth.fhir.schema.app.Main SCHEMA CHANGE: OK
Running update-schema task
2023-03-02 10:01:22.934 00000001  SEVERE forhealth.fhir.schema.app.Main bad argument

Invalid SQL object name: 'my-user'
java.lang.IllegalArgumentException: Invalid SQL object name: 'my-user'
	at org.linuxforhealth.fhir.database.utils.common.DataDefinitionUtil.assertValidName(DataDefinitionUtil.java:152)
	at org.linuxforhealth.fhir.schema.app.Main.parseArgs(Main.java:1180)
	at org.linuxforhealth.fhir.schema.app.Main.main(Main.java:2006)
Usage: 
[...]
2023-03-02 10:01:23.062 00000001  SEVERE forhealth.fhir.schema.app.Main SCHEMA CHANGE: BAD ARGS
```
Names seem to be verified against [this](https://github.com/LinuxForHealth/FHIR/blob/main/fhir-database-utils/src/main/java/org/linuxforhealth/fhir/database/utils/common/DataDefinitionUtil.java#L24) pattern.

**Environment**
Helm chart `lfh-helm/fhir-server:0.9.1` using `ghcr.io/linuxforhealth/fhir-schematool:5.1.1`

[...]

PR #4209, opened 2023-04-12
Schematool: allow `-` in database usernames

bauerjs1:
Solves #4193 

When there are dashes (-) in the Postgres username, the schematool fails with an exception. In Postgres, dashes are legal in usernames, so the verifying pattern is adapted accordingly.

### DIFF
--- a/fhir-database-utils/src/main/java/org/linuxforhealth/fhir/database/utils/common/DataDefinitionUtil.java
+++ b/fhir-database-utils/src/main/java/org/linuxforhealth/fhir/database/utils/common/DataDefinitionUtil.java
@@ -21,7 +21,7 @@ import org.linuxforhealth.fhir.database.utils.model.OrderedColumnDef;
  * Handles common syntax for generating DDL
  */
 public class DataDefinitionUtil {
-    private static final String NAME_PATTERN_RGX = "[a-zA-Z_]\\w*$";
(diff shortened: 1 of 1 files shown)


fhir-database-utils/src/main/java/org/linuxforhealth/fhir/database/utils/common/DataDefinitionUtil.java
```diff
@@ -21,7 +21,7 @@
  * Handles common syntax for generating DDL
  */
 public class DataDefinitionUtil {
-    private static final String NAME_PATTERN_RGX = "[a-zA-Z_]\\w*$";
+    private static final String NAME_PATTERN_RGX = "[a-zA-Z_][-\\w]*$";
```
lmsurpre, line 24: today i [learned](https://stackoverflow.com/questions/3697202/including-a-hyphen-in-a-regex-character-bracket) that you can use an unescaped `-` to match within a character class if its the first (or last) char... i always escape these :-)

lmsurpre (approved):
LGTM

fhir-database-utils/src/main/java/org/linuxforhealth/fhir/database/utils/common/DataDefinitionUtil.java
jobo-prod: I always just use [regex101](https://regex101.com/) for this 😄

